PR #4, opened 2026-04-06
SOL-148234: [Prod] Add golangci-lint config and CI job

solace-lootii:
Add golangci-lint config and CI lint job to catch issues early before the composite executor and registry work lands.

- .golangci.yml with errcheck, staticcheck, gosec, govet, revive, bodyclose, ineffassign, noctx
- Lint job in CI pinned to golangci-lint v2.11.4
- -race flag added to test step to catch data races on shared state (BrokerPool)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,6 +4,22 @@ on:
   push:
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v2.11.4
+
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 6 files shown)


cmd/server/main.go
```diff
@@ -62,13 +33,14 @@ func main() {
...
+			http.Error(w, "failed to write response", http.StatusInternalServerError)
+		}
 	})
 
-	// 7. Start server
-	addr := fmt.Sprintf(":%s", cfg.Port)
-	log.Printf("MCP server listening on %s", addr)
-	if err := http.ListenAndServe(addr, mux); err != nil {
+	addr := fmt.Sprintf(":%s", port)
+	log.Printf("MCP server listening on %s", addr) //nolint:gosec // addr is derived from config, not external user input
```
amitmorade: did gosec flag this log statement as an issue?

.github/workflows/build-and-test.yml
```diff
@@ -25,7 +41,10 @@ jobs:
         run: go vet ./...
```
bczoma, line 41: Is this step already covered in golangci ?

cmd/server/main.go
solace-lootii: @amitmorade yes it did, gosec flagged it as G706 (log injection). I believe the concern is that a tainted string could end up in a log statement and inject fake log entries. In this case addr is just ":9090" derived from config, so there's no real risk. The nolint suppresses the false positive with an explanation of why it's safe.

.github/workflows/build-and-test.yml
solace-lootii: Oh yes, good catch @bczoma , govet in golangci-lint covers go vet (redundant). It's harmless to keep but can be removed to avoid the duplication

bczoma (approved):
LGTM

PR #9, opened 2026-04-16
SOL-148431: Implement Queue and Client Detail Monitoring Tools

solace-lootii:

Implements three composite monitoring tools for SOL-148431: `get-queue-metrics`, `get-client-details`, and `list-client-subscriptions`. These give an AI agent the ability to diagnose queue backlogs, identify slow subscribers, and inspect client subscription patterns by querying the SEMPv2 Monitor API.

Also fixes a URL path encoding bug discovered during live integration testing — client names containing forward slashes (e.g. `app/client-1`) were inserted raw into the SEMP URL path, causing SEMP to return `INVALID_PATH` errors. Fixed in `HTTPClient.buildURL` using `url.PathEscape`.

## How is this accomplished?

All three tools are implemented as YAML composite tool definitions — no new Go files. Each tool maps to a single SEMPv2 monitor operation via the existing composite engine:

- `get-queue-metrics`: calls `getMsgVpnQueue`, params: `msgVpnName` + `queueName`
- `get-client-details`: calls `getMsgVpnClient`, params: `msgVpnName` + `clientName`
- `list-client-subscriptions`: calls `getMsgVpnClientSubscriptions`, params: `msgVpnName` + `clientName` + optional `maxResults` (defaults to 100, passed as SEMP `count` query param)

The private SEMPv2 monitor spec (`__private_monitor__`) is used instead of the public monitor spec. This gives access to additional fields such as `bindCount` and `subscriptionCount` referenced in the story. The embedded spec file has been updated accordingly and `deriveSpecType` normalizes `__private_monitor__` → `monitor` so all tool definitions and downstream code are unchanged.

All tools use `strategy: collect` which returns the full raw SEMP response — no field filtering. This is intentional and gives the AI more data to reason with.

Pagination is handled by passing `count` to SEMP — the raw response includes `nextPageUri` when more results exist, which the AI can act on.

## Files changed

| File | Change |
|------|--------|
| `internal/composite/definitions/tools.yaml` | Added 3 new tool definitions |
| `internal/semp/sempv2/client.go` | `buildURL`: use `url.PathEscape` for path parameters |
| `internal/semp/sempv2/operation.go` | `deriveSpecType`: normalize `__private_monitor__` → `monitor` |
| `internal/semp/sempv2/specs/` | Replaced public monitor/action/config specs with private monitor spec |
| `internal/composite/executor_test.go` | Added 7 new tests + 2 new mock operations + 3 helper functions |
| `internal/composite/loader_test.go` | Added 3 new loader tests |
| `internal/semp/sempv2/operation_test.go` | Updated tests for private spec path prefix |
| `internal/semp/sempv2/client_test.go` | Added `TestClient_Execute_PathEncoding` |

## Anything reviewers should focus on / be aware of?

- **Private vs public spec**: Switched to `semp-v2-swagger-private-monitor-extended.json`. The `__private_monitor__` basePath is normalized to `monitor` at parse time so wire URLs use the private path automatically — confirmed `bindCount` field appears in live responses
- **`url.PathEscape` vs `url.QueryEscape`**: PathEscape is correct here since these are URL path segments, not query values. `QueryEscape` encodes spaces as `+` which breaks paths
- **`count` arg naming**: SEMP's `countQuery` ref resolves to a query param named `count`. The YAML passes it as `count` accordingly — found during live pagination testing where `maxResults` was being silently ignored

## What testing was performed?

Unit tests pass. Live integration tested against BookIt VMR (192.168.134.40):

- `get-queue-metrics`: confirmed backlog detection on a real queue with 20 unacked messages and no bound consumers — `bindCount: 0` confirmed (field only available in private API)
- `get-client-details`: confirmed client metrics for a WebSocket subscriber including slow subscriber status — URL encoding fix verified with a slash-containing client name
- `list-client-subscriptions`: confirmed subscription listing and pagination (`maxResults: 2` correctly returned 2 of 4 subscriptions with `nextPageUri` in response)
- Error handling: confirmed clean error propagation for a non-existent queue

Pasted Graphic 7

Pasted Graphic 6

Pasted Graphic 8

Pasted Graphic 9

Pasted Graphic 10


internal/semp/sempv2/client.go
amitmorade, line 154: This change affects URL encoding for every SEMP call so I think this needs a few unit tests for buildURL, something like cases for slashes, spaces, and plain values to catch regressions.

solace-lootii: @amitmorade created unit test for this and all test cases passed !

internal/composite/executor_test.go
```diff
@@ -461,6 +543,223 @@ func TestExecute_BrokerParamStripped(t *testing.T) {
...
+	if data["queueName"] != "Orders" {
+		t.Errorf("expected queueName %q, got %v", "Orders", data["queueName"])
+	}
+}
+
+func TestExecute_GetQueueMetrics_NotFound(t *testing.T) {
+	client := newMockClient()
+	client.errors["getMsgVpnQueue"] = &sempv2.SEMPError{
+		Operation:  "getMsgVpnQueue",
+		StatusCode: 404,
```
bczoma: This should be 400, consistent with the body. SEMP returns HTTP 400, not 404, for missing resources.

bczoma: Please also review and fix possible multiple occurences

bczoma (changes requested):

amitmorade (approved):
Looks good

bczoma (approved):
LGTM

PR #11, opened 2026-04-20
SOL-148421: Add Authentication Middleware with JWT Validation and Dev Mode

solace-wmaryam:
### What is the purpose of this change?

  This PR implements OAuth 2.1 authentication for the MCP server with support for both
  automated (Client Credentials) and interactive (Authorization Code + PKCE) flows. It
  adds three authentication modes: disabled (local dev), static dev token (shared
  dev/POC), and full JWT validation (production). It also adds the OAuth Protected
  Resource Metadata endpoint (RFC 9728) to enable MCP clients like Claude Code to
  discover the authorization server and initiate browser-based login flows.

### How is this accomplished?

  - Authentication middleware validates Bearer tokens using OIDC discovery and JWKS key
   caching via github.com/coreos/go-oidc/v3
  - Adds development_mode flag to control whether to use static tokens or
   full JWT validation
  - Protected Resource Metadata endpoint at /.well-known/oauth-protected-resource
  advertises the authorization server to MCP clients
  - WWW-Authenticate headers on 401 responses include resource_metadata URL for client
  discovery per MCP spec
  - Secure logging with LogValuer implementations to prevent credential leaks

### Anything reviewers should focus on/be aware of?

  - ClientAuthConfig added with issuer, audience,
  required_scopes, dev_token, and resource_url fields
  - AuthConfig renamed to BrokerAuthConfig (for broker auth vs client
  auth distinction)

### What testing if any was performed?

  - 13 unit tests covering all three auth modes, JWT validation paths, config
  validation, and WWW-Authenticate header format
  - Integration test with real Keycloak OIDC provider using Docker + Terraform for
  complete end-to-end verification
  - All existing tests updated and passing (registry, config, server tests)
  - Manual testing with Client Credentials flow verified


internal/config/config.go
```diff
@@ -42,6 +52,7 @@ var validLogLevels = []string{"debug", "info", "warn", "error"}
 type BrokerConfig struct {
 	URL                string     `yaml:"url"`                  // SEMP API base URL (e.g., "https://broker:1943")
 	InsecureSkipVerify bool       `yaml:"insecure_skip_verify"` // skip TLS cert verification (dev only, self-signed certs)
+	EnvPrefix          string     `yaml:"env_prefix"`      // prefix for credential env vars ({EnvPrefix}_USERNAME, {EnvPrefix}_PASSWORD)
```
amitmorade: can we please delete this field (EnvPrefix), this was removed in a previous PR

internal/config/config.go
```diff
@@ -79,10 +90,23 @@ func (b BrokerConfig) LogValue() slog.Value {
 	return slog.GroupValue(
 		slog.String("url", b.URL),
 		slog.Bool("insecure_skip_verify", b.InsecureSkipVerify),
+		slog.String("env_prefix", b.EnvPrefix),
```
amitmorade: also this one related to EnvPrefix

internal/registry/registry_test.go
```diff
@@ -17,12 +17,22 @@ func newTestPool() *semp.BrokerPool {
 	cfg := &config.ServerConfig{
 		Brokers: map[string]*config.BrokerConfig{
 			"dev": {
-				URL:  "http://localhost:8081",
-				Auth: config.AuthConfig{Mode: "basic", Username: "admin", Password: "admin"},
+				URL:       "http://localhost:8081",
+				EnvPrefix: "DEV",
```
amitmorade: and also this one related to EnvPrefix

internal/registry/registry_test.go
```diff
@@ -17,12 +17,22 @@ func newTestPool() *semp.BrokerPool {
...
+					Mode:     "basic",
+					Username: "admin",
+					Password: "admin",
+				},
 			},
 			"prod": {
-				URL:  "http://localhost:8082",
-				Auth: config.AuthConfig{Mode: "basic", Username: "admin", Password: "admin"},
+				URL:       "http://localhost:8082",
+				EnvPrefix: "PROD",
```
amitmorade: and this one for EnvPrefix

internal/config/config.go
solace-wmaryam: Will do.

internal/registry/registry_test.go
solace-wmaryam: Sounds good. I'm removing all references to this field and also updating the README and Architecture docs

internal/auth/middleware.go
```diff
@@ -0,0 +1,175 @@
...
+			Expiration: time.Now().Add(24 * time.Hour), // Dev tokens don't expire for 24 hours
+		}, nil
+	}
+}
+
+// createOIDCTokenVerifier creates a TokenVerifier that validates JWTs using OIDC.
+// It fetches public keys from the issuer's JWKS endpoint and handles automatic key rotation.
+func createOIDCTokenVerifier(cfg *config.ServerConfig) (sdkauth.TokenVerifier, error) {
+	// Create OIDC provider - fetches .well-known/openid-configuration and JWKS
+	oidcProvider, err := oidc.NewProvider(context.Background(), cfg.ClientAuth.Issuer)
```
amitmorade: Server won't start listening if the IdP is unreachable

oidc.NewProvider(context.Background(), ...) fetches {issuer}/.well-known/openid-configuration at startup. Checked the go-oidc source ([v3.10.0 oidc.go:87-93](https://github.com/coreos/go-oidc/blob/v3.10.0/oidc/oidc.go#L87-L93)) — it uses http.DefaultClient with no internal timeout, so the context we pass is the only deadline. context.Background() has none.

Effect: if the issuer is down, slow, or misconfigured, the call blocks until the OS eventually gives up (minutes). The process is alive but never reaches ListenAndServe — from the outside, the server looks stuck, and the slog.Error at main.go:74 only fires after that long wait.

internal/auth/middleware.go
```diff
@@ -0,0 +1,175 @@
...
+	verifier, err := NewTokenVerifier(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create token verifier: %w", err)
+	}
+
+	// Construct the metadata URL at the server root
+	// Extract base URL (scheme://host) from resource URL
+	var metadataURL string
+	if cfg.ClientAuth.ResourceURL != "" {
+		if parsedURL, err := url.Parse(cfg.ClientAuth.ResourceURL); err == nil {
```
amitmorade: A few concerns with this block:

- url.Parse doesn't fail on bad input like "not a url" — it returns no error but with empty scheme/host, so metadataURL becomes "://".
- When url.Parse does return an error, we ignore it and silently fall back to localhost.
- If the user explicitly set resource_url and it's malformed, falling back to localhost hides the issue instead of surfacing it.

Could this validation move to config.go? There's already a validateBrokerURL helper (in config.go) that checks scheme + host — reusing it for resource_url (and issuer) at config load would fail fast at boot and let middleware trust the values (or middleware can also have its defensive checks but that would just duplicate the effort and might not be useful). Curious what you think.

internal/registry/registry_test.go
amitmorade: Thank you!

internal/auth/middleware.go
solace-wmaryam: Good catch! I've fixed this.

internal/auth/middleware.go
solace-wmaryam: Yes, I agree. I'll move this validation to config.go. Better to fail earlier than later.

amitmorade:
Looks good to me, thanks for addressing those concerns, will let @bczoma approve it

bczoma (approved):
Looks good!

PR #13, opened 2026-04-22
SOL-148427: Implement Tool Manager Foundation

bczoma:
## Summary

Introduces the `internal/tools/` package with the Tool Manager foundation — the abstraction layer that validates parameters, resolves broker connections, routes tool calls to handlers, and produces structured MCP responses.

- **ToolHandler interface** with Handle, Schema, OutputSchema, Annotations, Description, Name methods
- **ToolManager** with CallTool, Route, Register for centralized tool dispatch
- **CompositeToolHandler** adapter wrapping YAML-driven composite tools
- **JSON Schema Draft 7 parameter validation** (gojsonschema) collecting all errors into a single message
- **Structured output** via both `structuredContent` and `TextContent` fallback per MCP spec 2025-06-18
- **Generic envelope output schema** with validation
- **Tool annotations** (ReadOnlyHint) on all monitoring tools
- **Destructive tool WARNING logging** (ready for post-MVP write tools)
- Absorbed `internal/registry/` into `internal/tools/`
- Documented LLM-optimized response filtering as tracked gap in `docs/todo-gaps.md`

## Design Decisions

These were made interactively during implementation:

| Decision | Choice | Rationale |
|---|---|---|
| Package structure | `internal/tools/` absorbs `internal/registry/` | Comment added noting shared plumbing (broker resolution, validation, structured output) should be extracted into a common package when MCP resources are introduced, to avoid duplication |
| ToolContext | Separate `context.Context` param + `*ToolContext` | Standard Go convention; cancellation already propagates end-to-end via existing context plumbing |
| Parameter validation | gojsonschema (JSON Schema Draft 7) | Story's suggestion; mature library. Typed `AddTool[In, Out]` from SDK rejected because composite tools are YAML-driven with dynamic schemas — Go structs per tool would defeat the data-driven design |
| Validation errors | Collect all errors into single message | More useful for AI agents — fix everything in one shot rather than one-by-one |
| Cacheable annotation | Skipped | Not in MCP SDK's `ToolAnnotations` (only ReadOnlyHint, DestructiveHint, IdempotentHint, OpenWorldHint). Caching implementation out of scope for MVP per story |
| Output schemas | Generic envelope schema | SEMP responses are dynamic and vary per operation. Envelope validates structure (object with step ID keys) without constraining SEMP field details |
| Structured output | Both `StructuredContent` + `TextContent` fallback | MCP spec compliance + backwards compatibility with older clients |
| Composite adapter | Single `CompositeToolHandler` type | All YAML tools share one adapter type; each instance wraps a different CompositeTool definition |
| `list-brokers` | Direct `Server.AddTool`, outside ToolHandler | Discovery tool that doesn't call SEMP or need broker resolution — doesn't fit the handler pattern |
| Description guidelines | YAML header comment in `tools.yaml` | Descriptions are authored in YAML, so guidelines belong where they're used |
| Annotations source | YAML field on each tool definition | Supports future write/action tools with different annotations via the same YAML mechanism |

## Tracked Gap: LLM-Optimized Response Filtering

Full SEMP responses are currently returned unfiltered. Added to `docs/todo-gaps.md`:
- **SEMPv2**: `select` query param available but not yet used
- **SEMPv1 / other sources**: No server-side filtering — MCP server must filter post-retrieval
- Requires a general-purpose transformation layer (deferred)

## Test Plan

- [x] `go build ./...` — compiles clean
- [x] `go vet ./...` — no warnings
- [x] `go test -race ./...` — 178 tests pass
- [x] `internal/tools/` coverage at 84.8% (>80% threshold)
- [ ] Peer review


internal/tools/manager.go
```diff
@@ -0,0 +1,192 @@
...
+
+	handler, err := m.Route(name)
+	if err != nil {
+		return nil, err
+	}
+
+	// Extract and resolve broker.
+	brokerAlias, _ := params["broker"].(string)
+	if brokerAlias == "" {
+		slog.Error("tool invoked",
```
amitmorade: There are around 6 error paths and each have their own slog.Error block with near-identical fields. I was thinking that a single deferred log at function exit, with error_type set per path, would give the same telemetry with less duplication. Curious if this was considered, is there a reason to keep them separate?

cmd/server/main.go
```diff
@@ -139,17 +139,18 @@ func main() {
 		Version: version.Version(),
 	}, nil)
 
-	// 7. Register composite tools
-	reg := registry.NewRegistry(server, pool, executor)
-	if err := reg.RegisterAll(tools); err != nil {
-		slog.Error("failed to register tools", slog.String("error", err.Error()))
-		os.Exit(1)
+	// 7. Create tool manager and register composite tools
+	mgr := tools.NewToolManager(pool)
```
amitmorade: feels like the tool registration details are leaking into the bootstrap. I was thinking maybe we could have like one helper so that main goes from describing mechanics back to describing intent. Curious if there's a reason you kept this logic here

internal/tools/manager.go
bczoma: No, this was simply taken over from in the original internal/registry/registry.go and extended.
It makes sense to havea single deferred log at function exit

cmd/server/main.go
bczoma: No specific reason, it got introduced by refactoring

amitmorade (approved):
LGTM

internal/tools/manager.go
amitmorade: yeah I just checked the deleted file, will be more careful, thank you!

PR #17, opened 2026-04-27
SOL-148425: Rate Limiting, Retry Logic, and MCP Error Translation

bczoma:
## Summary

- Add shared `resilience.Sender` package wrapping `hashicorp/go-retryablehttp` with per-broker rate limiting, custom retry policy, and 401 re-auth — used by both SEMPv1 and SEMPv2 clients
- Custom `CheckRetry` policy: 401 basic auth (clear cookie jar + retry once), 401 bearer (fail fast), 429/503 (full retries), other 5xx (retry once)
- Per-broker rate limiting via `time.NewTicker` channel (configurable `RequestMinInterval`, default 100ms)
- MCP-compliant error translation: tool execution errors return `CallToolResult` with `IsError: true` (per MCP spec) instead of opaque JSON-RPC protocol errors, so the LLM can see errors and self-correct
- Parse SEMPv2 `meta.error` into structured fields (`Description`, `SEMPCode`, `SEMPStatus`) instead of discarding as raw body string
- `StructuredContent` on errors carries machine-readable fields (`retryable`, `status`, protocol-specific data); `Content` carries human-readable message using the broker's own error descriptions
- Only `RetriesExhaustedError` is marked `retryable: true`; all client errors and SEMPv1 envelope errors are `retryable: false`

## Key files

| File | Change |
|---|---|
| `internal/semp/resilience/sender.go` | Shared `Sender` struct: rate limiter, retryablehttp, error handler |
| `internal/semp/resilience/retry.go` | Custom `checkRetry` policy, retry state, `OperationIDKey` |
| `internal/semp/resilience/sender_test.go` | 16 tests for rate limiter, 401/429/503/5xx retry, error handler |
| `internal/semp/sempv2/client.go` | Compose `Sender`, parse `meta.error` into `SEMPError` fields |
| `internal/semp/sempv1/client.go` | Compose `Sender`, wire into `Execute()`, add `Close()` |
| `internal/tools/manager.go` | `buildErrorResult`, `isRetryable`, `buildErrorMessage` for MCP error translation |
| `internal/tools/manager_test.go` | Tests for error translation (SEMPv2, SEMPv1, retries exhausted, plain error, limit-error guidance) |
| `internal/semp/broker.go` | `Close()` calls both protocol clients |
| `docs/todo-gaps.md` | Updated deferred items |

## Test plan

- [x] 307 tests pass (`go test -race -count=1 ./...`)
- [x] `go vet ./...` — no issues
- [x] `golangci-lint run ./...` — 0 issues
- [x] 83.5% test coverage (meets >80% DoD)
- [ ] Verify rate limiting with DEBUG log output on a live broker
- [ ] Verify 429/503 retry behavior against a rate-limited broker
- [ ] E2E test suite passes


internal/semp/sempv2/client.go
```diff
@@ -149,6 +238,107 @@ func (c *HTTPClient) Execute(ctx context.Context, op *Operation, args map[string
...
+	switch {
+	case resp.StatusCode == http.StatusUnauthorized: // 401
+		if c.authMode == config.AuthModeBasic && !state.auth401Retried {
+			state.auth401Retried = true
+			// Stale session cookie is the most likely cause. Clear the jar so
+			// the retried request re-sends Basic Auth credentials from scratch.
+			c.jarMu.Lock()
+			jar, _ := cookiejar.New(nil)
+			c.httpClient.Jar = jar
+			c.jarMu.Unlock()
```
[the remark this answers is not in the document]
bczoma: No action. Too big of a change for a benign outcome (fresh jar vs stale jar, both functionally correct).

internal/semp/sempv2/client.go
```diff
@@ -92,25 +132,55 @@ func NewHTTPClient(brokerCfg *config.BrokerConfig, sempCfg *config.SEMPConfig) (
...
+	// When interval > 0, each Execute() blocks until the ticker fires.
+	// When interval == 0, the closed channel makes receives non-blocking (no rate limit).
+	if *sempCfg.RequestMinInterval > 0 {
+		client.rateTicker = time.NewTicker(*sempCfg.RequestMinInterval)
+		client.rateLimiter = client.rateTicker.C
+	} else {
+		ch := make(chan time.Time)
+		close(ch)
+		client.rateLimiter = ch
+	}
```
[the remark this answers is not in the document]
bczoma: Valid observation, but intentional behavior. Fix adds complexity for minimal benefit.

internal/semp/sempv2/client.go
```diff
@@ -74,9 +105,18 @@ func (c *HTTPClient) LogValue() slog.Value {
 	)
 }
 
+// Close releases resources held by the HTTPClient. It stops the rate limiter
+// ticker if one is running. Safe to call multiple times.
+func (c *HTTPClient) Close() {
+	if c.rateTicker != nil {
+		c.rateTicker.Stop()
+	}
+}
```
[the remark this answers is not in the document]
bczoma: Valid but low-priority. Close() is called during shutdown — no requests should follow.

internal/semp/resilience/sender.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("wrapping request: %w", err)
+	}
+
+	return d.retryClient.Do(retryReq)
+}
+
+// Close releases resources held by the Sender. Stops the rate limiter ticker
+// if one is running. Safe to call multiple times.
+func (d *Sender) Close() {
```
amitmorade, line 126: Question more than a request, was the goroutine lifetime here considered? `Ticker.Stop()` doesn't close the ticker channel (stdlib documents this to avoid spurious zero-value reads), so the forwarding goroutine in `New()` stays parked on `for t := range d.rateTicker.C` after `Close()` returns.

I realize this has zero impact in our current model since Senders live for the process lifetime, so this might be a deliberate "not worth the extra channel" call. Wanted to flag it in case the deployment model ever shifts toward runtime broker management — at that point every removed Sender would leak a goroutine with nothing in the code to hint that `Close()` isn't fully cleaning up. If it's intentional for now, a one-line comment saying so would make the constraint visible to whoever reads this next.

internal/semp/resilience/sender.go
```diff
@@ -0,0 +1,161 @@
...
+			slog.Int("status", resp.StatusCode),
+			slog.Int("attempts", numTries))
+		return nil, &RetriesExhaustedError{
+			StatusCode: resp.StatusCode,
+			Attempts:   numTries,
+		}
+	}
+
+	return nil, fmt.Errorf("request failed after %d attempts", numTries)
+}
```
amitmorade, line 161: The rate limiter here can admit up to 3 requests back-to-back after an idle period, even though the comment says only the first request should fire immediately.

Why: there are three storage slots in the pipeline that each hold one tick during idle — the ticker's internal channel buffer, the `t` variable in the blocked forwarding goroutine, and the seeded `ch` itself. When traffic resumes, all three drain into `ch` faster than the throttle interval, so three new requests get permission instantly before the throttle actually kicks in. Subsequent ticks from `time.Ticker` are dropped while the pipeline is full, so it caps at 3 per Sender — but the implicit burst of 3 isn't what the code intends.

For short intervals this is invisible. The case where it could actually matter is when `RequestMinInterval` is long and the broker is fragile — the throttle was presumably sized to space requests carefully, and the burst-of-3 briefly violates that spacing exactly when it's most needed. How tolerant is the broker to that? If "very," this can stay as-is or get a comment. If "not really," I'd suggest swapping the hand-rolled ticker+goroutine+seeded-channel construction for `golang.org/x/time/rate.Limiter` with an explicit burst size of 1 — smaller diff than the current code, and makes the burst policy a single visible argument instead of an emergent property of the pipeline.

internal/semp/resilience/sender.go
```diff
@@ -0,0 +1,161 @@
...
+	retryClient.ErrorHandler = d.errorHandler
+	retryClient.Logger = nil // manual logging in checkRetry
+	d.retryClient = retryClient
+
+	// Per-broker rate limiter: ticker-based interval enforcement.
+	// When interval > 0, each Do() blocks until the channel yields a token.
+	// The first request fires immediately (seeded token); subsequent requests
+	// wait for the ticker. A goroutine forwards ticker events into the channel.
+	// When interval == 0, the closed channel makes receives non-blocking (no rate limit).
+	if *sempCfg.RequestMinInterval > 0 {
```
amitmorade, line 80: Trying to understand the rate limiter design here, unless I'm misreading it, the pipeline can admit up to 3 requests back-to-back after an idle period rather than the 1 the seed comment suggests.

The trace I followed: there are three storage slots that can each hold one tick during idle — the ticker's internal channel buffer, the `t` variable in the blocked forwarding goroutine, and the seeded `ch` itself. When traffic resumes, all three drain into `ch` faster than the throttle interval, so three new requests get permission instantly before the throttle actually kicks in. Further ticks from `time.Ticker` are dropped while the pipeline is full, so it caps at 3 per Sender.

Was an implicit burst of 3 the intent? For short intervals it's invisible either way, so it might just not have come up. The case I'm less sure about is when `RequestMinInterval` is configured long for a fragile broker — the throttle was presumably sized to space requests carefully, and the burst briefly violates that spacing. If the broker tolerates it, this can stay as-is (maybe with a comment noting the actual burst behavior). If not, `golang.org/x/time/rate.Limiter` with an explicit burst size of 1 would be smaller than the current code and make the policy a visible argument rather than emergent. Curious what the original thinking was.

internal/semp/resilience/sender.go
bczoma: Ended up removing the goroutine. See next comment.

internal/semp/resilience/sender.go
bczoma: Thanks Amit, you brought up a valid concern.
The reason for introducing this construct was that there is a per-broker, per-sender 100ms first-request latency.
  - Once per broker after the first tool call targeting that broker                                                                                        
  - Never again for that broker — subsequent calls just wait for the next tick (which is the intended spacing)

This was an attempt to eliminate that but turned about too complex introducing a number of other issues (as you saw it too). I have reverted to the original simple behavior with the one-time request delay of 100ms per broker.

internal/semp/resilience/sender.go
bczoma: See the other comment, backed out this code.

internal/semp/resilience/retry.go
```diff
@@ -0,0 +1,105 @@
...
+	auth401Retried  bool // true after first 401 re-auth attempt (basic auth only)
+	other5xxRetried bool // true after first non-429/503 5xx retry
+}
+
+// OperationIDKey is the context key callers use to attach an operation
+// identifier for logging. The Sender reads it in errorHandler and checkRetry.
+type OperationIDKey struct{}
+
+// getRetryState retrieves the per-request retryState from the context.
+func getRetryState(ctx context.Context) *retryState {
```
amitmorade, line 37: Curious what you think on **getRetryState**, the fallback to a fresh **retryState{}** when the context key is missing means the "retry once" caps would silently turn into "retry up to RetryMax" if a future caller ever skips **Sender.Do**. Today only **Sender.Do** attaches the key, so it's latent, but there's nothing in the code (no doc, no test, no runtime signal) that would point a future contributor at the contract.

Worth either a doc comment on **getRetryState / retryStateKey** marking the contract, or making the missing-key path loud (panic, or force both flags to true so retries stop)? Worth a small note now if it ever bites, the trail back to the cause is pretty thin.

internal/semp/resilience/sender.go
```diff
@@ -0,0 +1,155 @@
...
+// Unwrap returns the underlying error so errors.Is/As can traverse the chain.
+func (e *RetriesExhaustedError) Unwrap() error { return e.Err }
+
+// Sender wraps retryablehttp.Client with per-broker rate limiting, custom retry
+// policy, and 401 re-auth. Both SEMPv1 and SEMPv2 clients compose this to get
+// shared HTTP resilience without duplication.
+//
+// Sender is safe for concurrent use from multiple goroutines. The cookie jar
+// replacement in checkRetry (401 re-auth) races with http.Client.Do reads of
+// the Jar field, but this is benign: the worst case is one retry using a stale
```
amitmorade, line 39: the comment notes the jar swap is a benign race because Basic Auth header alone succeeds. Is there a reason we cannot eliminate the race outright over keeping it?

solace-lootii (dismissed):
Looks good!

internal/semp/resilience/sender.go
bczoma: We acknowledge the race here in the comment, it is not expected to cause an issue. Eliminating the race would add a good chunk of code that would be more confusing.  Even removed the `d.jarMu.Lock()` guard for now as it doesn't fully prevent the race anyway.

internal/semp/resilience/retry.go
bczoma: Added a comment for now. No external caller can reach checkRetry - it's a private callback

amitmorade:
Looks good to me

internal/tools/manager.go
```diff
@@ -271,3 +270,98 @@ func (m *ToolManager) classifyBrokerError(alias string, err error) (string, erro
...
+			if sempv1Err.Kind == sempv1.ErrorKindLimit {
+				msg += ". Reduce the scope of the request."
+			}
+			return msg
+		}
+		if sempv1Err.Kind == sempv1.ErrorKindHTTP {
+			return fmt.Sprintf("SEMPv1 HTTP %d error", sempv1Err.StatusCode)
+		}
+		return fmt.Sprintf("SEMPv1 %s error (status=%d)", sempv1Err.Kind, sempv1Err.StatusCode)
+
```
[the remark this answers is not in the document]
bczoma: There is a  String() method to ErrorKind()

internal/tools/manager.go
```diff
@@ -271,3 +270,98 @@ func (m *ToolManager) classifyBrokerError(alias string, err error) (string, erro
...
+func buildErrorMessage(err error) string {
+	var retriesErr *resilience.RetriesExhaustedError
+	var sempv2Err *sempv2.SEMPError
+	var sempv1Err *sempv1.Error
+
+	switch {
+	case errors.As(err, &retriesErr):
+		return fmt.Sprintf(
+			"Request failed after %d attempts (HTTP %d). Internal retries exhausted.",
+			retriesErr.Attempts, retriesErr.StatusCode)
```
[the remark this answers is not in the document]
bczoma: Fixed

internal/semp/resilience/sender.go
```diff
@@ -0,0 +1,155 @@
...
+	return d
+}
+
+// Do sends an HTTP request through the rate limiter and retryablehttp client.
+// The caller is responsible for building the request and adding authentication.
+// On success, returns the HTTP response (body open for the caller to read).
+// On failure after retries, returns a RetriesExhaustedError or a wrapped error.
+func (d *Sender) Do(ctx context.Context, req *http.Request) (*http.Response, error) {
+	// Rate limit: wait for the per-broker interval before sending a new request.
+	// This does NOT apply to retries — retryablehttp handles those internally.
```
amitmorade: Curious about the rate limiter and retry interaction.

The doc comment already hints at this by noting that retries skip the limiter and rely on retryablehttp internally. Looking at it a bit more, two things compound: retries inside retryClient.Do skip our limiter, and retryablehttp v0.7.8's DefaultBackoff is non-jittered, so concurrent failures could retry in sync. Total broker traffic per interval can end up around (1 + RetryMax) times the configured rate.

Would it be worth gating retries through the limiter as well, or opting into LinearJitterBackoff? If the current behavior is intentional, would expanding the doc comment to spell out that consequence (rather than just the mechanic) be reasonable?

bczoma: Modified to use  `retryablehttp.RateLimitLinearJitterBackoff` - this takes also the `Retry-After header` into account.

For the total broker traffic per interval, retries are already spread over time by backoff (with jitter, even more so). The fix would overwrite retryableHttp behavior and I feel it would be over-engineering

amitmorade (approved):
LGTM

PR #23, opened 2026-05-01
SOL-148433: Implement Remaining Monitoring Tools (DMR, Message Rates, RDP)

solace-lootii:
## Summary

Adds 3 new composite monitoring tools as pure YAML definitions — no new Go files.

- **`get-message-rates`** — Returns the full VPN object including all throughput rate fields (rx/tx message and byte rates, 1-minute averages) via `monitor/getMsgVpn`
- **`get-dmr-status`** — Two-step tool (cluster sequential + links parallel) via `monitor/getDmrCluster` + `monitor/getDmrClusterLinks`. The composite engine has no aggregation capability so `upLinkCount`/`downLinkCount` are not pre-computed — the tool description instructs the AI to derive them by counting links where `up == true/false`
- **`list-rdps`** — Paginated listing of all REST Delivery Points in a VPN via `monitor/getMsgVpnRestDeliveryPoints`, follows pages up to `maxResults` (default 100, max 500), consistent with `list-queues` and `list-clients`

Server now loads `tool_count: 11`.

## Changes

- `internal/composite/definitions/tools.yaml` — 3 new tool definitions
- `internal/composite/loader_test.go` — 3 new loader tests
- `internal/composite/executor_test.go` — 3 new operations in `testOperations()`, tool helpers, and tests covering happy path, SEMP error propagation, single-page results, and `maxResults` truncation

## Testing

- All unit tests pass
- `get-message-rates` and `list-rdps` verified end-to-end via MCP against a live broker
- `get-dmr-status` not testable on dev broker (no DMR clusters configured) — unit tests cover the logic

## Note on `get-dmr-status` computed fields

[The story](https://sol-jira.atlassian.net/browse/SOL-148433) calls for pre-computing `upLinkCount`/`downLinkCount`. The composite engine returns raw SEMP data only — no aggregation. Rather than extending the engine or writing a Go-native tool for a simple count, the tool returns the raw cluster object and links array and the description instructs the AI to derive the counts. Open to feedback if a different approach is preferred.

### DIFF
--- a/internal/composite/definitions/tools.yaml
+++ b/internal/composite/definitions/tools.yaml
@@ -266,7 +266,6 @@ tools:
           dmrClusterName: "{{.Params.dmrClusterName}}"
       - id: links
         operation: monitor/getDmrClusterLinks
-        parallel: true
         args:
           dmrClusterName: "{{.Params.dmrClusterName}}"
     result:
--- a/internal/composite/definitions/tools.yaml
+++ b/internal/composite/definitions/tools.yaml
@@ -226,3 +226,75 @@ tools:
           count: "100"
     result:
       strategy: collect
+
+  - name: get-message-rates
+    description: >
+      Get current and average message and byte throughput rates for a Message
+      VPN. Use this to assess traffic volume and identify high-traffic VPNs.
+    annotations:
+      readOnly: true
+    parameters:
+      - name: msgVpnName
+        type: string
+        required: true
+        description: "The name of the Message VPN"
+    steps:
+      - id: rates
+        operation: monitor/getMsgVpn
+        args:
+          msgVpnName: "{{.Params.msgVpnName}}"
+    result:
+      strategy: collect
+
+  - name: get-dmr-status
+    description: >
+      Get the status of a DMR cluster including its enabled state, uptime, and
+      all link statuses. Returns the cluster object and a links array. To find
+      upLinkCount and downLinkCount, count links where up is true or false
+      respectively. Use this to diagnose mesh connectivity issues.
+    annotations:
+      readOnly: true
+    parameters:
+      - name: dmrClusterName
+        type: string
+        required: true
+        description: "The name of the DMR cluster"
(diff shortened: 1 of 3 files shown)


internal/composite/definitions/tools.yaml
```diff
@@ -226,3 +226,75 @@ tools:
           count: "100"
     result:
       strategy: collect
+
+  - name: get-message-rates
+    description: >
```
amitmorade, line 231: Just a note that this is exactly identical to the get-vpn-health tool. I understand that this could be because we are not filtering the response. Not sure what the story says? @solace-lootii @bczoma

internal/composite/definitions/tools.yaml
```diff
@@ -226,3 +226,75 @@ tools:
...
+        required: true
+        description: "The name of the DMR cluster"
+    steps:
+      - id: cluster
+        operation: monitor/getDmrCluster
+        args:
+          dmrClusterName: "{{.Params.dmrClusterName}}"
+      - id: links
+        operation: monitor/getDmrClusterLinks
+        parallel: true
```
amitmorade: **parallel: true** on links is a no-op here, since **cluster** is sequential and **links** is the only step after it, the batching logic in executor.go puts **links** alone in a parallel batch. It runs in an errgroup of one (functionally sequential, with a small amount of extra goroutine overhead).

Compare with get-rdp-status, where two **parallel: true** steps actually run concurrently — that's what the keyword usually signals in this file.

Suggest dropping **parallel: true** so the YAML matches actual execution. Not blocking.

internal/composite/definitions/tools.yaml
bczoma: Yes, noticed too and this shall be fine for this PR. The next step where we aim to filter to a subset of results shall make the difference

bczoma (commented):
I don't have additional comments

amitmorade (approved):
LGTM

PR #24, opened 2026-05-04
SOL-149343: Fix misleading list tool descriptions that claim to return all results

bczoma:
- List tools (list-vpns, list-queues, list-clients, list-client-subscriptions) described themselves as returning "all" results while silently defaulting to 100, causing LLM callers to miss the `maxResults` parameter and receive truncated results

## How is this accomplished?
- Removed misleading "all" from list tool descriptions
- Added "Returns up to 100 by default (max 500 via maxResults)" to all four list tool descriptions
- Replaced SEMP pagination implementation detail in list-client-subscriptions with the consistent pattern
- Added pagination limits guideline to Tool Description Guidelines header referencing executor.go constants

- Description-only change, no logic changes
- Verify the wording is clear for both LLM and human consumers

- All 102 tests in composite and tools packages pass


amitmorade (approved):
Looks good to me

PR #27, opened 2026-05-11
SOL-149341: OAuth Testing and Documentation

solace-wmaryam:
## Summary

This PR fixes some minor bugs found with the OAuth flow and adds user documentation in AUTHENTICATION.md providing instructions and guidance on how to set up the three authentication modes for the first hop (i.e. MCP client to server). It also updates the OAuth end-to-end test infrastructure to support browser-based auth without requiring custom scopes.

Fixes[ SOL-149341](https://sol-jira.atlassian.net/browse/SOL-149341)

## Type of Change

- [x] Documentation update

## Motivation and Context

During testing, two bugs were identified that were preventing authentication from working properly: scope enforcement in dev mode always rejected static tokens (which carry no scopes), and Go's default plain-text 404 responses caused JSON parse errors in MCP SDK clients probing OAuth discovery endpoints. 
Additionally, the Keycloak integration test setup assumed custom scopes were always required when they were actually optional. 
We also needed to add user-facing documentation explaining how to configure the various modes of authentication (dev mode with no authentication, dev mode using a static token, and production mode using the full OAuth with an identity provider).

## Changes Made

- Added AUTHENTICATION.md. This is a comprehensive guide covering all three auth modes (no auth, static dev token, OAuth/JWT), including general guidance for Dynamic Client Registration, audience mapper configuration, and troubleshooting
- Bug Fix: Added code to skip scope enforcement in development mode. Static dev tokens carry no scopes, so requiring them in dev mode would always reject the token with 403. Scopes are now only enforced in production mode.                                                                                
- Bug Fix: Added code to return JSON 404 for unregistered paths — MCP SDK clients probe multiple OAuth discovery endpoints during authentication (e.g. /.well-known/oauth-authorization-server, /authorize, /register). Go's default plain-text "404 page not found" caused JSON parse errors in those clients. A JSON response lets them fail gracefully.
- Updated the OAuth integration test to work without requiring custom scopes (solace:read, solace:write). This included removing per-client audience mappers from Terraform which was preventing Dynamic Client Registration from working since DCR clients didn't receive the audience claim. Instead, we are now configuring the Keycloak container post-Terraform in setup-keycloak.sh for DCR compatibility with an audience mapper on the basic scope (applies to all clients including DCR), an openid placeholder scope, and registration policy updates (allowed scopes, trusted hosts)                                  - Removes required_scopes from test-config.yaml and scope validation from test scripts

## Testing

**Test Environment:**
- Go version: 1.26.2
- OS: macOS (darwin/arm64)
- Broker version (if applicable): Latest main

**Tests Performed:**

**Test Coverage:**

-  Verified full OAuth flow end-to-end with Claude Code: Dynamic Client Registration, browser-based Authorization Code + PKCE login, token issuance with correct audience claim, and MCP server accepting the token                                                                        
- Tested dev mode works with no authentication, as well as with a static token configured
- Verified JSON 404 catch-all returns well-formed responses for unregistered OAuth discovery paths
- Verified dev mode with static token no longer rejects requests due to missing scopes

## Breaking Changes

There are no breaking changes. The `required_scopes` config field remains supported. The JSON 404 catch-all replaces Go's default plain-text response, which is a non-breaking improvement

**For Reviewers:**

**Focus Areas**: 

**Questions**:


cmd/server/main.go
```diff
@@ -277,12 +275,23 @@ func main() {
...
 		mux.Handle("/.well-known/oauth-protected-resource", metadataHandler)
 		slog.Info("registered OAuth protected resource metadata endpoint")
 	}
 
+	// Catch-all: return JSON 404 for any unregistered path. MCP SDK clients
+	// probe multiple OAuth discovery endpoints during (re-)authentication
+	// (e.g. /.well-known/oauth-authorization-server, /authorize, /register).
+	// Go's default plain-text "404 page not found" causes JSON parse errors
+	// in those clients. A JSON body lets them fail gracefully.
+	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
amitmorade, line 292: Nice fix for the JSON 404
could we log the unmatched path at debug level?

docs/client-authentication.md
```diff
@@ -0,0 +1,289 @@
...
+4. The client performs Dynamic Client Registration (RFC 7591) with the authorization server to register itself as an OAuth client
+5. The client initiates an OAuth 2.1 Authorization Code flow (with PKCE) — a browser window opens for the user to log in with the identity provider
+6. After successful login, the client receives a JWT access token and includes it in subsequent requests
+7. The MCP server validates each token's signature (via the IdP's JWKS endpoint), issuer, audience, and expiry
+- Tokens are automatically refreshed by the client when they expire
+- You'll see a log message: `"using JWT token for authentication — production mode"`
+
+### Setting up your identity provider
+
+The MCP server works with any OIDC-compliant identity provider. The MCP OAuth flow uses **Dynamic Client Registration** (RFC 7591) — the MCP client registers itself with the IdP at runtime rather than using a pre-configured client ID. This means your IdP must be configured to allow anonymous client registration.
```
amitmorade: @solace-wmaryam this line needs to be redefined for the reasons we discussed.

For others reviewing this PR, the context here is that MCP spec says:

Preregistration
MCP clients SHOULD support an option for static client credentials such as those supplied by a preregistration flow. This could be:

- Hardcode a client ID (and, if applicable, client credentials) specifically for the MCP client to use when interacting with that authorization server, or
- Present a UI to users that allows them to enter these details, after registering an OAuth client themselves (e.g., through a configuration interface hosted by the server).

[​](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#dynamic-client-registration)
Dynamic Client Registration
MCP clients and authorization servers MAY support the OAuth 2.0 Dynamic Client Registration Protocol [RFC7591](https://datatracker.ietf.org/doc/html/rfc7591) to allow MCP clients to obtain OAuth client IDs without user interaction. This option is included for backwards compatibility with earlier versions of the MCP authorization spec

bczoma (dismissed):
LGTM

amitmorade (approved):
LGTM

PR #31, opened 2026-05-12
SOL-149664: Fix .env parser: strip quoted values and warn on unreadable file

solace-aross:
## Summary

- **Strip matched quotes** — `.env` lines like `PASSWORD="secret"` now load as `secret`, not `"secret"`. Single quotes work the same way. Unbalanced or unquoted values are untouched. This matches the `godotenv`/Docker convention and fixes silent credential corruption causing broker auth failures.
- **Warn on unreadable file** — Only `os.ErrNotExist` is now treated silently. Any other read error (e.g. `permission denied`) logs `slog.Warn("env file unreadable", path=...)` so operators get a pointer to the real cause instead of a confusing downstream "env var not set" error.
- **Guard empty keys** — Lines like `=value` (no key) are now skipped to prevent `os.Setenv("", ...)`.
- **`stripMatchedQuotes` helper** — Extracted as a named function so it's directly unit-testable.

## Test plan

- [ ] `go test -race ./internal/config/... -run TestLoadEnvFile` — all three new tests pass
- [ ] `go test -race ./internal/config/...` — full suite green, no regressions
- [ ] Manual: create a `.env` with `FOO="bar"` and verify `${FOO}` in a broker config resolves to `bar` (not `"bar"`)
- [ ] Manual: create a `.env` with mode `0000` and verify a `WARN env file unreadable` line appears in server logs at startup
- [ ] Manual: remove the `.env` entirely and verify no warning appears (silent skip preserved)

Closes SOL-149664

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -295,10 +295,12 @@ func applyDefaults(cfg *ServerConfig) {
 
 // loadEnvFile loads a .env file into the process environment. It checks two
 // locations in order: the ENV_FILE environment variable (explicit path), then
-// a .env file in the same directory as the config file (convention). If neither
-// exists, it silently skips — .env files are optional. Variables already set in
-// the environment are not overwritten, so real env vars (e.g., from CI/CD)
-// take precedence over .env values.
+// a .env file in the same directory as the config file (convention). If the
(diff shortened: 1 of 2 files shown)


amitmorade (approved):
LGTM

PR #32, opened 2026-05-12
SOL-149665: Reject http:// broker and auth URLs in production mode

solace-aross:
## Summary

- `validateBrokerURL` now accepts a `productionMode bool` parameter and returns an error when scheme is `http://` and production mode is active — preventing plaintext transmission of basic-auth credentials and bearer tokens
- `validateBroker` and `validate()` thread `!cfg.DevelopmentMode` through to all URL validation call sites (broker URLs, `client_auth.issuer`, `client_auth.resource_url`)
- Removes stale "will be added with the dev/prod flag from Story 1B" comment — enforcement is now in place
- `http://` continues to work in `development_mode: true` (e.g. local broker, E2E tests)

## Test plan

- [ ] `go test -race ./internal/config/... -run "TestLoadConfig_Rejects|TestLoadConfig_Allows|TestLoadConfig_InvalidURLScheme"` — all pass
- [ ] `go test -race ./internal/config/...` — full suite green
- [ ] Manual: config with `development_mode: false` and `http://` broker URL → startup error naming alias and mentioning `development_mode`
- [ ] Manual: E2E config (uses `development_mode: true`) with `http://localhost` broker → still loads cleanly

Closes SOL-149665

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -503,8 +503,11 @@ func ValidatePort(port int) error {
 }
 
 // validateBrokerURL checks that s is a well-formed URL with an http or https
-// scheme and a host component. When productionMode is true, http:// is
-// rejected — credentials would otherwise be transmitted in plaintext.
+// scheme and a host component. This is structural validation only — it does
+// NOT verify that the host resolves, is reachable, or actually runs a SEMP
+// endpoint. Those are deliberately runtime concerns surfaced on the first
+// tool call. When productionMode is true, http:// is rejected — credentials
+// would otherwise be transmitted in plaintext.
 func validateBrokerURL(s string, productionMode bool) error {
 	u, err := url.Parse(s)
 	if err != nil {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -514,7 +514,7 @@ func validateBrokerURL(s string, productionMode bool) error {
 		return fmt.Errorf("url scheme must be http or https, got %q", u.Scheme)
 	}
 	if productionMode && u.Scheme == "http" {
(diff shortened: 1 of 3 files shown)


internal/config/config.go
```diff
@@ -502,20 +502,20 @@ func ValidatePort(port int) error {
...
+func validateBrokerURL(s string, productionMode bool) error {
 	u, err := url.Parse(s)
 	if err != nil {
 		return fmt.Errorf("url is not a valid URL: %w", err)
 	}
 	if u.Scheme != "http" && u.Scheme != "https" {
 		return fmt.Errorf("url scheme must be http or https, got %q", u.Scheme)
 	}
+	if productionMode && u.Scheme == "http" {
+		return fmt.Errorf("url scheme must be https when development_mode is false (got %q); set development_mode: true to allow http", s)
```
amitmorade: Should the error message suggest **set development_mode: true** as the remediation? Most of the time the correct fix is "use an https URL," not "lower the security gate." Curious whether the hint should steer toward the secure fix instead and leave the dev-mode escape hatch undocumented in the error text(users who genuinely need it will find it in the docs)

solace-aross: Good call — agreed that the message shouldn't steer users toward lowering the security gate. Reworded in 10b0f07:

`url scheme must be https to protect credentials in transit (got %q)`

No mention of `development_mode` in the error; folks who genuinely need it will find it in the docs. Test assertion updated to match. Thanks for the catch.

amitmorade (dismissed):
LGTM

bczoma (approved):
LGTM

PR #34, opened 2026-05-12
SOL-149667: Config code quality: sorted errors, comment fix, redundant guard, octal

solace-aross:
## Summary

Four mechanical fixes in `internal/config/`, all in the same file area:

- **Sorted broker error ordering** — `validate()` now iterates broker aliases via `slices.Sorted(maps.Keys(...))` so the joined validation error string always lists brokers alphabetically. Previously random map iteration made error messages non-deterministic across runs.
- **Remove redundant guard** — `cfg.SEMP.RetryMinInterval > 0` was checked again in the max-min comparison (`else if` at line 411), despite the preceding check at line 405 already guaranteeing `> 0`. Removed the redundant clause; `TestLoadConfig_RateLimit_MaxSmallerThanMin` still passes.
- **Fix misleading comment** — `applyEnvOverrides` doc said "runs after validation" but it runs *before* `validate()`. Corrected to "runs before validate() so overridden values are still range-checked."
- **Fix unprefixed octal** — `0644` → `0o644` in `config_test.go:32`. Swept rest of repo — this was the only occurrence.

## Test plan

- [ ] `go test -race ./internal/config/... -run TestValidate_BrokerErrorsAreSorted` — passes
- [ ] `go test -race ./internal/config/...` — full suite green
- [ ] `grep -rn --include="*.go" -E '\b0[0-9]{3}\b' .` — returns no matches
- [ ] `grep "after validation" internal/config/config.go` — returns no matches

Closes SOL-149667

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"maps"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


bczoma (dismissed):
LGTM

bczoma (approved):
LGTM

PR #37, opened 2026-05-12
SOL-149670: Reject unknown fields in composite YAML loader

solace-aross:
## Summary

- Switches `yaml.Unmarshal` to `yaml.NewDecoder` + `KnownFields(true)` in `internal/composite/loader.go`
- Typos in tool YAML (e.g. `paralel: true` instead of `parallel: true`) were previously silently dropped; they now return a clear parse error at server startup
- Adds `TestLoadTools_UnknownFieldRejected` covering both top-level and step-level unknown keys (two sub-tests)

## Test plan

- [ ] `go test -race ./internal/composite/...` passes
- [ ] Verify a tool YAML with a deliberate typo (e.g. `bogusField: oops`) causes startup to fail with a descriptive error
- [ ] Existing composite tool YAML (`internal/composite/definitions/`) loads without error

Fixes [SOL-149670](https://sol-jira.atlassian.net/browse/SOL-149670)

[SOL-149670]: https://sol-jira.atlassian.net/browse/SOL-149670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/composite/loader.go
+++ b/internal/composite/loader.go
@@ -15,6 +15,7 @@
 package composite
 
 import (
+	"bytes"
 	"fmt"
 	"io/fs"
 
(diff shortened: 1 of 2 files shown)


solace-lootii (approved):
Ran go test ./internal/composite/..., and confirmed the real definitions/tools.yaml (11 tools) parses cleanly with KnownFields(true). LGTM !

PR #38, opened 2026-05-12
SOL-149671: Add sync.RWMutex to ToolManager to guard concurrent handlers map access

solace-aross:
## Summary

- Adds `sync.RWMutex` field to `ToolManager` struct (`internal/tools/manager.go`)
- `Register()` takes a write lock; `Route()` and `Handlers()` take a read lock; `CallTool()` is protected transitively through `Route()`
- The handlers map was safe today only by single-threaded-startup convention — now explicitly safe for any future dynamic registration scenario

## Test plan

- [ ] `go test -race ./internal/tools/...` passes with zero data-race reports
- [ ] New `TestToolManager_ConcurrentRegisterAndRoute` explicitly exercises 20 parallel writers + 60 parallel readers under the race detector
- [ ] All existing `TestToolManager_*` and `TestCallTool_*` tests pass

Fixes [SOL-149671](https://sol-jira.atlassian.net/browse/SOL-149671)

[SOL-149671]: https://sol-jira.atlassian.net/browse/SOL-149671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/tools/manager.go
+++ b/internal/tools/manager.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log/slog"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/SolaceDev/solace-broker-mcp/internal/composite"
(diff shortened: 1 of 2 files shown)


amitmorade (approved):
LGTM

PR #40, opened 2026-05-13
SOL-149746: Skip retry for non-idempotent HTTP methods (POST/PATCH)

solace-aross:
## Summary
- `checkRetry` in `resilience/retry.go` previously had no HTTP method guard — POST and PATCH requests would retry on 503/5xx, risking double-writes (e.g. creating a queue twice)
- Added method check before all other retry logic; POST and PATCH return `(false, nil)` immediately regardless of status
- The HTTP method is stored in `retryState` at `Do()` time so the guard fires on connection errors too (where `resp` is nil)

## Test plan
- [x] `TestSender_NoRetry_POST_On503` — POST with 503 makes exactly 1 request
- [x] `TestSender_NoRetry_PATCH_On503` — PATCH with 503 makes exactly 1 request  
- [x] `TestSender_NoRetry_POST_On500` — POST with 500 makes exactly 1 request
- [x] All existing retry tests pass (GET/503, GET/429, 5xx-retry-once, 401 re-auth)
- [x] `go test -race ./internal/semp/resilience/...` passes
- [x] `golangci-lint run ./internal/semp/resilience/...` clean

### DIFF
--- a/internal/semp/resilience/retry.go
+++ b/internal/semp/resilience/retry.go
@@ -20,8 +20,9 @@ type retryStateKey struct{}
 // Each Do() call creates its own instance via context, so concurrent requests
 // to the same Sender are safe.
 type retryState struct {
-	auth401Retried  bool // true after first 401 re-auth attempt (basic auth only)
-	other5xxRetried bool // true after first non-429/503 5xx retry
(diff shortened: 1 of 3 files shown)


bczoma (approved):
LGTM

PR #41, opened 2026-05-13
SOL-149747: Replace os.Exit(1) in ListenAndServe goroutine with error channel

solace-aross:
## Summary
- `ListenAndServe` goroutine previously called `os.Exit(1)` on startup failure (port conflict, TLS error), bypassing graceful shutdown and any future deferred cleanup
- Extracted `startServer()` helper that returns a `<-chan error`; errors are sent to the channel instead of calling `os.Exit`
- `main()` now selects on both OS signal and server error; `httpServer.Shutdown()` runs in both cases before the process exits

## Test plan
- [x] `TestStartServer_PortConflict_SendsToChannel` — port conflict error arrives on channel (not os.Exit)
- [x] `TestStartServer_NormalShutdown_NoErrorSent` — clean Shutdown produces no error on channel
- [x] All existing health/route tests pass
- [x] `go test -race ./cmd/server/...` passes
- [x] `golangci-lint run ./cmd/server/...` clean

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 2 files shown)


amitmorade (approved):
LGTM

PR #43, opened 2026-05-13
SOL-149749: buildURL errors on unfilled path parameter placeholders

solace-aross:
## Summary
- `buildURL` previously substituted `{placeholder}` tokens from args but never checked whether any remained unfilled — a missing required path param would silently produce a wrong URL
- `buildURL` now returns `(string, error)`; after substitution it scans the path for any remaining `{...}` tokens and returns a descriptive error naming them
- `Execute` propagates the error before making any HTTP call
- Also fixed pre-existing `SA5011` staticcheck violations in `operation_test.go`

## Test plan
- [x] `TestClient_Execute_MissingPathParam_ReturnsError` — omitting `queueName` returns error mentioning `{queueName}`; server is never called
- [x] `TestClient_Execute_AllPathParamsProvided_NoError` — all params present succeeds as before
- [x] All existing path/query/body param tests pass
- [x] `go test -race ./internal/semp/sempv2/...` passes
- [x] `golangci-lint run ./internal/semp/sempv2/...` clean


bczoma (approved):
LGTM

PR #44, opened 2026-05-13
SOL-149750: Add pagination page-count cap to prevent infinite loops

solace-aross:
## Summary
- Adds a hard ceiling of 1000 pages (`maxPages`) to the `executePaginatedStep` pagination loop
- Guards against brokers that return a persistent `nextPageUri` indefinitely (broker bug or adversarial input)
- Tracks a `pageLimitHit` flag to emit a distinct `truncatedMessage` distinguishing cap-hit from normal item truncation
- Emits a `slog.Warn` when the page cap fires for operational visibility

## Why
The pagination loop previously had no page-count limit. A broker returning a perpetual `nextPageUri` would loop until it hit the item cap (`capMax=500`), but edge cases (e.g., extremely large pages bypassing item accumulation, or broker bugs returning empty pages with cursors in future schema changes) could cause unbounded iteration. The page cap is an explicit defense-in-depth safety net.

## Changes
- `internal/composite/executor.go`: Changed `maxPages` from `const` to `var` (same default of 1000) to allow test override; added `pageLimitHit` tracking; added page cap check at top of pagination loop; added `pageLimitHit` branch to truncation message switch
- `internal/composite/executor_test.go`: Added `TestExecute_ListQueues_PageCapTerminatesLoop` — temporarily sets `maxPages=3`, uses a mock that always returns 1 item with a cursor, and verifies the loop terminates after exactly 3 calls with the correct truncation message

## Test plan
- [x] `go test -race ./internal/composite/...` — all tests pass
- [x] `golangci-lint run ./internal/composite/...` — 0 issues
- [x] New test `TestExecute_ListQueues_PageCapTerminatesLoop` verifies page cap fires at `maxPages` iterations
- [x] Existing truncation tests (`TestExecute_ListQueues_TruncatesAtMaxResults`, `TestExecute_ListQueues_MaxResultsCappedAt500`) still pass unmodified — their message strings are unchanged

### DIFF
--- a/internal/composite/executor.go
+++ b/internal/composite/executor.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"log/slog"
 	"net/url"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


amitmorade (approved):
LGTM

PR #50, opened 2026-05-19
fix: require dev_token when development_mode is true [SOL-149921]

solace-aross:
## What was wrong

`internal/auth/middleware.go:34-37` short-circuited the auth chain and returned the unprotected next handler whenever `cfg.DevelopmentMode == true && cfg.ClientAuth.DevToken == ""`. The config validator (`internal/config/config.go:434-448`) documented this state as intentional via a comment ("When DevToken is empty, all requests pass through"). A single blank YAML field — `dev_token: ""`, or no `dev_token` line at all — silently removed client authentication, with only a WARN log to flag it. Exploitable auth bypass triggered by a one-line config omission.

## What the fix does

- Extends `validate()` to require `cfg.ClientAuth.DevToken != ""` whenever `cfg.DevelopmentMode == true`. The error names the offending field and the trigger condition.
- Removes the now-unreachable bypass branch in `NewAuthMiddleware` and the matching dead branch in `NewProtectedResourceMetadataHandler`. The downstream `Issuer == ""` check at `middleware.go:150` covers any state that bypasses validation.
- Updates 27 existing test fixtures that used `development_mode: true` without a `dev_token` so they continue to exercise their intended assertions under the stricter validator.
- Removes `Test_AuthDisabled` — it directly asserted the bypass behaviour we removed. Keeping it would defend a state that can no longer exist (skill guidance: don't keep dead defensive code).

## Tests

- Added `TestLoadConfig_DevModeRequiresDevToken` — feeds `development_mode: true` with no `client_auth` block, asserts the validation error mentions both `dev_token` and `development_mode`.
- Existing fixture updates: mechanical addition of `client_auth.dev_token: test` to YAML strings that used dev mode without a token. The bulk-update was done by a script that explicitly skipped the new failure-case test.
- Removed `Test_AuthDisabled` (was asserting `200 OK` for every request when `dev_token` is empty — the exact behaviour the fix removes).
- Revert-verify confirmed: stash the `config.go` change → new test fails; restore → it passes. The `middleware.go` changes are dead-code removal post-fix, so they don't have their own coupled test.

Deployments with `development_mode: true` and a missing or empty `dev_token` will now fail at startup with:

```
validating config: client_auth.dev_token is required when development_mode is true (set a non-empty token to enable static auth)
```

This is the intended outcome of the fix. Same severity / migration flavor as SOL-149927. Worth a CHANGELOG line for the EA.

## Choice of fix

The ticket sketched two options:
- **A.** Require `dev_token` non-empty in dev mode.
- **B.** Add an `allow_anonymous: true` opt-in flag for intentional no-auth dev mode.

Going with A — smallest possible diff for the security fix. If anyone genuinely needs anonymous-auth dev mode, B can be a separate feature ticket.

## Jira

[SOL-149921](https://sol-jira.atlassian.net/browse/SOL-149921) — part of the May 2026 robustness/security sweep tracked under epic [SOL-149480](https://sol-jira.atlassian.net/browse/SOL-149480).

[SOL-149921]: https://sol-jira.atlassian.net/browse/SOL-149921?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/auth/middleware.go
+++ b/internal/auth/middleware.go
@@ -31,11 +31,6 @@ import (
 )
 
 func NewAuthMiddleware(cfg *config.ServerConfig, next http.Handler) (http.Handler, error) {
-	if cfg.DevelopmentMode && cfg.ClientAuth.DevToken == "" {
-		slog.Warn("authentication disabled — development mode with no dev token — not for production use")
-		return next, nil
-	}
-
 	verifier, err := NewTokenVerifier(cfg)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create token verifier: %w", err)
@@ -144,11 +139,6 @@ func createOIDCTokenVerifier(cfg *config.ServerConfig) (sdkauth.TokenVerifier, e
 // and initiate browser-based OAuth flows (Authorization Code + PKCE).
 // Returns nil when there's no OAuth configuration to advertise.
 func NewProtectedResourceMetadataHandler(cfg *config.ServerConfig) http.Handler {
-	// Only provide metadata endpoint when JWT validation is active
-	if cfg.DevelopmentMode && cfg.ClientAuth.DevToken == "" {
-		return nil
-	}
-
 	// Skip metadata endpoint when there's no issuer configured.
 	// The endpoint's purpose is to advertise the OAuth authorization server,
 	// so serving it with an empty issuer would be misleading to clients.
--- a/internal/auth/middleware_test.go
+++ b/internal/auth/middleware_test.go
@@ -37,61 +37,6 @@ var dummyHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request)
 	w.Write([]byte("OK"))
 })
 
-// Test_AuthDisabled tests that all requests pass through when development_mode=true and dev_token=""
-func Test_AuthDisabled(t *testing.T) {
-	cfg := &config.ServerConfig{
-		Port:            9090,
-		DevelopmentMode: true,
-		ClientAuth: config.ClientAuthConfig{
-			DevToken: "", // Empty dev token means auth disabled
-		},
-	}
-
-	middleware, err := NewAuthMiddleware(cfg, dummyHandler)
-	if err != nil {
-		t.Fatalf("failed to create middleware: %v", err)
-	}
-
-	tests := []struct {
-		name         string
-		authHeader   string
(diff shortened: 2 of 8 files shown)


solace-wmaryam (commented):
Thanks for the thorough writeup and the clean diff — the security instinct here is right, but this removes intentional behavior.

## The design intent

The server has three deliberate auth modes:

| Mode | Config | Behavior |
|------|--------|----------|
| Mode 1 | `development_mode: true`, no `dev_token` | No auth — all requests pass through |
| Mode 2 | `development_mode: true`, `dev_token: "secret"` | Static token auth |
| Mode 3 | `development_mode: false` | OIDC/JWT (production) |

Mode 1 is intentional — it lets developers run the server locally without any auth ceremony. The WARN log is the signal that you're in this state.

## The concern with this PR

Making Mode 1 a startup error removes a valid use case. Developers who want a zero-friction local setup (no token, no OIDC) will now be blocked.

## Suggested path forward

Your PR description already names the right fix: **Option B** — an explicit `allow_anonymous: true` opt-in flag. That keeps Mode 1 available while making it an unambiguous choice rather than a silent default. The config validator can then enforce: `development_mode: true` requires either a non-empty `dev_token` **or** `allow_anonymous: true` — no more ambiguity about whether an empty token is deliberate.

That change also gives operators a clear error message if they genuinely forgot to set a token, which is the real risk you're protecting against.

solace-wmaryam:
If I understand correctly, Mode 1 (no authentication, i.e. dev mode set to true but no dev token configured) is intentional 🤔 The WARN log is the signal that you're in this state.

Did we want to consider **Option B** — an explicit `allow_anonymous: true` opt-in flag? That keeps Mode 1 available while making it an unambiguous choice rather than a silent default. The config validator can then enforce: `development_mode: true` requires either a non-empty `dev_token` **or** `allow_anonymous: true` — no more ambiguity about whether an empty token is deliberate.

That change also gives operators a clear error message if they genuinely forgot to set a token, which is the real risk you're protecting against.

solace-aross:
I'm going to close this PR, and Jira ticket, and implement the new design in a fresh ticket and PR. Thank you all so much for the great advice and feedback!

PR #51, opened 2026-05-19
fix: cap semp.max_concurrent_per_broker at 1024 [SOL-149924]

solace-aross:
## What was wrong

`internal/config/config.go:405-407` only rejected negative values for `semp.max_concurrent_per_broker`:

```go
if cfg.SEMP.MaxConcurrentPerBroker < 0 {
    errs = append(errs, fmt.Errorf("semp.max_concurrent_per_broker must be > 0, got %d", ...))
}
```

The value drives the per-broker semaphore **and** the HTTP transport's `MaxIdleConnsPerHost` plus `MaxIdleConns` (×2). Each broker gets its own transport, so the total allocation scales linearly with the broker count too. An operator setting `max_concurrent_per_broker: 1000000` (intentional misconfig, typo, or compromised config) would over-allocate fixed-size structures and OOM the process at startup or first broker access.

## What the fix does

- Adds `MaxConcurrentPerBrokerCeiling = 1024` in `internal/defaults` with a comment explaining the choice. The documented typical usage is 1–10 (per `DefaultMaxConcurrentPerBroker`); 1024 gives ~100× headroom for legitimate high-throughput scenarios while rejecting pathological values.
- Replaces the one-sided check in `validate()` with a single bounded check (`< 1 || > ceiling`). The error message names both bounds via the constant so a future change to the ceiling automatically updates the message.

## Tests

- Added `TestLoadConfig_MaxConcurrentPerBroker_OutOfRange` — table test with two sub-cases:
  - `negative` — `max_concurrent_per_broker: -1`, asserts the error names the field. (This subcase passes both before and after the fix; it's regression protection for the existing negative-rejection behavior.)
  - `above_ceiling` — `max_concurrent_per_broker: 1000000`, asserts the error names the field. (This subcase is red before the fix, green after.)
- Revert-verify confirmed: stash `config.go` → upper-bound subcase fails; restore → it passes.
- Full suite stays green: `go build ./... && go vet ./...` clean; `go test ./...` passes all 12 packages.

Configs setting `semp.max_concurrent_per_broker > 1024` will now fail at startup with:

```
validating config: semp.max_concurrent_per_broker must be between 1 and 1024, got <value>
```

This is the intended fix. Real-world configurations should comfortably stay below 1024 given the documented typical range. Same EA-release migration-note flavor as SOL-149927 / SOL-149921.

## Edge cases verified

- `max_concurrent_per_broker: 0` (or field omitted): `applyDefaults` rewrites to 10 (existing behavior, untouched). Passes validation.
- `1`, `1024` (inclusive boundary): pass.
- `0` (after applyDefaults: 10), `-1`, `1025`, `1000000`: fail with the new bounded error.

## Jira

[SOL-149924](https://sol-jira.atlassian.net/browse/SOL-149924) — May 2026 robustness/security sweep, epic [SOL-149480](https://sol-jira.atlassian.net/browse/SOL-149480).

[SOL-149924]: https://sol-jira.atlassian.net/browse/SOL-149924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -402,8 +402,8 @@ func validate(cfg *ServerConfig) error {
 		errs = append(errs, fmt.Errorf("log_level %q is invalid (must be one of %v)", cfg.LogLevel, validLogLevels))
 	}
 
-	if cfg.SEMP.MaxConcurrentPerBroker < 0 {
-		errs = append(errs, fmt.Errorf("semp.max_concurrent_per_broker must be > 0, got %d", cfg.SEMP.MaxConcurrentPerBroker))
+	if n := cfg.SEMP.MaxConcurrentPerBroker; n < 1 || n > defaults.MaxConcurrentPerBrokerCeiling {
+		errs = append(errs, fmt.Errorf("semp.max_concurrent_per_broker must be between 1 and %d, got %d", defaults.MaxConcurrentPerBrokerCeiling, n))
 	}
 
 	if cfg.SEMP.RequestTimeoutDuration <= 0 {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -769,6 +769,11 @@ func TestLoadConfig_MaxConcurrentPerBroker_OutOfRange(t *testing.T) {
 			value:       "1000000",
 			wantInError: "semp.max_concurrent_per_broker",
 		},
+		{
+			name:        "just above ceiling",
+			value:       "1025",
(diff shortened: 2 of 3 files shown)


solace-lootii (approved):
LGTM!

PR #52, opened 2026-05-19
fix(config): warn on insecure_skip_verify=true in production mode [SOL-149928]

solace-aross:
## What was wrong

`internal/config/config.go` `validateBroker()` takes a `productionMode` flag and uses it to reject `http://` broker URLs in production. It did **not** surface the symmetric risk for `BrokerConfig.InsecureSkipVerify`. A production deployment with `insecure_skip_verify: true` silently accepted any TLS certificate the broker presented — including an attacker-issued one on a hijacked path. The two SEMP clients log a one-time WARN at first broker access (`internal/semp/sempv1/client.go`, `internal/semp/sempv2/client.go`), which is easy to miss in operational triage.

## What the fix does

Emits a startup `slog.Warn` from `validate()` whenever `insecure_skip_verify: true` is paired with `development_mode: false`:

```go
if !cfg.DevelopmentMode && broker.InsecureSkipVerify {
    slog.Warn("INSECURE: insecure_skip_verify=true in production mode; broker TLS certificate will not be validated",
        slog.String("broker", alias))
}
```

Per review feedback (bczoma), this is a WARN rather than a hard rejection — parity with other tooling (e.g. Terraform providers) and keeps test/dev paths simple. The combination is dev-only by intent (see `DefaultInsecureSkipVerify` comment in `internal/defaults/defaults.go`), but operators occasionally need it in prod against a self-signed staging broker; the WARN puts that decision in triage logs at startup, not just on first SEMP call.

## Tests

- `TestLoadConfig_WarnsOnInsecureSkipVerifyInProductionMode` — production-mode + `insecure_skip_verify: true` loads successfully and emits a WARN naming the field.
- `TestLoadConfig_NoWarnOnInsecureSkipVerifyInDevelopmentMode` — development-mode + `insecure_skip_verify: true` loads with no WARN.
- Full suite stays green: `go build ./... && go vet ./...` clean; `go test ./...` passes all packages.

Production-mode deployments (`development_mode: false`) with any broker having `insecure_skip_verify: true` will now emit a startup WARN like:

```
level=WARN msg="INSECURE: insecure_skip_verify=true in production mode; broker TLS certificate will not be validated" broker=<alias>
```

Startup is **not** blocked. Same EA-migration-note flavor as SOL-149927 / SOL-149921 / SOL-149924.

## Jira

[SOL-149928](https://sol-jira.atlassian.net/browse/SOL-149928) — May 2026 robustness/security sweep, epic [SOL-149480](https://sol-jira.atlassian.net/browse/SOL-149480).

[SOL-149928]: https://sol-jira.atlassian.net/browse/SOL-149928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bczoma:
I see the concern but not sure I agree with the tight control. For one, it would make testing very hard.
The same issue exists for example with many Terraform providers, where the option is there without restriction, to set this same flag or a variation. It is generally expected a policy check tool would flag this when scanning configs.
A middle ground remediation could be to log it at the WARN level if set or have a security warning for each user access which would be visible at observability.

solace-aross:
Good points, thanks. Reworked per your middle-ground suggestion in 0bf6e68:

- Removed the hard rejection in `validateBroker`.
- `validate()` now emits a one-shot `slog.Warn` per offending broker at config load when `development_mode: false` and `insecure_skip_verify: true`. Operators see it in startup/triage logs without scraping the existing per-broker SEMP-client warns (which still fire on first access for request-time visibility).
- Test for production mode now asserts the WARN fires and config loads cleanly. Dev-mode test asserts the warn stays silent (it's the expected default there).

Title still says "reject" — happy to retitle to something like `warn on insecure_skip_verify=true in production mode` before merge if you'd prefer.

bczoma:
Agree with the coding, the only suggestion is to align the log messages.

*More details*

Current logs:
```
config.go:597     "INSECURE: insecure_skip_verify=true in production mode; broker TLS certificate will not be validated"  broker=<alias>
...
sempv1/client.go:69    "INSECURE: TLS verification disabled for broker"                                                    url=<URL>
...
sempv2/client.go:93    "INSECURE: TLS verification disabled for broker"                                                    url=<URL>
```

The actual problem — two schema mismatches

1. Different message strings. Operators grepping "insecure_skip_verify=true in production mode" see only the startup hit; operators grepping "TLS verification disabled for broker" see only the first-access hits. The common token is "INSECURE" — too broad to be useful as a filter. A log-aggregation rule for "alert me if any broker has TLS verification off" has to OR over two distinct message strings.

2. Different identifier keys for the same broker. Startup WARN uses broker=<alias>, e.g. broker=prod-us. Per-access WARNs use url=<URL>, e.g. url=https://broker-prod-us.internal.example.com:943. Joining "all WARNs for the same broker" requires the operator to mentally map alias → URL because no structured field is common to both. In a log pipeline (Splunk/Datadog/Loki), this means you can't group by broker without a parser that knows the mapping.

Add the schema asymmetry the security agent flagged: url=<URL> leaks internal hostnames/ports (topology disclosure), while broker=<alias> doesn't.

Recommended simple fix:

Align the schema (≈10 lines). Same message string in all three sites + drop `url=` in favor of `broker=alias` in the SEMP clients. Pass DisplayName() into the SEMP clients (or just BrokerConfig.DisplayName()).

solace-aross:
Aligned in aab5b79.

- Same message in all three sites: `INSECURE: TLS verification disabled for broker`
- All three identify the broker via `broker=<DisplayName()>`; SEMP clients drop `url=<URL>`, which also closes the topology-disclosure asymmetry
- Updated config tests to assert the new message and the `broker=<alias>` key

`go build ./... && go vet ./...` clean; `go test ./...` green.

bczoma (approved):
Looks good!

PR #53, opened 2026-05-19
fix(server): set ReadTimeout and IdleTimeout on the MCP HTTP server [SOL-149914]

solace-aross:
## What was wrong

The MCP server's HTTP listener (`cmd/server/main.go:331-335`) was configured with only `ReadHeaderTimeout`. Two attack/exhaustion vectors remained open:

- **Slow body**: a client can send headers within 10 s, then trickle the request body forever. Goroutine + connection held indefinitely.
- **Idle keep-alive exhaustion**: after a response completes, an idle HTTP/1.1 keep-alive connection sits open until the kernel reaps the fd. A modest attacker can open thousands of idle sockets, send one request each, and exhaust the server's fd budget.

## What the fix does

- Adds `DefaultReadTimeoutSeconds = 30` and `DefaultIdleTimeoutSeconds = 120` to `internal/defaults/defaults.go` with documented-decision comments (assumption / reasoning / trade-off, matching the existing style).
- Extracts `newHTTPServer(addr, handler)` in `cmd/server/main.go` so the production timeout posture is in one testable place. `main()` now calls it instead of building `&http.Server{...}` inline.
- Wires up all three timeouts: `ReadHeaderTimeout`, `ReadTimeout`, `IdleTimeout`. **`WriteTimeout` is deliberately not set** — the `/mcp` endpoint serves long-lived MCP streamable HTTP / SSE responses, and a server-wide write deadline would cut legitimate streams. The trade-off is captured in the constant's doc comment.

## Tests

- Added `TestNewHTTPServer_TimeoutsAreSet`: locks in the posture (`ReadHeaderTimeout=10s`, `ReadTimeout=30s`, `IdleTimeout=120s`, `WriteTimeout=0`). If a future change removes a timeout or sets `WriteTimeout`, this fails loudly.
- Revert-verify confirmed: stash `cmd/server/main.go` → test build-fails (proves the test couples to the new function); restore → it passes.

- A client whose request body takes longer than 30 s will be dropped. MCP request bodies are JSON-RPC payloads of a few KB; this is generous in practice.
- An idle keep-alive connection is closed after 120 s. MCP clients that reuse connections within that window are unaffected; clients past 120 s will reconnect.

Same EA-CHANGELOG flavor as the other tickets in this sweep.

## Why no WriteTimeout

`http.Server.WriteTimeout` bounds the entire response duration (from when reading the request body finishes to when the response write completes). For SSE / MCP streamable HTTP, the response is long-lived by design — a 1 minute or 5 minute deadline would cut active sessions. Mitigations for the slow-read attack on streaming endpoints would need to be per-handler via `http.ResponseController` rather than a blanket server setting; out of scope for this ticket.

## Jira

[SOL-149914](https://sol-jira.atlassian.net/browse/SOL-149914) — May 2026 robustness/security sweep, epic [SOL-149480](https://sol-jira.atlassian.net/browse/SOL-149480).

[SOL-149914]: https://sol-jira.atlassian.net/browse/SOL-149914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -134,6 +134,22 @@ func healthConfigFromFile() healthConfig {
 	return healthConfig{Port: cfg.Port, Scheme: scheme}
 }
 
+// newHTTPServer builds the MCP server's *http.Server with the production
+// timeout posture from internal/defaults. Extracted so timeout fields can be
+// asserted in a unit test without spinning up a real listener. See the
+// constants in internal/defaults for the rationale on each timeout value;
+// WriteTimeout is deliberately zero to preserve long-lived MCP streamable
+// HTTP / SSE responses.
+func newHTTPServer(addr string, handler http.Handler) *http.Server {
+	return &http.Server{
+		Addr:              addr,
+		Handler:           handler,
+		ReadHeaderTimeout: time.Duration(defaults.DefaultReadHeaderTimeoutSeconds) * time.Second,
+		ReadTimeout:       time.Duration(defaults.DefaultReadTimeoutSeconds) * time.Second,
+		IdleTimeout:       time.Duration(defaults.DefaultIdleTimeoutSeconds) * time.Second,
+	}
+}
+
 // startServer starts httpServer in a background goroutine and returns a channel
 // that receives any startup/runtime error (excluding http.ErrServerClosed, which
 // is the normal result of Shutdown). The channel is buffered so the goroutine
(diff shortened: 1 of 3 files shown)


solace-lootii (approved):
LGTM overall, but just something I noticed in the `DefaultIdleTimeoutSeconds` comments.. it credits `IdleTimeout` with mitigating slow-read attacks, but `IdleTimeout `only closes connections between requests - a slow reader is active, not idle so only the TCP send buffer really applies there.

PR #54, opened 2026-05-19
fix(semp): cap broker response body at 16 MiB [SOL-149920]

solace-aross:
## What was wrong

`internal/semp/sempv1/client.go:137` and `internal/semp/sempv2/client.go:145` both read SEMP response bodies with `io.ReadAll(resp.Body)` and no size cap. The `http.Client.Timeout` (60 s default) bounds *duration* but does not bound *bytes* — a peer streaming 100 MB/s for 60 s delivers ~6 GB into the MCP server's heap. A single oversized response is enough to OOM-kill the entire MCP process, taking down all tools across all configured brokers.

The SEMPv2 pagination contract (500 items/page) keeps real responses comfortably small, but a compromised broker, a man-in-the-middle, or a misbehaving build of the broker can sidestep that.

## What the fix does

- Adds `MaxSEMPResponseBytes = 16 * 1024 * 1024` (16 MiB) to `internal/defaults/defaults.go`. Sized to comfortably exceed any realistic SEMPv2 page response (~15 MB at the extreme of 30 KB per item × 500 items) while bounding worst-case heap usage. Documented-decision comment with assumption / reasoning / trade-off, matching the existing style.
- Adds `resilience.ReadCappedBody(r, limit)` plus typed `ErrResponseTooLarge` in a new `internal/semp/resilience/io.go`. The helper reads `limit+1` bytes via `io.LimitReader` and rejects over-cap reads with the typed error so callers can branch via `errors.Is`.
- Swaps the `io.ReadAll(resp.Body)` calls in both SEMP clients for the helper.

## Tests

- `internal/semp/resilience/io_test.go` covers the helper directly: under-cap returns body, at-cap (inclusive) returns body, over-cap returns `ErrResponseTooLarge` and a nil body, well-over-cap (1 MiB source vs 1 KiB cap) is rejected, EOF-is-not-an-error, and underlying read errors are wrapped and not classified as `ErrResponseTooLarge`.
- `internal/semp/sempv1/client_test.go` adds `TestExecute_OversizedResponseBody_ReturnsTypedError`: spins up an `httptest.Server` that streams `> MaxSEMPResponseBytes` bytes (64 KiB chunks) and asserts `errors.Is(err, resilience.ErrResponseTooLarge)` surfaces through the full `Execute` pipeline.
- `internal/semp/sempv2/client_test.go` adds the parallel test.
- Revert-verify confirmed: stash both `client.go` files → both integration tests fail (sempv1 returns `"err = sempv1: unknown (status=200)"` because the old code happily read the oversized body); restore → both pass.

A broker that ever returns more than 16 MiB in a single response will now produce an error wrapping `resilience.ErrResponseTooLarge` instead of silently buffering the entire payload. The SEMP pagination contract makes this implausible in practice. Same EA-CHANGELOG flavor as the rest of this sweep.

## Jira

[SOL-149920](https://sol-jira.atlassian.net/browse/SOL-149920) — May 2026 robustness/security sweep, epic [SOL-149480](https://sol-jira.atlassian.net/browse/SOL-149480).

[SOL-149920]: https://sol-jira.atlassian.net/browse/SOL-149920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/defaults/defaults.go
+++ b/internal/defaults/defaults.go
@@ -29,6 +29,22 @@ const DefaultShutdownTimeoutSeconds = 30
 // Terraform provider convention.
 const DefaultSEMPRequestTimeoutDuration = time.Minute
 
+// MaxSEMPResponseBytes caps the in-memory buffering of a broker's SEMP
+// response body. Without this cap, a misbehaving broker or a man-in-the-
+// middle could stream gigabytes of data within the request timeout and OOM
+// the MCP server.
+//
+// Decided: 16 MiB.
+// Reasoning: SEMPv2 list responses are bounded by the broker's page size
+// (typically 500 items). At an extreme of ~30 KB per item, a single page
+// sits around 15 MB — 16 MiB gives a safety margin above any realistic
+// page response while bounding worst-case allocation to a few percent of
+// process memory.
+// Trade-off: a broker that legitimately returned > 16 MiB in a single
+// response would now fail with a typed "response too large" error. The
+// SEMP pagination contract makes this implausible in practice.
+const MaxSEMPResponseBytes = 16 * 1024 * 1024
+
 // DefaultMaxConcurrentPerBroker is the maximum number of concurrent SEMP
 // requests allowed per broker, enforced via a per-broker semaphore.
 //
--- a/internal/semp/resilience/io.go
+++ b/internal/semp/resilience/io.go
@@ -1,0 +1,44 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resilience
+
+import (
+	"errors"
+	"fmt"
+	"io"
+)
+
+// ErrResponseTooLarge is returned by ReadCappedBody when the source reader
+// yields more than the configured cap. Callers branch on this with errors.Is
+// to distinguish "broker misbehaved / MITM" from generic I/O errors.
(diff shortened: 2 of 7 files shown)


solace-aross:
## FOSSA Licensing block — looks like a false positive

The failing required check is `FOSSA SCA Licensing` flagging `gopkg.in/yaml.v3@v3.0.1` as `unlicensed_dependency` (`License: None`). Two facts argue against this being a real new issue introduced by this PR:

1. **This PR doesn't touch dependencies.** Files changed are all under `internal/...` plus `defaults.go`. No `go.mod`/`go.sum` edits — the helper uses stdlib `io.LimitReader`.
2. **`gopkg.in/yaml.v3 v3.0.1` is already on `main`** (direct dep in `go.mod`, hash in `go.sum`). The PR branch and main have an identical dep graph for `yaml.v3`.

Yet FOSSA reports `1 new, 1 total (0 in base)` against main SHA `8e0438a1…` — same dep, two verdicts. Likely causes (all FOSSA-side):

- License-DB drift: FOSSA's metadata for `yaml.v3@v3.0.1` may have flipped to "None" between the cached base scan (ID `106231516`) and the live PR scan (ID `106554610`). Upstream `yaml.v3 v3.0.1` is dual-licensed MIT + Apache-2.0, so `None` is a detection failure, not a real classification.
- Stale base scan that under-resolved the dep graph at the time main was scanned.
- Transient resolver failure when FOSSA tried to fetch the license file on the PR-side scan.

### What I've done

- Re-ran the failed `FOSSA Scan / SCA Scan` job (run `26112363333`) on the off chance it's a transient detection failure.

### If the re-run still fails

If yaml.v3 still shows as `License: None` on the FOSSA dashboard, this is a project-wide misclassification affecting `main` too (just not surfaced there because main isn't diff-gated against itself). The fix is a FOSSA license override on `gopkg.in/yaml.v3@v3.0.1` by the FOSSA project owner — not something this PR can address.

Other checks (build, lint, e2e, e2e-oauth, FOSSA SCA Vulnerabilities) are all green.

solace-lootii (approved):
LGTM! Clean defensive fix. Single chokepoint via ReadCappedBody, typed error surfaces cleanly through errors.Is, and the integration tests exercise the real Execute pipeline end-to-end

PR #55, opened 2026-05-19
fix(semp): wrap cookie jar in atomic SafeCookieJar to remove 401 race [SOL-149922]

solace-aross:
## What was wrong

`internal/semp/resilience/retry.go:88-91` cleared session cookies on a 401 by allocating a fresh `*cookiejar.Jar` and assigning it directly to `httpClient.Jar`:

```go
jar, _ := cookiejar.New(nil)
d.httpClient.Jar = jar
```

This races concurrent `http.Client.Do` reads of the `Jar` field. The interface value is a two-word (type pointer + data pointer); Go's memory model makes torn reads undefined. The Sender's own doc comment characterised the race as "benign", but that was speculative — `go test -race` flags it, and even on architectures where torn reads do not panic, callers may observe inconsistent jar state across a swap.

## What the fix does

Adds `resilience.SafeCookieJar` (new file `cookiejar.go`): an `http.CookieJar` wrapper around `*cookiejar.Jar` whose inner pointer is stored in an `atomic.Pointer`. `http.Client.Jar` is pinned to a stable `*SafeCookieJar` value for the lifetime of the client, so concurrent `http.Client.Do` reads always see the same interface value. The inner jar is swapped atomically when `Clear()` runs — concurrent `SetCookies`/`Cookies` calls load the pointer via `atomic.Load` and observe a consistent jar.

Plumbs the jar through `resilience.New`: callers (the SEMPv1 and SEMPv2 client constructors) build the `*SafeCookieJar` once, set it on `http.Client.Jar`, and pass it to `resilience.New(...)` so the `Sender` can call `Clear()` during the 401 re-auth path.

`retry.go` now calls `d.cookieJar.Clear()` instead of swapping the `Jar` field directly. The (unreachable) error path is logged with a WARN and short-circuits the retry rather than silently dropping it.

## Tests

- `internal/semp/resilience/cookiejar_test.go` (new):
  - `SetAndGet` / `Clear_EmptiesTheJar` — basic semantics.
  - `ConcurrentSetClear` — hammers SetCookies, Cookies, and Clear from many goroutines. Designed to fail under `go test -race` if the wrapper ever permits a torn read.
  - `ImplementsHTTPCookieJar` — runtime assertion that pairs with the compile-time `var _ http.CookieJar = (*SafeCookieJar)(nil)` interface assertion in `cookiejar.go`.
- Updated `sender_test.go` helper to construct a `*SafeCookieJar`, set it on `httpClient.Jar`, and pass it to `resilience.New`. Inline test sites that constructed bare clients are updated for the new signature.
- Full suite passes under `go test -race ./...`. The existing `TestSender_Retry_401_BasicAuth_ClearsCookiesAndRetries` still exercises the end-to-end 401 retry behaviour and still passes; the implementation of "clearing" is now atomic.

## Behavior change

None operator-visible. The cookie-clearing semantic is unchanged; only the underlying mechanism is now race-free.

## Jira

[SOL-149922](https://sol-jira.atlassian.net/browse/SOL-149922) — May 2026 robustness/security sweep, epic [SOL-149480](https://sol-jira.atlassian.net/browse/SOL-149480).

[SOL-149922]: https://sol-jira.atlassian.net/browse/SOL-149922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/semp/resilience/cookiejar.go
+++ b/internal/semp/resilience/cookiejar.go
@@ -1,0 +1,90 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resilience
+
+import (
+	"net/http"
+	"net/http/cookiejar"
+	"net/url"
+	"sync/atomic"
+)
+
+// SafeCookieJar is a thread-safe wrapper around *cookiejar.Jar that supports
+// concurrent SetCookies / Cookies calls plus an atomic Clear() that swaps the
+// inner jar.
+//
+// The standard library jar is itself thread-safe for read/write operations
+// but exposes no way to clear stored cookies. The 401 re-auth path in the
+// Sender wants to drop session state and force fresh credentials on the
+// retry. Replacing the entire jar via direct assignment to
+// http.Client.Jar — the previous implementation — races with concurrent
+// http.Client.Do reads of the interface value (two-word: type pointer + data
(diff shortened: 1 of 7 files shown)


bczoma (approved):
Looks good. I was looking into this myself as well in the past, from a different angle. These changes make sense.

PR #56, opened 2026-05-19
fix(config): redact userinfo in broker-URL validation errors [SOL-149923]

solace-aross:
## What was wrong

`internal/config/config.go:539, 542` formatted the raw broker URL into validation errors via `%q`. A misconfigured broker URL like `https://admin:hunter2@broker.example.com:1943` wrote the password into the error message. That error propagated through `errors.Join` to `cmd/server/main.go:231` and was logged via `slog.Error("failed to load config", slog.String("error", err.Error()))`. The slog handler's `ReplaceAttr` redactor matches on attribute *key* (`"error"`), not on value content, so the credential passed through untouched and landed in stderr → journald → log aggregators → SIEM, with retention windows that typically outlive operator awareness.

## What the fix does

- `validateBrokerURL` rejects URLs with non-empty `u.User` with a clear error directing operators to the `auth` block. Credentials belong in `AuthConfig`, never inline in the URL. The error itself names the broker alias and the right fix location, with the URL routed through `sanitizeURL` so the userinfo never reaches the message.
- The two existing `%q` formatters (http-scheme-in-production, missing-host) now also route the URL through the new `sanitizeURL(u *url.URL)` helper. Defense in depth: even if the userinfo-rejection path is ever changed or bypassed, the user:pass component never reaches the error message.

## Tests

- `TestLoadConfig_RejectsBrokerURLWithCredentials_AndRedactsThemInError`: feeds a broker URL with `admin:hunter2-super-secret-must-not-leak@…`, asserts (a) validation fails, (b) the password substring does **not** appear anywhere in `err.Error()`, and (c) the error mentions both `prod-us` (the broker alias) and `auth block` (the right place to put credentials).
- Revert-verify confirmed: stash `config.go` → the test fails on "expected error for broker URL with embedded credentials"; restore → it passes.

Configs that currently embed `user:pass@` in a broker URL will now fail at startup with:

```
validating config: broker "<alias>": url must not include credentials (use the auth block instead), got "<sanitized URL>"
```

Operators should move credentials into the broker's `auth` block. Same EA-CHANGELOG flavor as the rest of the sweep.

## Jira

[SOL-149923](https://sol-jira.atlassian.net/browse/SOL-149923) — May 2026 robustness/security sweep, epic [SOL-149480](https://sol-jira.atlassian.net/browse/SOL-149480).

[SOL-149923]: https://sol-jira.atlassian.net/browse/SOL-149923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -98,10 +98,13 @@ func (a AuthConfig) LogValue() slog.Value {
 
 // LogValue implements slog.LogValuer for BrokerConfig. It exposes connection
 // metadata (URL, TLS settings, auth method) but excludes credentials.
+// The URL is routed through sanitizeURLString so any userinfo is stripped
+// before reaching the log — defense in depth against logging a BrokerConfig
+// before validateBrokerURL has had a chance to reject credentialed URLs.
 // See docs/secure-logging-rules.md Rule 2.
 func (b BrokerConfig) LogValue() slog.Value {
 	return slog.GroupValue(
-		slog.String("url", b.URL),
+		slog.String("url", sanitizeURLString(b.URL)),
 		slog.Bool("insecure_skip_verify", b.InsecureSkipVerify),
 		slog.String("auth_mode", b.Auth.Mode),
 	)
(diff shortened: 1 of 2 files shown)


internal/config/config.go
solace-wmaryam, line 111: Note that we log the raw URL string here. Since this PR has now added a check to prevent credentials from appearing in the URL, this shouldn't technically be a problem, but if any future code path logs a `BrokerConfig` before validation has run, we could risk emitting credentials here. We can consider replacing the raw URL with the sanitized version instead.

solace-wmaryam: There are other places in the SEMP layer that log the URL but these can only be reached after config validation so should be okay!

solace-aross: Good catch, Wajiha — agreed that LogValue shouldn't lean on validation ordering. Pushed 08a94fb: BrokerConfig.LogValue, ClientAuthConfig.LogValue (Issuer and ResourceURL too, for parity), and a small sanitizeURLString helper that handles the unparseable-input case. Two new tests pin the LogValue output so a regression here would fail loudly.

solace-wmaryam (approved):

PR #57, opened 2026-05-19
SOL-149478: MCP minimum documentation to support release

solace-wmaryam:
## Summary

Restructures and expands the authentication documentation (docs/authentication.md) to make Mode 3 (OAuth/JWT) easier to follow for new users, with step-by-step setup instructions and Keycloak-specific examples.    

Fixes SOL-149478

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The existing Mode 3 documentation was structured in a non-linear way — IdP setup instructions appeared after server configuration, making it difficult for new users to follow. Keycloak-specific steps were also missing.

## Changes Made

- Restructured Mode 3 into a linear step-by-step flow: IdP setup → server config → start server → add to Claude Code                                     
- Made all instructions generic (any OIDC provider) with Keycloak-specific callouts inline at each step                                                

## Testing

**Test Environment:**
- Go version: 1.26.2
- OS: macOS
- Broker version (if applicable): N/A

## Breaking Changes

N/A - documentation only.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -128,7 +128,7 @@ Before starting, decide how your MCP client will register with your identity pro
 | IdP requirements | Standard OAuth client setup | Must support anonymous DCR (RFC 7591) |
 | Best for | Most setups; more control | Zero-config end-user experience |
 
-Both options use the same browser-based OAuth 2.1 Authorization Code flow with PKCE. The difference is only in how the MCP client obtains its client ID.
+Both options use the same browser-based OAuth 2.1 Authorization Code flow with PKCE. The difference is in how the MCP client obtains its client credentials (and whether it uses DCR) — Option A may also involve a client secret for confidential clients, while Option B never does.
 
 ### Step 1: Set up your identity provider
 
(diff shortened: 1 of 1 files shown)


solace-aross:
Hi Wajiha, read through this and the new structure is a real improvement over what was there. The linear flow lands well, the Keycloak callouts are placed where readers actually need them, and the DCR policy guidance (openid scope placeholder, trusted hosts) will save people real debugging time.

Two of the Copilot points worth addressing before merge:

1. Step 2 prose says to set `resource_url` to "the URL your MCP server is listening on." For production deployments behind a reverse proxy or ingress, that wording is misleading. The table immediately below has the right framing ("public URL of your MCP server endpoint"). Worth aligning the two.

2. "The difference is only in how the MCP client obtains its client ID" undersells the distinction. Option A can require a client secret; Option B never does. "client credentials" reads more accurately.

A few smaller things from my own pass:

- Step 1.3 asks the user to set a Client ID but doesn't suggest a value. Step 4 then uses `mcp-client` in the example. A short pointer (e.g. "you'll pass this as \`--client-id\` later") would close the loop for first-time readers.

- The Step 2 YAML uses `your-idp.example.com` and `your-mcp-server.example.com` while the rest of the flow uses `localhost`. The Keycloak callout right below corrects course, but the production-style example is what readers see first. Either reorder so the localhost variant is primary, or add a one-line note that the example shows production values.

- PR description points to `docs/AUTHENTICATION.md` (uppercase) but the file is `docs/authentication.md`. Quick fix to the description for case-sensitive filesystems.

None of this is blocking. Address the two prose items above and this is good to go.

bczoma (approved):
LGTM

PR #58, opened 2026-05-19
fix(semp): set granular timeouts on the SEMP HTTP transport [SOL-149925]

solace-aross:
## What was wrong

`internal/semp/resilience/transport.go:34-40` built the per-broker `*http.Transport` with `MaxIdleConnsPerHost`, `MaxIdleConns`, `IdleConnTimeout`, and `TLSClientConfig` — but **no** `TLSHandshakeTimeout`, **no** `ResponseHeaderTimeout`, **no** `ExpectContinueTimeout`. The outer `http.Client.Timeout` (60 s) bounded the entire request lifecycle, but a broker stuck in TLS handshake (stalled cert validation, HSM latency, partial network split) or one that accepted the TCP connection then never returned response headers would tie up a `MaxConcurrentPerBroker` semaphore slot for the full request-timeout window before failing. With concurrent SEMP calls targeting the same broker, the per-broker concurrency saturated within seconds and other operations queued or failed.

## What the fix does

- Adds three constants to `internal/defaults/defaults.go` with documented-decision comments (assumption / reasoning):
  - `DefaultTLSHandshakeTimeoutSeconds = 10`
  - `DefaultResponseHeaderTimeoutSeconds = 30` — must stay strictly less than the existing `DefaultSEMPRequestTimeoutDuration = 60 s` so the granular timeout actually fires before the outer client.Timeout. Pinned by a test.
  - `DefaultExpectContinueTimeoutSeconds = 1` — SEMP doesn't use 100-continue, but standard HTTP-client hygiene.
- Wires all three into `NewTunedTransport` alongside the existing pool-tuning fields.

## Tests

- New `internal/semp/resilience/transport_test.go`:
  - `TestNewTunedTransport_GranularTimeoutsAreSet` — asserts the returned transport has all three timeouts set to the corresponding constants.
  - `TestResponseHeaderTimeout_StrictlyLessThanRequestTimeout` — pins the contract the constant comment relies on. If a future change pushes `ResponseHeaderTimeout` past the request timeout, this fails loudly.
- Revert-verify implicit: the new test would fail to compile without the transport.go change (the fields would default to zero and the asserted values would not match).

## Behavior change

None operator-visible. These are tighter granular timeouts that fail faster on already-broken broker connections; no legitimate broker handshake or header response approaches the new ceilings.

## Jira

[SOL-149925](https://sol-jira.atlassian.net/browse/SOL-149925) — May 2026 robustness/security sweep, epic [SOL-149480](https://sol-jira.atlassian.net/browse/SOL-149480).

[SOL-149925]: https://sol-jira.atlassian.net/browse/SOL-149925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/defaults/defaults.go
+++ b/internal/defaults/defaults.go
@@ -41,19 +41,6 @@ const DefaultSEMPRequestTimeoutDuration = time.Minute
 // failure window to a small fraction of the request timeout.
 const DefaultTLSHandshakeTimeoutSeconds = 10
 
-// DefaultResponseHeaderTimeoutSeconds bounds the time the SEMP transport
-// waits after sending a request before the broker returns response headers.
-// Must stay strictly less than DefaultSEMPRequestTimeoutDuration so the
-// granular timeout actually fires before the outer client.Timeout. A broker
-// that accepts the TCP connection then never sends headers would otherwise
-// hold the per-broker semaphore slot for the full request timeout.
-//
-// Assumption: 30 seconds is comfortably under the 60s request timeout.
-// Reasoning: SEMP operations send response headers as soon as the broker
-// dispatches the request. Anything past 30s indicates a stuck broker, not
-// a slow one — fail fast.
-const DefaultResponseHeaderTimeoutSeconds = 30
-
 // DefaultExpectContinueTimeoutSeconds is the time the transport waits for a
 // "100 Continue" intermediate response after sending a request with an
 // "Expect: 100-continue" header before sending the body. SEMP requests do
--- a/internal/defaults/defaults.go
+++ b/internal/defaults/defaults.go
@@ -29,6 +29,39 @@ const DefaultShutdownTimeoutSeconds = 30
 // Terraform provider convention.
 const DefaultSEMPRequestTimeoutDuration = time.Minute
 
+// DefaultTLSHandshakeTimeoutSeconds bounds how long the SEMP transport will
+// wait for a TLS handshake to complete with a broker. Without this granular
+// timeout, a broker stuck in handshake (stalled certificate validation, HSM
+// latency, partial network split) ties up a MaxConcurrentPerBroker semaphore
+// slot for the full DefaultSEMPRequestTimeoutDuration window before failing.
+//
+// Assumption: 10 seconds is generous for a real TLS handshake.
+// Reasoning: Even under congested networks a TLS 1.2/1.3 handshake completes
+// in well under a second; 10s tolerates extreme outliers while bounding the
+// failure window to a small fraction of the request timeout.
+const DefaultTLSHandshakeTimeoutSeconds = 10
+
+// DefaultResponseHeaderTimeoutSeconds bounds the time the SEMP transport
+// waits after sending a request before the broker returns response headers.
+// Must stay strictly less than DefaultSEMPRequestTimeoutDuration so the
+// granular timeout actually fires before the outer client.Timeout. A broker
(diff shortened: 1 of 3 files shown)


bczoma:
These proposed additional timeouts are constants vs. user configurable.
Some of the constant values may not help achiving this fix' purpose if the user changes certain user configurable ones.

AI says:
  internal/semp/resilience/transport_test.go:62-65 — TestResponseHeaderTimeout_StrictlyLessThanRequestTimeout pins DefaultResponseHeaderTimeoutSeconds < DefaultSEMPRequestTimeoutDuration, but    
  transport.go:40 hard-codes the 30 s ResponseHeaderTimeout while sempv1/client.go:76 and sempv2/client.go:99 set http.Client.Timeout = sempCfg.RequestTimeoutDuration (per-broker,                
  operator-configurable).                                                                                                                                                                          
                                                                                            
  If an operator sets semp.request_timeout_duration: 10s in broker-config.yaml, the granular 30 s timeout never fires — the outer 10 s client timeout wins, and the regression the PR claims to fix
   (stuck-headers broker holding a semaphore slot) is back. config.go:409 only validates > 0, so this configuration is accepted today.
                                                                                                                                                                                                   
  Why it matters: the PR description states "Behavior change: None operator-visible. These are tighter granular timeouts that fail faster on already-broken broker connections." That promise only 
  holds when the operator hasn't customized request_timeout_duration to a value below 30 s.
                                                                                                                                                                                                   
  Recommended fix (pick one):                                                               
  1. (looks the better option to me) Derive the granular timeout from config, e.g. ResponseHeaderTimeout: sempCfg.RequestTimeoutDuration / 2 (or - 5*time.Second), so the relationship is preserved no matter what the operator
  configures. This mirrors how the pool sizing already derives from MaxConcurrentPerBroker.                                                                                                        
  2. Validate the relationship in config.go:409 — reject request_timeout_duration ≤ DefaultResponseHeaderTimeoutSeconds (or warn), and document the lower bound in the YAML field tag at
  config/config.go:140.

solace-aross:
Thanks @bczoma — good catch. Took option 1 in 0da8392: `ResponseHeaderTimeout` is now derived as `sempCfg.RequestTimeoutDuration / 2`, so the "strictly less than the outer client timeout" relationship holds regardless of what an operator sets for `semp.request_timeout_duration` in `broker-config.yaml`. Mirrors how `MaxIdleConnsPerHost` already derives from `MaxConcurrentPerBroker`.

Dropped the now-unused `DefaultResponseHeaderTimeoutSeconds` constant. Updated tests to cover both the 60s default and an operator-set 10s aggressive timeout case (which would have masked the regression with the previous hardcoded 30s).

Left `TLSHandshakeTimeout` (10s) and `ExpectContinueTimeout` (1s) as constants — those bound connection-establishment phases, not the response-wait window, so they're not on the same axis as `RequestTimeoutDuration`.

solace-aross:
Follow-up cleanup: moved the two new TLS/Expect-Continue timeouts to `transport.go` as unexported constants (`tlsHandshakeTimeout`, `expectContinueTimeout`) since they're not operator-configurable. Keeps `internal/defaults` strictly for override-able defaults. `ResponseHeaderTimeout` derivation from operator config is unchanged.

bczoma (approved):
approved

PR #60, opened 2026-05-19
SOL-149904: Skip env var substitution in YAML comments

solace-aross:
## Summary

Jira: [SOL-149904](https://sol-jira.atlassian.net/browse/SOL-149904)

`substituteEnvVars` ran the `${VAR_NAME}` regex against raw YAML bytes via `regexp.ReplaceAllFunc`, with no awareness of YAML comment syntax. A `${VAR}` reference inside a `#` comment was treated identically to a live field, so an unset env var on a commented-out line aborted config loading even though the line has no effect on the parsed config.

Example: the shipped `broker-config.example.yaml` itself triggers this — an operator copying the example and using basic auth would still see config loading fail with `environment variables not set: BROKER_TOKEN` because line 50 of the example file contains the commented-out bearer-token reference.

## Approach

Process input line by line. Run the existing regex only on the portion of each line before any unquoted `#` comment marker. Add a small `splitYAMLComment` helper that tracks single/double quote state so a `#` inside a quoted value remains part of the value.

- Pure byte transformation — no behavior change for live fields.
- Block scalars (`|`, `>`) are not handled by the heuristic; the config schema doesn't use them. Documented inline.
- No new dependencies. No public API change. No logging change.

## Files changed

- `internal/config/config.go` — modified `substituteEnvVars`; added `splitYAMLComment` helper. (+58, -15)
- `internal/config/config_test.go` — added 3 tests. (+72)

## Test plan

- [x] `go build ./...` — passes
- [x] `go test ./...` — all 12 packages green
- [x] New tests for whole-line and inline comment cases (failing before the fix, passing after)
- [x] Regression guard test for `#` inside a quoted value (passing before and after)
- [x] Revert-verify performed: stashed implementation → new tests went red; restored → green
- [x] `/check-logs` on the diff — 0 violations
- [x] `gofmt` / `go vet` — clean

[SOL-149904]: https://sol-jira.atlassian.net/browse/SOL-149904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -19,6 +19,7 @@
 package config
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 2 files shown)


solace-wmaryam (approved):
This looks good overall. One minor concern is that the yaml parsing logic seems a bit complex and difficult to understand at first glance. Might be difficult to maintain on our end. I did a quick search though and looks like there aren't any good alternative utility functions available that would be suitable for our use-case. So it's probably fine!

PR #61, opened 2026-05-20
SOL-148432: Implement Advanced Monitoring Tools (Replication, Discards, Slow Subscribers)

solace-lootii:
## Summary

Adds three composite monitoring tools (`get-replication-status`, `list-slow-subscribers`, `get-discard-stats`) to complete the operator-facing diagnostics surface for SOL-148432.

Fixes SOL-148432

## Type of Change

## Motivation and Context

The MCP server's operator-facing toolset previously covered VPN health, queues, clients, and RDPs but had gaps around replication state, slow-subscriber detection, and per-queue discard accounting. Without these tools, an LLM-assisted operator couldn't answer common diagnostic questions ("is replication sync-eligible?", "which clients are slow?", "where are messages being dropped?") without falling back to the broker UI or raw SEMP calls.

These three tools close that gap and finish the advanced-monitoring scope of SOL-148432.

## Changes Made

- **`get-replication-status`** — single-step composite over `monitor/getMsgVpn` with a curated 18-field `select` covering role, enabled/sync-eligible/transaction-mode state, both bridge directions, queue binding, and active sync/async queued counters (including `replicationActiveSyncQueuedAsAsyncMsgCount`).
- **`list-slow-subscribers`** — paginated composite over `monitor/getMsgVpnClients` using broker-side predicate filter `where=slowSubscriber==true`, avoiding a full-client scan over the network. Returns 10 fields per matching client (client/queue identity + slow-subscriber-relevant counters).
- **`get-discard-stats`** — paginated composite over `monitor/getMsgVpnQueues` returning per-queue discard counters across 12+ categories (TTL-expired, max-redelivery, spool-quota-exceeded, DMQ-routed, destination-group errors, client-profile-denied, etc.).
- All three are pure YAML composite tools — no new Go code required.
- Each tool uses a curated `select` list so responses are bounded and only carry fields relevant to the question the tool answers.
- Tests added in `loader_test.go` (schema/parameter shape) and `executor_test.go` (single-page, truncation, SEMP-error paths, and a `where=slowSubscriber==true` assertion).

## Testing

**Test Environment:**
- Go version: 1.25.0
- Broker version: lab-129-78 (Solace PubSub+ software broker)

**Tests Performed:**

**Test Coverage:**
- `TestLoadTools_GetReplicationStatus`, `TestLoadTools_ListSlowSubscribers`, `TestLoadTools_GetDiscardStats` — verify each tool loads with expected name, step ID, operation, `followPages` flag, and parameter shape.
- `TestExecute_GetReplicationStatus_ReturnsData` + `_SEMPError` — happy path returns replication object; SEMP error propagates as `*SEMPError`.
- `TestExecute_ListSlowSubscribers_ReturnsFiltered` — asserts the `where=slowSubscriber==true` predicate is actually passed through to the SEMP client (not silently dropped).
- `TestExecute_ListSlowSubscribers_TruncatesAtMaxResults` — pagination + truncation honored.
- `TestExecute_GetDiscardStats_SinglePage` + `_TruncatesAtMaxResults` — single-page response and `maxResults`-bounded truncation.
- Full suite (`go test -count=1 ./...`) passes; `go vet ./...` clean; composite package coverage 92.5%.
- Live tested against the lab-129-78 broker via MCP from Claude Code:
  - `get-replication-status default` → returned cluster object, replication disabled, LLM correctly rendered N/A for derived fields.
  - `list-slow-subscribers default` → empty array (broker-side `where` predicate working — no slow subscribers on an idle lab broker).
  - `get-discard-stats default` → empty array (no queues with traffic), confirming clean request/response wiring.


internal/composite/definitions/tools.yaml
```diff
@@ -317,3 +317,84 @@ tools:
           select: "clientName, enabled, lastFailureReason, lastFailureTime, msgVpnName, restDeliveryPointName, up"
     result:
       strategy: collect
+
+  - name: get-replication-status
+    description: >
+      Get replication state for a Message VPN including role, sync eligibility,
+      bridge status, transaction mode, and queued message counts. Use this to
+      diagnose replication lag and sync issues. Note: lagSeconds is not exposed
+      by SEMP; use replicationActiveSyncIneligiblePeakTime as a proxy.
```
solace-aross: Small framing concern: `replicationActiveSyncIneligiblePeakTime` is a *historical worst-case duration in which sync was ineligible*, not a current-lag signal. It can be high while replication is currently healthy, or low while replication is currently lagging — so as a hint to an LLM doing live diagnosis, calling it a "lag proxy" may mislead.

Suggested reframe: *"use `replicationActiveSyncIneligiblePeakTime` as a historical indicator of past sync stalls; SEMP does not expose current lag"*. If you want a "has replication been flaky recently?" signal, `replicationActiveTransitionToSyncIneligibleCount` (already in the select list) is probably the more honest pointer.

solace-aross (commented):
Nicely scoped PR — three pure-YAML tools that close real diagnostic gaps without adding new executor machinery, plus solid unit coverage on the happy path, truncation, and the SEMP-error branch.

**What I particularly liked:**
- The `where`-predicate-passthrough assertion is exactly the right defensive test. Verifying the filter actually reaches the SEMP client (rather than being silently dropped by the YAML → args plumbing) is the kind of small, deliberate test that pays off later.
- Curated `select` lists keep payloads bounded.
- Honest documentation about SEMP limitations (no `lagSeconds`, no pre-aggregated discards) — small honesty that saves a future debugging round-trip.
- Including `replicationActiveSyncQueuedAsAsyncMsgCount` is a thoughtful operator-experience touch — easy to miss but operationally meaningful.

**A few notes here in the summary (specific suggestions left as inline comments):**

1. **The live "predicate working" inference doesn't quite hold.** The PR description reads: *"list-slow-subscribers default → empty array (broker-side where predicate working — no slow subscribers on an idle lab broker)."* An empty array would come back whether the predicate is valid or not — the broker just returns no matches. The unit test nails the passthrough, but the live run doesn't actually confirm the broker accepts the syntax. Two cheap ways to firm this up: (a) produce a slow subscriber on the lab broker once, or (b) send a deliberately-malformed `where` and confirm the broker errors. Worth doing once before this is relied on for triage.

2. **Optional: an empty-result test for `list-slow-subscribers`.** This is going to be the most common production return value (most VPNs don't have slow subscribers most of the time). The truncation and 3-item paths are covered; a `pageResult([], "")` case would lock in the empty-data behavior. The executor already handles it correctly via the `len(items) == 0 → break` branch, so it's small belt-and-suspenders.

3. **Tiny convention nit:** each step pins `count: "100"` even when the user's `maxResults` is smaller. Output is correct because the executor truncates in-memory, and this matches existing tools — so don't change it here. Just flagging for a future "tighten page size to maxResults" pass if you ever do one.

**Risk:** Low. Additive, read-only, routes through existing tested machinery. The main residual risks are operator/LLM confusion from the doc nuances flagged inline (the lag proxy and the `get-`/`list-` naming) — both fixable in description text.

Nice work.

internal/composite/definitions/tools.yaml
```diff
@@ -317,3 +317,84 @@ tools:
...
+        followPages: true
+        args:
+          msgVpnName: "{{.Params.msgVpnName}}"
+          count: "100"
+          where: "slowSubscriber==true"
+          select: "clientAddress, clientName, clientUsername, msgVpnName, platform, rxMsgRate, slowSubscriber, txDiscardedMsgCount, txMsgRate, uptime"
+    result:
+      strategy: collect
+
+  - name: get-discard-stats
```
solace-aross: Naming-convention thought: this tool returns an array of per-queue rows — same shape as `list-queues`, `list-clients`, and the new `list-slow-subscribers` — but it's named `get-*`, which in this catalog usually implies a single object (cf. `get-vpn-health`, `get-queue-metrics`, `get-rdp-status`).

Copilot already flagged the "aggregate" wording in the parameter doc; this is the same confusion at the tool-name level. `list-queue-discards` would line up with the rest of the catalog and make the LLM's mental model a bit easier. Not blocking — just worth a thought before this name is baked in and tool-rename costs go up.

internal/tools/sempv1/discardstats/handler.go
```diff
@@ -0,0 +1,270 @@
...
+// manager's errors.As path can extract *sempv1.Error. XML decode errors are
+// wrapped with the tool-name + step prefix.
+func (h *Handler) Handle(
+	ctx context.Context,
+	tc *tools.ToolContext,
+	params map[string]any,
+) (*tools.ToolResult, error) {
+	if vpnRaw, ok := params["vpnName"]; ok {
+		vpnName, _ := vpnRaw.(string)
+		if vpnName != "" {
```
amitmorade, line 146: **`get-discard-stats` silently downgrades `vpnName: ""` to broker-wide mode.**

`handler.go:142-148` — when `params["vpnName"]` is present but empty, the `vpnName != ""` check fails and we fall through to `handleBrokerWide` with no error. JSON Schema's `"type": "string"` accepts `""` as valid, so empty slips past `ValidateParams`. An LLM that intended VPN-scoped stats receives broker-wide totals and reports them as VPN-scoped.

Suggested fix — add `"minLength": 1` to the schema so empty `vpnName` is rejected before `Handle` runs:

```go
"vpnName": map[string]any{
    "type":      "string",
    "minLength": 1,
    "description": "...",
},
```

Note: this is to avoid silently misleading an LLM that sends an empty `vpnName`. Worth testing end-to-end to confirm the validation error surfaces to the LLM as an unambiguous message it can act on.

internal/tools/sempv1/discardstats/handler.go
```diff
@@ -0,0 +1,270 @@
...
+		Description: "Returns pre-aggregated message discard counters from " +
+			"the broker. Without vpnName, returns broker-wide totals across " +
+			"two categories: client-level discards (ingress/egress — no " +
+			"subscription match, topic-parse errors, msg-too-big, TTL " +
+			"exceeded, transmit congestion, client-not-connected, etc.) " +
+			"and spool-level discards (max-redelivery exceeded, TTL " +
+			"expired to DMQ, spool/queue over quota, max-msg-size " +
+			"exceeded, replication-standby discards, etc.). With vpnName, " +
+			"returns only client-level discards scoped to that VPN — the " +
+			"broker does not expose per-VPN spool discards via SEMPv1. " +
```
amitmorade: **Tool description could lead with the per-VPN spool-discard gap.**

The current description mentions *"the broker does not expose per-VPN spool discards via SEMPv1"* in the middle of the paragraph, after listing spool-level discard categories. An LLM doing retrieval on "VPN spool discards" could match this tool, parse left-to-right, see spool categories first, and call it with `vpnName=X` expecting both client- and spool-level counters — receiving only client-level with no indication the spool question went unanswered.

Suggested reframe: move the asymmetry to the top, name the question the tool can't answer ("how many messages did VPN X drop from the spool?"), and explicitly route the LLM to list-queue-discards for that case. Curious what you think.

bczoma (changes requested):
@solace-lootii as discussed in Slack: [SOL-150328](https://sol-jira.atlassian.net/browse/SOL-150328) doesn't consider your PR because it looked at main branch at the time it was raised. You could just independently update the PR branch with your suggested improved description

internal/tools/sempv1/discardstats/handler.go
```diff
@@ -0,0 +1,274 @@
...
+)
+
+// vpnStatsXML returns the per-VPN client-stats RPC. vpnName is XML-escaped to
+// neutralize any embedded markup before interpolation — the broker's parser
+// would reject most adversarial inputs, but escaping at the boundary is the
+// right defensive posture.
+func vpnStatsXML(vpnName string) string {
+	return fmt.Sprintf(
+		`<rpc><show><message-vpn><vpn-name>%s</vpn-name><stats/></message-vpn></show></rpc>`,
+		html.EscapeString(vpnName),
```
bczoma: Claude recommends this shall be `sempv1.Escape(vpnName)`

bczoma (approved):
Looks good, thank you!

PR #62, opened 2026-05-21
SOL-149989: Consolidate client auth config to single client_auth.mode enum

solace-aross:
## Summary

Replaces the `development_mode` + `dev_token` interaction with a single required `client_auth.mode` enum. Closes the SOL-149921 silent auth-bypass through structural simplification rather than patching.

- `client_auth.mode` is required (`disabled` | `static` | `oauth`)
- Operational profile (`https://` enforcement, etc.) derives from mode via `IsProductionMode()`
- Boot banner from `cmd/server/main.go` provides refactor-robust insecure-mode signal
- `development_mode` field accepted (so old configs parse), logs deprecation warning, otherwise ignored

Design spec: \`docs/superpowers/specs/2026-05-20-client-auth-mode-design.md\`
Jira: [SOL-149989](https://sol-jira.atlassian.net/browse/SOL-149989) (supersedes SOL-149921; PR #50 closed)

## Test plan

- [x] \`go test ./...\` green
- [x] \`go vet ./...\` clean
- [ ] \`/check-logs\` 0 CRITICAL/HIGH on diff
- [ ] Manual smoke: start server with each of the three modes, observe banner + endpoint behavior
- [x] FOSSA SCA + vulnerability scans pass
- [x] CODEOWNERS review

## Follow-ups (separate tickets to be filed)

- Restore e2e OAuth test path (TLS for test Keycloak)
- Investigate lightweight local OAuth server (Mark's vision — long-term replacement for `mode: disabled`/`static`)

[SOL-149989]: https://sol-jira.atlassian.net/browse/SOL-149989?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/auth/banner.go
+++ b/internal/auth/banner.go
@@ -51,14 +51,13 @@ func LogStartupBanner(cfg *config.ServerConfig) {
 const disabledBanner = `
 ============================================================
   INSECURE MODE: client_auth.mode = disabled
-  Client authentication is DISABLED.
-  All MCP requests pass through without verification.
-  This is development mode — NOT FOR PRODUCTION USE.
+  Client → MCP server auth is OFF. Broker auth is unaffected.
+  Development mode — NOT FOR PRODUCTION USE.
 ============================================================`
 
 const staticBanner = `
 ============================================================
   INSECURE MODE: client_auth.mode = static
-  Authentication uses a shared static dev token.
-  This is development mode — NOT FOR PRODUCTION USE.
+  Client → MCP server auth uses a static dev token. Broker auth is unaffected.
+  Development mode — NOT FOR PRODUCTION USE.
 ============================================================`
--- a/internal/auth/banner.go
+++ b/internal/auth/banner.go
@@ -1,0 +1,64 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"log/slog"
+
+	"github.com/SolaceDev/solace-broker-mcp/internal/config"
+)
+
+// LogStartupBanner emits a prominent, refactor-robust signal of the configured
+// client authentication mode. It lives in the boot path (called from
+// cmd/server/main.go) — NOT in NewAuthMiddleware/NewTokenVerifier — for three
+// reasons:
+//
+//  1. Single emission point — any future transport (stdio, gRPC) goes through
+//     main and picks up the signal automatically.
+//  2. Refactor-robust — surviving auth-path restructures means the signal
+//     cannot vanish silently.
+//  3. Visually loud — multi-line banner with INSECURE MODE callouts won't
+//     get lost in request-log noise.
+//
+// For mode: disabled and mode: static the banner emits at WARN. For mode:
+// oauth a single INFO log line is emitted; production is the unremarkable case.
+//
+// Do not move this into middleware. See SOL-149989 design spec at
+// docs/superpowers/specs/2026-05-20-client-auth-mode-design.md.
+func LogStartupBanner(cfg *config.ServerConfig) {
+	switch cfg.ClientAuth.Mode {
+	case config.AuthModeDisabled:
+		slog.Warn(disabledBanner)
+	case config.AuthModeStatic:
+		slog.Warn(staticBanner)
+	case config.AuthModeOAuth:
+		slog.Info("client auth: OAuth/OIDC", slog.String("issuer", cfg.ClientAuth.Issuer))
+	}
+}
+
+const disabledBanner = `
+============================================================
+  INSECURE MODE: client_auth.mode = disabled
+  Client authentication is DISABLED.
+  All MCP requests pass through without verification.
+  This is development mode — NOT FOR PRODUCTION USE.
+============================================================`
+
+const staticBanner = `
+============================================================
+  INSECURE MODE: client_auth.mode = static
+  Authentication uses a shared static dev token.
+  This is development mode — NOT FOR PRODUCTION USE.
+============================================================`
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -40,21 +40,13 @@ import (
 // ServerConfig holds the complete MCP server configuration, including all
 // configured brokers and SEMP client settings.
 type ServerConfig struct {
-	Brokers         map[string]*BrokerConfig // broker alias → config
-	SEMP            SEMPConfig               // SEMP client settings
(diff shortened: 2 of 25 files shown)


internal/auth/banner.go
```diff
@@ -0,0 +1,64 @@
...
+	case config.AuthModeOAuth:
+		slog.Info("client auth: OAuth/OIDC", slog.String("issuer", cfg.ClientAuth.Issuer))
+	}
+}
+
+const disabledBanner = `
+============================================================
+  INSECURE MODE: client_auth.mode = disabled
+  Client authentication is DISABLED.
+  All MCP requests pass through without verification.
```
amitmorade: The banner says "All MCP requests pass through without verification" — but only the client-to-server auth is disabled. The server-to-broker connection is still authenticated. Worth scoping the wording so operators don't read it as global:

INSECURE MODE: client_auth.mode = disabled
Client → MCP server auth is OFF. Broker auth is unaffected.
Development mode — NOT FOR PRODUCTION USE.

internal/auth/banner.go
```diff
@@ -0,0 +1,64 @@
...
+  This is development mode — NOT FOR PRODUCTION USE.
+============================================================`
+
+const staticBanner = `
+============================================================
+  INSECURE MODE: client_auth.mode = static
+  Authentication uses a shared static dev token.
```
amitmorade: same for this line:

Client → MCP server auth uses a static dev token. Broker auth is unaffected.

internal/config/config.go
```diff
@@ -44,17 +44,29 @@ type ServerConfig struct {
...
 	LogLevel        string                   // slog level name: "debug", "info", "warn", "error"
-	DevelopmentMode bool                     // use static dev token if true
+	// DevelopmentMode is no longer used. Retained only so external Go callers
+	// that referenced this field continue to compile; the YAML key is parsed
+	// (so old configs don't fail with "unknown field") and a deprecation
+	// warning logs at boot if present. Operational profile and auth backend
+	// are now derived from ClientAuth.Mode. See IsProductionMode().
+	//
+	// Deprecated: use ClientAuth.Mode and ServerConfig.IsProductionMode().
+	DevelopmentMode bool
```
amitmorade: Given we're pre-release and internal-only, is the `DevelopmentMode` shim earning its keep? Removing the field entirely (and the YAML parse-and-warn path) would be a few lines deleted, a quick Slack heads-up, and one CI run. Carrying a misleading public field pre-1.0 feels like paying compatibility cost without compatibility benefit. Curious whether this was deliberate practice for handling breaking changes, or just caution we could drop here.

One thing that pulls me toward removal: the field is declared on `ServerConfig (config.go:54)`, but the struct literal in `LoadConfig` (config.go:275-283) doesn't seem to copy `raw.DevelopmentMode` into it — so `cfg.DevelopmentMode` ends up as `false` regardless of the YAML value.

internal/config/config.go
```diff
@@ -479,32 +512,49 @@ func validate(cfg *ServerConfig) error {
...
-			errs = append(errs, fmt.Errorf("client_auth.resource_url is required when development_mode is false"))
+			errs = append(errs, fmt.Errorf("client_auth.resource_url is required when client_auth.mode is %q", AuthModeOAuth))
 		}
+	default:
+		errs = append(errs, fmt.Errorf("client_auth.mode %q is invalid (must be one of %v)", cfg.ClientAuth.Mode, validAuthClientModes))
 	}
 
-	// Validate issuer structure if set
+	// Validate issuer structure if set (under mode: oauth — required; under
+	// other modes — ignored if present, as documented in the spec).
```
amitmorade: Under mode: `static` or `disabled`, OAuth fields are described as ignored at runtime — but the validator still runs `validateBrokerURL` on `issuer` and `resource_url` if they're set (config.go:549-560). Worth noting this pattern also exists on main, so it's pre-existing rather than introduced here. Still curious if it's intentional, because it reads as a small contradiction with the rest of the switch: the "is it set" checks are mode-gated, but the "is it a valid URL" checks aren't.

There's a fair argument for validating fields whenever they're present — catches stale config early so a future mode flip doesn't fail at runtime. If that's the intent, totally reasonable.

The thing that feels off is the asymmetry: under `mode: static`, a missing `issuer` is silently accepted, but a malformed one is rejected. Either the field matters (validate fully) or it doesn't (ignore fully). Curious whether you'd lean toward making both checks mode-gated, or making both checks eager — either feels more consistent than the current half-and-half.

internal/config/config.go
amitmorade: If we do drop it, those `DevelopmentMode: true/false` assignments still sitting in `middleware_test.go` (lines 87, 171, 291, etc.) would also need to go — they'd stop compiling once the struct field is removed. Either way (drop the field or keep it), they read as if the field still matters, which it doesn't.

internal/auth/banner.go
solace-aross: Fixed in 8add83a — banner now reads: "Client → MCP server auth is OFF. Broker auth is unaffected."

internal/auth/banner.go
solace-aross: Fixed in 8add83a — banner now reads: "Client → MCP server auth uses a static dev token. Broker auth is unaffected."

internal/config/config.go
solace-aross: Removed entirely in 99c603f (plus stale-comment cleanup in a63f6ed and 07788be).

Verified your read on the LoadConfig literal — `raw.DevelopmentMode` was indeed never copied into `cfg.DevelopmentMode`, so the field was silently always `false`. Worst-of-both-worlds shim. Dropped the Go field; kept the YAML parse-and-warn path (`yamlConfig.DevelopmentMode *bool` and `TestLoadConfig_DevelopmentModeDeprecationWarning` unchanged) so old configs still load with a friendly deprecation message — that's the actual operator-facing concern.

internal/config/config.go
solace-aross: Mode-gated both checks in 7f391f4 — `validateBrokerURL` on `issuer` and `resource_url` now lives inside `case AuthModeOAuth:` as `else if` clauses. Missing-and-malformed are treated symmetrically per mode now (both rejected under `mode: oauth`, both ignored under `static`/`disabled`).

Added a paired regression test in 808ca99: `TestLoadConfig_StaticMode_IgnoresOffModeOAuthFields` (proves malformed off-mode fields are accepted under static) and `TestLoadConfig_OAuthMode_RejectsMalformedIssuer` (proves the symmetric rejection under oauth still holds — guarding against a future refactor accidentally weakening either direction).

internal/config/config.go
solace-aross: All 11 `DevelopmentMode:` literals deleted in 99c603f. Production middleware already gated on `cfg.ClientAuth.Mode` exclusively, so those test fields were redundant scaffolding rather than load-bearing setup. Two table-test structs also lost a redundant `devMode bool` field.

Renames:
- `Test_DevelopmentModeFlexibility` → `Test_StaticMode_AllowsMissingIssuerAndAudience`
- `TestLoadConfig_AllowsHTTPBrokerInDevelopmentMode` → `TestLoadConfig_AllowsHTTPBrokerInStaticMode`

Two follow-up commits (a63f6ed, 07788be) caught stale `"development mode"` phrasing in a doc comment and a `t.Errorf` message that survived the initial sweep.

amitmorade (approved):
Looks good to me

PR #67, opened 2026-05-25
SOL-148426: Implement Basic Health Endpoints

solace-wmaryam:
## Summary

Implements /health and /ready HTTP endpoints for MCP server liveness and  broker readiness checks. The readiness probe uses a lightweight TCP dial instead of a full SEMP API call, confirming broker reachability without putting load on the broker. 

Fixes https://sol-jira.atlassian.net/browse/SOL-148426

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

Platform operators deploying the MCP server need health and readiness endpoints to integrate with broker orchestration  and monitor MCP server health. The readiness endpoint must verify broker connectivity, but a full SEMP API call (authentication + HTTP round-trip + broker query) is unnecessarily heavy for a probe that runs on every check interval. A TCP dial to the broker's SEMP port confirms the broker is up and accepting connections with essentially zero load.

## Changes Made

- Added /health endpoint returning {"status":"healthy"} with HTTP 200
- Added /ready endpoint that concurrently TCP-dials each configured broker's SEMP host:port (2s timeout per broker), returning {"ready":true,"brokers":[...]} on HTTP 200 or {"ready":false,"errors":[...]} on HTTP 503
- Refactored buildMux to accept aliases and ping as injected dependencies, decoupling the route logic from broker connectivity implementation and enabling unit testing without a real broker
- Updated /health response body from {"status":"ok"} to {"status":"healthy"} 
-  Added unit and integration tests                        

## Testing

**Test Environment:**
- Go version: 1.26.2
- OS: macOS 26.3.1
- Broker version (if applicable): latest

**Tests Performed:**

**Test Coverage:**

- Added unit tests for /ready: method not allowed, all brokers reachable (200), one broker unreachable (503)                                       - Added standalone e2e test test/e2e/health/test_health.sh that manages its own docker containers and MCP server, covering: /health happy path, /ready with both brokers reachable, and /ready with an unreachable broker

## Breaking Changes

**Impact:**

None. The /health response body changed from {"status":"ok"} to {"status":"healthy"} — any client already polling /health and parsing the status field will see a different string value. The existing test_standalone.sh assertion has been updated accordingly.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -127,14 +127,20 @@ func buildMux(aliases func() []string, ping func(ctx context.Context, broker str
 		wg.Wait()
 		close(results)
 
-		// reduce: partition into ready brokers and errors
-		var readyBrokers []string
-		var errs []string
+		// Collect goroutine results into a map, then emit in the original broker
+		// order so the JSON response is deterministic regardless of goroutine
+		// scheduling. Initialize as empty slices so they marshal as [] not null.
+		resultsByBroker := make(map[string]error, len(brokers))
 		for res := range results {
-			if res.err != nil {
-				errs = append(errs, fmt.Sprintf("%s: %s", res.broker, res.err.Error()))
+			resultsByBroker[res.broker] = res.err
+		}
+		readyBrokers := []string{}
+		errs := []string{}
+		for _, broker := range brokers {
+			if err := resultsByBroker[broker]; err != nil {
+				errs = append(errs, fmt.Sprintf("%s: %s", broker, err.Error()))
 			} else {
-				readyBrokers = append(readyBrokers, res.broker)
+				readyBrokers = append(readyBrokers, broker)
 			}
 		}
 
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -251,7 +251,7 @@ func startServer(srv *http.Server, tlsCertFile, tlsKeyFile string) <-chan error 
 // production implementation of buildMux's probeBroker parameter.
 func newBrokerReachabilityProbe(cfg *config.ServerConfig) func(context.Context, string) error {
 	return func(ctx context.Context, broker string) error {
-		brokerCfg, ok := cfg.Brokers[broker]
+		brokerCfg, ok := cfg.Broker(broker)
 		if !ok {
 			return fmt.Errorf("unknown broker %q", broker)
 		}
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -81,8 +81,8 @@ func newSlogHandler(level slog.Level) slog.Handler {
 // Both main() and tests use this function to avoid route drift.
 //
 // aliases returns the list of broker names to check.
-// ping tests connectivity to a single broker; it returns nil on success.
-func buildMux(aliases func() []string, ping func(ctx context.Context, broker string) error) *http.ServeMux {
+// probeBroker tests connectivity to a single broker; it returns nil on success.
+func buildMux(aliases func() []string, probeBroker func(ctx context.Context, broker string) error) *http.ServeMux {
 	mux := http.NewServeMux()
 
 	mux.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
@@ -104,6 +104,8 @@ func buildMux(aliases func() []string, ping func(ctx context.Context, broker str
 		}
 		w.Header().Set("Content-Type", "application/json")
 
+		// Config validation ensures there is at least one broker configured, so aliases()
+		// is never empty in production.
 		brokers := aliases()
 		type brokerResult struct {
 			broker string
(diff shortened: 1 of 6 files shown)


solace-aross (commented):
Nice work on this. The dependency injection into `buildMux` is a clean refactor, the test coverage on the new endpoint is solid, and the TCP dial as a lightweight readiness signal is a reasonable trade-off that the PR description explains well. A few things I think are worth addressing before merge below. Copilot's three earlier comments all look resolved in the latest push, so I won't re-raise those.

### 🟡 Extract the TCP-dial closure out of `main()` for testability

`main.go:389-416` builds a 28-line closure inline to satisfy the `ping` parameter. It does real work (URL parsing, port defaulting, dial, logging) but nothing exercises it directly. The unit tests cover `buildMux` with a fake ping, and the e2e test exercises the integrated path, but the URL parsing and port defaulting in between has no unit coverage. If `url.Hostname()` returns empty for an IPv6-bracketed URL, or the scheme is something unexpected, we'd find out in production.

Suggestion: pull it out as a named function, e.g. `func newBrokerReachabilityProbe(cfg *config.ServerConfig) func(context.Context, string) error`, and add focused unit tests for the URL parsing and port resolution cases. `main()` then reads as `mux := buildMux(pool.Aliases, newBrokerReachabilityProbe(cfg))`.

### 🟡 Move the 2-second timeout to `internal/defaults`

`main.go:117` hardcodes `2*time.Second` as a magic number inside the handler. Every other operational timeout in this server lives in `internal/defaults` (DefaultShutdownTimeoutSeconds, DefaultReadHeaderTimeoutSeconds, DefaultSEMPRequestTimeoutDuration, etc.) with a comment explaining the rationale. A new `DefaultReadinessProbeTimeoutDuration` constant there keeps the convention and gives operators a single place to find tunables.

### 🟡 Rename the `ping` parameter

The signature `buildMux(aliases func() []string, ping func(ctx, broker) error)` reads like ICMP. It's actually "is this broker's SEMP port accepting TCP." Something like `probeBroker` or `checkBrokerReachable` documents intent at the call site without needing to read the closure body.

### 🟡 Handler returns `ready=true` when zero brokers are configured

If `aliases()` returns an empty slice the handler skips the loop, never hits the error branch, and emits `{"ready":true,"brokers":[]}` with HTTP 200. Config validation probably prevents this, but readiness is a defense-in-depth signal. Worth either explicitly rejecting empty broker lists with a 503 here, or asserting elsewhere that the config validator guarantees at least one broker is configured.

### 🟡 `/ready` errors leak per-broker network details to an unauthenticated endpoint

`main.go:339` shows `/mcp` is wrapped in auth middleware, but `/ready` is not. The 503 body returns strings like `\"broker-a: dial tcp 10.0.0.5:1943: connect: connection refused\"`, which gives anyone who can reach the server a list of broker aliases, resolved IPs, and ports. K8s-internal exposure makes this fine; if `/ready` is ever reachable from outside the cluster (e.g. via an ingress that doesn't filter probe paths) it becomes recon-friendly. At minimum worth a comment at the handler noting the assumption that `/ready` is on the internal probe network only; better to return generic per-broker status (`\"broker-a: unreachable\"`) and log the detailed reason instead.

### 🟢 PR description path doesn't match the diff

The description says the new e2e lives at `test/e2e/health/test_health.sh` but the file lands at `test/health/test_health.sh`. The script also sources helpers via `../e2e-basic-mcp/helpers.sh`, which cross-couples two top-level test directories. Either consolidate under `test/e2e/health/` to match the description and avoid the relative path hop, or update the description. Minor, but easier to fix now than to debug later.

### 🟢 Consider whether the `/health` body change is worth the break

Changing `{\"status\":\"ok\"}` to `{\"status\":\"healthy\"}` is called out as a breaking change but I don't see a corresponding gain. Any external poller parsing the field has to update. If there's a downstream consumer that requires `\"healthy\"`, ignore this; if not, keeping `\"ok\"` saves us the breakage for no functional difference.

### 🟢 Minor: the `// map: ...` / `// reduce: ...` comments

The map/reduce terminology in the handler feels heavier than the code warrants. A one-line "run pings concurrently, then collect in the configured order" would be lighter. Cosmetic.

Thanks for putting this together; the bones of the change are good and the trade-offs are well-reasoned.

solace-aross (approved):
Approving. The substantive items from the last round are addressed, verified against the pushed code (build and `go test ./cmd/server/... ./internal/auth/...` pass locally):

- TCP-dial logic extracted to `newBrokerReachabilityProbe`, with focused unit tests for the URL parsing and 443/80 port defaulting.
- Readiness timeout moved to `defaults.DefaultReadinessProbeTimeoutSeconds`.
- `ping` renamed to `probeBroker`.
- 503 body now returns `"<broker>: unreachable"` and logs the dial detail, so resolved IPs and ports no longer leak from an unauthenticated endpoint.
- Results collected in configured broker order; `brokers`/`errors` init to `[]` so the arrays are never `null`.
- Tests now assert the `brokers` and `errors` fields and `Content-Type`.

Two non-blocking follow-ups, neither gating this merge:

1. A zero-broker config still returns `ready:true` / 200. You documented that the validator guarantees at least one broker, which is acceptable. If you want the defense-in-depth, a one-line `if len(brokers)==0 { 503 }` closes it later.
2. The health test lives at `test/health/` and sources helpers from `../e2e-basic-mcp/`. It works, but the cross-directory coupling I flagged is still there. Worth tidying in a future pass.

Good iteration on this.

PR #72, opened 2026-05-26
SOL-150064: Restore e2e OAuth integration test — add TLS to test Keycloak

solace-wmaryam:
## Summary

Restore the e2e OAuth integration test by adding TLS to the local Keycloak instance so its issuer URL satisfies the `https://` requirement introduced in SOL-149989.

Fixes [SOL-150064](https://sol-jira.atlassian.net/browse/SOL-150064)

## Type of Change

## Motivation and Context

PR #62 (SOL-149989) consolidated client auth into `client_auth.mode` and enforced `https://` on the `issuer` URL under `mode: oauth`. The e2e OAuth test ran Keycloak on plain `http://localhost:8090`, so it was disabled with `if: false` in CI. Rather than weaken the validator, this PR puts TLS in front of the test Keycloak using its built-in HTTPS listener.

## Changes Made

- **Self-signed TLS cert/key** (`test/oauth/certs/`) — `CN=localhost`, SAN `DNS:localhost,IP:127.0.0.1`, checked in as test fixtures
- **Docker Compose** — mount cert/key into Keycloak, set `KC_HTTPS_CERTIFICATE_FILE`/`KC_HTTPS_CERTIFICATE_KEY_FILE`, expose port 8443 for HTTPS alongside 8090 for the admin API
- **Terraform outputs/variables** — all OIDC endpoint URLs switched from `http://localhost:8090` to `https://localhost:8443`; new `keycloak_https_port` variable
- **`oidcHTTPClient()` in `internal/auth/middleware.go`** — when `SSL_CERT_FILE` is set, builds an `http.Client` with a custom root CA pool so OIDC discovery trusts the test cert. On macOS/Windows, Go's default TLS skips `SSL_CERT_FILE`; the explicit `RootCAs` pool makes behavior consistent across platforms
- **Test scripts** (`setup-keycloak.sh`, `test-oauth.sh`) — `curl --cacert` for all OIDC calls, `export SSL_CERT_FILE` for the MCP server process, fixed `PROJECT_ROOT` paths after the directory restructure in SOL-150126
- **Test config** (`test-config.yaml`) — `issuer` updated to `https://localhost:8443/realms/solace`, removed the TODO/known-broken disclaimer

## Testing

**Test Environment:**
- Go version: 1.24
- OS: macOS / Linux (CI)

**Tests Performed:**

**Test Coverage:**
- OAuth e2e test exercises a full OIDC client-credentials flow against TLS-enabled Keycloak

## Related Issues/PRs

- Follow-up from #62 (SOL-149989)
- Related to [SOL-150126](https://sol-jira.atlassian.net/browse/SOL-150126) (directory restructure)

**For Reviewers:**

**Focus Areas**: `oidcHTTPClient()` in `internal/auth/middleware.go` — this is the only production code change. Review the `SSL_CERT_FILE` handling and the fallback to `x509.SystemCertPool()`.

[SOL-150064]: https://sol-jira.atlassian.net/browse/SOL-150064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-150126]: https://sol-jira.atlassian.net/browse/SOL-150126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/auth/middleware.go
```diff
@@ -146,6 +154,34 @@ func createOIDCTokenVerifier(cfg *config.ServerConfig) (sdkauth.TokenVerifier, e
...
+	if err != nil {
+		return nil
+	}
+	pool, err := x509.SystemCertPool()
+	if err != nil {
+		pool = x509.NewCertPool()
+	}
+	pool.AppendCertsFromPEM(certPEM)
+	return &http.Client{
+		Transport: &http.Transport{
```
solace-aross: Building a bare `&http.Transport{TLSClientConfig: ...}` drops `ProxyFromEnvironment`, HTTP/2 enablement, connection pool tuning, and the default dial timeouts. Since `SSL_CERT_FILE` is the production knob for trusting a private CA, this path can run in real deployments where those defaults matter (corporate proxies in particular). Clone the default instead:

```go
t := http.DefaultTransport.(*http.Transport).Clone()
t.TLSClientConfig = &tls.Config{RootCAs: pool}
return &http.Client{Transport: t}
```

Seconding Copilot's suppressed note here.

test/oauth/docker-compose.yml
```diff
@@ -3,14 +3,20 @@ services:
...
+      KC_HTTPS_CERTIFICATE_FILE: /opt/keycloak/conf/certs/keycloak.crt
+      KC_HTTPS_CERTIFICATE_KEY_FILE: /opt/keycloak/conf/certs/keycloak.key
+    volumes:
+      - ./certs/keycloak.crt:/opt/keycloak/conf/certs/keycloak.crt:ro
+      - ./certs/keycloak.key:/opt/keycloak/conf/certs/keycloak.key:ro
     command:
       - start-dev
     healthcheck:
-      test: ["CMD-SHELL", "timeout 2 bash -c '</dev/tcp/localhost/8080' || exit 1"]
+      test: ["CMD-SHELL", "timeout 2 bash -c '</dev/tcp/localhost/8443' || exit 1"]
```
solace-aross: Healthcheck only probes 8443, but `setup-keycloak.sh` and the Terraform provider talk to Keycloak over the 8080 admin API. The container can mark healthy before the admin path is ready. Today it passes, but it's an order-of-startup race waiting to flake. Prefer Keycloak's `/health/ready` endpoint over a TCP probe so the signal matches what consumers depend on. Also Copilot's note.

solace-aross (commented):
Well-scoped restoration of the e2e OAuth path. The TLS-on-test-Keycloak approach is the right call, better than weakening the validator. CI runs `e2e-oauth` again and it's green.

Inline comments are ordered roughly by impact. The two I'd treat as merge-blockers are (1) the bare `http.Transport{}` losing production defaults and (2) the healthcheck only probing 8443 while consumers still talk to 8080. Both also flagged by Copilot.

One ask not tied to a line: `test/oauth/certs/` could use a short README saying "test-only self-signed pair, never reuse, regenerate with `openssl req ...`". Prevents someone copying these into another fixture.

Copilot's earlier `if: false still gated` note was against the prior commit; current HEAD removes the gate. No action.

internal/auth/middleware.go
```diff
@@ -146,6 +154,34 @@ func createOIDCTokenVerifier(cfg *config.ServerConfig) (sdkauth.TokenVerifier, e
...
+	}
+	certPEM, err := os.ReadFile(filepath.Clean(certFile))
+	if err != nil {
+		return nil
+	}
+	pool, err := x509.SystemCertPool()
+	if err != nil {
+		pool = x509.NewCertPool()
+	}
+	pool.AppendCertsFromPEM(certPEM)
```
solace-aross: `AppendCertsFromPEM` returns a bool that's ignored. If `SSL_CERT_FILE` points at a corrupted or non-PEM file, the certs are silently skipped and the function still returns a custom client backed only by the system pool. An operator who set the env var expecting their private CA to be trusted then sees an opaque TLS error from OIDC discovery, with no breadcrumb back to the config.

At minimum, log a warning when the append returns false. Better: return `nil` from the helper so the upstream init path surfaces the failure cleanly.

internal/auth/middleware.go
```diff
@@ -146,6 +154,34 @@ func createOIDCTokenVerifier(cfg *config.ServerConfig) (sdkauth.TokenVerifier, e
 	}, nil
 }
 
+// oidcHTTPClient returns a custom HTTP client that trusts the CA bundle in
+// SSL_CERT_FILE, or nil when that variable is unset/unreadable. On macOS and
+// Windows, Go's default TLS verification delegates to the OS-native trust
+// store and ignores SSL_CERT_FILE; an explicit RootCAs pool bypasses that
+// delegation so the env var works consistently on every platform.
+func oidcHTTPClient() *http.Client {
```
solace-aross: The new helper has several branches (env empty, unreadable file, non-PEM contents, system-pool failure, success) and only the env-empty path is exercised by the e2e. Worth a small table-driven unit test in `internal/auth`. The `SystemCertPool` failure branch is hard to force on macOS/Linux; skip that one.

test/oauth/test-config.yaml
```diff
@@ -10,24 +10,18 @@
...
-  resource_url: "http://localhost:9091/mcp"
+  resource_url: "https://localhost:9091/mcp"
 
 # Broker configuration
 # Note: This broker is only needed for config validation.
 # The OAuth test doesn't actually connect to the broker.
 brokers:
   test-broker:
-    url: http://localhost:8080
+    url: https://localhost:8080
```
solace-aross, line 25: Broker URL flipped from `http://localhost:8080` to `https://`, but the comment above still says this broker isn't contacted by the test. Was this required by a validator change, or cosmetic? If validator-driven, a one-line in-file note would save the next person a trace.

test/oauth/test-oauth.sh
```diff
@@ -100,6 +101,7 @@ echo "✓ Using test configuration: ${TEST_CONFIG}"
 # Start MCP server
 echo ""
 echo "Starting MCP server with OAuth..."
+export SSL_CERT_FILE="${SCRIPT_DIR}/certs/keycloak.crt"
 export CONFIG_FILE="${TEST_CONFIG}"
 "${PROJECT_ROOT}/bin/mcp-test" > /tmp/mcp-oauth-test.log 2>&1 &
 MCP_SERVER_PID=$!
```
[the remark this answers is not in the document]
solace-wmaryam: Currently, this won't fail on a clean checkout as `go build` will create the directory if it does not exist. That being said, it's a simple fix and worth adding as a defensive measure.

internal/auth/middleware.go
solace-wmaryam: Added a unit test covering the failure branches along with an integration test confirming the SSL_CERT_FILE serves its purpose correctly.

test/oauth/test-config.yaml
solace-wmaryam: Correct, this was validator-driven. I have added a comment mentioning this.

test/oauth/docker-compose.yml
solace-wmaryam: The keycloak container being used for the test is minimal and doesn't have support for curl. I've updated the heathcheck to check both ports instead.

solace-aross (approved):
Approving. Both merge-blockers from the last round are resolved, and I verified each against the pushed code rather than the replies:

- OIDC client clones the default transport (`middleware.go:182`), so `ProxyFromEnvironment`, HTTP/2, and pool/dial defaults survive. This was the main one.
- `AppendCertsFromPEM` is now checked and returns an error on a bad bundle (`middleware.go:179`). That is stricter than the logged warning I suggested, and the better call: a misconfigured `SSL_CERT_FILE` fails loudly instead of silently falling back to the system pool.
- Healthcheck probes 8443 and 8080 via bash TCP redirection (`docker-compose.yml:19`), which sidesteps the no-curl-in-the-image constraint cleanly.
- Helper now has a table-driven test over the empty / unreadable / non-PEM / valid branches, plus an integration test against an `httptest` TLS server.
- `mkdir -p bin`, the validator-driven comment in `test-config.yaml`, the `certs/README.md`, and the CI job re-enable are all in.

CI is green again. Solid restoration.

PR #88, opened 2026-06-10
SOL-148462: Implement Action Tools (Queue Actions, Client Actions)

solace-lootii:
## Summary

Implements `execute-queue-action` and `execute-client-action` — two SEMPv2 native MCP tools that execute broker operational actions (`deleteMsgs`/`clearStats` on queues, `disconnect`/`clearStats` on clients).

Fixes SOL-148462

## Type of Change

## Motivation and Context

First write/action tools in the MCP server. Customer-facing workflows like clearing a dead-letter backlog or disconnecting a slow subscriber are now possible from an LLM agent. Per @bczoma review refinement, the operative confirmation control is the description text instructing the LLM to obtain explicit user confirmation before invoking — the MCP `requiresConfirmation` annotation is advisory only and not all agents honor it, and the server-side WARNING log is an audit record, not a user prompt.

## Changes Made

- `internal/tools/sempv2/queueaction/` — handler + unit tests + build-tagged integration test (`integration` tag, env-gated, skip-on-missing).
- `internal/tools/sempv2/clientaction/` — same shape for the client-side actions.
- `internal/tools/types.go` — new `RequiresConfirmation` field on `tools.Annotations` (internal advisory; SDK has no wire-protocol equivalent).
- `cmd/server/main.go` — `registerSEMPv2Tools` registration alongside the existing SEMPv1 helper.
- `README.md` + `docs/user-guide.md` — Actions row / Actions section, tool count bumped 17 → 19.

## Testing

**Test Environment:**
- Go: 1.25
- OS: Linux (dev3-205)
- Broker: lab-129-78 (SEMP 100.0main.0.7305)

**Tests Performed:**
- [x] Unit tests added (5 per package: metadata, both action paths, invalid-action defense-in-depth, SEMP error passthrough)
- [x] Integration tests added — `clearStats` on queue + client, live against `lab-129-78`
- [x] `go test -race ./...` clean

**Test Coverage:**
- Happy paths for both actions on both tools (mocked unit tests + live integration tests for `clearStats`)
- Negative paths: invalid `action` enum, empty `msgVpnName`/`queueName`/`clientName`, non-existent target (SEMP 404 surfaces as `*sempv2.SEMPError`)
- Destructive WARN log auto-fires from the manager via `Annotations.Destructive: true` — verified live
- Mode-2 LLM-driven confirmation eval against Claude (via Claude Code's MCP integration): **6/6 prompts PASS** including two adversarial bypass attempts. The eval caught a real description-text bypass on Prompt 3 ("just clear it out" was treated as preemptive consent); tightened the description with a "MUST come as a separate user reply received AFTER you state the target and effect" clause and re-tested to confirm the fix.

Pasted Graphic 6
Pasted Graphic 7
Pasted Graphic 8
Pasted Graphic 13

**After seperate user reply change in the description:**

Pasted Graphic 14
Pasted Graphic 16

**Adversarial Prompt: preemptive consent**
Pasted Graphic 17

## Breaking Changes

None.

### Code Quality
- [x] Self-review completed
- [x] Comments explain *why*, not *what*
- [x] No commented-out code

### Security
- [x] No credentials in code
- [x] No new log calls added (action WARN reuses centralized manager-side logging)

### Testing
- [x] Edge cases + error paths covered

### Documentation
- [x] README.md updated
- [x] docs/user-guide.md updated
- [x] godoc comments on exported types/funcs

### Git & CI
- [x] Commit has clear, descriptive message
- [x] Branch verified merge-clean against `main` via `git merge-tree`
- [x] No merge conflicts

## Additional Notes

**Mode-2 eval coverage gap:** the AC requires verification across the model × agent matrix. Claude is fully verified (6/6 PASS). The other three matrix LLMs (MAI-Code-1-Flash, GPT-5 mini, Raptor mini preview) are pending Copilot Pro access. Will land as a follow-up once access is granted.

**Live publish-and-delete proof:** Story 16 DoD requires only non-destructive integration tests, which are committed and passing live. The destructive path is verified end-to-end via the Mode-2 eval (Claude → `execute-queue-action deleteMsgs` → broker 200 OK + destructive WARN) on the empty `awtest` queue. Have yet to test on a queue that is not empty.

## Related Issues/PRs

- Implements SOL-148462
- Builds on the SEMPv1 native-handler pattern (file layout, mock-client tests, strict output schema, manager registration); first SEMPv2 native handlers in the codebase. similar to the native-handler pattern established by `discardstats` (PR #61)

---

**For Reviewers:**

**Focus Areas:**
- Tool description wording — the description IS the operative confirmation control, not the annotation. Both descriptions instruct restate-target + name-effect + require-separate-user-reply. Worth a careful read.
- `RequiresConfirmation` annotation is internal-only; the wire-protocol confirmation hint is the standard `DestructiveHint`.

# Follow-up: `enable_write_tools` gating (review response)

Added in response to review discussion (@bczoma, @amorade, @aross) on this PR: these are the first write tools, and read-vs-write AI-access isn't a mechanism yet. Even after the OAuth admission grant lands, developer-mode trials (`client_auth.mode: disabled`/`static`) would leave write tools wide open. Per the agreed approach — land the feature secured from day one — write tools are now gated behind a server-level flag, default off, enforced at registration so disabled tools don't even appear in `tools/list`.

### Changes Made

- `internal/config/config.go` — new `enable_write_tools` YAML key → `ServerConfig.EnableWriteTools` (default `false`, secure-by-default).
- `internal/tools/register.go` — `RegisterWithServer` now takes `enableWriteTools bool` and skips any tool whose `Annotations.Destructive` is true when the flag is off. New `isDestructive()` helper treats nil `DestructiveHint` as not-destructive (matches the spec's safer default). This is the single chokepoint for destructive registration — native or composite.
- `cmd/server/main.go` — passes `cfg.EnableWriteTools` through; logs the resolved value at registration.
- `internal/tools/types.go` — **removed** the `RequiresConfirmation` annotation field added earlier in this PR. Per the review thread it was non-operative (never reached the wire, no agent consumed it); the description-text instruction remains the sole operative confirmation control. Handlers and their metadata tests updated to drop it.
- Docs — `docs/configuration.md` (config table row), `README.md` + `docs/user-guide.md` (gating note on the Actions section), `broker-config.example.yaml` (commented example).

### Why gate at registration (not request time)

Skipping registration means the tools are never advertised in `tools/list` and cannot be invoked — independent of `client_auth.mode`. An authenticated client still can't reach them when the flag is off, because the server never registered them. Checking at request time would still expose the tools in discovery.

### Testing

- Unit: `enable_write_tools` defaults to false when omitted, reads through when set (`config_test.go`); `isDestructive` table test (nil / explicit-false / explicit-true); gated-registration smoke test over both flag values (`register_test.go`).
- Live against `lab-129-78`:
  - Flag off (and flag-absent): startup logs `destructive tool not registered` for both tools; `tools/list` returns **17 tools**, action tools absent.
  - Flag on: `tools/list` returns **19 tools**; `execute-queue-action clearStats` on `default/awtest` → `status: ok`; destructive WARN fires for the destructive path, not for `clearStats`.
- `go vet`, `gofmt`, and `go test ./...` clean on Go 1.25.

### Notes

- `enable_write_tools` is a separate control from the OAuth read-vs-write admission grant discussed on this PR — it's the developer-mode safety net so trial deployments are gated regardless of auth mode. (Tracked separately for GA gating; see thread.)
- Default stays `false` — operators must deliberately opt in to expose action tools.

enable_write_tools false
enable_write_tools true
client auth


solace-aross (commented):
Strong PR. First write path in the server and it shows real care: strict output schemas, defense-in-depth action validation, env-gated integration tests that never run the destructive path live, and an honest disclosure of the eval-matrix gap. The confirmation-via-description design is well reasoned and the prior thread already settled it; not relitigating that.

Four things, two of them in inline comments. None is a correctness bug. My one hold before approval is the dead field below.

**1. First write tools register unconditionally (architecture / sequencing).** `registerSEMPv2Tools` in `cmd/server/main.go` wires these in for every deployment, so any instance now exposes `deleteMsgs` and `disconnect` to the agent. Broker SEMP RBAC backstops actual authorization, which is the right authority. But the read-vs-write AI-access level — a separate admission grant the MCP owns — isn't represented yet. Fine to land the feature ahead of that mechanism; the ask is a tracked dependency so write tools don't reach GA ungated. Is there a ticket linking these to the AI-access work?

**2. `RequiresConfirmation` is inert (the one I'd resolve before approving).** The new `Annotations.RequiresConfirmation` field is set `true` by both handlers and asserted in tests, but nothing reads it: the destructive WARN keys off `Destructive` (`manager.go`), and `register.go` maps only `DestructiveHint` to the wire. As the PR body notes, it never propagates. A flag that looks like a control but actuates nothing invites a future reader to trust it. Either drop it and let `Destructive` carry the signal, or add a one-line "reserved, no consumer yet" comment. 30-second change.

Happy to approve once #2 is resolved or explained, with #1 captured as a tracked GA dependency.

internal/tools/sempv2/queueaction/handler.go
```diff
@@ -0,0 +1,199 @@
...
+func (h *Handler) Handle(
+	ctx context.Context,
+	tc *tools.ToolContext,
+	params map[string]any,
+) (*tools.ToolResult, error) {
+	msgVpnName, _ := params["msgVpnName"].(string)
+	queueName, _ := params["queueName"].(string)
+	action, _ := params["action"].(string)
+
+	if action != actionDeleteMsgs && action != actionClearStats {
```
solace-aross: Endorsing Copilot's point here, with a sharper framing. `Handle()` already runs a runtime guard on `action` just below "for direct Go callers that bypass the manager" — but skips the same check on `msgVpnName`/`queueName`. Same stated threat model, uneven coverage. With empty values the handler builds `/SEMP/v2/action/msgVpns//queues//deleteMsgs`; the broker 404s, so it's not dangerous, but a non-empty guard belongs right next to the action check rather than relying on upstream schema validation for these two and runtime validation for the third. Low severity, worth the consistency.

internal/tools/sempv2/clientaction/handler.go
```diff
@@ -0,0 +1,190 @@
...
+func (h *Handler) Handle(
+	ctx context.Context,
+	tc *tools.ToolContext,
+	params map[string]any,
+) (*tools.ToolResult, error) {
+	msgVpnName, _ := params["msgVpnName"].(string)
+	clientName, _ := params["clientName"].(string)
+	action, _ := params["action"].(string)
+
+	if action != actionDisconnect && action != actionClearStats {
```
solace-aross: Same as the queue handler: the `action` guard below is justified as defense for callers that bypass the manager, but `msgVpnName`/`clientName` get no equivalent non-empty check. Add the guard here too so the rationale is applied evenly. Low severity (broker 404s on the empty-segment path), consistency fix.

internal/tools/sempv2/clientaction/handler.go
```diff
@@ -0,0 +1,190 @@
...
+			{Name: "msgVpnName", In: "path", Type: "string", Required: true},
+			{Name: "clientName", In: "path", Type: "string", Required: true},
+			{Name: "body", In: "body", Type: "object", Required: true},
+		},
+	}
+}
+
+// capitalize returns s with its first ASCII letter upper-cased. Used to map
+// action constants (camelCase) to the SEMPv2 operationId suffix (PascalCase).
+func capitalize(s string) string {
```
solace-aross: `capitalize` is duplicated verbatim in `queueaction/handler.go` and here. Fine for two copies; if a third SEMPv2 native handler lands, lift it into a shared `internal/tools/sempv2` helper before the drift sets in. Optional / follow-up.

internal/tools/sempv2/queueaction/handler.go
```diff
@@ -0,0 +1,205 @@
...
+					"enum": []string{actionDeleteMsgs, actionClearStats},
+					"description": "The action to execute. deleteMsgs deletes " +
+						"all spooled messages (irreversible); clearStats " +
+						"resets statistics counters (non-destructive).",
+				},
+			},
+			"required": []string{"msgVpnName", "queueName", "action"},
+		},
+		OutputSchema: outputSchema(),
+		Annotations: tools.Annotations{
```
solace-awaheed: Ran this PR through Claude and there was a question of if per action tools are better than per category: 

1. Are tool level annotations the way to go? Current state (tool-level annotation): Destructive: true and Idempotent: false are pinned at the tool level, so they apply to every invocation regardless of the action parameter. Result: clearStats invocations are mis-annotated as destructive + non-idempotent, and the manager fires a destructive WARN log on every clearStats call. The WARN line doesn't include the action value

2. Descriptions stop forking. "BEFORE INVOKING THIS TOOL, obtain confirmation … for deleteMsgs, name the irreversible message loss explicitly" is conditional logic the LLM has to evaluate; "BEFORE INVOKING THIS TOOL, obtain confirmation — this permanently deletes all spooled messages" is unconditional. The Mode-2 adversarial bypass on Prompt 3 was exactly a fork-misread.

docs/user-guide.md
```diff
@@ -150,6 +150,15 @@ The server exposes 17 read-only tools. All broker-querying tools require a `brok
...
 
+### Actions
+
+These tools modify broker state via the SEMPv2 action API. They are marked destructive via the MCP `destructiveHint` annotation, and their descriptions instruct the calling LLM to obtain explicit user confirmation — restating the target (broker, VPN, queue or client) and the effect — before invocation. The tool manager also logs a WARNING line on every destructive invocation for audit.
+
+| Tool | Description |
+|---|---|
+| `execute-queue-action` | Execute an action on a queue: `deleteMsgs` (irreversible — permanently deletes all spooled messages) or `clearStats` (resets statistics counters; non-destructive). Use after operator-confirmed intent to drain a queue or reset stats during testing. |
+| `execute-client-action` | Execute an action on a connected client: `disconnect` (service-impacting — terminates the session; client must reconnect) or `clearStats` (resets per-connection statistics; non-destructive). Common use: disconnect a slow subscriber identified via `list-slow-subscribers` or `get-client-details`. |
```
solace-wmaryam: It's not clear here whether `disconnect` is also considered a destructive operation. It would be good to explicitly tabulate which tool type is destructive/non-destructive.

internal/tools/sempv2/clientaction/handler.go
solace-wmaryam: Seconding this. I think it would be worth following up as part of this commit.

solace-wmaryam:
@solace-lootii I've been going through your PR in preparation for the work I intend to do for the management tools: https://sol-jira.atlassian.net/browse/SOL-148460. I was wondering what the rationale was for choosing `execute-queue-action` and `execute-client-action` as the tool names. My story involves implementing tools to manage vpns and queues but I'm concerned about the inconsistency in naming between these tools and the ones in your story. Having both `execute-queue-action` and `manage-queue` would be confusing to the end-user. Alternatively, I can lump mine under the execute action tools but that leaves us with a very general tool name. We can discuss this in more detail once you're back from vacation.

amitmorade:
The comments above cover the concerns I had — once those are addressed it looks good to me. One small thing: branch seems behind main (README diff rolls back a few things like mcp_client_auth and the tool count), probably just needs a rebase before merge.

internal/tools/sempv2/clientactions/handler.go
```diff
@@ -0,0 +1,212 @@
...
+		return s
+	}
+	b := []byte(s)
+	if b[0] >= 'a' && b[0] <= 'z' {
+		b[0] -= 32
+	}
+	return string(b)
+}
+
+func ptr(b bool) *bool { return &b }
```
helen-zhu-solace, line 142: This is duplicated as well in `queueactions/handler.go:148`, can fold at same time

internal/tools/register_test.go
```diff
@@ -109,96 +110,63 @@ func TestRegisterWithServer(t *testing.T) {
...
 	// that AddTool was called without error.
 }
 
-// TestRegisteredHandlerPanicReturnsErrorResult reproduces SOL-150685: the SDK
-// invokes tool handlers on its own goroutine with no recover(), so a panic in
-// any handler kills the whole server process. The test drives a panicking
-// handler through a real client session over the in-memory transport and
-// expects a sanitized IsError result back — and the server still answering a
-// follow-up call — instead of a crash.
-func TestRegisteredHandlerPanicReturnsErrorResult(t *testing.T) {
```
helen-zhu-solace, line 125: `TestRegisteredHandlerPanicReturnsErrorResult` removal — `register_test.go`

The surviving `TestPanicAuditedAsError` checks the audit log line but discards the response. Two assertions from the deleted test aren't covered anywhere now:

- `res.IsError == true` — that the client actually sees an error result, not a success.
- The raw panic string doesn't appear in `res.Content` — that panic detail can't leak into agent-facing text.

Both can be added to the existing test in a few lines by keeping the `res` from `CallTool` instead of discarding it. Low priority — narrow regression shapes — but cheap to close.

solace-aross (dismissed):
Approving. Excellent work, Lootii. This is the first write path in the server and it shows real care end to end: secure-by-default gating at registration, accurate per-tool annotations, strict output schemas, defense-in-depth guards, and an honest disclosure of the eval-matrix gap. You took the review thread seriously and the design is stronger for it. Splitting into one tool per action resolved the annotation and confirmation-fork problems cleanly.

Everything the reviewers raised is addressed in the current code. A few small things to land or track, none a correctness issue:

- One pre-merge to-do: rebase on main. The branch is behind, so the README diff rolls back `mcp_client_auth` and the tool count (amitmorade's note).
- The `capitalize`/`run`/schema scaffolding is duplicated across `queueactions` and `clientactions`. wmaryam, Helen, and I flagged the `capitalize` copy; the fold is a bit broader than that one function, so a small shared helper folds it all in one move. Fine as a same-commit tidy-up or a fast follow.
- Low-severity heads-up on `register.go`: the write-tool gate correctly fails closed, but a future read-only composite tool that forgets `readOnly: true` would silently drop from `tools/list` with only a misleading "write tool not registered" log. Worth a one-line note in `tools.yaml`. Not a blocker.

Thanks for the thoroughness, especially the Mode-2 adversarial eval. Enjoy the vacation.

internal/tools/sempv2/queueactions/integration_test.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
```
amitmorade: I have a question about the integration_test.go files. They're guarded by the `integration` build tag and env vars, which means:

- CI skips them (line 74 doesn't include -tags=integration)
- Local `go test ./...` skips them silently
- They only run if someone manually sets env vars and invokes them with the flag

So right now they're providing zero validation — on every PR, in CI, or in normal local test runs. What's the intended use case? Are they meant to be run somewhere, or are they aspirational/WIP?

If they're meant to validate the tools, they should run in CI. They hit a live external broker, so they're e2e tests — I'd suggest moving them to test/e2e-basic-mcp/ where the harness handles broker setup and they'll actually run on every PR.

amitmorade (changes requested):

internal/tools/sempv2/queueactions/integration_test.go
amitmorade: This comment relates to both the integration_test.go files introduced in this PR

solace-lootii: Agreed, these run nowhere (no -tags=integration in CI, and the build tag keeps them out of go build/vet too), I've removed them here and the real wire-level coverage is already scoped as [SOL-150455](https://sol-jira.atlassian.net/browse/SOL-150455), which reuses the e2e-monitoring scaffold and is gated on this PR merging. This work will be done next

amitmorade (approved):
LGTM

PR #89, opened 2026-06-10
SOL-150660: Cap inbound /mcp request body at 4 MiB

solace-aross:
## Summary

Closes a memory-exhaustion vector on the `/mcp` endpoint ([SOL-150660](https://sol-jira.atlassian.net/browse/SOL-150660)).

The MCP go-sdk's `StreamableHTTPHandler` buffers the entire POST body in memory with `io.ReadAll` (streamable.go:427, :1098). The server never bounded the inbound body — `ReadTimeout` (30s) bounds duration, not bytes — so a client could stream a multi-GB JSON-RPC POST and OOM the server. The outbound direction was already hardened (`MaxSEMPResponseBytes` = 16 MiB); this adds the inbound counterpart.

## Changes

- `limitRequestBody(next, maxBytes)` middleware wraps the authed `/mcp` handler; production wires `defaults.MaxMCPRequestBytes` (4 MiB — MCP tool calls are KB-scale, so ~1000x headroom).
- A request that declares an oversized `Content-Length` is rejected with 413 before the SDK runs. Bodies without a declared length (or that lie) are bounded by `http.MaxBytesReader`; the SDK surfaces the failed read to the client as 400 "failed to read body".
- Root-anchors the `server` .gitignore pattern, which was matching the `cmd/server/` source directory instead of just the built binary (blocked adding the test file).

## Testing

- Mimic-handler tests prove the security property: oversized body fails the downstream read with `*http.MaxBytesError` and is never fully buffered; under-limit body passes through intact; declared-oversize never reaches the handler.
- Real-SDK tests (review feedback from @amitmorade) run an actual `StreamableHTTPHandler` behind the middleware at a 1 KiB cap and pin the client-observable contract: 200 for a normal initialize, 400 "failed to read body" for streamed oversize, 413 for declared oversize. An SDK upgrade that changes body buffering fails these tests loudly.
- Full `go test ./...` green; `/check-logs` clean (no log calls added).

[SOL-150660]: https://sol-jira.atlassian.net/browse/SOL-150660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/server/body_limit_test.go
+++ b/cmd/server/body_limit_test.go
@@ -1,0 +1,92 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/SolaceDev/solace-broker-mcp/internal/defaults"
+)
+
+// readAllHandler mimics the MCP SDK's StreamableHTTPHandler, which buffers
+// the entire request body with io.ReadAll before processing. It records the
+// read outcome so tests can assert what the SDK would observe.
+type readAllHandler struct {
+	readErr error
+	read    int
+}
+
+func (h *readAllHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	body, err := io.ReadAll(r.Body)
+	h.read = len(body)
+	h.readErr = err
+	if err != nil {
+		w.WriteHeader(http.StatusRequestEntityTooLarge)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+func TestLimitRequestBody_UnderLimitPassesThrough(t *testing.T) {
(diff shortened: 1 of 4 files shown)


cmd/server/body_limit_test.go
```diff
@@ -0,0 +1,92 @@
...
+	readErr error
+	read    int
+}
+
+func (h *readAllHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	body, err := io.ReadAll(r.Body)
+	h.read = len(body)
+	h.readErr = err
+	if err != nil {
+		w.WriteHeader(http.StatusRequestEntityTooLarge)
```
amitmorade: The test handler writes its own 413, so we're not verifying what the MCP SDK does when io.ReadAll returns *http.MaxBytesError — the HTTP status and body a real client would observe.

If limitRequestBody took the cap as a parameter, a test could set it to 1 KB and exercise the real SDK handler with a small body.

Would love to know what you think.

solace-aross: Fair point, and it goes further than the test shape. I checked the SDK source (streamable.go, go-sdk v1.5.0): on a ReadAll failure both POST paths return `http.Error(w, "failed to read body", http.StatusBadRequest)`. So the 413 my mimic handler wrote asserted a status the real stack never produces.

Took your suggestion in 4546ffd: `limitRequestBody` takes the cap as a parameter, and new tests run a real `StreamableHTTPHandler` behind it at a 1 KiB cap. They pin the client-observable contract: 200 for a normal initialize, 400 "failed to read body" when `MaxBytesReader` trips inside the SDK read. If an SDK upgrade changes how it buffers bodies, these tests fail loudly instead of silently weakening the mitigation.

Your comment also surfaced a UX gap: a client that declares its oversized Content-Length deserves a 413, not a generic 400. The middleware now rejects those up front with 413; `MaxBytesReader` still covers undeclared or lying bodies. Ready for another look.

amitmorade (approved):
LGTM

PR #91, opened 2026-06-10
SOL-150662: Contain tool handler panics to the failing call

solace-aross:
## Summary

Adds the missing panic containment layer on the tool-call path ([SOL-150662](https://sol-jira.atlassian.net/browse/SOL-150662)).

The MCP go-sdk v1.5.0 dispatches each JSON-RPC request on a bare goroutine (internal/jsonrpc2/conn.go:715-719) with no recover, and there was no `recover()` anywhere on our request path. A single latent nil-deref or map-write regression in any tool would crash the entire long-lived multi-session server instead of failing one tool call. The risk is not hypothetical: `identity.go`'s doc comment records that an earlier shipped version panicked on a verifier contract violation.

## Changes

- New `withPanicRecovery(toolName, handler)` middleware: logs the panic value and stack server-side, returns a generic `internal error in tool "<name>"` to the agent (no internal detail leaks to the client).
- Applied at both `AddTool` sites in `register.go` (`RegisterWithServer` and `RegisterListBrokers`) — the single translation boundary, so every tool is covered.

## Testing

- New tests: a nil-map-write panic is recovered into an error naming the tool without echoing panic internals; normal results and errors pass through unchanged. Panic test fails (crashes) without the wrapper.
- Full `go test ./...` green; `/check-logs` reviewed (panic/stack in structured server-side fields, generic agent-facing error).

[SOL-150662]: https://sol-jira.atlassian.net/browse/SOL-150662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/tools/recover.go
+++ b/internal/tools/recover.go
@@ -1,0 +1,49 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tools
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"runtime/debug"
+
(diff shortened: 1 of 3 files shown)


solace-aross:
Closing in favour of #96, which fixes the same defect (duplicate tickets: SOL-150662 here, SOL-150685 there). #96 returns a sanitized in-band IsError tool result rather than a protocol-level error, which is the idiomatic MCP shape for tool execution failures and matches the error translation pattern from SOL-148434. It also carries an integration test that drives a panicking handler through a real client session and verifies the server keeps serving. SOL-150662 is closed as Won't Do with the same rationale.

PR #97, opened 2026-06-10
fix(semp): enforce MaxConcurrentPerBroker with shared in-flight semaphore [SOL-150116]

solace-aross:
## Summary
`semp.max_concurrent_per_broker` now actually bounds in-flight SEMP requests: `Sender.Do` acquires a slot from a per-broker semaphore before sending and releases it when the request completes.

## Why
`semp.max_concurrent_per_broker` is documented as the per-broker concurrency cap. Before #94 it only sized the transport's idle-connection pools, so in-flight requests were unbounded. #94 set `MaxConnsPerHost = MaxConcurrentPerBroker`, which bounds in-flight per protocol client at N (HTTP/1.1, one request per connection), making the per-broker worst case 2N across SEMPv1+SEMPv2. This PR makes the documented cap the enforced bound: one shared semaphore per broker admits at most N in-flight requests across both protocol clients. The semaphore is explicit where the transport bound is incidental (it depends on HTTP/2 staying disabled), and its `len()` feeds the SOL-149791 broker-pool in-flight gauge (capability B11.1).

## Changes
- `internal/semp/resilience/semaphore.go` (new) — `Semaphore` type (`chan struct{}`); buffered channel over `x/sync` `Weighted` so `len()` can feed the SOL-149791 in-flight gauge later. `NewSemaphore` guards `n < 1` to the default (defense in depth; config validates 1..1024).
- `internal/semp/resilience/sender.go` — `Do` acquires after the rate-limiter tick (context-aware), releases in a defer; retryablehttp's internal retries hold the slot, since the request is still in flight against the broker. `New` gains a required `sem` parameter and panics on nil: a Sender-private fallback would silently recreate the per-protocol-client 2× shape.
- `internal/semp/broker.go` — `NewBrokerClient` creates **one** semaphore per broker and passes it to both protocol clients. This is the load-bearing decision: each broker has two Senders (SEMPv1 + SEMPv2), so a per-Sender semaphore would have allowed 2× the configured cap.
- `internal/semp/sempv{1,2}/client.go` — `NewHTTPClient` forwards the shared semaphore.
- `internal/defaults/defaults.go`, `internal/semp/resilience/transport.go` — refreshed #94-era comments that described the transport as the enforcement point ("there is no separate semaphore", "2× this value"); the semaphore is now the enforcing bound and the transport limit is pool sizing.

## Testing
- `TestSenderDo_SharedSemaphoreEnforcesPerBrokerCap` — two Senders sharing a cap-3 semaphore, 12 server-blocked requests: exactly 3 reach the server, no 4th admitted until release. Red-green and revert-verified (fails with "in-flight requests exceeded the cap of 3" without the enforcement hunk).
- `TestSenderDo_SemaphoreWaitRespectsContextCancel` — a request waiting for a slot honors context cancellation.
- `TestNew_PanicsOnNilSemaphore` — pins the required-`sem` constructor contract.
- Full `go build`, `go vet`, `go test -race ./...` green. `/check-logs` clean (no new log calls).

## Review Focus
- Acquire placement in `internal/semp/resilience/sender.go` (after rate-limiter tick, before send) and slot-held-across-retries semantics.
- Shared-semaphore wiring in `internal/semp/broker.go:33`.

## Compatibility
`resilience.New` and both `NewHTTPClient`s gain a parameter (internal packages, no external API change). No config or behavior change for operators beyond the cap now being enforced as documented.

## Related
- #94 (SOL-150665, merged) bounds connections via `MaxConnsPerHost`; that transport sizing stays per the ticket. The comment rebase predicted there landed here (`defaults.go`, `transport.go`).
- Saturation observability (gauge/logging on acquire-wait) is intentionally out of scope — belongs to SOL-149791.

## Jira
[SOL-150116](https://sol-jira.atlassian.net/browse/SOL-150116)

[SOL-150116]: https://sol-jira.atlassian.net/browse/SOL-150116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/semp/resilience/sender_test.go
```diff
@@ -771,3 +772,158 @@ func TestSender_NoRetry_400(t *testing.T) {
...
+		go func() {
+			defer wg.Done()
+			resp, err := sender.Do(context.Background(), req)
+			if err != nil {
+				errs <- err
+				return
+			}
+			resp.Body.Close()
+		}()
+	}
```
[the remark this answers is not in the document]
solace-aross: Not applicable here. sender and req are declared inside the loop body (fresh variables per iteration), not loop variables captured by the goroutine; the only loop variable is i, which the goroutine does not reference. Go 1.22+ per-iteration loop variable semantics apply as well (go.mod targets 1.25). Verified with go test -race -run TestSenderDo_SharedSemaphoreEnforcesPerBrokerCap -count=3: passes with no race reports.

amitmorade:
One small thing I noticed while reading through — feel free to push back if I'm reading it wrong.

In `internal/defaults/defaults.go`, the doc comment on `DefaultMaxConcurrentPerBroker` (added by #94) currently says:

> *"Each broker has one transport per protocol client, so the worst-case in-flight bound per broker is 2× this value."*

Once this PR's shared semaphore lands, I think the sem becomes the binding per-broker cap and the 2× caveat no longer applies — the worst case becomes N, not 2N.

Might be worth a small update in this PR so the comment doesn't go stale the moment it merges?

amitmorade:
Thanks for the detailed write-up — it helped me trace through what's happening.

Wanted to align my mental model with how this fits alongside #94, which just merged. My current understanding:

- After #94, `MaxConnsPerHost = MaxConcurrentPerBroker` already bounds in-flight at the transport, per protocol client.
- What this PR adds on top is: (a) tightening the per-broker worst case from 2N down to N by sharing one sem across SEMPv1+v2, and (b) picking a sem type (`chan struct{}`) whose `len()` will be readable by the SOL-149791 gauge later — no gauge code in this PR itself.

One question: **is 10 a strict ceiling where 20 (the 2N worst case after #94) is genuinely a problem for the broker management plane?** Or is the bigger driver here the observability prep for SOL-149791, with the 2N→N tightening being a useful side effect?

Either way, I think the description should acknowledge #94's role. Reading "in-flight requests per broker were unbounded" on its own, I had to go back to #94 to reconcile it — a quick mention of where #94 fits in would save the next reviewer that step.

internal/semp/resilience/sender.go
```diff
@@ -57,10 +58,20 @@ type Sender struct {
...
 // sempCfg.Retries and sempCfg.RequestMinInterval must be non-nil.
-func New(httpClient *http.Client, cookieJar *SafeCookieJar, sempCfg *config.SEMPConfig, authCfg config.AuthConfig, brokerURL string) *Sender {
+//
+// sem bounds in-flight requests and is expected to be shared by both protocol
+// Senders of the same broker so the cap is per-broker (see semp.NewBrokerClient).
+// A nil sem falls back to a Sender-private semaphore sized from
+// sempCfg.MaxConcurrentPerBroker — bounded, but per protocol client rather
+// than per broker.
+func New(httpClient *http.Client, cookieJar *SafeCookieJar, sempCfg *config.SEMPConfig, authCfg config.AuthConfig, brokerURL string, sem Semaphore) *Sender {
+	if sem == nil {
```
amitmorade, line 71: One thing I want to surface on the `nil sem` fallback.

Looking at the test diff, this fallback exists so the 8 existing test sites can keep working with just `, nil` appended — which makes sense as a way to keep the diff small.

The tension: the PR description calls the shared-per-broker sem "the load-bearing decision" because anything per-Sender allows 2× the cap. This fallback creates exactly that per-Sender shape. So a future caller constructing a Sender outside `BrokerClient` would silently get the 2N behavior again, with nothing in the API surface flagging that they did.

Suggestion: make `sem` required and have tests build one through a small helper. That turns the load-bearing invariant into a guarantee instead of a convention. Adds a bit of test churn here but removes the ambiguity permanently. Curious what you think.

solace-aross: Agreed, the fallback undercut the exact invariant this PR exists to create. Fixed in 21bc129: `sem` is now required and `New` panics on nil with a pointer to `NewBrokerClient`. Tests build semaphores explicitly, and `TestNew_PanicsOnNilSemaphore` pins the contract. The panic is unreachable through production wiring (both protocol clients are constructed by `NewBrokerClient`), and worst case it lands inside the SOL-150685 panic containment.

solace-aross:
Fair call, and it was bigger than that one comment: `transport.go` still said "this is what protects the broker management plane, there is no separate semaphore", which would have been flatly wrong the moment this merged. Both are rewritten in 21bc129: the shared semaphore is the enforcing per-broker bound; `MaxConnsPerHost` is pool sizing from the same value.

solace-aross:
Your mental model matches mine, with one addition. Direct answer first: no, 2N at the default is not a known broker problem; the KA range is a sizing heuristic, not a hard ceiling. Three drivers, in order:

1. Contract honesty. The config key is documented as the per-broker cap. An operator who sets 2 for a fragile appliance gets 4 today. The knob should mean what it says at every value, not just the default.
2. The transport bound is incidental. `MaxConnsPerHost` bounds in-flight only because our custom `TLSClientConfig` disables Go's automatic HTTP/2, making it one request per connection. If h2 ever gets enabled, one connection multiplexes many streams and that bound evaporates silently. The semaphore survives transport evolution.
3. Observability. Requests queueing inside the transport for a connection slot are invisible; the semaphore's `len()` is the admission point the SOL-149791 gauge will read.

Agreed on the description. I've rewritten the Why section to position against #94: unbounded before #94, 2N per broker after, N enforced and observable here.

amitmorade (approved):
LGTM

PR #100, opened 2026-06-11
SOL-150689: Fix flaky e2e-monitoring F4 txMsgRate assertion (get-message-rates)

helen-zhu-solace:
## Summary

Fix flaky `e2e-monitoring` F4 `txMsgRate` assertion in `get-message-rates`. Splits rx/tx thresholds and polls 5× for peak instead of asserting a single SEMP sample.

Fixes [SOL-150715](https://sol-jira.atlassian.net/browse/SOL-150715).

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The F4 sustained-traffic check `txMsgRate ≥ 80` flakes on unrelated branches (`wmaryam/SOL-148434`, `hzhu/sol-150248`). Observed `txMsgRate` values: **57, 68, 78, 88** — straddling the 80 line. `rxMsgRate` (~1150) has huge headroom and never flakes.

Root cause: `get-message-rates` returns a single short-window SEMP-averaged sample. `txMsgRate` (delivery to the F3 receiver) is inherently lower and noisier than the publish-side aggregate `rxMsgRate`, and was not empirically characterized in SOL-150024 — only `rxMsgRate` was.

## Changes Made

- Split F4 thresholds: `F4_RX_THRESHOLD=80` (unchanged behaviour), `F4_TX_THRESHOLD=40` (covers observed 57 floor with margin).
- Replace single-read assertion with **peak-of-5 polling** (5 samples × 1 s ≈ 5 s window) in both `test/e2e-monitoring/tool-tests.sh` and `test/e2e-monitoring/verify-fixtures.sh`.
- Log line now includes all samples and computed peaks for diagnosability.
- README updated: F4 fixtures-table row and "F4 — message rates" section now document the empirical 57–88 tx band, the lower tx threshold, and the polling rationale (cites SOL-150715).

## Testing

**Test Environment:**
- Go version: N/A (bash test-suite change only)
- OS: Linux
- Broker version (if applicable): same as e2e-monitoring suite

**Tests Performed:**
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated (if applicable)
- [x] E2E tests added/updated (if applicable)
- [ ] Tested locally with `go test -race ./...`
[...]

PR #101, opened 2026-06-12
SOL-150749: Re-arm F6 slowSubscriber flag before Tool 10 assertions (fix flaky e2e-monitoring)

solace-aross:
## Summary
Fixes the flaky `e2e-monitoring` Tool 10 (`list-slow-subscribers`) assertions by re-polling the F6 `slowSubscriber` flag immediately before each assertion, using the existing `wait_for_slow_subscriber` helper.

Fixes [SOL-150749](https://sol-jira.atlassian.net/browse/SOL-150749)

## Root cause
The F6 fixture SIGSTOPs a direct subscriber under a 3000 msg/s flood; the broker flips the per-client `slowSubscriber` flag on TCP egress back-pressure, computed over a rolling ~1 min window (the fixture comment and README both note this, and it is why `wait_for_slow_subscriber` polls instead of reading once). Tool 10, however, read the flag once.

On a healthy runner the gap between the verify-fixtures F6 check and Tool 10 is ~26 s. On 2026-06-12 two slow runners stretched it to ~63 s and the flag decayed on one broker in each run:

| Run | Branch | Failure |
| --- | --- | --- |
| [27424785855](https://github.com/SolaceDev/solace-broker-mcp/actions/runs/27424785855) | main | broker-a F6 client absent |
| [27425041949](https://github.com/SolaceDev/solace-broker-mcp/actions/runs/27425041949) | PR #97 | broker-b F6 client absent, pagination got 0 |

In the PR #97 run the flag was confirmed `true` at 15:24:22 and gone at 15:25:25, 63 s later, while the flood was still publishing (F4 rx ≈ 1500/s on both brokers seconds earlier). The ~15 s of `get-message-rates` sampling added by #100 sits in exactly this gap, which is why the flake surfaced today. Same class of bug as SOL-150715: a single read of a noisy, windowed broker metric.

## Changes Made
- `test_list_slow_subscribers` and `test_list_slow_subscribers_pagination` call `wait_for_slow_subscriber` (broker-direct SEMP poll, up to `F6_FLAG_TIMEOUT`) before invoking the MCP tool. The poll returns in 0 s when the flag is already set, so green-path runtime is unchanged.
- Wrappers pass the broker URL and F6 client name through.

If a future run shows the poll timing out (flag never returns once decayed), the escalation is to SIGCONT/SIGSTOP-cycle the subscriber to regenerate back-pressure events; the poll's per-attempt logging will make that failure mode unambiguous.

## Testing
- `bash -n` passes; change is confined to the e2e test scripts (no Go code).
- The full e2e-monitoring job runs in CI on this PR; that is the real verification.

[SOL-150749]: https://sol-jira.atlassian.net/browse/SOL-150749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/e2e-monitoring/tool-tests.sh
+++ b/test/e2e-monitoring/tool-tests.sh
@@ -447,10 +447,18 @@ test_get_queue_metrics_slow_consumer_b() { test_get_queue_metrics_slow_consumer 
 # entry — it confirms the cap is honored but, like list-rdps, cannot demonstrate
 # multi-page truncation (no second slow subscriber exists to drop).
 # Envelope: {"slowSubscribers":{"data":[...],"truncated":bool}}.
+#
+# Precondition re-arm: slowSubscriber is computed over a rolling ~1 min window
+# (see wait_for_slow_subscriber) and can drop back to false between the
+# verify-fixtures F6 check and this test — observed verified-true then gone
+# 63 s later on a slow runner (2026-06-12, runs 27424785855 / 27425041949).
+# Each test re-polls the flag via the broker before asserting on the tool;
+# the poll returns immediately when the flag is already set.
 
 test_list_slow_subscribers() {
-    local broker="$1" own="$2" other="$3"
+    local broker="$1" broker_url="$2" own="$3" other="$4"
     local response content
+    wait_for_slow_subscriber "$broker_url" "$broker" "$own" || return 1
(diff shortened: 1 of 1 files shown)


helen-zhu-solace (approved):
LGTM!

PR #105, opened 2026-06-15
Authenticator interface and SEMP protocol-client migration (SOL-150794, SOL-150795)

amitmorade:
## Summary

Behavior-preserving refactor that introduces an `auth.Authenticator` interface for broker-side authentication (Hop 2) and migrates both SEMP protocol clients to consume it. Foundation for the upcoming OAuth Hop 2 work (SOL-150070); ships only basic and bearer implementations.

Tickets: [SOL-150794](https://sol-jira.atlassian.net/browse/SOL-150794), [SOL-150795](https://sol-jira.atlassian.net/browse/SOL-150795)

## Type of Change

## Motivation and Context

Today, SEMP protocol clients call a package-level `auth.AddAuth(ctx, req, cfg)` function that switches on `cfg.Mode` for every request. Adding a third mode (OAuth) requires a per-broker runtime dependency (`*Exchanger`, cookie jar) that the package-level function cannot hold.

This PR introduces an `Authenticator` interface where each implementation holds the per-broker state it needs as fields, set once at construction. The protocol clients hold the interface; they never branch on auth mode. Adding OAuth in a follow-up becomes one new implementation plus one new case in the dispatcher — no protocol-client changes.

## Changes Made

**Production**
- New `auth.Authenticator` interface with `AddAuth(ctx, req) error`.
- New `BasicAuthenticator` and `BearerAuthenticator` implementations; fields set at construction, never mutated.
- New `auth.NewAuthenticator(cfg)` dispatcher.
- `semp.NewBrokerClient` is now the single builder: it constructs one Authenticator per broker and passes the same pointer to both protocol clients.
- `sempv1.HTTPClient` and `sempv2.HTTPClient` drop the `authCfg config.AuthConfig` field, gain an `authenticator auth.Authenticator` field, and call `c.authenticator.AddAuth(ctx, req)` at the call site.
- Protocol-client constructors panic on a nil Authenticator. Rationale: the failure is reachable only through programmer error (ignoring the error from `NewAuthenticator`, or passing nil explicitly). Returning an error would conflate this with genuine runtime failures and let a delayed nil dereference surface inside `AddAuth` at request time. Matches the existing pattern in `resilience.New`, which panics on a nil semaphore.
- Deleted the deprecated package-level `auth.AddAuth` function and its test file; no callers remain.
- Sender's existing `authCfg.Mode` 401-handling path is **untouched** — that refactor is deferred to a follow-up sub-ticket.

**Tests**
- New 16-goroutine immutability test on both Authenticator types under `-race`: struct fields are unchanged after concurrent calls, and every goroutine produces a correct header (base64 round-trip for basic, exact equality for bearer).
- New in-process integration test `TestCredentialsAreIsolatedPerBroker` in `test/integration/broker_credential_isolation_test.go`: three brokers with distinct credentials (mixed basic and bearer), asymmetric scheduled load (10/20/30 requests interleaved across goroutines), runs under `-race`. Asserts two invariants: no broker ever received a foreign Authorization header, and each broker received exactly its scheduled count.
- All pre-existing unit and race tests pass unchanged. Test setup adjusted where helpers used to take `AuthConfig`; helpers now take `auth.Authenticator`.

**Docs**
- `docs/oauth/token-exchange-SOL-150070/architecture-plan.md`: the upstream spec for SOL-150070 (13 decisions covering the full OAuth Hop 2 design). Included here so reviewers can see the "why" behind the abstraction shape.
- `docs/superpowers/plans/2026-06-14-sol-150070-implementation-decisions.md`: implementation-time decisions made during this PR — where the Authenticator is built (only in `NewBrokerClient`), what `NewHTTPClient` accepts (a pre-built Authenticator, not `AuthConfig`), why nil triggers panic instead of error, and the new `test/integration/` directory convention.

**New directory**
- `test/integration/` — in-process integration tier between unit tests (in `internal/`) and shell-driven E2E suites (under `test/e2e-*/`, `test/oauth/`). Runs as part of `go test ./...` under `-race`. Naming convention: one file per qualitatively distinct invariant. See `test/integration/README.md`.

## Testing

**Test Environment:**
- Go version: 1.24
- OS: macOS (darwin)
- Broker version: N/A locally; CI covers Solace broker E2E.

**Tests Performed:**
- [x] Integration tests added/updated (new in-process tier under `test/integration/`)
- [ ] E2E tests added/updated — none needed; CI runs the existing E2E suites unchanged
- [ ] Tested with real Solace broker — deferred to CI's E2E job

**Test Coverage:**
- Concurrent-safety of Authenticator types under `-race` (immutability invariant).
- Cross-broker credential isolation under concurrent multi-broker load with mixed basic/bearer modes (new integration test).
- All pre-existing unit and race tests pass unchanged, exercising the same protocol paths through the migrated abstraction.

## Breaking Changes

None. Behavior on the wire is identical: same Authorization headers from the same credentials, same retry/timeout posture, same config schema, same logs. No public API changes from `semp` (`pool.GetSEMPv1`, `pool.GetSEMPv2`, `BrokerClient.SEMPv1()`, `BrokerClient.SEMPv2()` are unchanged). A server running on this branch is interchangeable with one on `main`.

The `sempv1.NewHTTPClient` and `sempv2.NewHTTPClient` *signatures* gained an `auth.Authenticator` parameter, but these constructors are only called from `semp.NewBrokerClient` and tests in the same module — no downstream consumers.

### Code Quality
- [x] Code follows the project's coding standards
- [x] Self-review completed
- [x] Comments explain "why" (panic-vs-error reasoning, shared-pointer rationale, immutability invariant)

### Security
- [x] No credentials in code
- [x] `/check-logs` clean (no CRITICAL/HIGH violations)
- [x] Credential-carrying types implement `slog.LogValuer` (unchanged from main; protocol clients' LogValue already excludes credentials)
- [x] No new security vulnerabilities

### Testing
- [x] Edge cases covered (nil-Authenticator panic, concurrent access, multi-broker isolation, mixed auth modes)

### Documentation
- [x] `docs/` updated (architecture plan, implementation decisions)
- [ ] README.md unchanged (no user-facing behavior change)
- [ ] CHANGELOG.md — N/A (project does not maintain one yet)

### Git & CI
- [x] Branch is current with main
- [x] No merge conflicts
- [ ] CI checks — will run after this PR opens

## Additional Notes

**Sender still holds `authCfg`.** Per scope of SOL-150795, the resilience.Sender's 401 handling continues to read `authCfg.Mode` directly. Both consumers (the protocol clients' new Authenticator and the Sender's `authCfg`) read from `brokerCfg.Auth` at construction, so they agree by construction. A follow-up sub-ticket (T7b in the decomposition) will fold the Sender into the Authenticator abstraction.

**New integration test tier.** `test/integration/` is the in-process layer the project did not previously have. It runs as part of `go test ./...` (no build tag, no separate Make target) — the directory itself is the marker. The decisions doc explains why this layer was carved out now rather than waiting. A future cleanup may gate it behind `//go:build integration` once the suite is large enough to justify the friction.

## Related Issues/PRs

- Implements [SOL-150794](https://sol-jira.atlassian.net/browse/SOL-150794) and [SOL-150795](https://sol-jira.atlassian.net/browse/SOL-150795)
- Sub-tasks of parent epic [SOL-150070](https://sol-jira.atlassian.net/browse/SOL-150070) (OAuth Token Exchange Hop 2)
- Follow-ups (out of scope for this PR): SOL-150796 (T2 — config), SOL-150797+ (T3–T8)

---

**For Reviewers:**

**Focus Areas**:
1. **`NewBrokerClient` as the single builder.** Confirm that no other production code path constructs an Authenticator. (`grep` confirms; happy to discuss the rationale.)
2. **Panic vs error on nil Authenticator** in protocol-client constructors. The decisions doc covers the reasoning at length; happy to revisit if the team prefers an error return.
3. **The new `test/integration/` directory.** First test in a new tier — convention, naming rule, and rationale are in `test/integration/README.md` and the decisions doc. Would value input on whether this matches the team's expectations for an in-process integration layer.
4. **Sender's lingering `authCfg`.** Deliberately untouched per scope; will be addressed in the T7b follow-up. Worth confirming this scoping is acceptable.

**Questions**: None blocking. Marked draft to allow CI to settle before requesting review.

[SOL-150794]: https://sol-jira.atlassian.net/browse/SOL-150794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-150795]: https://sol-jira.atlassian.net/browse/SOL-150795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-150794]: https://sol-jira.atlassian.net/browse/SOL-150794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/semp/broker.go
```diff
@@ -16,31 +17,47 @@ import (
...
+// SEMPv1 and SEMPv2 receive the same pointer at construction and share it for
+// the broker's lifetime — see Decision 7 in
+// docs/oauth/token-exchange-SOL-150070/architecture-plan.md for the rationale.
 type BrokerClient struct {
-	sempV1Client *sempv1.HTTPClient // SEMPv1 protocol client (concrete for Close)
-	sempV2Client *sempv2.HTTPClient // SEMPv2 protocol client (concrete for Close)
-	alias        string             // broker alias (for error messages)
+	sempV1Client  *sempv1.HTTPClient // SEMPv1 protocol client (concrete for Close)
+	sempV2Client  *sempv2.HTTPClient // SEMPv2 protocol client (concrete for Close)
+	authenticator auth.Authenticator // one per broker, shared by both protocol clients
```
solace-aross: `authenticator` is set at construction but none of `BrokerClient`'s methods read it — both protocol clients already hold the same pointer directly. If this is pre-staging for the accessor that T7b (Sender 401 migration) will need, a brief field comment to that effect is enough: something like `// retained for T7b: Sender migration needs per-broker Authenticator access`. Without it, staticcheck will flag it as unused and future readers won't know why it's there.

internal/semp/sempv1/client_test.go
```diff
@@ -446,20 +439,14 @@ func TestExecute_InputValidation(t *testing.T) {
...
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))
 	defer srv.Close()
 
-	client := newTestClientWith(t, srv, config.AuthConfig{
-		Mode:     config.AuthModeBasic,
-		Username: secretUser,
-		Password: secretPass,
-		Token:    secretToken, // not used by basic mode, but set to prove it also doesn't leak
-	})
+	client := newTestClientWith(t, srv, auth.NewBasicAuthenticator(secretUser, secretPass))
```
solace-aross: `TestHTTPClient_LogValue_ExcludesCredentials` only constructs a `BasicAuthenticator`. Add a second sub-test with `auth.NewBearerAuthenticator("SECRET_TOKEN_VAL")` that logs the client the same way and asserts the token string is absent. The test name says "credentials" but only covers one type; if bearer were accidentally added to `LogValue`, this test wouldn't catch it. Copilot flagged this (suppressed); it's a real gap.

solace-aross (dismissed):
Clean, well-constructed foundation. The interface shape is minimal and correct, the concurrency model (structurally immutable types, nothing written after construction) is the right call, and `NewBrokerClient` as the single construction site prevents exactly the drift problem the PR motivates. The integration test covers the right failure mode — cross-broker credential leakage under concurrent load is exactly the bug that matters and is hard to catch any other way.

Two inline comments below for issues not already flagged. Also confirming: Copilot's three comments on `architecture-plan.md` are accurate — all three sections describe Authenticator construction happening in `pool.getOrCreate`, but the actual construction site is `NewBrokerClient`. Worth fixing those before merge; the architecture plan is the wiring map for T2–T8 and pointing at the wrong layer will mislead whoever implements them. The Copilot comment on `defer fb.mu.Unlock()` in the integration test assertion loop is also valid — copying `seen` under the lock and releasing before asserting is the safer pattern.

The scope boundary (Sender untouched for T7b) is the right call. Approved pending the doc fixes and the two inline comments.

internal/semp/broker.go
amitmorade: Kept the field and added a comment explaining it's pre-staging for Sender migration work — the field is now annotated to make the deferred use explicit, so staticcheck has the context and future readers understand the intent rather than guessing. Fixed in cd42431.

internal/semp/sempv1/client_test.go
amitmorade: Good catch — the test name promised more than it covered. Restructured TestHTTPClient_LogValue_ExcludesCredentials as a table-driven test with one row per auth mode (basic, bearer). The bearer row constructs a BearerAuthenticator with a known token string and asserts the token does not appear in the slog output. Extending to OAuth in future becomes one more row. Fixed in cd42431.

solace-aross (approved):
All four items from the prior review are done in cd42431:

- Architecture plan: six stale `pool.getOrCreate` references corrected to `NewBrokerClient`
- `BrokerClient.authenticator` field: comment explains the pre-staging for T7b Sender migration
- `TestHTTPClient_LogValue_ExcludesCredentials`: restructured as table-driven with a bearer row; the gap is closed
- Integration test assertion loop: lock correctly dropped after `wg.Wait()` with an explanatory comment

FOSSA checks pass. Good to merge.

PR #108, opened 2026-06-17
SOL-150878: Fix flakey slow-subscriber, clean up e2e-monitoring and gitignore

helen-zhu-solace:
## Summary

Self-heal the F6 slow-subscriber fixture in the `e2e-monitoring` suite when the broker reaps the SIGSTOPped client mid-suite, eliminating an intermittent `list-slow-subscribers` failure and the leaked-subscriber-process that accompanied it. Includes drive-by cleanup of `e2e-monitoring` scripts and `.gitignore` entries.

Fixes [SOL-150878](https://sol-jira.atlassian.net/browse/SOL-150878)

## Type of Change

- [ ] Documentation update
- [x] Refactoring (no functional changes) — see "Cleanup" below
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The `e2e-monitoring` job intermittently fails Tool 10 (`list-slow-subscribers`) on broker-a while broker-b passes with identical setup. Example failure: [run 27643027686](https://github.com/SolaceDev/solace-broker-mcp/actions/runs/27643027686/job/81748105826?pr=106).

**Root cause.** The F6 fixture creates a direct subscriber, SIGSTOPs it to force TCP back-pressure, and verifies `slowSubscriber=true`. The fixture is created once during setup and must survive ~2 minutes of unrelated Tool 1–9 runs before Tool 10 executes. During that window the broker can disconnect the SIGSTOPped client (keepalive timeout, egress buffer/discard threshold, etc.). When Tool 10 then polls SEMP for the client, the broker returns HTTP 400 (NOT_FOUND), `wait_for_slow_subscriber` logs empty observations for the full 60s budget, and the test fails.

**Secondary issue — leaked subscriber process.** When the broker disconnects the client, the local subscriber is SIGSTOPped and therefore cannot observe the disconnect or exit on its own. It sits as a frozen zombie holding RAM and a PID for the remainder of the suite, only reaped at teardown by `stop_broker_drivers` via the pidfile glob. Every retry of a failed CI run compounded this — on a self-hosted runner or a long iteration, the leaks accumulate. The fix in this PR also closes that leak (the respawn helper explicitly SIGCONT + SIGTERMs the stale PID before spawning a fresh one).

Timeline from the linked run:
- 19:39:23 — fixture created on broker-a, `slowSubscriber=true`
- 19:40:16 — F6 re-check: still alive, flag=true, txDiscardedMsgCount=66712
- 19:41:21 — Tool 10 first poll: HTTP 400 (client gone — local PID still frozen)
- 19:42:38 — 60s timeout → FAIL (and PID still leaking until suite teardown)

The existing caveat at `tool-tests.sh:451-456` already acknowledged this "verified-true then gone" race; this PR closes it.

## Changes Made

**Fix (the flake + the leak):**
- `test/e2e-monitoring/helpers.sh`
  - `wait_for_slow_subscriber` detects `http=400` during polling and respawns the subscriber once per call. A `recreated` flag prevents re-entry; the attempt counter resets so the new subscriber gets a full `F6_FLAG_TIMEOUT` budget.
  - New respawn_slow_subscriber_on helper: SIGCONT the still-SIGSTOPped local PID and call kill_gracefully to wait for it to actually exit (the broker tore down its end but the process is alive — this is the leak), then spawn a fresh subscriber directly onto the canonical pidfile path and SIGSTOP it. Waiting for the old PID to exit before reusing the path closes the race against the driver's defer os.Remove(*pidfile) — which would otherwise delete the new pidfile out from under the next kill -STOP. Best-effort /proc/$pid/exe check guards against signalling a recycled PID.
  - Flood publisher is intentionally **not** respawned — spawning a second one with the same client-name would let the broker evict the live one and kill the egress pressure.
  - Re-labeled the comment above the `curl` block: `http_status` is load-bearing for the self-heal branch, not just diagnostic.
- `test/e2e-monitoring/tool-tests.sh`
  - Replaced the stale "verified-true then gone" anecdote with a pointer to `respawn_slow_subscriber_on`.

**Cleanup (drive-by):**
- `.gitignore` — removed dead `!test/e2e/.env` and `!test/e2e/broker-config.yaml` exceptions. `test/e2e/` no longer exists; the suite now lives under `test/e2e-basic-mcp/` and `test/e2e-monitoring/`, so these lines were no-ops carrying false signal.
- `test/e2e-monitoring/start-server.sh` — install the EXIT/INT/TERM/HUP trap **before** `mktemp`/`build_server`/`wait_for_all_brokers` run, so a Ctrl-C or build failure still cleans up the temp config. Previously the trap was only installed after state was created, leaving a window where an abort leaked `/tmp/e2e-config-*.yaml`.
- `test/e2e-monitoring/verify-fixtures.sh` — minor signature alignment for the new `wait_for_slow_subscriber` arg.

## Testing

**Test Environment:**
- Go version: project pin (`go.mod`)
- OS: Linux (Fedora 43)
- Broker version: per `test/e2e-monitoring/docker-compose.yml`

**Tests Performed:**

**Test Coverage:**

Manual reproduction of the race against live Dockerized brokers:
1. `create_slow_subscriber_on` → fixture healthy, `slowSubscriber=true` after 5s.
2. `PUT __private_action__/.../clients/.../disconnect` → broker drops the client; GET returns HTTP 400. Local subscriber PID confirmed still alive in state `T` (stopped) — this is the leak.
3. `wait_for_slow_subscriber` observes `http=400`, logs `broker reaped it; respawning subscriber`, kills the old SIGSTOPped PID, spawns + SIGSTOPs a new subscriber, polls, flag flips to `true` after 6s, returns 0. Old PID confirmed gone via `kill -0`.

Re-entry guard verified by code inspection — wait_for_slow_subscriber only takes the respawn branch while recreated=0. A second HTTP 400 in the same call falls through to normal polling and times out via the existing log_fail rather than respawning again. (respawn_slow_subscriber_on does not call back into wait_for_slow_subscriber, so there's no recursion to guard.)

## Breaking Changes

None.

**Impact:** N/A

**Migration Guide:** N/A

### Code Quality

### Security
- [x] No credentials in code
- [x] Credential-carrying types implement `slog.LogValuer` (N/A — shell only)

### Testing
- [x] Edge cases covered by tests (respawn path exercised manually; recursion guard reviewed)
- [x] Error paths tested

### Documentation
- [ ] README.md updated (not user-facing)
- [ ] docs/ updated (no architecture/design change)
- [x] godoc comments added/updated for exported functions (N/A — shell only)
- [ ] CHANGELOG.md updated (test-infra-only fix)
- [ ] Examples updated (N/A)

### Git & CI
- [x] No merge conflicts
- [x] CI checks passing (lint, build, test, E2E) — pending push

- [ ] N/A

Excerpt from a local repro run showing the self-heal path:

[INFO]    poll [broker-a/e2e-monitoring-slow-subscriber-a] t=0s http=400 flag=<empty> ...
[INFO]    client e2e-monitoring-slow-subscriber-a absent on broker-a (http=400) — broker reaped it; respawning subscriber
[INFO]    SIGSTOP sent to respawned slow-subscriber (PID=2742343) on broker-a
[INFO]    poll [broker-a/e2e-monitoring-slow-subscriber-a] t=6s http=200 flag=true ...
[INFO]    slowSubscriber=true for e2e-monitoring-slow-subscriber-a on broker-a (6s)

## Additional Notes

Considered (and rejected) lowering `F6_FLOOD_RATE` / `F6_FLOOD_SIZE` to reduce the probability of a broker disconnect. That only mitigates one specific trigger (discard-driven), requires re-tuning per broker version, and slows the initial flag-flip on the happy path. The self-heal approach covers any cause of client absence (keepalive timeout, egress threshold, transient broker hiccup), reaps the leaked subscriber process at the moment it becomes stale, and only does extra work when a reap actually occurs.

The cleanup items are intentionally bundled rather than split — they're all in `test/e2e-monitoring/` and each is too small to be worth its own PR.

---

**For Reviewers:**

**Focus Areas**:
- `respawn_slow_subscriber_on` in `helpers.sh` — correctness of the wait-for-exit sequence on the old PID (SIGCONT to unblock pending SIGTERM, then kill_gracefully to wait, then reuse of the canonical pidfile path) The recursion guard pattern → The single-shot re-entry guard pattern
- The recursion guard pattern (`recreated` local, reset on respawn, `continue` instead of `return`).

[SOL-150878]: https://sol-jira.atlassian.net/browse/SOL-150878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/e2e-monitoring/helpers.sh
```diff
@@ -697,16 +697,25 @@ F6_FLAG_TIMEOUT=60     # max seconds to wait for slowSubscriber to flip true
 # Polls the broker until a client's slowSubscriber flag reads true, or fails
 # after F6_FLAG_TIMEOUT. The flag is computed over a rolling ~1 min window, so
 # a single read just after SIGSTOP flakes — poll until it settles true.
-#   $1 broker_url   $2 label   $3 client_name
+#
+# Self-healing: if the broker reports the client absent (HTTP 400 NOT_FOUND),
+# the SIGSTOPped subscriber was reaped by the broker (keepalive timeout or
+# egress threshold) — respawn it once and keep polling. The `recreated` flag
+# allows exactly one respawn per call so a second reap fails fast instead of
```
solace-wmaryam: Minor wording fix for the comment. The comment says that if the broker kills the subscriber a second time, the test "fails fast." It doesn't. After the second failure, the code goes back to its normal once-per-second polling and keeps trying for the full 60-second timeout before giving up. The real benefit of the guard is to stop the test from respawning the subscriber over and over in an endless loop. Consider rewording the comment to reflect this.

test/e2e-monitoring/helpers.sh
```diff
@@ -729,13 +738,86 @@ wait_for_slow_subscriber() {
...
+respawn_slow_subscriber_on() {
+    local label="$1" broker_letter="$2" client_name="$3"
+    local sub_pidfile="$BIN_DIR/broker-driver-f6-sub-$broker_letter.pid"
+    local sub_logfile="$BIN_DIR/broker-driver-f6-sub-$broker_letter.log"
+    log_info "  client $client_name absent on $label (http=400) — broker reaped it; respawning subscriber"
+    # The old subscriber is still alive locally (SIGSTOPped, just disconnected
+    # from the broker). Reap it BEFORE reusing the pidfile path: the driver
+    # has `defer os.Remove(*pidfile)` (see broker-driver/slow_subscriber.go),
+    # so a late-exiting old process would otherwise delete the new pidfile we
+    # rename into the canonical path.
```
solace-wmaryam: Minor nitpick, there's not really a renaming happening here - use the word write instead.

solace-wmaryam (dismissed):
LGTM...I've left some minor comments but they are in no way blocking.

solace-wmaryam (approved):

PR #109, opened 2026-06-17
SOL-150796: add Hop 2 OAuth config schema (mcp_client_auth rename)

amitmorade:
## Summary

Adds the YAML configuration schema the MCP server will use to obtain broker-bound access tokens from an Identity Provider via RFC 8693 token exchange — the "Hop 2" exchange in the OAuth flow where the MCP server, having already authenticated the agent (Hop 1), needs a broker-specific token from the IdP before calling SEMP. The runtime that performs the exchange ships in later sub-tickets; this PR only puts the schema in place, validated at startup, so the contract is fixed before runtime code is written against it.

Two changes operators see:

1. **New top-level `broker_oauth:` block** in `broker-config.yaml` — holds the IdP coordinates (token URL, the MCP server's `client_id`, its token-endpoint client-auth method, grant type, audience parameter name). Accepted by the loader, validated for shape, but the runtime is not wired yet: any broker configured with `auth.mode: oauth` is rejected at startup with a clear operator-facing banner and a per-broker validation error.
2. **`client_auth:` → `mcp_client_auth:` rename (BREAKING)** — the existing top-level Hop 1 client-auth block is renamed to disambiguate it from the new nested `broker_oauth.mcp_server_client_auth` sub-block. Every existing config must rename one key.

Tracked under [SOL-150796](https://sol-jira.atlassian.net/browse/SOL-150796), part of the [SOL-150070](https://sol-jira.atlassian.net/browse/SOL-150070) OAuth Token Exchange epic.

## Type of Change

- [x] New feature (non-breaking change which adds functionality) — `broker_oauth:` block
- [x] Breaking change — top-level `client_auth:` → `mcp_client_auth:` rename
- [x] Documentation update — decisions doc, CHANGELOG, sam-integration, e2e-testing

## Motivation and Context

The architecture plan for SOL-150070 has the MCP server obtaining broker-bound access tokens from the IdP via RFC 8693 token exchange (Hop 2). The runtime that performs that exchange lands in later sub-tickets; this PR puts the *schema* in place first so:

1. The shape of the OAuth config can be reviewed independently of the runtime that will consume it.
2. Subsequent sub-tickets can build the exchanger, the per-broker `Authenticator`, and the cookie-jar wiring against a fixed YAML contract — no schema churn during runtime work.
3. Operators who add an `auth.mode: oauth` broker before the runtime ships get a clear, actionable startup error rather than silent acceptance or a confusing per-request 401.

[...]

PR #110, opened 2026-06-18
SOL-150796: add OAuth broker config schema (mcp_client_auth rename, banner consolidation)

amitmorade:
## Summary

Adds the YAML configuration schema the MCP server will use to obtain broker-bound access tokens from an Identity Provider via OAuth token exchange — the outbound (server → IdP) hop in the OAuth flow that delivers a broker-scoped token. The runtime that performs the exchange ships in later sub-tickets; this PR puts the schema in place, validated at startup, so the contract is fixed before runtime code is written against it.

Three operator-visible things this PR ships:

1. **New top-level `broker_oauth:` block** in `broker-config.yaml` — holds the IdP coordinates (token endpoint, the MCP server's `client_id`, its token-endpoint client-auth method, grant type, audience parameter name). Accepted by the loader, validated for shape, but the runtime is not wired yet: any broker configured with `auth.mode: oauth` is rejected at startup with a clear operator-facing banner and a per-broker validation error.

2. **`client_auth:` → `mcp_client_auth:` rename (BREAKING)** — the existing top-level client-auth block is renamed to disambiguate it from the new nested `broker_oauth.mcp_server_client_auth` sub-block. Every existing config must rename one key.

3. **New `internal/banner` package** — operator-facing banners (Hop 1 startup signal, V1 OAuth-not-supported guard, Hop 1 / Hop 2 alignment) live in one canonical home rather than spread across `internal/auth` and `internal/config`.

Tracked under [SOL-150796](https://sol-jira.atlassian.net/browse/SOL-150796), part of the [SOL-150070](https://sol-jira.atlassian.net/browse/SOL-150070) OAuth Token Exchange epic.

## What the new config looks like

**Before (today's main):**

```yaml
client_auth:
  mode: oauth
  issuer: "https://idp.example.com"
  audience: "mcp-server"
  resource_url: "https://mcp.example.com/mcp"

brokers:
  prod-east:
    url: "https://broker.example.com:943"
    auth:
      mode: basic
      username: "${BROKER_USER}"
      password: "${BROKER_PASS}"
```

**After this PR — minimum operator change (the BREAKING rename):**

```diff
- client_auth:
+ mcp_client_auth:
    mode: oauth
    issuer: "https://idp.example.com"
    audience: "mcp-server"
    resource_url: "https://mcp.example.com/mcp"

  brokers:
    prod-east:
      url: "https://broker.example.com:943"
      auth:
        mode: basic
        username: "${BROKER_USER}"
        password: "${BROKER_PASS}"
```

That's the only required change for existing deployments. The new `broker_oauth:` block is opt-in.

**After this PR — opting into OAuth broker auth (schema only in V1):**

```yaml
mcp_client_auth:                          # client → MCP server (unchanged behavior; just renamed key)
  mode: oauth
  issuer: "https://idp.example.com"
  audience: "mcp-server"
  resource_url: "https://mcp.example.com/mcp"

broker_oauth:                             # NEW — IdP coordinates for broker token acquisition
  idp_token_endpoint: "https://idp.example.com/oauth/token"
  mcp_server_client_id: "mcp-server-prod"
  mcp_server_client_auth:                 # discriminated union — populate exactly ONE method sub-block
    client_secret_basic:                  # RFC 6749 §2.3 — secret in HTTP Basic header
      secret: "${MCP_CLIENT_SECRET}"
    # client_secret_post:                 # alternative; pick one or the other, not both
    #   secret: "${MCP_CLIENT_SECRET}"
  grant_type: "urn:ietf:params:oauth:grant-type:token-exchange"
  audience_parameter_name: "audience"     # one of: audience | scope | resource (per IdP convention)

brokers:
  prod-east:
    url: "https://broker.example.com:943"
    auth:
      mode: oauth                         # NEW per-broker mode (alongside basic, bearer)
      audience: "https://broker-prod-east.example.com"  # optional; per-broker
      scopes:                             # optional; per-broker
        - "semp:read"
        - "semp:write"
```

> **V1 behavior**: configs with any broker on `auth.mode: oauth` are rejected at startup with a clear banner — the schema is accepted for validation but the runtime that performs token exchange ships in a follow-up sub-ticket. Operators can stage the config and validate its shape today; they cannot run against it until the runtime lands.

## Type of Change

- [x] New feature (non-breaking change which adds functionality) — `broker_oauth:` block
- [x] Breaking change — top-level `client_auth:` → `mcp_client_auth:` rename
- [x] Refactoring (no functional changes) — banner subsystem consolidated into `internal/banner`
- [x] Documentation update — decisions doc, CHANGELOG, sam-integration, e2e-testing

## Motivation and Context

The architecture plan for SOL-150070 has the MCP server obtaining broker-bound access tokens from the IdP for SEMP requests (RFC 8693 token exchange). The runtime that performs that exchange lands in later sub-tickets; this PR puts the schema in place first so:

1. The shape of the OAuth config can be reviewed independently of the runtime that will consume it.
2. Subsequent sub-tickets can build the exchanger, the per-broker `Authenticator`, and the cookie-jar wiring against a fixed YAML contract — no schema churn during runtime work.
3. Operators who add an `auth.mode: oauth` broker before the runtime ships get a clear, actionable startup error rather than silent acceptance or a confusing per-request 401.

The original `client_auth:` block name was already overloaded in design discussions — operators reading a YAML with both inbound client auth and outbound (server → IdP) client auth couldn't tell which `client_auth:` belonged to which side. The rename to `mcp_client_auth:` fixes that ambiguity at the schema layer, before the nested block ships.

A naming principle (explicit/qualified names by default, reserve generic names for future use) drove the four `broker_oauth:` field names: `idp_token_endpoint`, `mcp_server_client_id`, `mcp_server_client_auth`, `audience_parameter_name`. Rationale lives in the decisions doc.

## Changes Made

Commit-by-commit, the branch is structured so each commit does one thing and is reviewable in isolation:

1. **[`97f6012`](https://github.com/SolaceDev/solace-broker-mcp/commit/97f6012)** — rename `client_auth` → `mcp_client_auth` (BREAKING). Pure rename: YAML tag, Go type `ClientAuthConfig` → `MCPClientAuthConfig`, field `ServerConfig.ClientAuth` → `MCPClientAuth`, error strings, comments, every inline-YAML test fixture, all E2E shell scripts, all operator-facing docs. No new functionality.

2. **[`1f34e6f`](https://github.com/SolaceDev/solace-broker-mcp/commit/1f34e6f)** — whitespace-trim hardening for basic/bearer credentials. `${VAR}` substitutions that resolve to whitespace-only strings are now rejected at startup with the "required for X auth" error rather than failing every SEMP request with a 401 at runtime. Matches the existing whitespace-validation on oauth scopes.

3. **[`a33aa5f`](https://github.com/SolaceDev/solace-broker-mcp/commit/a33aa5f)** — `broker_oauth` config schema (schema-only). `BrokerOAuthConfig` struct with the four qualified field names; `BrokerClientAuth` discriminated union (`client_secret_basic` / `client_secret_post` sub-blocks per IANA "OAuth Token Endpoint Authentication Methods" registry); all validators; `LogValue` implementations that exclude the secret; the V1 not-yet-supported guard (`oauthBrokerCount` counter, per-broker rejection error, banner emission). Decisions doc carries the full rationale.

4. **[`292dc70`](https://github.com/SolaceDev/solace-broker-mcp/commit/292dc70)** — banner subsystem consolidated into a new `internal/banner` package. Mechanical relocation: emitters take primitives instead of `*config.ServerConfig`, so the package has no internal dependencies and neither config nor auth hits an import cycle calling into it. End-to-end verified byte-identical banner output before vs. after.

5. **[`6358244`](https://github.com/SolaceDev/solace-broker-mcp/commit/6358244)** — Hop 1 / Hop 2 OAuth alignment invariant. New permanent banner + validator: if any broker uses `auth.mode: oauth`, `mcp_client_auth.mode` must also be `oauth`. The MCP server needs the agent's token (received via `mcp_client_auth`) to obtain a broker token, so client auth on the inbound side must be OAuth. Banner emission is gated behind the V1 guard so V1 operators see one headline; the validation error always lands in the joined error blob so the invariant is enforced and tested today.

6. **[`628ad49`](https://github.com/SolaceDev/solace-broker-mcp/commit/628ad49)** — CHANGELOG entries (Added: `broker_oauth`; Changed-BREAKING: `client_auth` rename; Changed: whitespace-trim).

## Testing

**Test Environment:**
- Go version: 1.24 (per `go.mod`)
- OS: macOS (Darwin 24.6.0); CI runs on Linux
- Broker version: PubSub+ Standard latest (E2E)

**Tests Performed:**
- [x] Unit tests added/updated — every new validator, the discriminated union, `LogValue` redaction, the V1 banner singular/plural/none cases, the alignment validator (called directly so the future-runtime path is exercised today), and the V1 alignment-banner suppression.
- [x] E2E tests updated — `test/e2e-basic-mcp/helpers.sh`, `test/e2e-monitoring/helpers.sh`, `test/health/test-health.sh` heredoc YAML updated for the `mcp_client_auth:` rename.
- [x] Tested locally with `go test -race ./...` — every package passes.
- [x] End-to-end startup verification against the built server binary across seven scenarios (happy path, BREAKING rename enforcement, V1 OAuth-not-supported guard, Hop 1 / Hop 2 mismatch with banner suppression, strict-fields enforcement on `broker_oauth`, discriminated-union rejection, whitespace-trim rejection). All seven behave as expected — no regressions.

**Test Coverage:**
- `TestLoadConfig_BrokerOAuth_*` — every required field, the discriminated-union "exactly one populated" rule, the grant-type allowlist, the audience-parameter-name allowlist, `${VAR}` substitution + post-substitution emptiness, production-mode HTTPS requirement.
- `TestLoadConfig_EmitsOAuthNotSupportedBanner` — singular vs plural wording when 1 / N / 0 brokers use `auth.mode: oauth`.
- `TestValidateHop1Hop2Alignment_Direct` — the alignment invariant exercised in isolation, including the cases that become primary when the V1 guard is removed.
- `TestLoadConfig_Hop1Hop2AlignmentError_AppearsInJoinedError` + `TestLoadConfig_Hop1Hop2AlignmentBanner_SuppressedInV1` — the V1 gating order.
- `TestLoadConfig_CredentialsTrimmedBeforeEmptyCheck` — whitespace-only credentials rejected.
- `TestBrokerOAuthConfig_LogValue` — slog never leaks the populated client-secret value.

## Breaking Changes

**Impact:** Every config file with a top-level `client_auth:` block must rename it to `mcp_client_auth:`. The server fails fast at startup with a precise "field client_auth not found" error if the rename is missed. No field semantics change inside the block.

**Migration Guide:**

Operator config — rename the key:

```diff
- client_auth:
+ mcp_client_auth:
    mode: static
    dev_token: "..."
```

Go API (internal, only used inside this module):
- Type `ClientAuthConfig` → `MCPClientAuthConfig`
- Field `ServerConfig.ClientAuth` → `MCPClientAuth`

## Additional Notes

**Schema-only, not feature-flagged — the validator IS the gate.** V1 ships the schema without the runtime, and there is no `if FeatureFlag.OAuthEnabled { ... }` toggle anywhere in the code. The mechanism is simpler: the config validator hard-rejects any broker with `auth.mode: oauth` at startup with an actionable error and a loud banner. That means:

- The server cannot start with any oauth-mode broker — there's no path past `validate()` for that config shape.
- No code beyond validation runs the oauth-mode path. Operators cannot accidentally land in an "oauth half-works" state.
- When the runtime ticket lands, the gate is removed by deleting one rejection branch in `validateBroker`, the per-broker `oauthBrokerCount` accumulator, and the V1 banner emission — all three flagged with a single `LIFECYCLE — REMOVE WHEN THE OAUTH RUNTIME LANDS` comment so the cleanup is mechanical.

A feature flag would have been the wrong tool here: it would imply the runtime exists and can be toggled on, when in fact the runtime is unimplemented. The validator-as-gate is honest about that — the config is accepted for *shape* validation (so operators can stage YAML against V1 today and trust it'll parse the same after T6 lands), but rejected for *use*. Rationale recorded in `docs/superpowers/plans/oauth-token-exchange/SOL-150796-T2-config-schema.md`.

**Operator-facing wording avoids internal vocabulary.** The new banners and validator errors use the YAML keys operators wrote (`mcp_client_auth.mode`, `auth.mode: oauth`) and the corrective action. Internal mental-model terms (Hop 1/Hop 2, RFC 8693, `subject_token`) stay in code comments and the decisions doc — same audience split as every other implementation detail.

**Lifecycle markers on the temporary code.** The V1 not-yet-supported guard (banner + counter + validate() call site) carries an explicit `LIFECYCLE — REMOVE WHEN THE OAUTH RUNTIME LANDS` comment naming the three pieces that get removed together. The Hop 1 / Hop 2 alignment banner carries `LIFECYCLE — KEEP` so future readers don't conflate the two.

## Related Issues/PRs

- Tracks [SOL-150796](https://sol-jira.atlassian.net/browse/SOL-150796) under the [SOL-150070](https://sol-jira.atlassian.net/browse/SOL-150070) OAuth Token Exchange epic
- Builds on the Authenticator refactor merged in [#105](https://github.com/SolaceDev/solace-broker-mcp/pull/105) ([SOL-150794](https://sol-jira.atlassian.net/browse/SOL-150794) / [SOL-150795](https://sol-jira.atlassian.net/browse/SOL-150795))
- Supersedes the closed [#109](https://github.com/SolaceDev/solace-broker-mcp/pull/109) — same end-state, restructured into a clean per-concern commit history

---

**For Reviewers:**

**Focus Areas:**
- The discriminated-union shape for `mcp_server_client_auth` — is the "populated sub-block selects the method" pattern clear from the YAML, or does it want an explicit `method:` discriminator key alongside the sub-block?
- Field naming under the explicit-names principle — the four names are spelled out in the decisions doc. If any of `idp_token_endpoint`, `mcp_server_client_id`, `mcp_server_client_auth`, or `audience_parameter_name` reads as overcorrection, now is the time to flag it (the names are operator-facing and a V2 change is breaking).
- The Hop 1 / Hop 2 alignment validator's V1 gating logic — the banner is suppressed in V1 but the validation error always lands in the joined blob. Does the test pair `TestLoadConfig_Hop1Hop2Alignment*` cover the cases you'd want covered before the runtime ticket starts landing against this?
- The `internal/banner` package boundary — keeping it free of internal dependencies means callers pass primitives. Does the `LogStartupAuthMode(mode, issuer)` signature shape feel right, or does it want a small struct argument?

**Questions:**
- Any concerns with shipping the alignment invariant before the runtime that primarily exercises it? The validator runs unconditionally and is tested directly, so the code path isn't dead — but if you'd rather see it land with T6, happy to defer.

[SOL-150796]: https://sol-jira.atlassian.net/browse/SOL-150796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amitmorade:
Looking at copilot comments

internal/config/config.go
```diff
@@ -728,21 +999,201 @@ func validateBroker(broker *BrokerConfig, productionMode bool) []error {
...
+	// the sub-block was set. Belt-and-suspenders: handle nil cleanly anyway.
+	if cfg == nil {
+		errs = append(errs, fmt.Errorf("broker_oauth.mcp_server_client_auth.%s: sub-block is empty", method))
+		return errs
+	}
+	if cfg.Secret == "" {
+		errs = append(errs, fmt.Errorf(
+			"broker_oauth.mcp_server_client_auth.%s.secret is required (empty after ${VAR} substitution if used)",
+			method))
+	}
```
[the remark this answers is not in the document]
solace-awaheed: Local review also caught this same issue

internal/config/config.go
```diff
@@ -728,21 +999,201 @@ func validateBroker(broker *BrokerConfig, productionMode bool) []error {
...
-		if broker.Auth.Token == "" {
+		// Trim before comparing — see basic-auth note above.
+		if strings.TrimSpace(broker.Auth.Token) == "" {
 			errs = append(errs, fmt.Errorf("broker %q: token is required for bearer auth", alias))
 		}
+	case AuthModeOAuth:
+		// Per-broker OAuth params. The global broker_oauth block (IdP
+		// coordinates) is validated separately in validateBrokerOAuthConfig.
+		//
+		// Audience is optional in V1: the broker's OAuth profile may have
```
solace-awaheed, line 1026: The scopes loop below trims and rejects whitespace-only entries, but broker.Auth.Audience itself isn't checked. A ${VAR} that resolves to "   " passes startup and ends up as the audience claim in every future token-exchange request.                                                                                                                                                                                                                 
                                               
Claude suggested fix, just before the scopes loop:                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                
```
                if broker.Auth.Audience != "" && strings.TrimSpace(broker.Auth.Audience) == "" {
                        errs = append(errs, fmt.Errorf("broker %q: auth.audience is empty or whitespace-only", alias))                                                                                                                                          
                }                                                                                                                                                                                                                                               
  
                for i, s := range broker.Auth.Scopes {      
```

amitmorade: On it, thanks!

amitmorade: fixed in [68fdf0f](https://github.com/SolaceDev/solace-broker-mcp/commit/68fdf0f) (force-pushed; the SHA in the PR's commit list is current). Applied the exact pattern you suggested:

if broker.Auth.Audience != "" && strings.TrimSpace(broker.Auth.Audience) == "" {
    errs = append(errs, fmt.Errorf("broker %q: auth.audience is empty or whitespace-only", alias))
}
right before the scopes loop, as you suggested. The != "" guard preserves the field's optional semantics — absent audience stays valid; only set-but-whitespace is rejected.

Pinned by a new test case in TestLoadConfig_WhitespaceTrimmedAcrossRequiredStrings so a future refactor can't collapse it to a bare TrimSpace == "" check (which would incorrectly reject absent-audience configs).

While I was in there, I also extended the same trim-before-empty rule to every other operator-supplied required string in the validator

internal/config/config.go
```diff
@@ -663,50 +852,137 @@ func validate(cfg *ServerConfig) error {
...
 
+	// Hop 2 OAuth IdP block. Runs after per-broker validation so any oauth-mode
+	// broker has already been examined; this validator inspects the global
+	// block and its relationship with the brokers' auth modes.
+	errs = append(errs, validateBrokerOAuthConfig(cfg)...)
+
+	// Hop 1 / Hop 2 alignment invariant: any broker with auth.mode: oauth
+	// requires mcp_client_auth.mode: oauth, because RFC 8693 token exchange
+	// consumes the agent's Hop 1 token as the subject_token. Without Hop 1
+	// OAuth there is no subject_token to exchange.
```
solace-aross: In V1, this config shape surfaces two errors with opposite remediation: `validateBroker` emits `auth.mode "oauth" is recognized but not yet supported ... use basic or bearer`, and this alignment check emits `mcp_client_auth.mode ... must be oauth`. An operator who follows the second (flip `mcp_client_auth.mode` to oauth) still hits the first.

The banner for this invariant is already suppressed in V1 for exactly this noise reason. Worth considering whether the alignment *error* should be gated the same way, surfacing only once the runtime lands and the not-yet-supported guard is removed. Low severity, the V1 banner carries the real signal, but it's the one spot where the operator-facing guidance points two directions for a feature that does not run yet. A conscious call either way is fine.

solace-aross (commented):
Reviewed the full diff, existing comments, and test coverage. Solid, merge-leaning PR. The correctness gaps already flagged (whitespace-only secret, whitespace-only audience) are the right catches; no need to restate them. One non-blocking inline note added on the V1 double-error remediation.

Answering the two questions in the PR description:

**Ship the alignment invariant before the runtime?** Yes. `TestValidateHop1Hop2Alignment_Direct` exercises it across all hop1 (oauth/static/disabled) x hop2 (present/absent, singular/plural) cases, and it runs unconditionally in `validate()`. The path is enforced and tested today, not dead code. No reason to defer to T6.

**Does the `LogStartupAuthMode(mode, issuer)` primitives signature feel right?** Yes. Primitives-only is what keeps `banner` free of `config`/`auth` and breaks the import cycle cleanly. A struct arg would re-couple it for no gain. Keep it.

One approve-gating item from my side: the two credential-trim fixes (secret + audience) are real validation holes, not nits, since they defer a bad config to a runtime 401. Small fixes, both acknowledged. Happy to approve once they land.

amitmorade:
Thanks for the careful pass — both points addressed. Quick rundown:

**1. Whitespace-trim on per-broker `auth.audience` (your earlier comment with the Claude-suggested patch)** — applied verbatim in [`68fdf0f`](https://github.com/SolaceDev/solace-broker-mcp/commit/68fdf0f). The `!= "" && TrimSpace == ""` guard preserves the field's optional-but-set semantics. The same commit also extended the rule to every other operator-supplied required string in the validator (`client_secret_basic/post.secret`, `dev_token`, `issuer`, `resource_url`, `idp_token_endpoint`, `mcp_server_client_id`, `broker.url`) — same class of bug, same fix.

**2. Alignment error and banner gating asymmetry** — gated the error the same way as the banner in [`2e0c385`](https://github.com/SolaceDev/solace-broker-mcp/commit/2e0c385). Both now live inside the existing `else if oauthBrokerCount == 0` branch in `validate()`. While the not-yet-supported guard is active, the alignment branch is structurally unreachable, so operators see one remediation path in the startup logs for the misconfiguration, not two pointing in opposite directions. Verified end-to-end against a Hop1=static + 2 oauth brokers config: the startup log contains only the not-yet-supported banner + per-broker errors; the `mcp_client_auth.mode must be oauth` sentence no longer appears in either the joined error string or the standalone banner.

The alignment validator function (`validateHop1Hop2Alignment`) stays callable and is exercised today via `TestValidateHop1Hop2Alignment_Direct`, so the invariant logic doesn't rot while it sleeps in `validate()`. When the OAuth runtime ships and the `if oauthBrokerCount > 0` arm is deleted, the `else if` collapses to an unconditional call and the alignment check becomes the operator-facing surface — no further code changes needed at that point.

Tracked the cleanup as a dedicated acceptance criterion on [SOL-150800 (T6)](https://sol-jira.atlassian.net/browse/SOL-150800), the sub-ticket that wires the OAuth dispatcher. T6's PR is the natural place for the guard removal because that's the moment the dispatcher's `oauth` branch starts actually working.

solace-aross (approved):
Approving. Second pass — every thread from the first round is resolved in code, and I verified each against the branch rather than the commit messages:

- **`BrokerOAuth` field comment** (`96b76a9`) — now states the provided-but-unused case logs a WARN, matching the validator.
- **Whitespace-only client secret + `auth.audience`** — `TrimSpace`-before-empty guards in place, and the rule was extended to every operator-supplied required string, not just the two flagged. `TestLoadConfig_WhitespaceTrimmedAcrossRequiredStrings` pins it against a future collapse back to a bare check.
- **Plan doc `client_auth_method` mismatch** (`68fdf0f`) — the doc now carries an explicit design-pivot note; the discriminated union is the single source of truth.
- **Hop 1 / Hop 2 alignment gating** (`2e0c385`) — the alignment error is gated behind the same guard as the banner, so operators get one remediation path while OAuth-on-brokers is unsupported. `validateHop1Hop2Alignment` stays a pure function exercised by `TestValidateHop1Hop2Alignment_Direct`, and the guard-removal step is tracked as an acceptance criterion on SOL-150800 (T6). Good call routing the cleanup to the ticket that turns the runtime on.

CI is green across build, lint, analyze, all three E2E suites, and FOSSA. Schema-only with the runtime rejecting `auth.mode: oauth` at startup is the right way to fix the contract ahead of the runtime. Nice work on the defensive hardening pass.

solace-awaheed (approved):
LGTM - thanks for the changes!

PR #111, opened 2026-06-22
SOL-150797: capture raw subject token on ctx after SDK validation

amitmorade:
## Summary

Adds `InjectRawSubjectToken` middleware that captures the raw bearer token onto `ctx` after the MCP SDK has validated it, so the future OAuth `Authenticator` can use it as the `subject_token` in RFC 8693 token exchange (Hop 2).

Jira: [SOL-150797](https://sol-jira.atlassian.net/browse/SOL-150797).

## Type of Change

## Motivation and Context

The MCP SDK's `RequireBearerToken` validates the incoming bearer token and places `*TokenInfo` (parsed claims) on the request context — but it discards the raw signed JWT string. Hop 2 token exchange needs that raw string as the `subject_token` in the exchange request to the IdP (RFC 8693 §2.1, Microsoft Entra OBO, Keycloak token-exchange all agree: the full signed token, verbatim, not a claim).

Without this middleware, the raw token is unreachable to any downstream code — the SDK passes only `ctx` (no `*http.Request`) into tool handlers and the outbound `Authenticator`.

## Changes Made

**Production code:**
- `internal/auth/raw_subject_token.go` (new):
  - `InjectRawSubjectToken(next http.Handler) http.Handler` — middleware that copies the bearer token onto `ctx` under an unexported key when the Authorization header is `Bearer <non-empty>`. No-op on any other shape (rejection is the responsibility of whatever component sits in front of us in the chain; we only capture).
  - `RawSubjectTokenFromContext(ctx) (string, bool)` — read-side accessor. Returns `("", false)` when no token is stashed, when the stored value is empty, or when the stored value is the wrong type (defensive).
  - Unexported `rawSubjectTokenKey struct{}` type prevents cross-package collision.
  - One-shot startup `slog.Debug("InjectRawSubjectToken middleware installed")` from inside the constructor.
- `internal/auth/middleware.go`: wrap `next` with `InjectRawSubjectToken` so the chain becomes `RequireBearerToken → InjectRawSubjectToken → next`.

**Tests:**
- `internal/auth/raw_subject_token_test.go` (new) — unit tests against the middleware in isolation.
- `test/integration/raw_subject_token_capture_test.go` (new) — integration tests against the composed chain and against the production wiring path.
- `test/integration/README.md` — index entry for the new integration test file.

**Documentation:**
- `docs/superpowers/plans/oauth-token-exchange/2026-06-21-SOL-150797-T3-raw-subject-token.md` (new): design, decisions, testing plan.

## Position decision

`InjectRawSubjectToken` runs **after** `sdkauth.RequireBearerToken`. That means a value reaching `ctx` under `rawSubjectTokenKey{}` has already been validated by whatever upstream validator is wired (today: the SDK, which checks signature, issuer, audience, expiry). Downstream code reads it without re-validating. This was a deliberate choice over the alternative of running before validation; the reasoning is in the plan doc under Decision 1.

## Responsibility boundary

The middleware does **not** reject missing or malformed Authorization headers. Capture is its responsibility; rejection is the responsibility of whatever component sits in front of it in the chain. This keeps the middleware composable with or without an upstream validator and avoids duplicating gating logic. See Decision 7 in the plan doc.

## Why no wrapper refactor

This was considered (wrapping the SDK's auth in our own `Authenticator` interface) and rejected. The MCP SDK is the substrate we live inside, not a library we call. Wrapping its types for replaceability defeats the point of using it. The leak is small (two files read four fields off `*sdkauth.TokenInfo`) and contained at a clean boundary (`NewIdentityFromTokenInfo`). The wrapper refactor was ~800 lines of churn for marginal benefit. Details in the plan doc under Decision 2.

## Why no per-request logging

The middleware is plumbing — it does not connect to anything, manage state, or fail in interesting ways. A per-request "stashed token" log line would:
- offer no operational signal beyond what the future OAuth Authenticator already provides on failure,
- create a side-channel leak (presence of the log line reveals a valid bearer header was seen on a given request).

The single startup line is enough to confirm the middleware is installed; runtime correctness surfaces via downstream token-exchange success or failure. See Decision 4.

## Testing

**Test Environment:**
- Go version: 1.26.0 (darwin/arm64)
- OS: macOS

**Tests Performed:**
- [x] Unit tests added (5 functions in `internal/auth/raw_subject_token_test.go`)
- [x] Integration tests added (2 functions in `test/integration/raw_subject_token_capture_test.go`)
- [x] Tested locally with `go test -race ./...` (all packages pass)
- [x] `/check-logs` scan: zero violations on the changed files

**Unit-level coverage (in `internal/auth/raw_subject_token_test.go`):**

- `TestInjectRawSubjectToken_HeaderShapes` — 13 subtests covering valid Bearer, case-insensitive scheme matching, missing header, empty token, wrong scheme (Basic/DPoP), token-only, extra-fields, whitespace/tab handling.
- `TestInjectRawSubjectToken_ConcurrentRequests` — 64 goroutines × 1000 iterations × `-race`. Pins the no-shared-state invariant so any future change introducing shared mutable state fails CI.
- `TestInjectRawSubjectToken_RequestIsolation` — original Authorization header is not mutated; parent ctx is not polluted; next handler still sees the header.
- `TestRawSubjectTokenFromContext_AccessorContract` — background ctx, non-string value, empty string, exact round-trip.

**Integration-level coverage (in `test/integration/raw_subject_token_capture_test.go`):**

- `TestRawSubjectTokenCapture_AfterSDKValidation` — chain-order invariant. Wires `sdkauth.RequireBearerToken(verifier)(auth.InjectRawSubjectToken(probe))` and exercises two scenarios:
  - **Valid token**: round-trips byte-for-byte from the wire through the SDK verifier to ctx readable by downstream.
  - **Invalid token**: SDK short-circuits with 4xx/5xx; our middleware is never invoked (proves the position decision).
- `TestRawSubjectTokenCapture_WiredByNewAuthMiddleware` — production-wiring regression guard ([added in response to Copilot review](https://github.com/SolaceDev/solace-broker-mcp/pull/111#discussion_r2659876870)). Builds a static-mode `*config.ServerConfig`, calls `auth.NewAuthMiddleware(cfg, downstream)`, sends a request with the configured dev token, and asserts the downstream handler observes the same token via `RawSubjectTokenFromContext`. Catches the failure mode where a future refactor silently drops `InjectRawSubjectToken(next)` from `internal/auth/middleware.go`.

### Code Quality
- [x] Self-review completed
- [x] Code is well-commented (explains why, not what)
- [x] Complex logic has explanatory comments (position decision, responsibility boundary, parse-mirrors-SDK note)

### Security
- [x] No credentials in code or logs
- [x] Followed secure logging rules — `/check-logs` clean
- [x] Token bytes never appear in any log call
- [x] Unexported ctx key type prevents cross-package read

### Testing
- [x] Regression guard for production wiring path

### Documentation
- [x] godoc comments on all exported symbols
- [x] Plan doc tracked under `docs/superpowers/plans/oauth-token-exchange/`
- [x] README entry under `test/integration/`

### Git & CI
- [x] Commit messages clear and descriptive
- [x] Branch fresh from main (no merge conflicts expected)
- [x] CI lint passing (initial commit failed `noctx`; fixed in commit `7abdc26` by switching to `httptest.NewRequestWithContext`)

## For Reviewers

**Focus Areas:**

1. **Position decision** — is wrapping `InjectRawSubjectToken(next)` inside `sdkauth.RequireBearerToken(...)` the right chain order? Plan doc covers the rationale.
2. **Responsibility boundary** — agree that this middleware should silently pass through missing/malformed headers rather than duplicate the upstream validator's rejection logic?
3. **Per-request logging absence** — agree we should not log per-request token presence?
4. **Parse mirroring** — `raw_subject_token.go` mirrors `sdkauth.verify`'s parse logic (case-insensitive Bearer, `strings.Fields`). Comment in the file flags this dependency for future SDK upgrades.

**Questions:**

None blocking. Plan doc captures every decision and rejected alternative.

[SOL-150797]: https://sol-jira.atlassian.net/browse/SOL-150797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/superpowers/plans/oauth-token-exchange/2026-06-21-SOL-150797-T3-raw-subject-token.md
+++ b/docs/superpowers/plans/oauth-token-exchange/2026-06-21-SOL-150797-T3-raw-subject-token.md
@@ -77,6 +77,23 @@ The Hop 2 prototype carried a `BMS_DEMO_LOG_TOKENS` env-flag for demo logging. T
 - All per-request data lives on the stack of the request goroutine and on the request's ctx.
 - The unexported key type ensures no other package can collide with our slot.
 
+### 7. Rejection is the upstream component's responsibility, not ours
+
+When the Authorization header is missing or malformed, `InjectRawSubjectToken` does **not** reject the request. It calls `next.ServeHTTP(w, r)` unchanged — no error, no log, no 401, no `WWW-Authenticate` header.
+
+The principle is positional, not coupled to any specific upstream component:
+
+- **Capture is our responsibility.** If a valid Bearer header is present, copy the token onto ctx.
+- **Rejection is the responsibility of whatever component sits in front of us in the chain.** Today that is `sdkauth.RequireBearerToken`; tomorrow it could be a different validator (e.g. an opaque-token introspection middleware), a custom auth layer, or a hand-built chain in tests.
+
+Two consequences:
+
+1. **No duplication of upstream gating.** Reasonable upstream auth components already produce well-formed 401 responses with the right `WWW-Authenticate` headers, RFC 9728 metadata pointers, and so on. Reimplementing that here would risk subtle disagreement between layers.
+
+2. **Composable in any chain.** The middleware can be wired with or without an upstream validator without becoming an accidental auth gate. In production it sits behind a validator; in some tests it is exercised in isolation. Behaviour stays consistent.
+
+This design intentionally trades "fail fast at the HTTP boundary" for "single responsibility." In production, the upstream validator will reject missing/malformed headers before our middleware ever sees them, so the silent-pass branch is effectively unreachable in real traffic. It exists so the middleware composes safely in non-production paths (tests, future configurations) without quietly turning into an auth enforcer.
+
 ---
 
 ## Files touched
@@ -123,7 +140,7 @@ This section covers production-code work only. Test work is in the next section.
      - Field 1 must be `"Bearer"` matched case-insensitively (`strings.EqualFold`). This matches the SDK's `strings.ToLower(fields[0]) != "bearer"` check.
      - Field 2 must be non-empty.
      - When all three conditions hold: derive a new ctx with `context.WithValue(r.Context(), rawSubjectTokenKey{}, token)` and call `next.ServeHTTP(w, r.WithContext(ctx))`.
-     - When any condition fails: call `next.ServeHTTP(w, r)` unchanged. **No rejection, no error, no log.** The SDK middleware upstream already rejected this case if it was a security issue; we only run when the SDK approved.
+     - When any condition fails: call `next.ServeHTTP(w, r)` unchanged. **No rejection, no error, no log.** Rejection is the responsibility of whatever component sits in front of this middleware in the chain (in production today, that is `sdkauth.RequireBearerToken`). This middleware only captures — it does not duplicate the upstream component's gating role.
 
 4. **Accessor** `RawSubjectTokenFromContext(ctx context.Context) (string, bool)`:
    - `v := ctx.Value(rawSubjectTokenKey{})`.
--- a/docs/superpowers/plans/oauth-token-exchange/2026-06-21-SOL-150797-T3-raw-subject-token.md
+++ b/docs/superpowers/plans/oauth-token-exchange/2026-06-21-SOL-150797-T3-raw-subject-token.md
@@ -1,0 +1,317 @@
+# SOL-150797 (T3): Capture the raw subject token on ctx after SDK validation
+
+**Status**: planned
+**Date**: 2026-06-21
+**Branch**: `amorade/SOL-150797`
+**Parent epic**: SOL-150070 (OAuth token exchange / Hop 2)
+**Predecessors**: SOL-150794 (Authenticator interface), SOL-150795 (SEMP clients consume Authenticator), SOL-150796 (broker_oauth config schema + Hop 1/Hop 2 alignment guard)
+
+---
+
+## Why this ticket exists
+
+When `broker_oauth` is enabled, the MCP server performs RFC 8693 token exchange against the customer's IdP per request. The exchange request needs the user's original signed JWT as the `subject_token` parameter. That JWT arrives in the `Authorization: Bearer <jwt>` header of the incoming MCP request, but the MCP Go SDK's `sdkauth.RequireBearerToken` middleware validates the token and forwards only the parsed `*TokenInfo` on ctx — the raw signed string is dropped.
+
+A later ticket will add the OAuth `semp/auth.Authenticator` implementation that calls the IdP. That implementation needs the raw JWT but has only `ctx` to work with. T3 closes this gap: a small middleware stashes the raw bearer token on ctx under an unexported key, and a typed accessor returns it.
+
+---
+
+## Key design decisions (locked)
+
+### 1. Position in the middleware chain: **after** the SDK's `RequireBearerToken`
+
+Chain order at runtime:
+
+```
+sdkauth.RequireBearerToken  →  InjectRawSubjectToken  →  next handler
+       (validates)              (stashes raw JWT)         (MCP SDK)
+```
+
+**Rationale.** Placing `InjectRawSubjectToken` after SDK validation gives the value on ctx a stronger contract: *if a token is present on ctx, the SDK has already validated its signature, issuer, audience, and expiry*. Downstream code (the future OAuth Authenticator) can rely on this invariant without re-validating.
+
+This is encoded in the wiring by wrapping `next` with our middleware first, then letting the SDK middleware wrap that:
+
+```go
+middleware := sdkauth.RequireBearerToken(verifier, &opts)
+return middleware(InjectRawSubjectToken(next)), nil
+```
+
+At request time, the SDK's middleware runs first. If validation fails, it writes 4xx and returns — our middleware is unreachable. If validation succeeds, the SDK calls `handler.ServeHTTP`, which is our middleware. The ordering is guaranteed by Go's control flow in `RequireBearerToken`'s body, not by configuration.
+
+### 2. No `Authenticator` wrapper refactor
+
+We considered building an `auth.AuthResult` + `auth.Authenticator` wrapper around the SDK's auth surface so the raw token would be a field on `AuthResult` (collapsing T3 into validator output). We rejected this for now:
+
+- The MCP SDK is a **framework** the server lives inside, not a **library** it calls. Standard "abstract at the boundary" advice is written for libraries. SDK types appearing in our code where the SDK delivers them is the normal mode of being inside an SDK.
+- The actual leak is small: `*sdkauth.TokenInfo` appears in two non-auth files (`internal/tools/identity.go`, `internal/tools/register.go`), where it is immediately converted to our `Identity` type at the call to `NewIdentityFromTokenInfo`. That conversion *is* the boundary.
+- The wrapper would still have to "shadow-write" `*TokenInfo` on ctx so the SDK's `StreamableHTTPHandler` session-affinity check (`streamable.go:306`) keeps working. The wrapper does not free us from the SDK's auth contract; it only renames it.
+- Cost of the wrapper (≈900 lines touched across production and tests) is disproportionate to the benefit when there is no second consumer of `*sdkauth.TokenInfo` in sight.
+
+If a third consumer of `*sdkauth.TokenInfo` appears outside `internal/auth/` in the future, that is the trigger to reconsider (rule of three). Until then, T3's small middleware is the right tool.
+
+### 3. Naming: `rawSubjectToken`, not `rawToken`
+
+The value's role is RFC 8693's `subject_token`. Naming it after that role makes the contract self-documenting and reserves room for other raw tokens (none today) without name collision.
+
+- File: `internal/auth/raw_subject_token.go`
+- Type: `type rawSubjectTokenKey struct{}` (unexported)
+- Middleware: `func InjectRawSubjectToken(next http.Handler) http.Handler`
+- Accessor: `func RawSubjectTokenFromContext(ctx context.Context) (string, bool)`
+
+The ticket's title and prior commit history use `rawToken`; the PR description will note the rename and the reasoning.
+
+### 4. No per-request logging; one startup log
+
+- **No log of the token bytes at any level**, not even at DEBUG. Token never appears in any log line.
+- **No per-request "token stashed" log.** A per-request log would be noise on a busy server, would constitute a side-channel ("this request carried a valid bearer header"), and the real failure mode (token missing on ctx when downstream needs it) is already loudly observable via the future OAuth Authenticator's error.
+- **One DEBUG line per installation**: `"InjectRawSubjectToken middleware installed"`. Emitted from inside `InjectRawSubjectToken`'s constructor (not from `NewAuthMiddleware`), so the log is colocated with the thing it describes. Today this fires once at startup because there is exactly one call site; if a future configuration installs the middleware on a second endpoint, the log will correctly emit a second line. No `sync.Once`, no package-level state.
+
+### 5. No demo or instrumentation hooks
+
+The Hop 2 prototype carried a `BMS_DEMO_LOG_TOKENS` env-flag for demo logging. T3 does not port that flag or any equivalent. Per parent ticket SOL-150070 acceptance criteria.
+
+### 6. Stateless, ctx-only data flow
+
+- No package-level mutable state.
+- No caches, sync primitives, or maps.
+- All per-request data lives on the stack of the request goroutine and on the request's ctx.
+- The unexported key type ensures no other package can collide with our slot.
+
+---
+
+## Files touched
+
+### New
+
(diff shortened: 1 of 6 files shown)


internal/auth/middleware.go
```diff
@@ -69,7 +69,14 @@ func NewAuthMiddleware(cfg *config.ServerConfig, next http.Handler) (http.Handle
 		ResourceMetadataURL: metadataURL,
 	})
 
-	return middleware(next), nil
+	// InjectRawSubjectToken is wrapped INSIDE the SDK middleware so it runs
+	// only after sdkauth.RequireBearerToken has validated the bearer token
+	// (signature, issuer, audience, expiry). A value reaching ctx under
+	// rawSubjectTokenKey{} therefore carries the SDK-validated invariant,
+	// which the future OAuth Authenticator relies on as the subject_token
```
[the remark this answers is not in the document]
amitmorade: Good catch — agreed, and added.

7c7341e introduces TestRawSubjectTokenCapture_WiredByNewAuthMiddleware in test/integration/raw_subject_token_capture_test.go. It does exactly what you suggested:

Builds a *config.ServerConfig with MCPClientAuth.Mode = AuthModeStatic and a fixture dev token.
Calls auth.NewAuthMiddleware(cfg, downstream) so the production wiring path runs end-to-end.
Sends a request with Authorization: Bearer <devToken>.
Asserts the downstream handler observes the same <devToken> via auth.RawSubjectTokenFromContext(ctx).
The failure mode this pins: a future refactor that drops InjectRawSubjectToken(next) from middleware.go:80 and returns middleware(next) instead. Today no existing test would catch that — the chain-order tests build the chain by hand. This new test now does.

Static mode was chosen because it's the smallest production-supported MCPClientAuth.Mode that activates the auth chain (disabled returns next directly without wrapping), and it doesn't require IdP infrastructure to exercise. Validator correctness for OIDC mode remains covered by internal/auth/middleware_test.go.

helen-zhu-solace (dismissed):
Nice work on this. A small note, not blocking.

### 1. Rejection test exercises the 500 path, not the 401 path it's meant to

`test/integration/raw_subject_token_capture_test.go:97` uses `errors.New("...: " + err.Error())` instead of `fmt.Errorf("...: %w", sdkauth.ErrInvalidToken)`. The current form makes the SDK return 500 rather than the 401 production produces. Worth wrapping with `%w` and tightening the assertion to expect 401 exactly.

helen-zhu-solace (dismissed):

amitmorade:
> Nice work on this. A small note, not blocking.
> ### 1. Rejection test exercises the 500 path, not the 401 path it's meant to
> 
> `test/integration/raw_subject_token_capture_test.go:97` uses `errors.New("...: " + err.Error())` instead of `fmt.Errorf("...: %w", sdkauth.ErrInvalidToken)`. The current form makes the SDK return 500 rather than the 401 production produces. Worth wrapping with `%w` and tightening the assertion to expect 401 exactly.

You were right — both points landed. The test was on the wrong SDK branch, and accepting 401 || 500 masked that. Your comment is what got this whole thread moving.

Applied the fix as suggested in 19bd111 (wrapped with %w, tightened to == 401). With that change in place I took another look at what the assertion was claiming, and stepped back to the bigger question: the test's name and load-bearing assertion are about our middleware not running on rejection, not about which SDK branch produces the rejection. Pinning == 401 from this test's frame turned out to be false precision — it coupled the assertion to an SDK internal while the property the test names is unrelated to that internal.

So 92a4ea1 walks in a different direction than your original suggestion:

- Verifier returns plain errors.New(...) — no %w.
- Status assertion relaxed to >= 400 as a sanity guard.
- probeInvoked == false stays as the load-bearing assertion.
- Added comments at the function doc and inline explaining the property-focused intent and noting why the tighter assertion was tried and walked back, so a future reader doesn't re-introduce it.

Your nudge is what surfaced the question in the first place. Happy to discuss if the trim feels too loose now.

helen-zhu-solace (dismissed):
LGTM!

solace-aross (dismissed):
Clean, well-scoped PR. Design is sound, test coverage is strong, docs are thorough. Approving.

I verified the load-bearing risk myself: the middleware's parse is identical to `sdkauth.verify` in go-sdk v1.5.0 (`auth/auth.go:101-103`) — same `Authorization` header via `Header.Get`, same `strings.Fields` split, same case-insensitive `bearer` match. So the SDK-validated token and the captured token are the same bytes. The "under-capture is the safe drift" reasoning holds, and the in-file comment already flags it for future SDK bumps.

Also confirmed:
- **Position** — wrapping `InjectRawSubjectToken(next)` inside `RequireBearerToken` is correct; the "present on ctx ⇒ validated" invariant is enforced by Go control flow, not config.
- **Concurrency/isolation** — stateless, ctx-only, unexported key. The 64×1000 `-race` tripwire is the right way to pin "no shared state."
- **Logging** — no token bytes anywhere; single startup DEBUG line.

Both prior threads are already resolved (Copilot's `NewAuthMiddleware` regression test; the rejection-test status-code discussion landed correctly on asserting the property, not the SDK's status code). Not re-raising either.

One optional, non-blocking note: `raw_subject_token_test.go` hand-rolls `itoa` to avoid importing `strconv`. `strconv.Itoa` is standard library, zero-cost, and already idiomatic in the repo's tests — I'd just use it rather than carry ~20 lines a reader has to verify. Minor, won't hold up the merge.

Nice work.

amitmorade:
> One optional, non-blocking note: `raw_subject_token_test.go` hand-rolls `itoa` to avoid importing `strconv`. `strconv.Itoa` is standard library, zero-cost, and already idiomatic in the repo's tests — I'd just use it rather than carry ~20 lines a reader has to verify. Minor, won't hold up the merge.
> 
> Nice work.

Fixed in 701e3a9. Thanks for the catch.

docs/superpowers/plans/oauth-token-exchange/2026-06-21-SOL-150797-T3-raw-subject-token.md
solace-wmaryam, line 1: This is something I've been wondering about for a while and unrelated to the PR implementation. Do we want to commit architecture decision docs in a GitHub repo that's going to be publicly available?

internal/auth/raw_subject_token.go
```diff
@@ -0,0 +1,112 @@
...
+// package can collide with it or read the value except through
+// RawSubjectTokenFromContext.
+type rawSubjectTokenKey struct{}
+
+// InjectRawSubjectToken returns middleware that copies the bearer token
+// from the Authorization header onto the request context.
+//
+// The middleware is intended to run AFTER the MCP SDK's RequireBearerToken
+// in the chain, so the token it captures has already been validated by
+// the SDK. The position is established at wiring time in NewAuthMiddleware.
```
solace-wmaryam: Nitpick. There is some comment duplication here. There is a lot of similarity here with the comments on lines 17-32..maybe we can consolidate?

internal/auth/raw_subject_token.go
```diff
@@ -0,0 +1,112 @@
...
+		// go-sdk v1.5.0 (auth/auth.go: strings.Fields + case-insensitive
+		// "bearer" check). Because the SDK runs upstream of us, anything
+		// it rejected we never see; anything it accepted should match
+		// these conditions. If the SDK ever changes its parsing rules,
+		// this code may under-capture (we'd silently fail to stash a
+		// token the SDK approved) — the safe drift direction, but worth
+		// revisiting on SDK upgrades.
+		authHeader := r.Header.Get("Authorization")
+		fields := strings.Fields(authHeader)
+		if len(fields) == 2 && strings.EqualFold(fields[0], "Bearer") && fields[1] != "" {
```
solace-wmaryam, line 77: Just wanted to confirm - I think we already have a test to verify the parsing behaviour? This would be useful if there is some change in the sdk version affecting the parsing output.

internal/auth/raw_subject_token_test.go
```diff
@@ -0,0 +1,238 @@
...
+				t.Errorf("ok = %v, want %v (%s)", ok, tc.wantOK, tc.description)
+			}
+			if gotToken != tc.wantToken {
+				t.Errorf("token = %q, want %q (%s)", gotToken, tc.wantToken, tc.description)
+			}
+		})
+	}
+}
+
+// TestInjectRawSubjectToken_ConcurrentRequests is the security-critical
```
solace-wmaryam, line 92: Is this test worth adding right now. Do we expect this design to change and for us to encounter the shared state situation? If not, maybe it's not worth the code bloat?

solace-wmaryam (dismissed):

internal/auth/raw_subject_token_test.go
amitmorade: Fair question. The test isn't catching a bug today — by construction the middleware has no shared state, so it'll always pass. Its value is as a tripwire: if a future change introduces shared state (a cache, a sync.Map, a package-level counter), the test fails immediately under -race rather than silently shipping a cross-user token leak

internal/auth/raw_subject_token.go
amitmorade: Yeah, the parse cases are covered — TestInjectRawSubjectToken_HeaderShapes runs 13 of them (valid Bearer, case variants, missing, malformed, Basic/DPoP, whitespace/tab, etc.). All green under -race.

The comment is actually for a different thing, which I think is what your second sentence is hinting at. We intentionally mirror sdkauth.verify's parse from SDK v1.5.0. Our tests pin our own parse, but they can't catch drift on the SDK side — if a future SDK release tightens or loosens what it accepts, our middleware would silently miss tokens it should be capturing. Safe direction to drift in, but still worth knowing about. The comment is really a heads-up for whoever does the next SDK bump.

So the tests and the comment are doing different jobs: tests guard against regressions in our code, the comment flags the coupling to the SDK's parse.

docs/superpowers/plans/oauth-token-exchange/2026-06-21-SOL-150797-T3-raw-subject-token.md
amitmorade: Good question, and I think it goes wider than just this PR. Probably worth a proper discussion since it applies to every plan doc we have under docs/superpowers/plans/, not just this one. Let me start a thread in Slack so we can think through it together with the team.

For this PR I'd suggest we keep the existing convention for now since other plan docs are already in the repo, and revisit once we land somewhere on the broader question.

internal/auth/raw_subject_token.go
amitmorade: Yeah, fair point. The function godocs ended up restating what's already in the package doc, especially around the position invariant and the "validated by SDK" narrative. Trimmed both in 0f11b22 so they focus on per-function behaviour and point at the package doc for the bigger picture. Net -5 lines.

docs/superpowers/plans/oauth-token-exchange/2026-06-21-SOL-150797-T3-raw-subject-token.md
solace-wmaryam: Thanks for taking that initiative. And agreed, you don't need to change the convention for this PR

internal/auth/raw_subject_token.go
solace-wmaryam: I see. I wonder if there's some sort of test we can have to determine an SDK change if it happens. Otherwise, we might forget about this in the long-term and only realize when failures start occurring. Basically, I wonder if there's a better way to document this behaviour that won't escape the team's notice when the need arises.

solace-aross: The answer here is a test, not a comment. Run the same header shapes through both `sdkauth.RequireBearerToken` and `InjectRawSubjectToken`, then assert per row that the two agree: the SDK admitting a shape and us capturing it always match. The `HeaderShapes` table and the SDK wiring already exist in the suite, so it's a small addition.

That gives you the property you're after. A future SDK parse change that diverges from ours fails CI immediately, instead of drifting silently behind a comment nobody re-reads at upgrade time.

One nuance: the drift fails closed. If the SDK loosens its parse and we under-capture, token exchange fails visibly rather than leaking a token. So this is a debuggability guard, not a security gate. Reasonable as a fast-follow rather than a blocker here.

solace-aross (approved):
Re-approving against 0f11b226.

The change since my last approval is cosmetic: `strconv.Itoa` in place of the hand-rolled helper, and the godoc trim. Both are what was asked for, both are clean. The logic, position invariant, and isolation properties hold from the prior review.

Wajiha's SDK parse-drift point is worth a follow-up, not a blocker. I left a concrete suggestion on her thread: a table-driven test that pins "SDK admits a shape" against "we capture it," so a future SDK bump that changes parsing fails CI instead of drifting silently. Good candidate for a fast-follow.

Nice work.

internal/auth/raw_subject_token.go
amitmorade: Makes sense. The integration test on HeaderShapes catches SDK parse divergence immediately. I'll add that as a fast-follow PR to verify our parse stays in sync with the SDK's on each shape.

helen-zhu-solace (approved):

PR #112, opened 2026-06-23
SOL-151052: Implement composite-tool postprocessor + first handler for list-queues

helen-zhu-solace:
## Summary

Productionizes Approach C from the SOL-149390 spike: composite tools can declare `postProcess: <handlerName>` in `tools.yaml`, and a Go handler computes a `summary` map (e.g. aggregate counts) that the executor merges alongside the raw step results. Adds the first handler for `list-queues` and a startup validator that fails boot if a handler's required SEMP fields aren't selected.

Fixes [SOL-151052](https://sol-jira.atlassian.net/browse/SOL-151052)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

SOL-149390 chose Approach C (Go postprocessor) over Approach A (YAML aggregation DSL) for count-style aggregations; SOL-150904 prototyped both and confirmed the call. This story lands C in the real codebase. The payoff: LLM clients get pre-computed counts (queues with no consumers, congested, backlogged, near-full) instead of having to scan the raw list themselves.

The `nearFullCount` ratio aggregation is included specifically to demonstrate expressivity Approach A could not match.

## Changes Made

**New `postProcess` strategy**
- New `postProcess` value for `result.strategy` in `tools.yaml`, with a `postProcess: <name>` field naming a registered handler.
- Executor merges the handler's output under a top-level `summary` key alongside the raw step results (`internal/composite/executor.go`).
- New `internal/composite/postprocess/` package: registry, `Handler` shape (`Fn` + `RequiredSteps` + `RequiredFields`), self-registration via `init()`. `cmd/server/main.go` blank-imports the handlers package so the registry is populated before startup validation runs.

**YAML schema: structured `select:`**
- Step schema gains a structured `select:` list, replacing the old comma-joined string under `args.select`. The executor joins it into the SEMP `select` query parameter at execute time (`applySelect`).
- The structured form is what makes the boot-time field cross-check possible without re-parsing strings.
- The loader rejects setting both `args.select` and `select:` on the same step (they would race, since `applySelect` overwrites `args["select"]`).

**Boot-time safety nets** (in `loader.go` / `ValidatePostProcess`)
- Every handler's `RequiredFields` must appear in the union of its tool's step `select:` clauses — catches SEMP field-name drift (`bindCount` vs `bindSuccessCount` class of bug from the spike report).
- Every handler's `RequiredSteps` must match an actual step ID in the tool — catches YAML step renames that would otherwise surface only at first call.
- The step ID `"summary"` is reserved under the `postProcess` strategy so a step can't silently shadow the handler output.
- All four checks fail the process at boot with a uniform error template.

**First handler: `list-queues`**
- `internal/composite/postprocess/handlers/list_queues.go` return `noConsumerCount`, `congestedCount`, `backloggedCount`, `nearFullCount`, plus `scanned` (items observed) and (when the paginator truncated) `truncated: true`. Surfacing truncation in the summary keeps the LLM from treating partial counts as global.
- `tools.yaml` `list-queues` switched to `strategy: postProcess` with the five required fields in `select:`.

**Tests**
- Handler counts, edge cases, missing-field errors, truncation surfacing.
- Loader cross-field rules (postProcess/collect mutual exclusion, reserved step ID, args.select/select coexistence guard).
- `ValidatePostProcess` boot-time cross-check (happy path, missing field, unregistered handler).
- Executor merge contract: `collect` results pass through, `summary` lands next to them.
- Regression test pinning `applySelect` on the parallel-batch path.
- New `postprocesstest` helper package for test-scoped register/unregister
  (lives in a sibling package so production binaries don't import `testing`).

Architecture docs updated in `docs/internal/architecture.md`.
## Testing

**Test Environment:**
- Go version: go1.25.11
- OS: Linux 7.0.10-101.fc43.x86_64
- Broker version (if applicable): N/A for unit tests; E2E uses Dockerized brokers via `make e2e-all`

**Tests Performed:**

**Test Coverage:**
- `list-queues` handler: each of `noConsumerCount`, `congestedCount`, `backloggedCount`, `nearFullCount` over representative queue sets, including the near-full ratio edge cases.
- Loader cross-check: tool with `postProcess` whose `select:` is missing a required field fails to load with the documented error template.
- Executor merge: `collect` results pass through under their existing key while the handler output lands under `summary`.
- E2E: `list-queues` against a real broker returns the merged `{summary, queues}` shape.

## Breaking Changes

N/A — `postProcess` is additive; the existing `collect` strategy is unchanged.

**Impact:**

**Migration Guide:**

### Code Quality

### Security

### Testing
- [x] Error paths tested

### Documentation

### Git & CI
- [x] No merge conflicts

## Related Issues/PRs

- Related to [SOL-149390](https://sol-jira.atlassian.net/browse/SOL-149390) (Approach C decision)
- Related to [SOL-150904](https://sol-jira.atlassian.net/browse/SOL-150904) (prototype validation)

---

**For Reviewers:**

Output:
image

**Focus Areas**: The executor merge contract in `ApplyResultStrategy` (top-level `summary` key alongside the existing `collect` result), and the startup validator's error path — that's the SEMP-version-drift safety net (`bindCount` vs `bindSuccessCount` class of bug from the spike report).

[SOL-151052]: https://sol-jira.atlassian.net/browse/SOL-151052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-149390]: https://sol-jira.atlassian.net/browse/SOL-149390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/composite/postprocess/handlers/list_queues.go
+++ b/internal/composite/postprocess/handlers/list_queues.go
@@ -30,9 +30,16 @@ import (
 // risk of taking SPOOL_OVER_QUOTA discards if their consumers don't catch up.
 const nearFullThreshold = 0.8
 
+// listQueuesStepID is the step ID this handler keys into. Declared as a const
+// so the init-time RequiredSteps registration and the runtime lookup cannot
+// drift out of sync — and so the boot-time check in ValidatePostProcess
+// catches a YAML rename of the step.
+const listQueuesStepID = "queues"
+
 func init() {
 	postprocess.Register("listQueues", postprocess.Handler{
-		Fn: ListQueues,
+		Fn:            ListQueues,
+		RequiredSteps: []string{listQueuesStepID},
 		RequiredFields: []string{
 			"bindCount",
 			"spooledMsgCount",
@@ -49,10 +56,16 @@ func init() {
 //   - backloggedCount: queues with spooledMsgCount > 0 (any backlog at all)
 //   - nearFullCount:   queues with msgSpoolUsage/maxMsgSpoolUsage >= 0.8
 //     (skips queues with maxMsgSpoolUsage == 0 — unbounded or unset)
+//
+// The counts are over what the paginator actually returned. Under truncation
+// (followPages stopped at maxResults / capMax / maxPages), the summary also
+// includes scanned (the count of items observed) and truncated: true so the
+// LLM sees the partial-scan reality next to the counts rather than only on
+// the raw data block.
 func ListQueues(stepResults map[string]map[string]any) (map[string]any, error) {
-	step, ok := stepResults["queues"]
+	step, ok := stepResults[listQueuesStepID]
 	if !ok {
-		return nil, fmt.Errorf("step %q not in results", "queues")
+		return nil, fmt.Errorf("step %q not in results", listQueuesStepID)
 	}
 	items, ok := step["data"].([]any)
 	if !ok {
@@ -97,12 +110,17 @@ func ListQueues(stepResults map[string]map[string]any) (map[string]any, error) {
 			nearFull++
 		}
 	}
-	return map[string]any{
+	out := map[string]any{
 		"noConsumerCount": noConsumer,
 		"congestedCount":  congested,
 		"backloggedCount": backlogged,
 		"nearFullCount":   nearFull,
-	}, nil
+		"scanned":         len(items),
+	}
+	if t, _ := step["truncated"].(bool); t {
+		out["truncated"] = true
+	}
+	return out, nil
 }
 
 // numField accepts any of the numeric shapes JSON decoders produce: float64
--- a/internal/composite/postprocess/handlers/list_queues.go
+++ b/internal/composite/postprocess/handlers/list_queues.go
@@ -52,7 +52,7 @@ func init() {
 
 // ListQueues aggregates the queue list into four counts:
 //   - noConsumerCount: queues with bindCount == 0 (nothing reading)
-//   - congestedCount:  queues whose lowPriorityMsgCongestionState == "active"
+//   - congestedCount:  queues whose lowPriorityMsgCongestionState == "congested"
 //   - backloggedCount: queues with spooledMsgCount > 0 (any backlog at all)
 //   - nearFullCount:   queues with msgSpoolUsage/maxMsgSpoolUsage >= 0.8
 //     (skips queues with maxMsgSpoolUsage == 0 — unbounded or unset)
@@ -95,7 +95,7 @@ func ListQueues(stepResults map[string]map[string]any) (map[string]any, error) {
 		if bindCount == 0 {
 			noConsumer++
 		}
-		if state == "active" {
+		if state == "congested" {
 			congested++
 		}
 		if spooled > 0 {
--- a/internal/composite/postprocess/handlers/list_queues.go
+++ b/internal/composite/postprocess/handlers/list_queues.go
@@ -88,7 +88,7 @@ func ListQueues(stepResults map[string]map[string]any) (map[string]any, error) {
 		spooled, ok3 := numField(q, "spooledMsgCount")
 		usage, ok4 := numField(q, "msgSpoolUsage")
 		maxUsage, ok5 := numField(q, "maxMsgSpoolUsage")
-		if !(ok1 && ok2 && ok3 && ok4 && ok5) {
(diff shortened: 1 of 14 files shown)


solace-aross:
Early feedback on the draft, not a formal review, and nothing here blocks you.

This is a clean landing of Approach C. The standout is the boot-time `RequiredFields` / `RequiredSteps` cross-check: catching SEMP field drift at startup instead of at first invocation is exactly the right place to fail.

Heads-up so you don't relitigate the Copilot threads. Most look already closed in the current tree:
- Registry concurrency: `registryMu` (RWMutex) guards every access. Covered.
- Parallel steps skipping `select`: `executeBatch` applies it at `executor.go:391`. Covered.
- `select` vs `args.select` race: rejected at load in `loader.go:90-94`. Covered.
- Doc template mismatch: the `architecture.md` template string matches the code error. Covered.

Two things from my read:

1. **(Medium, worth a decision) One bad item fails the whole call.** `numField`/`stringField` return an error on any unexpected type, including `nil`, and `ApplyResultStrategy` propagates it. So a single queue with a null `msgSpoolUsage` (possible on some queue types or broker versions even when the field is selected) aborts `list-queues` entirely and drops the raw list too, not just the summary. That's a robustness step down from `collect`, which would still return the data. Options: count the odd item as unknown and keep going, or fall back to returning raw results with the summary omitted. If hard-fail is intentional, a one-line comment on `numField` saying so would settle it.

2. **(Low, clarity) `postProcess` + `args.select`-only fails boot with a confusing message.** A tool that supplies fields via a templated `args.select` instead of structured `select:` passes loader validation, then trips `ValidatePostProcess` with "reads X but it is not in select" at boot. Technically correct, but the real cause is "postProcess requires structured select." Copilot raised a version of this; rejecting `args.select` outright when strategy is `postProcess` would make the requirement explicit.

Build, vet, and the composite + postprocess + handlers tests pass locally on the branch.

internal/composite/postprocess/handlers/list_queues.go
```diff
@@ -0,0 +1,151 @@
...
+		usage, ok4 := numField(q, "msgSpoolUsage")
+		maxUsage, ok5 := numField(q, "maxMsgSpoolUsage")
+		if !ok1 || !ok2 || !ok3 || !ok4 || !ok5 {
+			skipped++
+			continue
+		}
+		if bindCount == 0 {
+			noConsumer++
+		}
+		if state == "active" {
```
solace-aross: The `select:` plumbing and the boot-time cross-check are clean, and this counter is one word away from working.

`lowPriorityMsgCongestionState` only ever returns `disabled`, `not-congested`, or `congested` (SEMP monitor spec, `semp-v2-swagger-private-monitor-extended.json:48758`). There is no `active`, so `congestedCount` stays 0 against a real broker. Swapping `"active"` → `"congested"` fixes it.

The test feeds `"active"` too (`list_queues_test.go:36`, `:168`), so it passes on the wrong value. Pinning those fixtures to the real enum values would let the test catch this.

One thing worth noting: it slipped past the new validator because the cross-check matches field names, not values, so string-valued comparisons like this are the one spot it cannot guard.

solace-aross:
Thanks for the quick turnaround. Both items from my earlier pass landed well: odd rows now skip into `skipped` and keep the raw list, so one null field no longer drops the whole call, and rejecting `args.select` under `postProcess` at load makes the structured-select requirement explicit instead of a confusing boot error.

The one open thread on `congestedCount` (the `lowPriorityMsgCongestionState` value) is the only thing left before I approve. Resolve that and I'm good.

solace-aross (approved):
Approving. The `congestedCount` fix landed: `state == "congested"` against the real enum, and the test fixtures are pinned to `congested`/`not-congested` so they'd catch a regression on the value next time. That was my last open item.

The rest of the earlier passes holds up too. Odd rows skip into `skipped` and keep the raw list, `args.select` is rejected under `postProcess` at load, and the boot-time `RequiredFields` cross-check is the right place to catch SEMP field drift. CI is green across build, lint, and all three E2E suites.

Nice work landing Approach C cleanly.

PR #115, opened 2026-06-26
SOL-151278: Observability package skeleton and flag plumbing

solace-aross:
Part of the **Broker MCP Observability & Telemetry** epic ([SOL-150251](https://sol-jira.atlassian.net/browse/SOL-150251)) → MUST child epic [SOL-149791](https://sol-jira.atlassian.net/browse/SOL-149791). Story: [SOL-151278](https://sol-jira.atlassian.net/browse/SOL-151278).

## What this does
Lays the foundation every later observability story builds on. **Skeleton only — no behavior is wired into the request or tool path yet.**

- New `internal/observability/` packages (correlation, recover, metrics, audit, tracing, health), each with a per-capability flag accessor.
- `ObservabilityConfig`: env-driven `OBS_*_ENABLED` flags with v1 "door-closing" defaults (correlation + panic-recovery on; metrics, audit, tracing, saturation off; auth-failure-counter follows metrics). Numeric tunables via the existing `${VAR}` YAML substitution.
- `Principal` context-key scaffold (`internal/auth/principal.go`) for later identity work.
- Schema-version constants (metrics/audit = `1.0`).
- One INFO startup log line summarizing enabled capabilities.

## Why it's first
It's the serialization gate for the MUST epic: every other story adds its capability behind one of these flags, in one of these packages.

## Testing
Unit tests cover flag defaults, env overrides (both directions), auth-failure-counter follow-metrics, numeric defaults, and the principal zero-value. `go build`, `go vet`, `go test ./...`, and `-race` all pass. Reviewed by a principal-architect pass before opening.

## Notes for reviewers
- `health` exposes only `SaturationEventsEnabled`; `/livez` and `/readyz` are unconditional and deliberately not flag-gated.
- Capability flags are env-only (not YAML); numeric tunables are YAML with `${VAR}` support.
- Plan: `docs/plans/2026-06-26-obs-story1-story7.md`.

[SOL-150251]: https://sol-jira.atlassian.net/browse/SOL-150251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-151278]: https://sol-jira.atlassian.net/browse/SOL-151278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/config/observability.go
```diff
@@ -0,0 +1,141 @@
...
+//     "door-closing" defaults ship correlation + panic recovery ON and
+//     everything else OFF.
+//
+//   - Numeric tunables (the int fields, YAML-tagged) parse from the YAML
+//     config like every other field, so they inherit the existing ${VAR}
+//     substitution for free (substituteEnvVars runs over the raw bytes before
+//     decode). Defaults are applied in applyDefaults alongside the rest.
+type ObservabilityConfig struct {
+	// Capability flags. Loaded from OBS_* env vars in applyObservabilityEnv,
+	// NOT from YAML — hence no yaml tags. See the type doc for why.
```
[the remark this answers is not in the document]
solace-aross: Good catch — the comment contradicted itself. Fixed in 0003153: it now states the `yaml:"-"` tags are intentional, excluding the capability flags from YAML so the OBS_* env vars stay the single source of truth. No behavior change.

internal/config/observability.go
```diff
@@ -0,0 +1,141 @@
...
+	o := &cfg.Observability
+	if o.SaturationThresholdMs == 0 {
+		o.SaturationThresholdMs = defaults.DefaultSaturationThresholdMs
+	}
+	if o.ProgressSignalThresholdMs == 0 {
+		o.ProgressSignalThresholdMs = defaults.DefaultProgressSignalThresholdMs
+	}
+	if o.OTelSelfStatsIntervalS == 0 {
+		o.OTelSelfStatsIntervalS = defaults.DefaultOTelSelfStatsIntervalS
+	}
```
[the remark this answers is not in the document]
solace-aross: Valid. Changed the `== 0` checks to `<= 0` in `applyObservabilityDefaults` so a stray negative value from YAML is treated as "omitted" and re-defaulted rather than surviving into a future consumer, and added a regression test (`TestObservability_NonPositiveNumericsAreReDefaulted`). Fixed in 0003153.

docs/plans/2026-06-26-obs-story1-story7.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Story 1 — scope
+
+New packages, each with an `Enabled() bool` stub gated on its flag:
+`internal/observability/{correlation,recover,metrics,audit,tracing,health}/`, plus `internal/observability/schema/version.go` (`MetricsSchemaVersion`, `AuditSchemaVersion = "1.0"`).
+
+- `internal/auth/principal.go` — `PrincipalKey{}` context-key type, empty `Principal` struct, `PrincipalFrom(ctx) Principal` (population deferred to Story 20).
+- `internal/config/observability.go` — `ObservabilityConfig`: env-driven `OBS_*_ENABLED` flags with door-closing v1 defaults (correlation + panic-recovery on; metrics, audit, tracing, saturation off; auth-failure-counter follows metrics; **no** `OBS_READYZ_STRICT_ENABLED`), and numeric thresholds (`saturation_threshold_ms`=10, `progress_signal_threshold_ms`=5000, `otel_self_stats_interval_s`=60) as `${VAR}`-substitutable YAML fields.
+- Wire into `internal/config/config.go`, add constants to `internal/defaults/defaults.go`, and log `observability: enabled=[...]` at INFO in `cmd/server/main.go` startup.
+- Unit tests: flag defaults, env overrides both directions, principal zero-value.
```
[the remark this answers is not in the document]
solace-aross: Right. The code intentionally uses an unexported `principalKey{}` (the correct context-key idiom, matching `rawSubjectTokenKey` in `raw_subject_token.go`), so I aligned the plan doc to the code rather than the reverse, and corrected the startup-log-string reference. Fixed in 0003153.

helen-zhu-solace (commented):
**`internal/observability/recover/` — package name shadows the Go builtin** (`internal/observability/recover/recover.go`)

Go's builtin `recover()` is what panic-recovery middleware needs to call. Inside a package named `recover`, the identifier `recover` refers to the package, not the builtin — so the future middleware that lands here will hit a name collision and have to alias the import or work around it. Worth renaming now while there are no consumers; `recovery` or `panicrecover` both work. If `recover` was a deliberate choice, a one-line note in the package doc would help.

**Test env-var leakage** (`internal/config/observability_test.go`)

`TestObservability_FlagDefaults` asserts the door-closing defaults without first clearing the `OBS_*` env vars. If a developer or CI runner happens to have any of them exported, the test fails with no obvious cause. `t.Setenv` only sets and restores — to unset for the test you need `os.Unsetenv` paired with `t.Cleanup`. A small helper that wipes the whole `OBS_*` set at the top of any default-asserting test would prevent the whole class of flake.

internal/config/observability.go
```diff
@@ -0,0 +1,144 @@
...
+	envObsTracingEnabled            = "OBS_TRACING_ENABLED"
+	envObsSaturationEventsEnabled   = "OBS_SATURATION_EVENTS_ENABLED"
+	envObsAuthFailureCounterEnabled = "OBS_AUTH_FAILURE_COUNTER_ENABLED"
+)
+
+// envBool reads name from the environment and parses it as a boolean. When the
+// var is unset (or set to a value strconv.ParseBool rejects), it returns def.
+// This keeps an operator typo from silently flipping a capability — a bad value
+// falls back to the documented default rather than to Go's zero value.
+func envBool(name string, def bool) bool {
```
amitmorade, line 87: `envBool` silently falls back to the default when an operator sets an `OBS_*` var to an unparseable value (e.g. `OBS_METRICS_ENABLED=yse`).

The "unset" and "set but invalid" cases both hit `return def`, but they mean different things:
- **Unset** → operator didn't configure this (correct to use default)
- **Set but invalid** → operator tried to configure this and we silently ignored them

The startup log shows the resolved flag values but doesn't distinguish between the two, so there's no way to tell your override was dropped.

For comparison, `MCP_SERVER_PORT` fails hard on a bad value in the same `applyEnvOverrides` phase.

Not a blocker for the skeleton since nothing consumes the flags yet, but curious whether a `slog.Warn` on the bad-parse path (or a hard error) would be worth adding before real behavior lands behind these flags.

internal/observability/audit/audit.go
```diff
@@ -0,0 +1,29 @@
...
+// policy) — operators opt in. Emitted records will carry
+// schema.AuditSchemaVersion.
+package audit
+
+import "github.com/SolaceDev/solace-broker-mcp/internal/config"
+
+// Enabled reports whether the audit log is turned on, reading the
+// OBS_AUDIT_LOG_ENABLED flag off the loaded config. Later wiring consults this
+// before emitting audit records.
+func Enabled(cfg *config.ServerConfig) bool {
```
amitmorade: The `Enabled` accessors across the observability sub-packages (`audit`, `correlation`, `metrics`, `recover`, `tracing`, `health`) all take `*config.ServerConfig` to read a single bool:

```go
func Enabled(cfg *config.ServerConfig) bool {
    return cfg.Observability.AuditLogEnabled
}
```

This gives every observability package a mutable pointer to the entire server config — broker credentials, OAuth settings, port, write-tool flags, everything. Today these are one-liner accessors so the risk is theoretical, but as later stories build out middleware and emitters in these packages, that surface stays wide open.

One scenario worth considering: as AI-assisted development becomes more common across the team, a contributor debugging a tricky issue in the metrics package could end up with AI-generated code that mutates an unrelated config field — and it would compile cleanly. In a large PR with real behavioral changes, that kind of cross-concern mutation is easy to miss in review.

Narrowing the parameter to `*config.ObservabilityConfig` (or even just a `bool`) would make the Go compiler enforce the boundary. Code like `cfg.Port = 9999` wouldn't just be wrong — it wouldn't compile. The blast radius shrinks from "everything the server knows" to "just the observability flags and tunables."

I recognize that passing `*ServerConfig` through helper functions is a pattern we've used elsewhere in the codebase — so this isn't a departure from what's already here. But as the observability tree grows (6+ packages, each with growing logic), this might be a good place to start tightening that pattern. The skeleton is the cheapest point to make this change since no consumers exist yet.

Curious what you think — would narrowing to `*ObservabilityConfig` work for the accessor signatures, or is there a reason they'll need the full config later?

docs/superpowers/plans/2026-06-26-obs-story1-story7.md
amitmorade, line 1: This file should be moved under docs/superpowers/plans

internal/auth/principal_test.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+// TestPrincipalFrom_RoundTrip is defensive: only this package can construct
+// principalKey{}, so this guards against a future bug in this package that
+// stores a Principal and then fails to read it back. Today Principal is empty,
+// so the round-trip value equals the zero value, but the path (value present →
+// accessor returns it) is exercised.
+func TestPrincipalFrom_RoundTrip(t *testing.T) {
+	t.Parallel()
```
amitmorade, line 51: This test exercises a different code path than `TestPrincipalFrom_EmptyContext` — test 1 hits the type-assertion failure branch (`ok == false`), while this one hits the success branch (`ok == true`). So the code being executed is genuinely different.

The gap is in the assertion: since `Principal` is an empty struct, both branches produce `Principal{}`, so the assertion can't distinguish "found the value on the context and returned it" from "type assertion failed, returned zero value." If someone broke the success branch (e.g., hardcoded `return Principal{}` without the type assertion), both tests would still pass green.

When the story that adds fields to `Principal` lands, this test will need updating with distinct field values to actually prove the success branch works. The risk of keeping it as-is: a future contributor sees `TestPrincipalFrom_RoundTrip` passing and assumes the round-trip path is verified, so the real test with meaningful assertions never gets written.

Two options, curious which you'd prefer:

1. Remove this test now, add a proper round-trip test when `Principal` gets fields (with distinct field values that actually prove the type assertion succeeded, not just returned zero value)
2. Keep it as-is — just noting that the assertion provides no additional signal until fields exist

solace-aross: Good call — you're right that with `Principal` empty, the round-trip assertion can't distinguish the value-present branch from the type-assertion-failure branch (both yield `Principal{}`). I kept the test (it still exercises the success-branch path and documents the accessor contract) and made the limitation explicit in the doc comment, plus added a `TODO(Story 20 – principal population)` to set distinct non-zero field values once `Principal` gains fields — so the real assertion gets written rather than forgotten. Fixed in 075ddf9.

amitmorade (approved):
LGTM - all feedback/comments are addressed, thank you!

helen-zhu-solace (approved):
LGTM as well!

PR #116, opened 2026-06-26
SOL-151283: /livez endpoint with /health as alias

solace-aross:
Part of the **Broker MCP Observability & Telemetry** epic ([SOL-150251](https://sol-jira.atlassian.net/browse/SOL-150251)) → MUST child epic [SOL-149791](https://sol-jira.atlassian.net/browse/SOL-149791). Story: [SOL-151283](https://sol-jira.atlassian.net/browse/SOL-151283).

> **Stacked on #115** (Story 1 skeleton). This PR targets the `aross/sol-151278-observability-skeleton` branch; review #115 first. It will retarget to `main` once #115 merges.

## What this does
Adds a dedicated **`/livez`** process-alive endpoint and makes **`/health`** a body-identical alias of it.

- `GET /livez` → `200 {"status":"alive"}`; non-GET → `405`.
- `/health` becomes a true alias (same handler instance), so its body changes from `{"status":"healthy"}` to `{"status":"alive"}`.
- Handler lives in `internal/observability/health`; `/livez` is unconditional (not flag-gated).

## Compatibility (the body change)
Every in-repo consumer that asserts the `/health` **body** was reconciled: the two e2e shell suites and the internal docs/README. Status-only consumers are unaffected — the `--health` CLI flag and Docker `HEALTHCHECK` check the HTTP status only, and the k8s `httpGet` probes can't assert a body. K8s probes still point at `/health` here; repointing liveness→`/livez` is a later story.

## Testing
Unit tests for `/livez` (200 + body + 405) and `/health` alias-equivalence; a new `/livez` check in the health e2e suite. `go build`, `go vet`, `go test ./...`, `golangci-lint`, and `-race` all pass. Reviewed by a principal-architect pass (the review caught two e2e suites asserting the old body, now fixed).

[SOL-150251]: https://sol-jira.atlassian.net/browse/SOL-150251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-151283]: https://sol-jira.atlassian.net/browse/SOL-151283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,8 +177,8 @@ If the config file is named `broker-config.yaml` in the current directory, the s
 Verify:
 
 ```bash
-curl http://localhost:9090/health
-# {"status": "ok"}
+curl http://localhost:9090/livez
+# {"status":"alive"}
 ```
 
 The binary is statically linked with no external dependencies. It handles `SIGTERM` and `SIGINT` for graceful shutdown.
@@ -204,8 +204,8 @@ The container reads config from `/etc/mcp-server/config.yaml` by default. Pass t
 Verify:
 
 ```bash
-curl http://localhost:9090/health
-# {"status": "ok"}
+curl http://localhost:9090/livez
+# {"status":"alive"}
 ```
 
 The image includes a built-in Docker health check using the binary's `--health` flag (no shell or curl needed in the container). Check status with `docker inspect --format '{{.State.Health.Status}}' solace-broker-mcp`.
@@ -278,7 +278,7 @@ Create `broker-config.yaml` and `.env` in the repo root (both are gitignored). S
 go run ./cmd/server
 ```
 
-The MCP server listens on port `9090` by default and serves the MCP endpoint at `/mcp`. A health check endpoint is available at `/health`.
+The MCP server listens on port `9090` by default and serves the MCP endpoint at `/mcp`. A liveness endpoint is available at `/livez` (with `/health` retained as a backward-compatible alias).
 
 ### Configuration Options
 
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ Create `broker-config.yaml` and `.env` in the repo root (both are gitignored). S
 go run ./cmd/server
 ```
 
-The MCP server listens on port `9090` by default and serves the MCP endpoint at `/mcp`. A liveness endpoint is available at `/livez` (with `/health` retained as a backward-compatible alias).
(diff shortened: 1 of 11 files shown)


helen-zhu-solace (commented):
Nice clean diff and the alias-equivalence test is the right shape. One thing worth pushing on, plus two smaller notes.

### 1. `/health` body change vs. "backward-compatible alias" framing

The PR description and code comments call `/health` a "backward-compatible alias of `/livez`", but the body changed from `{"status": "healthy"}` to `{"status":"alive"}`. That's path-compatible, not body-compatible. A client doing something like `curl /health | jq -e '.status == "healthy"'` — perfectly reasonable code against the prior contract — now fails.

The justification given (e2e shells updated, `--health` reads status only, k8s `httpGet` can't read bodies) only covers consumers inside this repo. The reason to keep `/health` around at all is for *outside* consumers you can't see: user scripts, dashboards, monitoring checks, homegrown wrappers. For them, this is a silent break — docs promise compatibility, code breaks it.

Two clean options, either is fine, but pick one explicitly:

- **Keep `/health` body-compatible:** register a separate handler that still returns `{"status":"healthy"}`. ~5 lines and one test. "Backward-compatible" then actually means what it says.
- **Accept the break:** drop "backward-compatible" from the docs/comments, call `/health` a deprecated path alias with a removal milestone, and add a release note so external consumers see the body change.

What I'd push back on is the current middle state: the docs promise compatibility and the code breaks it. Someone's dashboard pages at 2am and nobody warned them.

### 2. `/ready` left inline while `/livez` moved to the `health` package

Minor architectural asymmetry — `/livez` is a clean package-level handler now, but `/ready` is still a big inline function in `main.go`, even though the `health` package doc explicitly names `/readyz` as the canonical readiness path. Fine to defer to a later story, but worth a one-line TODO at `/ready` pointing at the follow-up so the asymmetry doesn't get cemented.

### 3. README body drift predates this PR — flagging for context

Before this PR, the README documented `{"status": "ok"}` while the code actually returned `{"status": "healthy"}`. Docs and code had already drifted. The PR corrects the docs to `"alive"`, which is good. If you go with option 1 above (preserve old body), pick `"healthy"` — that's what the code actually shipped, so that's what real users will have integrated against.

amitmorade (dismissed):
LGTM as well

helen-zhu-solace (dismissed):

solace-awaheed (dismissed):
Nothing was flagged on my end, LGTM

amitmorade (approved):
LGTM

PR #120, opened 2026-06-29
SOL-151286: HTTP panic-recovery middleware (whole-mux)

solace-aross:
## What

Adds **whole-mux HTTP panic-recovery middleware** (`recovery.HTTPMiddleware`) — a `recover()` wrapper installed as the **outermost** layer around the entire HTTP mux. A panic in any handler (the `/mcp` chain, `/livez`, `/health`, `/ready`, the catch-all, future `/readyz` and `/metrics`) is caught, logged with a structured stack, and converted to a clean HTTP 500 — the process keeps running.

Part of the **Observability & Telemetry epic** ([SOL-149791](https://sol-jira.atlassian.net/browse/SOL-149791)). Story: [SOL-151286](https://sol-jira.atlassian.net/browse/SOL-151286).

## Why

Today a panic in a non-tool handler (auth, correlation, a probe) has no structured recovery. Wrapping the **whole mux** (not just the MCP chain) is the deliberate KA-review decision: a panic in `/readyz` that crashed the process would defeat the readiness design. Complements the existing tool-handler recovery (`withRecovery`, SOL-150685) which covers the SDK's own goroutine.

## How

- `recovery.HTTPMiddleware(next)` — `recover()` → ERROR log `event="panic_recovered"` with `panic_type` (Go type only, never the panic value's text) + `debug.Stack()`, then a clean `500 {"error":"internal_error","error_description":"..."}` (matches the repo's catch-all 404 body convention).
- Wired in `main()` as the outermost handler via the extracted `buildRootHandler`, gated by the existing `OBS_PANIC_RECOVERY_ENABLED` flag (default ON; when off, the middleware is not wired and panics propagate — today's behaviour).
- Sits **outside** `correlation.Middleware`; the 500 body deliberately omits `correlation_id` because the outer frame cannot see correlation's child context (documented in code).

## Tests

- Middleware: 500 + body shape; transparent pass-through; log shape proven to never leak the panic value's text; full panic-value domain (string/int/struct/pointer/error and `panic(nil)` → `*runtime.PanicNilError`); partial-write-then-panic (exact body pinned).
- Composition: `buildRootHandler` and the extracted `buildMCPEndpoint` are driven by the **real** production functions, so the chain order (recovery outermost; correlation inside, on `/mcp`) is pinned — a reorder or dropped correlation breaks the test (AC #5).
- `make check` green (build + vet + golangci-lint + `-race`).

## Notes / scope

- **Stacked on #118** (Story 2, correlation middleware) — base is `aross/sol-151279-correlation-middleware`, not `main`. The composition test asserts the chain *with* correlation present, so it's built on that branch. Will retarget to `main` once #118 merges.
- **Recovered-panic counter deferred to Story 15** (metrics infra doesn't exist yet) — log-only for now, with a TODO. No metrics API invented.
- **Known limits (documented in code):** panics on handler-spawned goroutines (e.g. `/ready` probes) are out of scope here; a partial-write-then-panic yields a best-effort concatenated body (wrapping the ResponseWriter to prevent it would break `http.Flusher`-based SSE on `/mcp`).

[SOL-151286]: https://sol-jira.atlassian.net/browse/SOL-151286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


helen-zhu-solace (commented):
## 1. `http.ErrAbortHandler` gets swallowed — please fix

**Where:** `internal/observability/recovery/middleware.go`, inside the `recover()` block.

**What's happening:** Go's `net/http` has a special sentinel value, `http.ErrAbortHandler`. When a handler panics with this value, it's a documented signal that means "abort this request, don't log anything, don't write a response." Your middleware catches it like any other panic and turns it into:
- a structured ERROR log saying a panic was recovered (false alarm), and
- a 500 response with a JSON body (defeats the "don't write a response" intent).

**Why it matters:** `/mcp` runs the MCP SDK's streaming/SSE path. If the SDK ever uses `ErrAbortHandler` for client-disconnect or shutdown — now or in a future version — we'll start emitting spurious 500s and noisy "panic_recovered" alerts that aren't real panics.

**Suggested fix** (one line, re-raise the sentinel so net/http handles it as designed):

```go
if rec := recover(); rec != nil {
    if rec == http.ErrAbortHandler {
        panic(rec)
    }
    // existing log + 500 code stays as-is
}
```

Worth a small test that asserts a handler panicking with http.ErrAbortHandler is NOT logged and does NOT produce a 500 body.

## 2. Partial-write test pins a malformed response as "expected"

Where: TestHTTPMiddleware_PartialWriteThenPanic in middleware_test.go.

What's happening: If a handler writes some body bytes and then panics, the middleware appends its error JSON onto those bytes. The wire output ends up looking like partial-body-{"error":"internal_error",...} — invalid JSON. Your test pins this exact string as the expected output.

I understand the rationale you wrote in the code: in-tree handlers write atomically today, and wrapping the ResponseWriter would supposedly break the http.Flusher type assertion that SSE needs on /mcp. The second point isn't quite right anymore — Go 1.20 added http.NewResponseController and the Unwrap() http.ResponseWriter convention specifically so middleware can wrap the writer and still preserve Flusher/Hijacker. That's the standard fix for exactly this problem.

Two concerns, in order:

1. Today's behavior on the wire is malformed JSON. "Atomic in-tree handlers" is true now but is a footgun for any future contributor who adds a streaming handler.
2. The test cements the bad behavior. A future PR that fixes this would have to delete the assertion — which makes the intent unclear. A test asserting "no append once a body has been committed" would describe the behavior we actually want.

Not a hard blocker — your rationale is documented — but I'd recommend either fixing it now with ResponseController, or at minimum changing the test to assert the recovered-panic-is-logged-and-process-survives part without pinning the malformed body as gospel.

## 3. Typo

middleware.go:34 — "" , has a stray space before the comma in the doc comment. Trivial.

solace-aross:
Thanks Helen — all three addressed in `c0b31de`.

1. **`http.ErrAbortHandler`** — fixed. The recover block now re-raises the sentinel as its first action, before any log or write, so net/http applies its documented suppression (no spurious 500, no false `panic_recovered`). Added a test asserting a handler panicking with `http.ErrAbortHandler` produces no log and no body and the panic escapes.

2. **Partial-write / wrapper** — you're right, and my rationale was wrong. I checked: the SDK flushes via `http.NewResponseController` (`mcp/streamable.go:1056`, `mcp/event.go:57`), not a direct `http.Flusher` assertion, so an `Unwrap()`-aware wrapper preserves SSE exactly as you said. The middleware now wraps the writer in a small `recoveryWriter` that tracks commitment and exposes `Unwrap()`; on a committed response it writes nothing (no more malformed append), and on the clean path it still returns the 500. The test now asserts the body is exactly the partial bytes with the panic logged and the process alive, and a separate test pins that `ResponseController.Flush()` still works through the wrapper (it fails if `Unwrap()` is removed). Removed the incorrect "wrapping breaks Flusher" comment.

3. **Typo** — fixed.

CI is green. #121 is rebased on top and keeps its approval.

internal/middleware/recovery/middleware.go
amitmorade, line 1: Two architectural concerns I wanted to raise — curious what you think:

**1. Package location:** `recovery` is under `internal/observability/` but its job is keeping the server alive (panic → clean 500), not observing the system. Correlation, metrics, tracing observe without changing behavior — recovery changes the outcome of a request. This should live in `internal/middleware/recovery/` or similar. Once it's in `observability/`, it sets the precedent — anything with an `OBS_*` flag lands there, and the package becomes a catch-all.

**2. Flag-gating:** Is there a real production scenario where an operator would want `OBS_PANIC_RECOVERY_ENABLED=false` — meaning one bad request crashes the process? Correlation/metrics/tracing have legitimate "off" reasons (overhead, missing infrastructure). Recovery has none. This should be unconditional — always on, no flag. Adding a flag now means removing it later is a config-breaking change.

The implementation itself is solid — the `recoveryWriter`, `ErrAbortHandler` exemption, secure logging, and test coverage are all well done. These two concerns are about where it lives and how it's wired, not the code quality.

amitmorade (changes requested):

internal/middleware/recovery/middleware.go
solace-aross: Agreed on both, and done.

1. Package: moved `recovery` from `internal/observability/` to `internal/middleware/recovery/`. You are right that it changes request behavior (panic to clean 500) rather than observing, so it should not sit with the observe-only packages.

2. Flag: removed entirely. `OBS_PANIC_RECOVERY_ENABLED` is gone (config field, env const, default, startup logs, and the `Enabled` accessor) and `HTTPMiddleware` is now wired unconditionally. There is no production reason to let one bad request crash the process, `debug.Stack()` already gives the diagnostic without dying, and a never-off flag would only be a footgun plus a config-breaking removal later.

This also means the tool-path PR (#121) is now moot — `withRecovery` was already unconditional from SOL-150685, so there is nothing left to gate. Closing it. Both in commit 843f822. Repo-wide grep for the flag/old path is clean.

internal/middleware/recovery/middleware_test.go
```diff
@@ -0,0 +1,308 @@
...
+// the middleware recovers regardless of the panic value's type, and panic_type
+// reflects the concrete Go type each time.
+func TestHTTPMiddleware_NonStringPanicValues(t *testing.T) {
+	// Not parallel: mutates the process-global slog default.
+	type custom struct{ Field int }
+	cases := []struct {
+		name      string
+		value     any
+		wantType  string
+		wantPanic bool // whether next actually panics (nil panic does not)
```
amitmorade, line 154: `wantPanic bool` in the test case struct is declared and set to `true` on every row, but never read in the test body — every case unconditionally runs through `panicHandler` and asserts a 500 regardless of the field's value.

A future contributor could add a case with `wantPanic: false` expecting the test to skip the recovery assertion — and it silently wouldn't. Could we remove the field and the `wantPanic: true` entries from each case to avoid that confusion?

amitmorade (approved):
LGTM

solace-aross:
Superseded by #122. GitHub auto-closed this PR when the correlation branch (#118) merged and its branch was deleted. The panic-recovery work continues in #122 (same commits, rebased onto main, all review feedback from here folded in). Thanks for the reviews — continuing there.

PR #121, opened 2026-06-29
SOL-151287: gate tool-path panic recovery behind OBS_PANIC_RECOVERY_ENABLED

solace-aross:
## What

Gates the existing tool-path panic recovery (`withRecovery`) behind the single **`OBS_PANIC_RECOVERY_ENABLED`** flag, so one flag now governs panic recovery on **both** the HTTP path (Story 12, #120) and the tool path.

Part of the **Observability & Telemetry epic** ([SOL-149791](https://sol-jira.atlassian.net/browse/SOL-149791)). Story: [SOL-151287](https://sol-jira.atlassian.net/browse/SOL-151287).

## Why

Tool-path recovery already shipped under SOL-150685 (the SDK runs each tool handler on its own goroutine with no `recover()`, so one panicking tool would otherwise crash the process). It was unconditional. The KA decision is a single flag for both paths with no per-tool sub-switches, so this brings the tool path under that flag:

- **ON (default):** unchanged — panic recovered, sanitized `IsError` result, logged with type + stack.
- **OFF:** the wrapper is not installed and a panic propagates (reintroducing the pre-SOL-150685 crash). A deliberate dev/debug affordance to get the real crash stack. Default-ON preserves the safety fix.

## How

- `withRecovery(toolName string, panicRecoveryEnabled bool, h mcp.ToolHandler)` — when `false`, returns `h` unchanged.
- `main()` computes `recovery.Enabled(cfg.Observability)` once and threads it into `RegisterWithServer` and `RegisterListBrokers`. No `internal/tools` → `recovery` coupling.
- Fixed the now-stale `manager.go` comments so they're accurate whether the flag is on or off (the `CallTool` defer still audits `error_type=panic` during unwinding either way; the panic then either gets sanitized by `withRecovery` or propagates).

## Tests

- Flag ON: panic → sanitized `IsError` result, logged; swept across string/int/struct/error and a real nil-deref panic; raw value never leaks.
- Flag OFF: deferred `recover()` asserts the panic actually propagates (original value), not just absence of an error result.
- Both `RegisterWithServer` and `RegisterListBrokers` honor the gate; `TestPanicAuditedAsError` still confirms the manager defer records the panic.
- `make check` green (build + vet + golangci-lint + `-race`).

## Notes / scope

- **Stacked on #120** (Story 12) — base is `aross/sol-151286-http-panic-recovery`. Merge #118 → #120 → this, in order.
- **Counter deferred to Story 15** (no metrics infra yet) — `TODO` only, no fabricated API.
- **Goroutine-safety AC is forward-looking:** the OTel exporter and self-stats emitter goroutines are later tracing stories and don't exist yet. Documented the discipline (every future background goroutine installs its own `recover()`); nothing wired now.

[SOL-151287]: https://sol-jira.atlassian.net/browse/SOL-151287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -41,6 +41,7 @@ import (
 	"github.com/SolaceDev/solace-broker-mcp/internal/defaults"
 	"github.com/SolaceDev/solace-broker-mcp/internal/observability/correlation"
 	"github.com/SolaceDev/solace-broker-mcp/internal/observability/health"
+	"github.com/SolaceDev/solace-broker-mcp/internal/observability/recovery"
 	"github.com/SolaceDev/solace-broker-mcp/internal/semp"
 	"github.com/SolaceDev/solace-broker-mcp/internal/semp/sempv2"
 	"github.com/SolaceDev/solace-broker-mcp/internal/semp/sempv2/specs"
@@ -160,6 +161,27 @@ func buildMux(aliases func() []string, probeBroker func(ctx context.Context, bro
 		_, _ = w.Write(resp)
 	})
 
+	return mux
+}
+
+// buildRootHandler returns the outermost HTTP handler for the server: the mux
+// wrapped with panic-recovery as the OUTERMOST layer (ADR-001 chain ordering),
+// gated on OBS_PANIC_RECOVERY_ENABLED. When the flag is off the mux is returned
+// unwrapped and panics propagate (today's behaviour).
+//
+// Recovery wraps the WHOLE mux, so it covers every route — the standalone
+// /livez, /health, /ready probes (and future /readyz, /metrics), the catch-all
+// 404, and the authenticated /mcp chain — sitting OUTSIDE the per-route
+// correlation.Middleware that main() installs only on /mcp. The two are
+// different layers and never conflict: recovery wraps the mux; correlation
+// wraps a handler inside it.
+//
+// Extracted from main() so the composition test can assert the assembled chain
+// order against the real wiring rather than a hand-rebuilt copy.
+func buildRootHandler(mux *http.ServeMux, obs config.ObservabilityConfig) http.Handler {
+	if recovery.Enabled(obs) {
+		return recovery.HTTPMiddleware(mux)
+	}
 	return mux
 }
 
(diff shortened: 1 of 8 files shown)


helen-zhu-solace (approved):
LGTM, Non-blocking comments:
## 1. Two adjacent booleans in RegisterWithServer — easy to swap at call sites (medium)

`internal/tools/register.go:108`

`func RegisterWithServer(mgr *ToolManager, server *mcp.Server, pool *semp.BrokerPool, enableWriteTools bool, panicRecoveryEnabled bool)`

At call sites this becomes `RegisterWithServer(mgr, server, pool, true, true)` (see `register_test.go:113, :167`). Two trailing bools of the same type are a classic source of silent-swap bugs — there's nothing the compiler can tell you if someone flips them.

Two low-cost options:
- A small` RegisterOptions` struct: `RegisterWithServer(mgr, server, pool, RegisterOptions{EnableWriteTools: ..., PanicRecoveryEnabled: ...})`. Named at every call site.
- Or pass `cfg.Observability` (the whole config) and call `recovery.Enabled(...)` inside `register.go`. The PR description argues against this to keep `internal/tools` decoupled from recovery, which is fair — but if you ever add a third feature flag (correlation, audit on/off, metrics on/off) this signature gets worse fast.

## 2. `TestRegisterWithServer_PanicGateOff / TestRegisterListBrokers_PanicGateOff` are near-no-ops (low)

`internal/tools/register_test.go:284, :294`

Both tests, by their own comments, verify only that "registration with the flag OFF is wired without our recover" — i.e. that the code compiles and doesn't panic on construction. The real behavior is already covered by `TestWithRecovery_DisabledPropagates`. These tests will pass even if `panicRecoveryEnabled` were silently ignored at registration. Consider either deleting them or strengthening them to actually call the registered handler through server.AddTool and assert propagation end-to-end.

## 3. runtime error case uses value: nil as a sentinel (low)

`internal/tools/register_test.go:225-233`

The table has `{"runtime error", nil}` but the body special-cases on `tc.name == "runtime error"`. Reading the table you'd expect a panic on nil, then the code branches on the name string. Either drop it from the table and make it its own `t.Run`, or carry a discriminator field. Minor readability nit.

amitmorade (changes requested):
Question on the overall direction here — `withRecovery` already shipped unconditional under SOL-150685 and it logs the Go type (`panic_type`) plus the full stack trace (`debug.Stack()`) in structured logs. So today, when a tool handler panics, the server survives and the operator gets the same diagnostic information they'd get from a crash.

This PR adds a flag to disable that, and the justification is a "dev/debug affordance to get the real crash stack." But given that `debug.Stack()` already captures the full stack without crashing — is there a scenario where an operator actually needs the process to die to debug a panic?

I also left a comment on #120 about whether the flag should exist at all on the HTTP path. If we end up agreeing it should be unconditional there, this PR's core change might not be needed — `withRecovery` was already doing the right thing.

Happy to be wrong if there's a use case I'm not seeing.

solace-aross:
Closing this — your call on #120 settles it, and you're right that the tool path was already handled.

Decision (with the PO): panic recovery becomes **unconditional** on both paths, so `OBS_PANIC_RECOVERY_ENABLED` is removed entirely (#120, commit 843f822). That makes this PR moot: `withRecovery` already shipped unconditional under SOL-150685 and logs `panic_type` + `debug.Stack()`, so a tool panic already survives with full diagnostics. Story 13's only change was to *gate* that behind the flag — and with no flag, there's nothing to do. Leaving `withRecovery` exactly as it is.

SOL-151287 will be closed as delivered-by-SOL-150685. Thanks for pushing on this — it removed a footgun and a PR.

PR #124, opened 2026-06-30
SOL-148460: Implement Core Management tools (VPN, Queue)

solace-wmaryam:
## Summary

This PR adds support for three different categories of management tools:

1. For Message VPNs:
  - create-message-vpn
  - update-message-vpn
  - delete-message-vpn

2. For Queues:
  - create-queue
  - update-queue
  - delete-queue

3. For Topic-endpoints:
  - create-topic-endpoint
  - update-topic-endpoint
  - delete-topic-endpoint

The tools are defined in yaml to leverage the automatic yaml parsing already available for read tools. According to the story specifications, we wanted users to have the flexibility of providing any desired config parameters when creating and updating objects. Hence, this PR adds additional logic to accept and parse a config object (i.e. msgVpnConfig, queueConfig, topicEndpointConfig) specified in yaml parameters, along with any required parameter names defined in the usual way (msgVpnName, queueName etc). 

Fixes #https://sol-jira.atlassian.net/browse/SOL-148460

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

To support broker management workflows, the MCP server needs configuration capability — provisioning and tearing down Message VPNs, queues, and topic endpoints over SEMPv2 config. This change adds those core management tools and the executor scaffolding required to turn tool inputs into SEMPv2 write request bodies, while guarding the mutative operations behind explicit user-confirmation wording.

## Changes Made

- Enabled SEMPv2 config spec support: registered the __private_config__ basePath as a valid spec type and embedded the private config Swagger spec (semp-v2-swagger-private-config-extended.json) so config operations resolve from the operation catalog.
- Added request-body construction in the composite executor: new constructRequestBody helper assembles the body for write operations — spreading object params (e.g. msgVpnConfig) into individual fields and setting scalar params directly, while skipping params the operation takes from the path/query.
- Added create/update/delete tools for Message VPNs (create-message-vpn, update-message-vpn, delete-message-vpn) as composite YAML definitions.
- Added create/update/delete tools for queues and topic endpoints (create-queue, update-queue, delete-queue, create-topic-endpoint, update-topic-endpoint, delete-topic-endpoint).
- Added confirmation wording to all write tool descriptions: each tool instructs the model to obtain explicit, separate user confirmation (restating broker/VPN/object target and effect) before invoking, and marks delete tools as destructive: true.

## Testing

**Test Environment:**
- Go version: go1.25.0
- OS: Linux 6.17.10-100.fc41.x86_64 (Fedora 41)
- Broker version (if applicable): 

**Tests Performed:**

**Test Coverage:**

Unit tests:
- loader_test.go: validates that the new VPN/queue/topic-endpoint tool definitions load, parse, and expose the expected parameters and annotations.
- executor_test.go: covers constructRequestBody — object params spread into the body, scalar params set directly, path/query params excluded from the body — and end-to-end execution of the create/update/delete steps against a fake client.
- operation_test.go: covers parsing of the __private_config__ spec type and its normalization to config.

Live broker testing: 
  - ran all 9 tools end-to-end (HTTP 200); confirmed updates PATCH only the supplied fields and create routes the name into the body. 
  - Verified broker-side rejections surface as tool errors (duplicate create, delete-nonexistent, delete-VPN-with-children, unknown attribute, bad enum, wrong type) and client-side rejections (missing required param, unknown broker)
  - Found that a name supplied in both the param and the config object resolved nondeterministically on the broker — now rejected before the call and covered by a unit test.
 
## Breaking Changes

N/A


internal/composite/definitions/tools.yaml
```diff
@@ -407,3 +407,305 @@ tools:
...
+      attributes are left unchanged. Use this to update VPN settings such as
+      enabled state, connection limits, or spool quotas.
+      BEFORE INVOKING THIS TOOL, obtain explicit user confirmation, restating
+      the target (broker, VPN name) and the attributes that will change. The
+      confirmation MUST come as a separate user reply received AFTER you state
+      the target and effect — do not treat the user's original request as a
+      pre-confirmation, even if it implies intent.
+    annotations:
+      readOnly: false
+      destructive: false
```
solace-aross: `update-*` tools are marked `destructive: false`, but they're service-affecting (medium). This applies to all three update tools (here, line 548 for update-queue, line 657 for update-topic-endpoint).

The MCP `destructiveHint` contract is roughly "may perform destructive updates vs. additive-only." `update-message-vpn` with `enabled: false` drops every client on the VPN; `update-queue` with `egressEnabled: false` halts delivery; lowering a spool quota can evict messages. Clients that surface `destructiveHint` as a confirmation prompt won't prompt for these.

`create` as `destructive: false` is correct (additive) and `delete` as `true` is correct. I'd flip the three update tools to `destructive: true`, or consciously document why partial updates are treated as non-destructive. Your call, but the current setting under-warns.

solace-aross (commented):
Strong PR. The design reuses the composite-YAML mechanism cleanly, `constructRequestBody` is well-scoped and well-commented, and the unit tests are thorough (path-param exclusion, scalar-vs-object spreading, no-body-on-delete, even asserting the `float64(100)` numeric type). Secure-by-default gating is reused correctly: every new tool sets `readOnly: false`, so `isWriteTool` keeps them behind `enable_write_tools`, and `broker` is stripped from params before body construction (executor.go:89), so it can't leak into the SEMP body.

Two things I'd want resolved before merge, plus a line note and a couple of optional items.

## 1. The write path has never touched a real broker (high)

The checklist leaves "Tested with real Solace broker" and "E2E tests added" unchecked; everything is validated against a fake client. This is the first code that builds SEMPv2 config request bodies, and that path has failure modes a fake client won't surface: SEMP's strict attribute typing, how integral float64 config values serialize, rejection of unknown attributes, PATCH-vs-PUT semantics on update. For provisioning and deletion tools, passing unit tests is a weaker guarantee than usual.

At minimum, run one manual smoke test per verb against a real broker (create, update, delete a throwaway VPN/queue/topic-endpoint) and note the broker version in the PR. Better, add a management E2E suite alongside the existing ones (docs/internal/e2e-testing.md).

## 2. Security-relevant docs now understate what `enable_write_tools` exposes (high)

Three docs enumerate the write-tool set explicitly and are now stale:

- docs/user-guide.md:94 says "17 read-only tools plus 4 action tools (21 total)". It's now 17 + 13 = 30.
- docs/configuration.md:41 lists the write tools as exactly the four delete/clear/disconnect tools.
- README.md:87 says "All four are write tools gated behind `enable_write_tools`."

This is more than doc drift: these tables describe the security posture. An operator deciding whether to set `enable_write_tools: true` would not learn that the flag now also exposes `delete-message-vpn`. Please update all three in this PR.

## Optional notes (non-blocking)

- Confirmation wording is a nudge, not a control. The "obtain explicit user confirmation" prose is defense-in-depth on the model, not enforcement. The real controls (`enable_write_tools` + `destructiveHint`) are correctly in place; just don't let it read as a guarantee.
- Body-key collision is order-nondeterministic. If the model ignores "do not include msgVpnName/queueName here" and also puts the name inside the config object, the winner depends on Go map iteration order (executor.go:486-498). Same value in practice, low risk, but a deterministic precedence would remove the sharp edge.
- Body is built from raw `execCtx.Params`, bypassing ResolveArgs/templating. Fine for these tools; worth noting a future write tool couldn't template a body field or pull one from a prior step result.

solace-awaheed (commented):
This isn’t blocking, but it’s worth noting for future write tools. I noticed that when using the create tools, we currently match the broker behaviour (leaving whatever was created disabled), but not the WebUI behaviour (enabled by default). For consistency, I think we should continue matching the broker behaviour.

Leaving this as a comment since I don’t think it should be a blocker, but it would be good to get input from @solace-aross and others to confirm whether that aligns with the intended behaviour.

(Still currently testing if anything else comes up)

internal/composite/definitions/tools.yaml
solace-wmaryam: I wasn't entirely sure if update tools would be considered destructive but you make a good point - I'll change the destructive field for these tools to true.

amitmorade:
Taking a look at this PR

amitmorade:
### `constructRequestBody` — future-proofing against tool params that shouldn't reach the SEMP body

Not blocking, and every tool in this PR is safe today. But I want to flag a latent trap while `constructRequestBody` is fresh, so the next contributor doesn't have to rediscover it.

**The trap**

`constructRequestBody` decides what goes into the body by process of elimination: any param not declared as `path`/`query`/`header` on the SEMP operation is treated as a body field (scalar) or spread into the body (object). That works today because every tool-level param on the new write tools maps to a real SEMP field.

But there's nothing preventing a future write tool from declaring a **tool-only** param — a `dryRun` flag, a `confirmToken`, a client-side `maxResults`-style knob — that SEMP doesn't know about. That value would silently land in the SEMP body, the broker would reject it with an "unknown attribute" 400, and the author would have to trace the failure back through the executor to figure out why. The invariant "every tool param maps to a SEMP field" only lives in a maintainer's head right now.

**Why I think it's worth doing here rather than later**

- Retrofitting later means auditing every write tool that exists at that point.
- The fix is cheap: the `op.Parameters` loop is already the right place. A follow-up option would be to teach `sempv2.ParseSpecs` to also capture the body schema's field names, so `constructRequestBody` can whitelist known body fields and reject unknowns with a precise error at execute time (rather than a broker 400).

Totally your call whether this is worth doing in #124 or as a follow-up — flagging it now so the decision is explicit either way.

solace-aross (dismissed):
Reviewed the changes since my 6/30 pass. Both items I'd flagged as resolve-before-merge are addressed, and the executor hardening closes Copilot's request-body concern. Approving.

**Resolved**

- **Real-broker testing.** The PR now documents all nine tools exercised end-to-end (HTTP 200), updates PATCHing only the supplied fields, and both broker-side (duplicate create, delete-nonexistent, delete-VPN-with-children, bad enum/type) and client-side rejections surfacing as tool errors. That was my main gap; the provisioning and delete paths have now touched a real broker.
- **Security docs.** user-guide.md (30 total = 17 read-only + 13 write), configuration.md, and README all now enumerate the full write-tool set, including delete-message-vpn. An operator can now see what `enable_write_tools` actually exposes.
- **constructRequestBody.** It now builds `nonBodyParams` and (a) skips them at the top level, (b) rejects a path/query name placed inside a config object, and (c) fails loudly when a field is set twice. That closes both Copilot's leak path and the order-nondeterministic collision I noted.
- **destructive flags.** All three `update-*` tools now carry `destructive: true`; creates stay additive.

**Non-blocking follow-up**

The manual smoke test is a reasonable bar for this PR, but a management E2E suite alongside the existing ones is worth a ticket so these write paths stay covered as the config surface grows.

On Ashar's disabled-by-default point: matching broker behaviour (created objects land disabled) is the right call over mirroring the WebUI default. Agreed with leaving it as is.

Nice work on this.

internal/composite/definitions/tools.yaml
```diff
@@ -407,3 +407,305 @@ tools:
...
+      - id: updateQueue
+        operation: config/updateMsgVpnQueue
+        args:
+          msgVpnName: "{{.Params.msgVpnName}}"
+          queueName: "{{.Params.queueName}}"
+    result:
+      strategy: collect
+
+  - name: delete-queue
+    description: >
```
amitmorade, line 572: ### A few concerns on the confirmation protocol

Reading all 9 write-tool descriptions together, I have a few things I'd like to think through — not blockers for #124.

My main concern is that I'm not sure the confirmation flow has actually been exercised end-to-end against the LLM. The PR mentions broker-side rejection testing, but I couldn't tell whether we've verified that the model actually pauses and waits for a separate user reply when asked something like "create a VPN named X". If that hasn't been checked, the "BEFORE INVOKING THIS TOOL..." block is doing nothing at runtime — the wording is right but the guardrail may not fire.

There's a real gotcha in this though — asking for confirmation before a tool call is already the default behaviour in Claude Code's ask mode, so testing there will always look fine. The interesting question is whether it still holds in **auto mode**, where Claude Code doesn't naturally pause for tool approval. If the description-embedded protocol is what has to carry the weight in auto mode, that's the case we actually need to verify.

Related — I'm also wondering how this holds up when the client isn't Claude Code. From my last work (roughly 4 months ago), SAM didn't have a native "ask the user before invoking a tool" capability the way Claude Code does — tool calls were more fire-and-forget. Might have changed since. If it hasn't, an instruction in the description telling the LLM to "obtain explicit user confirmation" won't behave the same on SAM as on Claude Code. Happy to keep this out of #124 and treat it as a follow-up, just wanted to flag it.

README.md
```diff
@@ -84,7 +84,10 @@ The server exposes read-only tools grouped by what they inspect, plus a small se
...
 | REST Delivery Points | `list-rdps`, `get-rdp-status` | List RDPs; inspect bindings, REST consumers, and last failure reason |
 | Discards | `get-discard-stats`, `list-queue-discards` | Broker-wide and per-VPN discard aggregates; per-queue discard counters |
-| Actions | `delete-queue-messages`, `clear-queue-stats`, `disconnect-client`, `clear-client-stats` | One tool per action. Destructive tools (`delete-queue-messages`, `disconnect-client`) are annotated `destructiveHint` and instruct the LLM to obtain user confirmation before invocation; the `clear-*-stats` tools are non-destructive. **All four are write tools gated behind `enable_write_tools: true` in the config — default off; not registered in `tools/list` when disabled.** |
+| Actions | `delete-queue-messages`, `clear-queue-stats`, `disconnect-client`, `clear-client-stats` | One tool per operational action. Destructive tools (`delete-queue-messages`, `disconnect-client`) are annotated `destructiveHint` and instruct the LLM to obtain user confirmation before invocation; the `clear-*-stats` tools are non-destructive. |
```
amitmorade: ### README line risks becoming an implicit safety guarantee

Related to my earlier comment about the confirmation protocol not being verified against actual clients (Claude Code auto mode, SAM) — this README line has the same underlying concern but with higher stakes because it's customer-facing. Noting up front that this wording pre-dates this PR (it was already in the README), but since #124 expands the destructive-tool surface, I think it's worth fixing here rather than leaving it in place.

> "Destructive tools ... instruct the LLM to obtain user confirmation before invocation"

This wording is technically true (the tool descriptions do contain that instruction), but I think it reads to a customer/auditor as "confirmation is enforced" — which is something we cannot actually guarantee, because the MCP protocol itself does not provide any such guarantee. `destructiveHint` is a *hint* to the client, and the confirmation text is a *hint* to the model. Neither is enforced by the protocol. Whether the model pauses depends on the client, its tool-approval mode (Claude Code auto vs ask), and — for other clients like SAM — whether the client supports a "wait for user reply before invoking" pattern at all.

If any of those layers doesn't fire, a destructive operation happens without the confirmation the docs seemed to promise. That's a bad shape for an incident write-up or a security review.

Suggest tightening to something like:
> "Destructive tools are annotated `destructiveHint` so MCP clients *can* prompt the user before invocation, and their descriptions ask the model to confirm. Neither is enforced by the MCP protocol — both are hints — so actual behaviour depends on the client and the model."

Puts the guarantee where it actually lives (nowhere, at the protocol level) and calls the description-embedded protocol what it actually is — a hint.

amitmorade (changes requested):

internal/composite/definitions/tools.yaml
solace-wmaryam: To address your first concern, I did test to see whether Claude Code prompted for confirmation before each write tool, and experienced no problems there. There was one case where it bypassed user permission while doing some other testing as it translated my initial testing prompt as confirmation. 

I agree that it would be worth testing the write tools in auto-mode to see how Claude behaves, in addition to testing on other clients in general (including SAM) to determine how our instructions hold up in other scenarios. If the behaviour is inconsistent, we might have to rethink our strategy and perhaps add explicit logic rather than relying solely on the LLM. I will open a spike for this.

solace-awaheed: FWIW I tested with SAM and was asked for confirmation each time on every type of tool (delete, update, create)

solace-wmaryam:
@amitmorade, I agree with the comment you raised here: https://github.com/SolaceDev/solace-broker-mcp/pull/124#issuecomment-4877476764. I'll be honest - I did consider this same issue but didn't think we'd introduce tool specific params in the future. That was probably presumptuous of me and I agree the fix is cheap. I'll implement it as part of this PR.

internal/composite/definitions/tools.yaml
amitmorade: But I think this is still only valid as long as the model remembers to ask, because I just checked and SAM currently does not have a functionality that will deterministically ask for the confirmation every single time even after context resets

amitmorade (approved):
LGTM, all comments addressed, changes are solid, thank you!

internal/semp/sempv2/operation.go
```diff
@@ -174,3 +179,49 @@ func extractParameters(opDef *openapi2.Operation, sharedParams map[string]*opena
...
 }
+
+// extractBodyFields returns the set of attribute names an operation's request
+// body accepts, resolved from its body parameter's schema. It returns nil when
+// the operation has no body parameter or the schema can't be resolved to a
+// concrete property set; callers treat nil as "unknown" and skip body
+// validation, so an operation we can't introspect keeps working rather than
+// having every field rejected. SEMP config definitions are flat objects (no
+// allOf composition), so the schema's top-level properties are the full set.
+func extractBodyFields(opDef *openapi2.Operation, sharedParams map[string]*openapi2.Parameter, definitions map[string]*openapi2.SchemaRef) map[string]bool {
```
solace-aross, line 190: Non-blocking, worth being conscious of: this makes the embedded config spec the source of truth for which attributes users can set, not just for read parsing. If a newer broker version adds a config attribute not in this embedded spec, `constructRequestBody` rejects it client-side (executor.go:551) even though the broker would accept it. Before this PR that value passed through. Fair tradeoff for catching tool-only params, but refreshing this spec is now part of supporting new broker config attributes. Might warrant a line in the spec's update process, or a follow-up to track spec currency.

solace-aross (approved):
Approving. I focused this pass on what changed since the last human approval and on Amit's requested change, since the whitelist commit post-dates both.

Verified resolved:

- **Body-field whitelist (Amit's request).** `constructRequestBody` now rejects any body field not in `op.BodyFields`, and `extractBodyFields` populates that set from the operation's body schema with a nil fallback when the schema can't be introspected (validation defers to the broker, as before). The tool-only-param trap is closed and tested: `TestExecute_UpdateMessageVPN_RejectsUnknownBodyField` rejects a `dryRun` field client-side with zero SEMP calls, and `TestParseSpecs_BodyFields` covers real-spec population. I also checked the "flat objects, no allOf" assumption in the `extractBodyFields` comment: the embedded config spec has zero `allOf` across all 475 definitions, and MsgVpn/MsgVpnQueue/MsgVpnTopicEndpoint are flat (123/34/31 top-level properties). The whitelist is complete and won't reject valid attributes on those types.
- **destructive on updates.** All three `update-*` tools now carry `destructive: true`; creates stay additive, deletes stay destructive.
- **Confirmation-protocol docs.** README/user-guide/configuration now frame `destructiveHint` and the confirmation wording as hints, not enforcement.

One new non-blocking note (inline): the whitelist makes the embedded config spec the source of truth for settable attribute names, not just for read parsing. Fair tradeoff, but it adds a maintenance coupling worth tracking.

Agreed on the two deferred follow-ups already in the thread: the auto-mode / SAM confirmation spike, and a management E2E suite so these write paths stay covered as the config surface grows.

Amit's CHANGES_REQUESTED is still formally on the PR; his re-review clears that. From my side this is good to merge once he's satisfied. Nice work.

solace-aross:
Raised SOL-151503 for the spec-currency point from my review (the inline note on operation.go). Nothing to do in this PR. The ticket tracks two decisions for later: how the embedded config spec gets refreshed as brokers add attributes, and whether an unknown attribute should hard-reject (current) or warn-and-pass-through.

https://sol-jira.atlassian.net/browse/SOL-151503

internal/semp/sempv2/operation.go
solace-wmaryam: Noted, but I wanted to mention that we are already dependent on the SEMP specs and have a story open in one of the later epics to determine how to automate pulling them in.

PR #127, opened 2026-06-30
SOL-151285: k8s split probes + startupProbe; alias /ready to /readyz

solace-aross:
## What
Operationalizes the decoupled readiness model in Kubernetes and retires the legacy broker-dialing `/ready`:
- `livenessProbe` → `/livez`, `readinessProbe` → `/readyz`, plus a cold-start `startupProbe` → `/readyz`.
- `/ready` is now a body-identical **alias of `/readyz`** (broker-dial removed); a broker being unreachable no longer marks the pod unready.

Part of the Observability & Telemetry epic ([SOL-149791](https://sol-jira.atlassian.net/browse/SOL-149791)). Story: [SOL-151285](https://sol-jira.atlassian.net/browse/SOL-151285).

## How
- `deploy/kubernetes/deployment.yaml`: probe paths repointed; `startupProbe` (period 2s × failureThreshold 30 ≈ 60s) covers normal cold start; liveness/readiness timings unchanged; `/livez` governs restart.
- `/ready` and `/readyz` share the same `health.ReadyzHandler` instance (identical body/status).
- `buildMux` simplified to `buildMux(readiness)` — the now-unused broker-dial code (`newBrokerReachabilityProbe`, probe goroutines, `DefaultReadinessProbeTimeoutSeconds`) removed; all callers updated.
- `/health` retained as the back-compat `{"status":"healthy"}` endpoint.

## Tests
`/ready` == `/readyz` (status+body+Content-Type, before and after init); decoupled-from-broker (200 with an unreachable broker); e2e shell suite updated to the new contract. `make check` green.

[SOL-151285]: https://sol-jira.atlassian.net/browse/SOL-151285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/kubernetes/deployment.yaml
+++ b/deploy/kubernetes/deployment.yaml
@@ -53,17 +53,34 @@ spec:
               mountPath: /etc/mcp-server/config.yaml
               subPath: config.yaml
               readOnly: true
+          # startupProbe gates liveness/readiness until the MCP server has
+          # finished cold-start init. /readyz reflects the server's OWN
+          # readiness (it makes no broker calls), so it flips ready as soon as
+          # init completes — there is no unbounded waiting on broker
+          # reachability. failureThreshold 30 x periodSeconds 2s allows ~60s
+          # for a normal cold start before Kubernetes treats startup as failed.
+          startupProbe:
+            httpGet:
+              path: /readyz
+              port: http
+            periodSeconds: 2
+            timeoutSeconds: 3
+            failureThreshold: 30
+          # /livez is the canonical liveness endpoint and governs restarts.
           livenessProbe:
             httpGet:
-              path: /health
+              path: /livez
               port: http
             initialDelaySeconds: 5
             periodSeconds: 15
             timeoutSeconds: 3
             failureThreshold: 3
+          # /readyz reflects MCP-server readiness only (decoupled from broker
+          # reachability); a not-ready pod is pulled from Service endpoints but
+          # not restarted.
           readinessProbe:
             httpGet:
-              path: /health
+              path: /readyz
               port: http
             initialDelaySeconds: 3
             periodSeconds: 10
--- a/deploy/kubernetes/deployment.yaml
+++ b/deploy/kubernetes/deployment.yaml
@@ -26,11 +26,11 @@ spec:
       #                                 + 5s buffer = 45
       # If you raise obs.shutdown_drain_delay_s, raise this value to match.
       #
-      # Note: the readinessProbe below still targets /health (always 200);
-      # rewiring it to /readyz so SetShuttingDown is probe-visible is Story 11
-      # (SOL-151285). Until then this in-process drain + grace period prevent
-      # 502s on their own — kube-proxy deregisters the pod on Terminating, and
-      # the sleep gives it time before in-flight requests are drained.
+      # The readinessProbe below targets /readyz, so on SIGTERM the server's
+      # SetShuttingDown flip makes the pod report not-ready and kube-proxy
+      # deregisters it from Service endpoints within one readiness period. The
+      # drain window above is sized to cover exactly that: orchestrator
+      # deregistration finishes before in-flight requests are drained.
       terminationGracePeriodSeconds: 45
       securityContext:
         runAsNonRoot: true
@@ -67,11 +67,12 @@ spec:
             timeoutSeconds: 3
             failureThreshold: 30
           # /livez is the canonical liveness endpoint and governs restarts.
+          # No initialDelaySeconds: startupProbe above gates liveness until
+          # cold-start init completes, so an initial delay here is never observed.
           livenessProbe:
             httpGet:
               path: /livez
               port: http
-            initialDelaySeconds: 5
             periodSeconds: 15
             timeoutSeconds: 3
             failureThreshold: 3
@@ -82,7 +83,6 @@ spec:
             httpGet:
               path: /readyz
               port: http
-            initialDelaySeconds: 3
             periodSeconds: 10
             timeoutSeconds: 3
             failureThreshold: 3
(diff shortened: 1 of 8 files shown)


test/health/test-health.sh
```diff
@@ -92,15 +97,37 @@ test_health() {
...
+test_readyz() {
+    # /readyz reflects MCP-server readiness only. Once the server is up it is
+    # ready (it makes no broker calls).
     local response
-    response=$(curl -sf "$MCP_URL/ready")
-    assert_json_field "$response" ".ready" "true" "/ready should return ready=true" || return 1
-    assert_contains "$response" "broker-a" "/ready should list broker-a" || return 1
-    assert_contains "$response" "broker-b" "/ready should list broker-b" || return 1
+    response=$(curl -sf "$MCP_URL/readyz")
+    assert_json_field "$response" ".status" "ready" "/readyz should return status=ready"
```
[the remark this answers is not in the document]
solace-aross: Fixed in 131b9dc — added a bounded `wait_for_ready` helper (polls /readyz for HTTP 200, ~20×0.5s) and call it before the readiness body assertions in test_readyz, test_ready_aliases_readyz, and test_ready_decoupled_from_broker, so the cold-start window (start_server only waits on /health) can no longer flake the assertions. The 200 + {"status":"ready"} / identity / decoupling assertions are unchanged.

amitmorade:
Looking at this one

amitmorade:
Not blocking this PR — surfacing something I noticed while reviewing that I
think is worth a separate follow-up ticket.

**The observation.** After this PR, `readinessProbe` in k8s targets `/readyz`,
which reads `ReadinessState.initialized`. Startup wires that flag in this
order (main.go:704-712):

1. `startServer(...)` spawns a goroutine that will call `ListenAndServe`
2. `readiness.SetInitialized()` flips `/readyz` to "ready"
3. main goroutine blocks on `select { <-done | <-serverErr }`

The gap: step 2 fires as soon as step 1 returns, but step 1 only guarantees
the goroutine was *spawned* — it does not wait for `net.Listen` to succeed.
So there's a semantic gap between what `SetInitialized` claims ("startup
complete and we are serving") and what it actually reflects ("we spawned
the goroutine that intends to serve"). If bind fails (port in use,
permission denied), `SetInitialized` briefly reports "ready" for a listener
that will never accept anything.

**Why it doesn't hurt today.**
- `startupProbe` polls until 200, so the brief window is absorbed on cold
  start.
- On bind failure, the process exits within microseconds via the `serverErr`
  channel, so no external observer sees the lie for long.

**Why this PR makes it worth caring about.** Before, nothing consumed
`SetInitialized` externally — `/ready` was broker-coupled and k8s probed
`/health`. After this PR, `/readyz` is what k8s actually uses **and** what
future required-listener probes (`RegisterListener`, already scaffolded in
`readiness.go`) will hang off of. So the accuracy of `SetInitialized`
matters more going forward than it did on `main` before.

**Suggested fix (for the follow-up ticket, not here).** Split the listen
step from the serve step so `startServer` only returns once the socket is
bound. The corrected sequence would be:

1. `net.Listen` runs synchronously in `startServer` — returns bind error
   fail-fast, before any readiness state is touched
2. Serving goroutine spawned on the already-bound listener
3. `startServer` returns to main **only** after step 1 succeeded
4. `readiness.SetInitialized()` — now honest: fires **iff** the socket is
   accepting
5. main goroutine blocks on `select` as before

Bind errors become synchronous (fail-fast, before `SetInitialized`); serve
errors stay async on the channel. Once `net.Listen` returns, the kernel is
queuing connects in the socket's accept backlog, so there's no race window
between `SetInitialized` and the first accepted connection.

**Ask.** Worth a small follow-up ticket? Leaving it up to you whether that
reads as a real gap or a purity nit.

deploy/kubernetes/deployment.yaml
amitmorade: Two small cleanups on this file worth catching before merge:

**1. The comment at lines 29–33 is now stale and contradicts the change
this PR makes.**

Current text says:

> Note: the readinessProbe below still targets /health (always 200);
> rewiring it to /readyz so SetShuttingDown is probe-visible is Story 11
> (SOL-151285). Until then this in-process drain + grace period prevent
> 502s on their own...

But this PR *is* Story 11 — `readinessProbe` a few lines below now targets
`/readyz`. The note reads as if the rewiring hasn't happened yet. Reading
the manifest cold-turkey, the two blocks contradict each other.

**Suggested rewrite** (something like):

> Note: readinessProbe below targets /readyz, so SetShuttingDown flips
> the pod to unready on SIGTERM and kube-proxy deregisters it from Service
> endpoints within one readiness period. That's what the drain window
> below is sized to cover — orchestrator deregistration finishes before
> in-flight requests are drained.

**2. Once startupProbe is in place, `initialDelaySeconds` on
liveness/readiness is dead weight.**

Kubernetes disables both liveness and readiness probes entirely until
`startupProbe` succeeds — so `livenessProbe.initialDelaySeconds: 5` and
`readinessProbe.initialDelaySeconds: 3` will never actually be observed
at the k8s level. Not harmful, just misleading to future readers ("we
picked 5s carefully" — no, we didn't, startupProbe owns that budget now).

Either drop them, or leave a one-line comment saying they're superseded
by startupProbe. I'd lean toward dropping.

Both are minor — happy to defer either if you want to keep the PR
tightly scoped to the routing change.

solace-aross: Both good catches — done in 576b2a0.

1. Rewrote the note: it now says the readinessProbe targets /readyz, so SetShuttingDown flips the pod not-ready on SIGTERM and kube-proxy deregisters it within one readiness period, and the drain window is sized to cover exactly that. (The stale text was carried over from the graceful-shutdown PR before this one did the rewiring.)
2. Dropped initialDelaySeconds from both liveness (5s) and readiness (3s), since the startupProbe gates them until cold-start init completes — left a one-line note on liveness saying so. Agreed it was misleading dead weight.

Manifest-only change, make check green.

amitmorade (approved):
LGTM, all comments are addressed, CI is green, thank you!

PR #129, opened 2026-06-30
SOL-151280: emit correlation_id as a slog attribute on every request-scoped log line

solace-aross:
## What
Every request-scoped log line now carries `correlation_id`, so an operator can `grep` all logs for one failed call.

Part of the Observability & Telemetry epic ([SOL-149791](https://sol-jira.atlassian.net/browse/SOL-149791)). Story: [SOL-151280](https://sol-jira.atlassian.net/browse/SOL-151280).

## How
- New `slog.Handler` decorator (`correlation.NewSlogHandler`) that, in `Handle(ctx, rec)`, reads `correlation.From(ctx)` and adds `correlation_id` when present — wired into `newSlogHandler` so both the bootstrap and user-level loggers get it.
- Preserves the existing `ReplaceAttr` credential redaction (the wrapper does no formatting; it delegates to the inner JSON handler that owns the filter).
- Does not overwrite an existing `correlation_id` attribute, and clones the record before adding so shared records are never mutated.
- Gated transitively: `From(ctx)` is empty when the correlation middleware isn't wired (`OBS_CORRELATION_ID_ENABLED` off) or for startup/non-request logs, so output matches today exactly when the capability is off.
- `WithGroup`/`WithAttrs` re-wrap correctly so `correlation_id` stays at the **root** of the log line even after `.WithGroup(...)`, while the record's own attrs nest under the group (a real slog gotcha, handled and tested across orderings incl. nested groups).

## Tests
ID added from context / absent when no ID / empty ID / not-overwritten / redaction preserved / `WithAttrs` + `WithGroup` (single, mixed, and nested) keep `correlation_id` at root. `make check` green (build + vet + lint + `-race`).

[SOL-151280]: https://sol-jira.atlassian.net/browse/SOL-151280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/observability/correlation/sloghandler.go
+++ b/internal/observability/correlation/sloghandler.go
@@ -86,10 +86,12 @@ func (h *slogHandler) Enabled(ctx context.Context, level slog.Level) bool {
 // non-empty correlation ID AND the record does not already carry a
 // correlation_id attribute, then delegates to next.
 //
-// The no-overwrite rule (AC #3) is enforced by scanning the record's attributes
-// for an existing correlation_id key before adding one; an explicitly logged
-// correlation_id always wins and is never duplicated. When there is no ID to add
-// and no recorded goa steps, the record is passed through untouched, so
+// The no-overwrite rule (AC #3) is enforced by checking for an existing
+// root-level correlation_id before adding one: either on the record itself
+// (hasCorrelationID) or bound on the logger at root via .With(...) before any
+// group was opened (hasRootCorrelationID). In both cases the caller's explicit
+// correlation_id wins and is never duplicated at the root. When there is no ID
+// to add and no recorded goa steps, the record is passed through untouched, so
 // absent-ID output matches the unwrapped handler exactly.
 //
 // With recorded goa steps, the delivery handler is rebuilt from the ungrouped
@@ -98,7 +100,7 @@ func (h *slogHandler) Enabled(ctx context.Context, level slog.Level) bool {
 // their grouping while correlation_id stays at the top level.
 func (h *slogHandler) Handle(ctx context.Context, rec slog.Record) error {
 	id := From(ctx)
-	addID := id != "" && !h.hasCorrelationID(rec)
+	addID := id != "" && !h.hasCorrelationID(rec) && !h.hasRootCorrelationID()
 
 	if len(h.goas) == 0 {
 		if !addID {
@@ -142,6 +144,35 @@ func (h *slogHandler) hasCorrelationID(rec slog.Record) bool {
 		return true
 	})
 	return found
+}
+
+// hasRootCorrelationID reports whether a correlation_id was already bound on the
+// logger at the ROOT level via .With(...)/WithAttrs, recorded in goas. The
+// wrapper always injects its correlation_id at the root (bound on the ungrouped
+// next before any goa step is replayed), so a duplicate root key arises only
+// when a root-level correlation_id already exists. Returning true here suppresses
+// that injection so the caller's explicit value wins with no duplicate (AC #3).
+//
+// Only attrs bound BEFORE the first WithGroup land at the root; attrs bound after
+// a group nest under that group (a different key path) and must NOT suppress the
+// root injection. So the scan walks goas in order and stops at the first group
+// step: any correlation_id in an attr step seen before that is root-level; a
+// correlation_id only appearing in attrs after a group is nested and is ignored
+// here (still inject at root). With no goas (the common case) this returns false.
+func (h *slogHandler) hasRootCorrelationID() bool {
+	for _, g := range h.goas {
+		if g.group != "" {
+			// First group opened: subsequent attrs nest under it, so no further
+			// step can introduce a root-level correlation_id.
+			return false
+		}
+		for _, a := range g.attrs {
+			if a.Key == attrKey {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 // WithAttrs returns a new slogHandler that records the attrs as a goa step. It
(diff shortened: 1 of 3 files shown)


internal/observability/correlation/sloghandler.go
```diff
@@ -0,0 +1,180 @@
...
+// absent-ID output matches the unwrapped handler exactly.
+//
+// With recorded goa steps, the delivery handler is rebuilt from the ungrouped
+// root: correlation_id is bound at the root first, then the goa steps are
+// replayed (groups reopened, attrs rebound) so the record's own attributes keep
+// their grouping while correlation_id stays at the top level.
+func (h *slogHandler) Handle(ctx context.Context, rec slog.Record) error {
+	id := From(ctx)
+	addID := id != "" && !h.hasCorrelationID(rec)
+
```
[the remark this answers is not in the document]
solace-aross: Good catch — fixed in db9b25b. Added `hasRootCorrelationID`, which scans the recorded `.With(...)`/WithAttrs steps (goas) and treats a root-level `correlation_id` bound there as already-present, so the context ID is not injected again — no duplicate root key. It stops at the first WithGroup step, so a `correlation_id` bound *after* a group nests under that group (different key path) and the root injection still happens. Two tests added (root-bound dedup; post-group nested + root injected).

internal/observability/correlation/sloghandler_test.go
```diff
@@ -0,0 +1,272 @@
...
+	line := buf.String()
+	for _, want := range []string{`"correlation_id":"abc"`, `"base":"b"`, `"ingroup":"x"`, `"rec":"r"`} {
+		if !strings.Contains(line, want) {
+			t.Errorf("missing %s in: %s", want, line)
+		}
+	}
+	// ingroup and rec must be inside group g; base and correlation_id must not.
+	if !strings.Contains(line, `"g":{"ingroup":"x","rec":"r"}`) {
+		t.Errorf("grouped attrs not nested correctly: %s", line)
+	}
```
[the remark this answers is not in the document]
solace-aross: Fixed in db9b25b — these assertions (and the nested-group test) now decode via the existing parseLine helper and assert on the nested map values instead of an ordered JSON substring, so they no longer depend on key emission order.

internal/observability/correlation/sloghandler_test.go
```diff
@@ -0,0 +1,335 @@
...
+	logger := newRedactingLogger(&buf)
+
+	ctx := With(context.Background(), "abc")
+	logger.InfoContext(ctx, "hello", slog.String("password", "hunter2"))
+
+	m := parseLine(t, &buf)
+	if got := m["password"]; got != "[REDACTED]" {
+		t.Errorf("password = %v, want %q", got, "[REDACTED]")
+	}
+	if got := m["correlation_id"]; got != "abc" {
```
solace-awaheed, line 203: I think this test might be giving us a false positive right now. Since `correlation_id` isn’t a redacted key, this assertion will pass whether it actually went through the redaction flow or was just passed through unchanged.

Maybe we can add a case where `correlation_id` gets rewritten to something like `"[CID_REDACTED]"` and assert on that instead? That would make it clear that the injected field is actually going through the redaction path.

internal/observability/correlation/sloghandler.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		// Clone before mutating: slog may pass the same Record to more than one
+		// handler, and rec.Add appends to the Record's own attribute slice, so
+		// cloning keeps our added correlation_id local to this delivery.
+		rec = rec.Clone()
+		rec.Add(slog.String(attrKey, id))
+		return h.next.Handle(ctx, rec)
+	}
+
+	delivery := h.next
```
solace-awaheed: I think we may be doing more work than needed here. For loggers created via .With(...) / .WithGroup(...), this ends up rebuilding the handler chain on every log call, which puts extra allocations into the hot path.

Could we build the child handler once in WithAttrs / WithGroup and reuse it here instead? That would avoid recreating the chain for every log line while keeping the same behaviour.

solace-awaheed (changes requested):

internal/observability/correlation/sloghandler_test.go
solace-aross: Good catch — it was a false positive. Fixed in 636b2d9: added TestSlogHandler_InjectedCorrelationIDFlowsThroughReplaceAttr, where next's ReplaceAttr rewrites correlation_id to [CID_REDACTED] and the test asserts that value in the emitted line. That proves the injected attr actually flows through next's ReplaceAttr — mutation-checked (making the inject path bypass next fails the test). Kept the credential-redaction test too, with a comment that presence alone proves nothing.

internal/observability/correlation/sloghandler.go
solace-aross: Fixed in 636b2d9. The handler now caches, once per derived handler (in WithAttrs/WithGroup via an internal newSlogHandler(next, goas)): built = next with the goas applied (no correlation_id), and rootHasID. Handle delivers the no-add case straight through the cached built — no per-call chain rebuild. One honest caveat: the inject-under-groups case (a .With/.WithGroup logger logging in a request context) still builds per record, because the ID comes from each record's context and must be bound at the root ahead of the caller's groups, and slog handlers are immutable so the group suffix can't be pre-applied over a per-record root attr. That path is now the only one that rebuilds, and it's documented inline. All the root-placement/no-overwrite/redaction behavior is unchanged (verified by mutation tests).

solace-awaheed (approved):
Thanks for the update! LGTM

PR #130, opened 2026-06-30
SOL-151288: SIGTERM in-process drain (/readyz 503 → sleep → graceful shutdown) + terminationGracePeriodSeconds

solace-aross:
## What
Makes pod shutdown drain cleanly on rolling restarts: on SIGTERM the server flips `/readyz` to 503, sleeps a propagation window so the orchestrator stops routing new traffic, then gracefully drains in-flight requests — and `terminationGracePeriodSeconds` is sized to cover the whole sequence so K8s never SIGKILLs mid-drain.

Part of the Observability & Telemetry epic ([SOL-149791](https://sol-jira.atlassian.net/browse/SOL-149791)). Story: [SOL-151288](https://sol-jira.atlassian.net/browse/SOL-151288). **This completes the MUST epic.**

## How
- New `drainAndShutdown(ctx, srv, readiness, drainDelay)` helper: `SetShuttingDown()` first (so `/readyz` → 503 immediately), then sleep `drainDelay`, then graceful `Shutdown`. Runs **only on the SIGTERM path** — a startup error shuts down without draining.
- The existing `Shutdown`→forced-`Close` fallback and `os.Exit(1)`-on-startup-error are preserved unchanged (extracted into `gracefulShutdown`).
- New `obs.shutdown_drain_delay_s` config tunable (default 10s), following the existing YAML-int pattern + `<=0` re-default.
- `deployment.yaml`: `terminationGracePeriodSeconds: 45` (= `shutdown_drain_delay_s` 10 + `DefaultShutdownTimeoutSeconds` 30 + 5 buffer). In-process drain, **no preStop** (distroless image has no shell).

## Dependency / scope note
The readiness *flip* (`SetShuttingDown` → `/readyz` 503) only becomes **probe-visible** once the K8s `readinessProbe` is rewired from `/health` to `/readyz` — that rewiring is **Story 11 (#127 / SOL-151285)**, which owns the probe blocks. I deliberately did **not** touch the probe paths here (avoids a conflict with #127). Until #127 lands, this PR still prevents 502s on its own: the in-process drain sleep + grace period give kube-proxy time to deregister the pod on `Terminating` before in-flight requests are drained. Once #127 merges, the two compose into full readiness-gated deregistration.

## Tests
`drainAndShutdown` ordering against a real server: readiness flips before the drain completes (sampled mid-drain), the server is still accepting mid-drain, and `Shutdown` runs only after the delay — with ms-scale delays (wide margin, race-clean). Zero-delay path still flips readiness. Config default/0/negative/explicit covered. `make check` green (build + vet + lint + `-race`).

[SOL-151288]: https://sol-jira.atlassian.net/browse/SOL-151288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -352,6 +352,18 @@ func drainAndShutdown(srv shutdowner, readiness *health.ReadinessState, drainDel
 	return gracefulShutdown(srv, shutdownTimeout)
 }
 
+// drainDelayForSignal returns the drain delay to apply for the received
+// shutdown signal. SIGTERM (orchestrator-initiated, e.g. Kubernetes) gets the
+// configured drain window so the orchestrator can deregister the pod before we
+// stop accepting. SIGINT (os.Interrupt, a local Ctrl-C) skips the drain
+// entirely — there is no orchestrator to wait for — so it returns 0.
+func drainDelayForSignal(sig os.Signal, configured time.Duration) time.Duration {
+	if sig == os.Interrupt { // syscall.SIGINT
+		return 0
+	}
+	return configured
+}
+
 // shutdowner is the subset of *http.Server that gracefulShutdown drives. It
 // exists so a test can substitute a fake that captures the deadline of the
 // context handed to Shutdown — proving the full timeout budget is granted at
@@ -683,15 +695,22 @@ func main() {
 
 	var startupErr error
 	var fromSignal bool
+	var receivedSignal os.Signal
 	select {
-	case <-done:
-		// SIGTERM (or Ctrl-C): drain before shutting down. The drain (flip
+	case sig := <-done:
+		// Shutdown signal received: drain before shutting down. The drain (flip
 		// /readyz, sleep the propagation window, then graceful shutdown) runs in
-		// drainAndShutdown below. The shutdown-timeout context is NOT built here;
-		// gracefulShutdown builds it after the drain sleep so Shutdown gets the
-		// full budget (see drainAndShutdown / gracefulShutdown).
+		// drainAndShutdown below. SIGTERM (orchestrator-initiated) honors the
+		// configured drain window; SIGINT (a local Ctrl-C) skips it via
+		// drainDelayForSignal — there is no orchestrator to wait for. The
+		// shutdown-timeout context is NOT built here; gracefulShutdown builds it
+		// after the drain sleep so Shutdown gets the full budget (see
+		// drainAndShutdown / gracefulShutdown).
 		fromSignal = true
-		slog.Info("server shutting down", slog.String("reason", "signal"))
+		receivedSignal = sig
+		slog.Info("server shutting down",
+			slog.String("reason", "signal"),
+			slog.String("signal", sig.String()))
 	case startupErr = <-serverErr:
 		// Error already logged in startServer. Capture it so future telemetry
 		// hooks (e.g., metrics flush, error reporting) have access to the value,
@@ -708,7 +727,7 @@ func main() {
 	shutdownTimeout := time.Duration(defaults.DefaultShutdownTimeoutSeconds) * time.Second
 
 	if fromSignal {
-		drainDelay := time.Duration(cfg.Observability.ShutdownDrainDelayS) * time.Second
+		drainDelay := drainDelayForSignal(receivedSignal, time.Duration(cfg.Observability.ShutdownDrainDelayS)*time.Second)
 		_ = drainAndShutdown(httpServer, readiness, drainDelay, shutdownTimeout)
 	} else {
 		_ = gracefulShutdown(httpServer, shutdownTimeout)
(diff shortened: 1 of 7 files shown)


cmd/server/main.go
```diff
@@ -610,28 +655,32 @@ func main() {
...
-		slog.Error("shutdown timed out, forcing close", slog.String("error", err.Error()))
-		if closeErr := httpServer.Close(); closeErr != nil {
-			slog.Error("forced close failed", slog.String("error", closeErr.Error()))
-		}
+	if fromSignal {
+		drainDelay := time.Duration(cfg.Observability.ShutdownDrainDelayS) * time.Second
+		_ = drainAndShutdown(ctx, httpServer, readiness, drainDelay)
+	} else {
+		_ = gracefulShutdown(ctx, httpServer)
 	}
```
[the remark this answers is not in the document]
solace-aross: Good catch — real bug, fixed in 45835a4. The shutdown timeout context is now created INSIDE gracefulShutdown, immediately before Shutdown, so it is built AFTER the drain sleep — Shutdown gets the full DefaultShutdownTimeoutSeconds (30s) regardless of the drain delay, matching the terminationGracePeriodSeconds budget. main() no longer pre-builds the context; drainAndShutdown sleeps then calls gracefulShutdown(srv, timeout). Added a test that captures the context deadline at the Shutdown call and asserts the remaining budget exceeds the old (timeout - drainDelay) ceiling (mutation-verified: it fails if the context is built before the sleep). Forced-Close fallback and the startup-error path are unchanged.

cmd/server/main.go
solace-awaheed, line 684: Not blocking:

The drain delay is applied to os.Interrupt too, so Ctrl-C during local dev now blocks ~10s before shutdown begins. Should the drain path be SIGTERM-only, with SIGINT skipping the drain (delay 0)? Right now fromSignal is handled the same way for both signals, so we'd need to track which signal was actually received from done.

cmd/server/main.go
```diff
@@ -610,28 +682,36 @@ func main() {
...
-	defer cancel()
 
-	if err := httpServer.Shutdown(ctx); err != nil {
-		slog.Error("shutdown timed out, forcing close", slog.String("error", err.Error()))
-		if closeErr := httpServer.Close(); closeErr != nil {
-			slog.Error("forced close failed", slog.String("error", closeErr.Error()))
-		}
+	if fromSignal {
+		drainDelay := time.Duration(cfg.Observability.ShutdownDrainDelayS) * time.Second
+		_ = drainAndShutdown(httpServer, readiness, drainDelay, shutdownTimeout)
```
solace-awaheed, line 731: Not blocking: 
Should we consider treating a second signal as a force-quit (srv.Close() immediately), e.g. by selecting on done during/around the drain period.

Right now, once drainAndShutdown() begins, the process is committed to sleep(drainDelay) followed by Shutdown(). signal.Notify remains active and done is buffered, but nothing reads from it again, so a second SIGTERM/Ctrl-C gets buffered and effectively ignored. In Kubernetes the grace-period SIGKILL eventually bounds this, but locally there's no "I mean it, exit now" path — an impatient Ctrl-C, Ctrl-C appears to do nothing.

solace-awaheed (commented):
The changes LGTM, I've left non blocking comments in the PR

cmd/server/main.go
solace-aross: Good call — fixed in ee27004. Captured the received signal and route the drain delay through a small pure helper drainDelayForSignal: os.Interrupt (Ctrl-C) → 0 (skip the drain), SIGTERM → the configured shutdown_drain_delay_s. So a local Ctrl-C no longer blocks for the drain window; k8s SIGTERM behavior is unchanged. Added a table test for the helper (SIGINT→0, SIGTERM→configured, SIGTERM+0→0).

cmd/server/main.go
solace-aross: Agreed this is a nice local-dev affordance. It's a bit more than a one-liner — to truly force-quit it needs to interrupt both the drain sleep and the graceful-Shutdown wait (run Shutdown in a goroutine and select on the signal channel), so I've split it into a follow-up rather than grow this PR's shutdown-concurrency surface: SOL-151437 (https://sol-jira.atlassian.net/browse/SOL-151437). Note that with #1 above SIGINT no longer drains, so the practical local hang is already gone; SOL-151437 adds the explicit second-signal force-quit on top.

solace-awaheed (approved):
LGTM !

solace-aross:
Rebased onto current main after #126 and #131 landed (strict branch-protection needs it up to date, which dismissed your approval). The only merge interaction was in internal/defaults/defaults.go — #131 added a token-exchange default and this PR adds DefaultShutdownDrainDelayS; git auto-merged them and make check is green (build, vet, lint, race). No other content change from what you approved. A quick re-approve lands it. Sorry for the round-trip.

PR #136, opened 2026-07-02
SOL-151468: lift SafeCookieJar to broker level, delegate 401 recovery to Authenticator

amitmorade:
## Why this PR exists (and why now)

This PR is **not** scope creep — it is a prerequisite for T6 (SOL-150800, OAuthAuthenticator).

T6 adds `OAuthAuthenticator`, which needs the `HandleAuthFailure` interface method to implement its own 401 recovery strategy. Under the old design:

1. **Two jars per broker** — SEMPv1 and SEMPv2 each created their own `SafeCookieJar`. Both clients talk to the same broker host, so a 401 clear on one was invisible to the other.
2. **Sender knew about auth modes** — `retry.go` switched on `AuthConfig.Mode` to decide 401 behavior. Every new auth mode would require modifying the Sender.
3. **Jar was unconditional** — every broker got a cookie jar, even though only basic auth uses session cookies (the broker never sets `Set-Cookie` for bearer/OAuth per [Solace docs](https://docs.solace.com/Admin/SEMP/SEMP-Security.htm)).

From solace docs: 

> Sessions are not created when authenticating with an OAuth token or tokens using HTTP Bearer authentication. If a session cookie is provided, it is ignored.

Shipping this separately from T6 keeps the OAuth PR focused on token exchange logic only.

### Basic auth broker
1. `NewBrokerClient` **conditionally** creates a `SafeCookieJar` — only when the broker's auth mode is basic
2. `NewAuthenticator` validates jar is non-nil for basic auth (hard error otherwise), gives it to `BasicAuthenticator`
3. Both SEMP clients get the **same jar** on `httpClient.Jar` — cookies are shared across V1 and V2
4. On 401 → Sender calls `authenticator.HandleAuthFailure` → BasicAuth clears the shared jar → returns `retry=true` → Sender retries once without stale cookies

### Bearer (or future OAuth) broker
1. `NewBrokerClient` **skips** jar creation entirely — `jar` stays `nil`, no allocation
2. `NewAuthenticator` doesn't pass jar to bearer (constructor doesn't accept one)
3. Both SEMP clients get `nil` jar — `httpClient.Jar` is never set, Go's `http.Client` silently ignores cookies
4. On 401 → Sender calls `authenticator.HandleAuthFailure` → Bearer returns `false` → no retry, failure propagates immediately

### Sender no longer knows about auth modes
It holds `auth.Authenticator` and delegates 401 decisions via `HandleAuthFailure`. Adding OAuth means implementing the method on `OAuthAuthenticator` — no Sender changes needed.

## Key design decisions

| Decision | Rationale |
|----------|-----------|
| Jar is conditional on `AuthModeBasic` | Only basic auth uses session cookies ([Solace docs](https://docs.solace.com/Admin/SEMP/SEMP-Security.htm)); no allocation for modes that don't need it |
| `NewAuthenticator` rejects nil jar for basic auth | Catches wiring bugs at construction, not silent misbehavior at runtime |
| `CookieJarClearer` interface in `auth` package | Consumer-side interface avoids import cycle (`resilience` → `auth`; `auth` must not → `resilience`) |
| `HandleAuthFailure` on `Authenticator` interface | Each auth mode owns its recovery strategy; Sender just delegates |

## Guards
- `NewAuthenticator`: basic auth + nil jar → hard error at construction
- `resilience.New`: nil authenticator → panic at construction
- `resilience.New`: nil semaphore → panic at construction

## What changed

| File | Change |
|------|--------|
| `auth/auth.go` | `HandleAuthFailure` on interface, `CookieJarClearer` interface, `NewAuthenticator` validates jar for basic |
| `auth/basic.go` | Stores jar, implements `HandleAuthFailure` (clear jar + retry) |
| `auth/bearer.go` | `HandleAuthFailure` returns `false` (no recovery) |
| `broker.go` | Conditional jar creation (basic auth only), passes to auth and both clients |
| `resilience/sender.go` | Holds `auth.Authenticator` instead of `AuthConfig` + jar; nil-authn panic guard |
| `resilience/retry.go` | 401 delegates to `authenticator.HandleAuthFailure` |
| `sempv1/client.go` | Accepts optional jar, conditional `httpClient.Jar` |
| `sempv2/client.go` | Same |
| `broker_test.go` | **New**: shared-jar integration test — 401 on V2 clears jar, V1 confirms absence |

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./internal/...` — clean
- [x] `go test -race -count=1` all affected packages — pass
- [x] `TestBrokerClient_SharedJar_401ClearVisibleAcrossProtocols` — proves jar sharing across V1/V2
- [x] `TestNewAuthenticator_BasicAuth_RequiresJar` — proves nil-jar guard
- [ ] CI pipeline (build, lint, test, e2e)


internal/semp/auth/auth.go
```diff
@@ -17,34 +17,51 @@ import (
...
-// mode (basic, bearer, and — added later — oauth). The SEMP transport
-// layer holds an Authenticator behind this interface and never branches
-// on auth mode; adding a new mode is one new implementation plus one
-// new case in NewAuthenticator.
+// SEMP request for a single broker and declares how to recover when
+// authentication fails. One implementation exists per auth mode (basic,
+// bearer, and — added later — oauth). The SEMP transport layer holds an
+// Authenticator behind this interface and never branches on auth mode;
+// adding a new mode is one new implementation plus one new case in
+// NewAuthenticator.
```
solace-aross, line 25: The concurrency contract here says safety is "achieved structurally — implementations hold only fields set at construction and never written again." True for Basic and Bearer, but it won't hold for the `OAuthAuthenticator` this PR is a prerequisite for: token refresh in `HandleAuthFailure` is exactly mutable, written-after-construction state that needs synchronization (mutex or atomic pointer).

As written, the contract points the T6 implementer at the wrong mental model — lock-free immutability — for the one implementation you're clearing the path for. Suggest softening, e.g.: "Current implementations achieve this structurally via immutable fields; implementations that mutate recovery state (e.g. OAuth token refresh) must provide their own synchronization." Cheap now, saves a subtle data race later.

internal/semp/auth/auth.go
```diff
@@ -17,34 +17,51 @@ import (
...
 //
 // Implementations must be safe for concurrent use: many goroutines may
-// call AddAuth simultaneously with their own ctx and req. This is
-// achieved structurally — implementations hold only fields set at
-// construction and never written again, so concurrent reads of struct
+// call AddAuth and HandleAuthFailure simultaneously with their own ctx.
+// This is achieved structurally — implementations hold only fields set
+// at construction and never written again, so concurrent reads of struct
 // state are safe by Go's memory model. Per-request data flows through
 // ctx and req; no per-request state lives on the Authenticator.
```
solace-aross: `HandleAuthFailure` receives `*http.Response`, and both current implementations ignore it. OAuth recovery will likely want to inspect the 401 (a `WWW-Authenticate` challenge, or an error body distinguishing "token expired" from "insufficient scope").

`checkRetry` runs inside retryablehttp, which owns the body lifecycle around the policy call. An implementation that reads or closes `resp.Body` here could interfere with the library's own draining. Since nothing consumes `resp` yet, worth documenting the contract on the interface method now: may an implementation read `resp.Body`? Must it not close it? A one-liner prevents a body-consumption bug in the OAuth PR.

internal/semp/resilience/retry.go
```diff
@@ -103,25 +102,15 @@ func (d *Sender) checkRetry(ctx context.Context, resp *http.Response, err error)
...
-			// the retried request re-sends Basic Auth credentials from scratch.
-			// SafeCookieJar.Clear performs an atomic swap of the inner jar,
-			// so concurrent http.Client.Do reads see either the old jar or
-			// the new one — never a torn read.
-			if err := d.cookieJar.Clear(); err != nil {
-				slog.Warn("retrying: 401 received but failed to clear cookie jar",
-					slog.String("broker", d.brokerURL),
-					slog.String("error", err.Error()))
-				return false, nil
+			retry := d.authenticator.HandleAuthFailure(ctx, resp)
```
solace-aross: Minor (observability, non-blocking): this logs only when recovery succeeds. When basic re-auth still fails after its one retry, or a bearer token is rejected outright, nothing is logged from the auth/retry layer — it propagates as a generic error. A persistent, unrecovered 401 is a security-relevant signal (rotated/expired credentials, or probing). Consider a `Warn` on the unrecovered-401 path so auth failures are greppable for monitoring. Fold in whenever you next touch this.

solace-aross (commented):
Security-focused review. Clean refactor — per-broker jar isolation holds, bearer/OAuth correctly allocates no jar, and the credential-clear-and-retry on 401 is safely bounded (the POST/PATCH idempotency guard in retry.go runs before the 401 case, so a stale-cookie retry can never replay a non-idempotent write). Two forward-looking comments on the interface contract you're setting for T6 (OAuth), plus one minor observability note. Nothing here blocks merge.

I've skipped the points Copilot already raised (nil-authenticator guard, which you've since addressed with the panic in resilience.New; log wording; and the test-wiring nil-jar notes).

amitmorade:
Thank you @solace-aross  for the feedback. Addressed all three points in 51dc0be:

1. **Concurrency contract (auth.go:25)** — Softened the doc to acknowledge that current implementations (basic, bearer) achieve safety structurally via immutable fields, but implementations that mutate state after construction (e.g. OAuth token refresh) must provide their own synchronization.

2. **HandleAuthFailure body-consumption risk (auth.go:32)** — Rather than documenting a "don't close the body" contract, narrowed the signature from `*http.Response` to `http.Header`. The status code is already known (checkRetry only calls this inside the 401 branch), and OAuth recovery only needs `WWW-Authenticate` (RFC 6750 §3). A type signature is a wall; a doc comment is a speed bump.

3. **Log unrecovered 401s (retry.go:107)** — Added WARN logs for both silent paths:
   - `"auth failure: 401 received, auth handler cannot recover"` — when HandleAuthFailure returns false (bearer, or basic jar clear failed)
   - `"auth failure: 401 persisted after recovery attempt"` — when a second 401 arrives after we already retried

   All 401 outcomes now log. Greppable by `"auth failure"` for monitoring.

solace-aross (approved):
Approving. Clean refactor and all prior feedback is resolved.

Verified on this pass:
- 401 recovery now delegates to the Authenticator, so the Sender no longer branches on auth mode. Adding OAuth is one implementation, no Sender change.
- The POST/PATCH idempotency guard still runs before the 401 case in checkRetry, so a stale-cookie clear-and-retry can never replay a non-idempotent write. The security property from my last review holds.
- All three of my earlier points addressed. Narrowing HandleAuthFailure to http.Header instead of *http.Response is the better fix than the doc contract I suggested: it kills the body-consumption risk at the type level.
- Copilot points (nil-authn panic guard, log wording) resolved.
- CI fully green: build, lint, all three e2e suites, FOSSA.

One forward-looking, non-blocking note for T6. In broker.go, `var jar *resilience.SafeCookieJar` is passed into the CookieJarClearer parameter of NewAuthenticator. For basic auth the jar is always real, so the `jar == nil` guard works today. But a typed nil pointer wrapped in an interface is not `== nil`. If OAuth ever receives a jar through this same interface and nil-checks it, the check would silently pass a non-nil interface holding a nil pointer. Not triggerable on this PR; worth keeping in mind when wiring T6 so the guard is on a concrete type or an explicit mode check, not the interface value.

Nice work.

amitmorade:
> One forward-looking, non-blocking note for T6. In broker.go, `var jar *resilience.SafeCookieJar` is passed into the CookieJarClearer parameter of NewAuthenticator. For basic auth the jar is always real, so the `jar == nil` guard works today. But a typed nil pointer wrapped in an interface is not `== nil`. If OAuth ever receives a jar through this same interface and nil-checks it, the check would silently pass a non-nil interface holding a nil pointer. Not triggerable on this PR; worth keeping in mind when wiring T6 so the guard is on a concrete type or an explicit mode check, not the interface value.

Will take care of this in T6's work, thank you!

PR #137, opened 2026-07-03
SOL-150260: report accurate live queue depth

solace-lootii:
## Summary

`get-queue-metrics` and `list-queues` were surfacing the broker's cumulative
`spooledMsgCount` as if it were live queue depth. This PR reports accurate
current depth instead.

Fixes SOL-150260

## Changes

- **list-queues** (`06cd984`): stop presenting cumulative `spooledMsgCount` as
  live depth in the post-process handler.
- **get-queue-metrics** (`e43fb1f`): reimplemented as a native mixed-protocol Go
  handler (`internal/tools/queuemetrics`) instead of a composite YAML tool — it
  merges the SEMPv2 monitor response with the SEMPv1 show-queue-detail
  `num-messages-spooled` count to report accurate live queue depth.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` all green (Go 1.25).
- Native handler unit tests drive the merge with a SEMPv1 `show queue detail`
  XML fixture (`internal/tools/queuemetrics/testdata`), including the
  live-vs-cumulative distinction.
- Verified live against `lab-129-78`: `get-queue-metrics` returns 200 from both
  the SEMPv2 monitor and SEMPv1 detail calls and merges `liveDepth` (SEMPv1
  `num-messages-spooled`) with `queueMetrics` (SEMPv2, incl. cumulative
  `spooledMsgCount`). A nonzero-depth scenario wasn't exercisable on the test
  host (messaging ports closed — no way to publish/spool), so the nonzero
  live-vs-cumulative gap is covered by the unit tests, not live.
- Rebased onto latest `main` (includes #88); 0 behind.
**- Still I'd like to confirm the behaviour against United's setup with more realistic queue/traffic patterns**

## Breaking Changes

None, same tool name and output shape; the depth value is now correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ The server exposes read-only tools grouped by what they inspect, plus a small se
 | Broker status | `get-broker-status`, `get-redundancy-status` | Snapshot of version, uptime, resources, spool, and HA and mate-link state |
 | Replication | `get-replication-status` | Replication role, sync eligibility, bridge status, transaction mode, and queued-message counts |
 | Message VPN | `list-vpns`, `get-vpn-health`, `get-message-rates` | List VPNs, check per-VPN service health, read message and byte rates |
-| Queues | `list-queues`, `get-queue-metrics` | List queues with depth and throughput; drill into spool, bindings, consumers |
+| Queues | `list-queues`, `get-queue-metrics` | List queues with cumulative spooled count and throughput; drill into spool, bindings, consumers |
 | Clients | `list-clients`, `get-client-details`, `list-client-subscriptions`, `list-slow-subscribers` | List connections, inspect per-client rates and discards, list subscriptions, filter for slow-subscriber-flagged clients |
 | REST Delivery Points | `list-rdps`, `get-rdp-status` | List RDPs; inspect bindings, REST consumers, and last failure reason |
 | Discards | `get-discard-stats`, `list-queue-discards` | Broker-wide and per-VPN discard aggregates; per-queue discard counters |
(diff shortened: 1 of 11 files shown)


solace-aross (approved):
Approving. This is a correct, well-tested fix for a real bug, and the tool descriptions and package comments are a model for the repo.

A few follow-ups, none blocking:

**Track the `list-queues` triage gap.** Dropping `backloggedCount` was the right call, but VPN-wide triage now has no way to answer "which queues are backed up right now." The only guardrail against misreading `spooledMsgCount` as depth is the description text, which is a weak guarantee for an LLM. Is the follow-up capability ticket filed and linked? I'd like it tracked before this drops off the radar.

**Confirm the new dual-plane dependency is intended.** `get-queue-metrics` was SEMPv2-only; it now fails whenever the SEMPv1 call fails. That policy is defensible, and SEMPv1 is already a baseline for `get-broker-status` and the redundancy and discard tools, so the real-world risk is low. Flagging it only so the added coupling is a deliberate choice, not a surprise on a broker where SEMPv1 is restricted but SEMPv2 works.

**Copilot's three comments are all worth acting on.** The `handler.go` one most of all: if SEMPv1 returns `ok` with no `<info>` block, `liveDepth` comes back `{}` and the tool still reports success, even though live depth is the entire point of the tool. Make a missing `currentMsgCount` a hard error, and add the test for it (SEMPv2 succeeds, SEMPv1 empty), since that path isn't covered today.

Nice work on the zero-vs-absent handling via pointers and `omitempty`, and on catching the MB-vs-bytes unit bug beyond the ticket's headline.

PR #138, opened 2026-07-03
SOL-150800, SOL-150801: OAuthAuthenticator, exchanger wiring, and exchange error visibility

amitmorade:
## Summary

Implements two decomposition tickets and a follow-up hardening change:

- **T6 (SOL-150800)**: Adds OAuthAuthenticator — the per-broker authenticator that exchanges agent tokens via RFC 8693
- **T7 (SOL-150801)**: Wires the Exchanger from main.go through BrokerPool so OAuth-mode brokers actually perform token exchange — the "feature turns on" moment
- **SOL-151507**: Fail fast at startup when exchanger construction fails (config-driven, deterministic — no reason to degrade gracefully)

Also surfaces exchange errors through `logToolResult`'s type gate so support engineers get diagnostics on a single correlated log line, and gates all broker OAuth behind `ENABLE_UNRELEASED_BROKER_OAUTH` until the feature ships.

## Type of Change

## Motivation and Context

When an OAuth token exchange fails (IdP rejects, transport error, invalid response), the `detail` field on the `msg="tool invoked" status=error` log line showed only the Go type name (`*fmt.wrapError`) instead of the diagnostic message. The exchange layer logged the error on a separate line with no shared request ID, forcing support engineers to manually correlate across log lines.

Additionally, a broken exchanger config (invalid token URL, bad client credentials) was logged as a warning at startup but the server continued running — every OAuth request would then fail at runtime with the same deterministic error. Since the server has no hot-reload, the operator must restart anyway, so failing fast is the correct behavior.

### T6: OAuthAuthenticator (SOL-150800)
- Add `OAuthAuthenticator` in `internal/semp/auth/oauth.go` for RFC 8693 token exchange
- Move authenticator dispatch from auth package to `broker.go` — `newAuthenticator` is a clean dispatcher, each constructor owns its own invariants
- Accept concrete `SafeCookieJar` in `newAuthenticator` to prevent typed-nil bypass
- `NewOAuthAuthenticator` panics on nil exchanger (wiring bug, not a runtime condition — the exchanger is built once at startup and cannot become nil at runtime)

### T7: Wire Exchanger in main; pool builds Authenticator per broker (SOL-150801)
- In `cmd/server/main.go`: construct the Exchanger once at startup, pass it through to `NewBrokerPool`
- Extract `newTokenExchanger` helper in `main.go`
- Modify `BrokerPool` to accept and hold the Exchanger reference
- `pool.getOrCreate` passes the exchanger to `newAuthenticator` (replacing the nil placeholder from T6)
- Update integration tests for `NewBrokerPool` exchanger parameter

### Fail-fast startup (SOL-151507)
- Change exchanger construction failure from `slog.Warn` + continue to `slog.Error` + `os.Exit(1)`
- Exchanger failures are deterministic and config-driven (no network calls, no transient failures) — if it fails at startup, every runtime exchange will fail identically

### Feature flag gate
- Gate broker OAuth behind `ENABLE_UNRELEASED_BROKER_OAUTH` env var
- When unset/false: `auth.mode: oauth` is rejected at config validation with "not yet supported"
- When `true`: validation bypass allows OAuth brokers for manual E2E testing
- Intentionally undocumented — testing hatch, not a feature toggle; remove when broker OAuth ships

### Exchange error visibility
- Add `ExchangeError` concrete struct type (`internal/tokenexchange/errors.go`) with `LogAttrs()` method
- Replace all `fmt.Errorf` sentinel returns with `&ExchangeError{}` in `response.go` and `exchange.go`
- Enrich `ExchangeError` with per-call context in `Exchange()` using struct copy (singleflight-safe)
- Remove `logExchangeError` — one log line per failure, not two
- Add `errors.As` branch in `logToolResult` and `buildErrorResult`
- Add agent-friendly messages in `buildExchangeErrorMessage`
- Mark `ErrExchangeTransport` as retryable

## Testing

**Test Environment:**
- Go version: 1.26.0
- OS: macOS (darwin)

**Tests Performed:**

**Test Coverage:**
- Integration test (`test/integration/oauth_error_visibility_test.go`) exercises full chain: config → pool → manager → SEMPv1 client → OAuth authenticator → exchanger → fake IdP → error propagation → logToolResult
- All three sentinel errors (ErrExchangeRejected, ErrExchangeTransport, ErrInvalidResponse) verified to surface diagnostic messages in the `detail` field
- "No subject token" case verified to remain type-only (unreachable in production)
- Existing exchanger unit tests updated — `logExchangeError` tests removed, `errors.Is` and `.Error()` assertions still pass
- Race detector clean across all packages
- `/check-logs` clean on all changed files

### Code Quality
- [x] Self-review completed

### Security
- [x] `/check-logs` passed on all changed files
- [x] ExchangeError.Message never contains subject_token, access_token, client_secret, or error_description

### Testing
- [x] No race conditions detected (struct copy in Exchange() prevents singleflight data race)
- [x] Error paths tested

### Git & CI

---

**For Reviewers:**

**Focus Areas**:
- `ExchangeError` type design and `LogAttrs()` method in `internal/tokenexchange/errors.go`
- Singleflight race safety via struct copy in `Exchange()` (`exchange.go:42-49`)
- The `errors.As` branches in `logToolResult` and `buildErrorResult` — same pattern as SEMP error types
- Agent-friendly messages in `buildExchangeErrorMessage` — no raw IdP detail leaked
- Constructor encapsulation: `NewOAuthAuthenticator` owns its nil-exchanger invariant via panic, `newAuthenticator` is a clean dispatcher
- Fail-fast vs degrade-gracefully tradeoff for exchanger construction in `main.go`
- Feature flag gate in `config.go` — intentionally undocumented, testing-only bypass

**Questions**:
- The "no subject token" case stays as `fmt.Errorf` (not `ExchangeError`) since it's unreachable in production due to config validation. Worth converting anyway for consistency?


solace-aross (commented):
Strong PR: clear decomposition, honest self-review, and the security claims hold up. I traced the no-leak assertions against the `ExchangeError` construction sites, the request builder, and the cookie-jar dispatch. No `subject_token` / `client_secret` / `error_description` leaks, HTTPS enforced, and the typed-nil cookie-jar bypass is genuinely closed.

Most of the feedback below clusters around one design decision and one testing gap. None of it blocks on its own; I'd want #1, #2, and #3 either addressed or explicitly deferred-with-rationale before merge. Copilot's five inline comments already cover the nil-exchanger panic, the casing nit, the test slog restore, `t.Setenv("")`, and the nil `ResponseWriter`, so I've left those alone.

## High impact

**1. Token exchange runs on every SEMP request with no cache, so the IdP is now on the hot path.** `internal/semp/auth/oauth.go` (`AddAuth`)
`AddAuth` does a full RFC 8693 round-trip per request. `ExpiresAt` is computed but never used to reuse a token, and singleflight only collapses concurrent identical keys, not steady sequential traffic. Net effect: IdP request volume scales 1:1 with broker traffic, and every request pays up to the 10s IdP timeout when the IdP is slow. An IdP sized for login events becomes a single point of failure for all broker traffic.
Suggestion: confirm exchange-per-request is an intentional v1 with caching as a scheduled follow-up. The unexported `tokenExchanger` interface is a clean seam for a caching/coalescing decorator (key `subjectToken+brokerAlias+audience`, TTL from `ExpiresAt`) without touching `OAuthAuthenticator`. If caching must wait, worth saying so in the PR and confirming IdP QPS = broker QPS is acceptable with the IdP owners.

**2. The agent-facing error surface and the retryable flag are untested.** `internal/tools/errors.go`
`buildExchangeErrorMessage` (all sentinel branches), `error_source="token_exchange"`, and the new `isRetryable(ErrExchangeTransport)==true` branch have no unit coverage; `errors_test.go` has zero exchange cases. The integration test discards the result (`_, _ = mgr.CallTool(...)`) and asserts only the operator log line, so the message the agent receives and the retryable flag are never checked. A flipped retryable flag or a wrong sentinel-to-message mapping would ship silently. Separately, the deleted `logToolResult` tests included an explicit no-secret-leak assertion (subject token / client secret never logged) that wasn't replaced, so that regression guarantee is now unguarded.
Suggestion: add a table test over the three sentinels asserting message, `isRetryable`, and `error_source`; restore one assertion that a constructed `ExchangeError` never contains a token or secret.

**3. The fail-fast plus feature-flag interaction can hard-exit a config that boots today.** `cmd/server/main.go` (`newTokenExchanger`) and `internal/config/config.go`
`newTokenExchanger` runs whenever a `broker_oauth:` block is present, gated only on `cfg.BrokerOAuth != nil`, not on `ENABLE_UNRELEASED_BROKER_OAUTH` and not on any broker actually using oauth mode. A `FromConfig` failure is now `os.Exit(1)`. Two validators can disagree: `config.validate` accepts some schema-valid values that `FromConfig` rejects as not-yet-implemented (for example `audience_parameter_name: scope`). So a staged `broker_oauth` block that passed validation and booted before can now refuse to start, even with the flag off and no broker using OAuth. Forward-looking: once the feature ships, an OAuth misconfig crashes the whole server (taking basic/bearer brokers down with it), and the flag is a test hatch rather than an operator kill switch, so a bad config is unrecoverable without a code change.
Suggestion: gate exchanger construction on an actual oauth broker being configured (not just block presence), or degrade only the affected brokers on construction failure. Worth deciding the blast-radius policy deliberately before removing the guard, and considering a real runtime kill switch.

## Medium impact

**4. The "no subject token" path stays `fmt.Errorf`, which is the exact opaque `%T` failure this PR exists to remove.** `internal/semp/auth/oauth.go`
This is a direct answer to your open question in the description. It returns a plain error, so `logToolResult`'s type gate misses it and `detail` falls back to the bare Go type with none of the `ExchangeError` attrs. It's unreachable only while a runtime invariant holds (Hop-1 middleware always populates the raw subject token). If middleware ordering ever changes, the operator gets `detail="*errors.errorString"` and zero diagnostics. My vote on your question: yes, convert it. Encoding "can't happen" into a test that asserts opaque output is the wrong side of this PR's own goal.

**5. `ErrExchangeTransport` is blanket-retryable and over-assigned.** `internal/tools/errors.go` and `internal/tokenexchange/response.go`
`response.go` classifies non-transient cases (proxy/WAF interception, oversized error code) as `ErrExchangeTransport`, which is marked retryable and surfaced to the agent. So the agent is told to retry failures that will never succeed, and transport retries fire exactly when the IdP is degraded, with no backoff or breaker on the exchange path, piling load on a struggling IdP.
Suggestion: narrow retryable to genuinely transient failures (network / IdP 5xx / 429); treat interception and oversized-code as non-retryable.

**6. Singleflight leader-context cancellation fails collapsed followers.** `internal/tokenexchange/exchange.go` (distinct from the struct-copy race Copilot raised)
The leader runs with the first caller's context. If that caller disconnects, all followers sharing the in-flight key fail with the leader's `ctx` error even though their own deadlines were fine.
Suggestion: consider running the exchange under a derived context with the exchange timeout, detached from any single caller's cancellation.

**7. No happy-path observability for the exchange.** `internal/tokenexchange/exchange.go`
`elapsed` is attached only to the error path; a successful exchange emits no metric or latency log, and the metrics package is still a skeleton. For a dependency now on every request's critical path, operators can't see IdP latency creeping up until requests fail. Partly acknowledged given metrics are pending. Worth at least logging success-path `elapsed` in the interim and tracking the metric as a follow-up.

**8. Hop-2 auth now imports Hop-1 `internal/auth`, so a package doc is now inaccurate.** `internal/semp/auth/oauth.go`
The `internal/semp/auth` package doc states `internal/auth` is "unrelated to SEMP traffic," but the OAuth authenticator now has a compile-time dependency on its context accessor. The coupling is inherent to token exchange (the Hop-1 token is the RFC 8693 subject), so it may well be fine. Worth making it honest: update the doc, or pass the subject token in via an explicit parameter rather than reaching across packages into a context key.

**Building on Copilot's nil-exchanger note (`oauth.go:54`):** the caller side in `broker.go` compounds it. `NewBrokerClient`'s doc says `newAuthenticator` "returns an error if an OAuth broker is configured without an exchanger," but the code panics via `NewOAuthAuthenticator(nil, ...)`, and because broker clients are built lazily that panic lands on the first request (recovered to 500), not at startup. Config validation makes it unreachable today; either honor the doc (return an error) or fix the doc to say it panics.

---

<sub>Reviewed with a multi-perspective pass (architect, QA, security, support, SRE/reliability lenses, then synthesized). Findings #1 and #3 were raised independently by three lenses each. Human-reviewed before posting.</sub>

amitmorade:
## Responses to Andrea's review findings

### #1 — Token exchange on every SEMP request, no cache

Correct — exchange-per-request is intentional v1. Caching is tracked in [SOL-150052](https://sol-jira.atlassian.net/browse/SOL-150052) (token cache with `ExpiresAt` TTL, keyed on `subjectToken+brokerAlias+audience`). The `tokenExchanger` interface is the seam for the caching decorator — designed for exactly this. IdP QPS = broker QPS is accepted for the initial rollout; we'll validate with IdP owners before enabling beyond dev.

### #2 — Untested agent-facing error surface

Addressed. Added in commit `0577afb`:
- `TestIsRetryable` — 5 exchange cases (transport=true, rejected=false, invalid=false, missing subject=false, wrapped transport=true)
- `TestBuildErrorMessage` — all 4 sentinel branches
- `TestBuildExchangeErrorMessage` — all 4 sentinels + unknown fallback
- `TestBuildErrorResult_ExchangeError_StructuredFields` — `error_source="token_exchange"` and retryable flag for rejected (false), transport (true), and missing subject (false)
- `TestExchangeError_NeverContainsSecrets` — asserts agent message, `Error()`, and `LogAttrs()` never contain token/secret values

### #3 — Fail-fast + feature-flag interaction

Deferring with rationale. Today OAuth broker support is unreleased — no docs, no examples, no supported configuration path. Even if someone reverse-engineers the config shape, the feature flag must be explicitly set. Adding deeper gating (e.g. only construct exchanger when a broker actually uses oauth mode, or degrade per-broker instead of fail-fast) is the right eventual design, but would be churn to add and then restructure when the feature ships. Will revisit blast-radius policy as part of the GA readiness work.

### #4 — Opaque `fmt.Errorf` for missing subject token

Addressed. Added `ErrExchangeMissingSubject` sentinel in commit `a86a306`. The "no subject token" path now returns a structured `*ExchangeError` that flows through `buildErrorResult` with `error_source="token_exchange"`, `retryable=false`, and a clear agent message: "Authentication failed: no identity token was found for the current session." The existing `TestOAuthAuthenticator_AddAuth_NoSubjectToken` now also asserts the sentinel.

### #5 — `ErrExchangeTransport` over-assigned and blanket-retryable

Agreed — the two 4xx cases (oversized error code at `response.go:172`, non-OAuth body at `response.go:185`) are misclassified as transport/retryable. Tracked in [SOL-151520](https://sol-jira.atlassian.net/browse/SOL-151520) along with the server-side retry evaluation. Once server-side retry lands, `isRetryable` for `ErrExchangeTransport` should flip to `false` since the server will have already exhausted retries before surfacing the error.

### #6 — Singleflight leader-context cancellation

Addressed. Commit `360c456` detaches the singleflight IdP call onto `context.WithTimeout(context.Background(), exchangeTimeout)` so no single caller's cancellation can abort the shared round-trip. Each caller checks its own `ctx.Err()` after `Do()` returns.

Remaining gap: `group.Do` blocks followers — a cancelled follower can't bail out early, it waits for the leader to finish. This is functionally correct (the follower gets the right context error), but wastes a goroutine for up to 10s. Evaluating `DoChan` migration for follower early-exit in [SOL-151521](https://sol-jira.atlassian.net/browse/SOL-151521) — performance optimization, not correctness.

### #7 — No happy-path observability

Addressed. Commit `8b0ca82` adds `slog.Debug("token exchange succeeded", broker, exchange_elapsed)` on the success path so operators can see IdP latency creeping up before requests fail. Metric instrument tracked in [SOL-151278](https://sol-jira.atlassian.net/browse/SOL-151278).

### #8 — Package doc inaccuracy + broker.go doc

Addressed in two commits:
- `c952592` — updated `internal/semp/auth` package doc to acknowledge the OAuth coupling with `internal/auth` (the Hop-1 token is the RFC 8693 subject, so the dependency is inherent).
- `e23bcd0` — fixed `NewBrokerClient` doc. `newAuthenticator` is a pure dispatcher; each Authenticator constructor owns its own precondition checks. The nil-exchanger guard lives in `NewOAuthAuthenticator`. The basic-auth jar check will also move out of `newAuthenticator` into its own constructor, tracked in [SOL-151506](https://sol-jira.atlassian.net/browse/SOL-151506).

[SOL-150052]: https://sol-jira.atlassian.net/browse/SOL-150052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-151520]: https://sol-jira.atlassian.net/browse/SOL-151520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

solace-aross (commented):
Second pass. Re-traced all eight findings against the current head. Seven resolved; one blocker.

Can't approve while the build is red. `go test -race ./...` fails on a test the #4 fix left behind.

**`test/integration/oauth_error_visibility_test.go:81` — `TestOAuthErrorVisibility/NoSubjectToken`**

Commit `a86a306` converted the missing-subject path to `*ExchangeError` (what #4 asked for), so `logToolResult`'s type gate now surfaces the diagnostic message on the `detail` line. The unit test was updated; this integration test was not. It still asserts the pre-conversion contract:

```go
if !strings.HasPrefix(detail, "*") {
    t.Errorf("[NoSubjectToken] expected Go type only (unreachable in production), got %q", detail)
}
```

CI:

```
expected Go type only (unreachable in production), got "executing tool
\"test-oauth-tool\": applying SEMPv1 auth: oauth auth: no subject token
on context ..."
```

Fix: flip the assertion to expect the diagnostic message now surfaces, and update the stale comment at lines 78-79 (and the helper doc) that still say the path is `fmt.Errorf`, not `ExchangeError`. The conversion made those wrong.

That's the only thing between this and an approve. The rest checks out:

- #2: table tests landed (`TestBuildExchangeErrorMessage`, `TestBuildErrorResult_ExchangeError_StructuredFields`, `TestIsRetryable`); the no-secret-leak guard is restored (`TestExchangeError_NeverContainsSecrets`).
- #4: converted; sentinel, agent message, and structured fields all present.
- #6: exchange detached onto `context.WithTimeout(context.Background(), exchangeTimeout)`, and each caller checks its own `ctx.Err()`. Correct.
- #7: success-path Debug log with `broker` and `exchange_elapsed`.
- #8: package doc and `NewBrokerClient` doc both corrected to match the pure-dispatcher design.
- #1 / #3 / #5: accepted as deferred with tickets (SOL-150052, GA-readiness, SOL-151520).

Also: the red "FOSSA / SCA Scan" check looks like the known-flaky SCA job. FOSSA Guard Licensing and Vulnerability both report 0 new. Re-run before digging in.

Push the test fix and I'll approve on green.

solace-aross (approved):
Approving. The build is green and the missing-subject fix is real: the assertion in oauth_error_visibility_test.go now checks the diagnostic message surfaces rather than the bare type, which is stronger, not looser. Everything from the last pass is addressed or ticketed as agreed (SOL-150052, SOL-151520, SOL-151278).

One leftover, not blocking: the comment at oauth_error_visibility_test.go:78-79 still says the missing-subject path is fmt.Errorf and "type-only", but it now returns *ExchangeError and the assertion just below checks the message. Worth dropping those two lines here or in a trivial follow-up.

PR #139, opened 2026-07-06
fix(safego): recover panics in errgroup worker goroutines [SOL-151514]

solace-aross:
## What was wrong
`errgroup` does not recover panics in the goroutines it starts, and the request-path recovery nets (`recovery.HTTPMiddleware`, `tools.withRecovery` in `internal/tools/register.go`) only trap panics on their own goroutine. A panic in any `g.Go` closure — e.g. `structToMap` in `queuemetrics`, or `applySelect`/`constructRequestBody` on user args in the composite executor — escaped both nets and crashed the entire multi-session server process. This is the spawned-goroutine gap the `recovery` package doc explicitly calls out (`internal/middleware/recovery/middleware.go:65-70`).

## What the fix does
Adds `internal/safego.Go(g, fn)`, which runs `fn` under `g.Go` inside a deferred `recover()` that converts a panic into the goroutine's returned error — logging the panic's Go type and stack only, never its value text (matching the secure-logging rule the other two nets follow). A worker panic now fails that one request instead of the whole process. Applied to all four `errgroup` call sites: `queuemetrics`, `discardstats`, `brokerstatus`, and the composite executor. The happy path is unchanged. A neutral leaf package was required because `internal/composite` cannot import `internal/tools` (import cycle).

## Tests
- Added `TestGo_ConvertsPanicToError`: a panic in a worker goroutine surfaces as `g.Wait()`'s error instead of escaping.
- Added `TestGo_PanicErrorOmitsValueText`: the panic value's text never reaches the returned error.
- Added `TestGo_PassesThroughNormalError`: a normal `fn` error flows through unchanged.
- Added `TestGo_SuccessReturnsNil`: the success path is unaffected.
- Revert-verify: with the recover removed, `TestGo_ConvertsPanicToError` fails (panic escapes and crashes the runner); with it restored, all pass. Full race-enabled suite green.

## Jira
[SOL-151514](https://sol-jira.atlassian.net/browse/SOL-151514)

[SOL-151514]: https://sol-jira.atlassian.net/browse/SOL-151514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/composite/executor.go
+++ b/internal/composite/executor.go
@@ -26,6 +26,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/SolaceDev/solace-broker-mcp/internal/composite/postprocess"
+	"github.com/SolaceDev/solace-broker-mcp/internal/safego"
 	"github.com/SolaceDev/solace-broker-mcp/internal/semp/sempv2"
 )
 
@@ -394,7 +395,7 @@ func (ce *CompositeExecutor) executeBatch(ctx context.Context, batch ExecutionBa
 
 	for _, step := range batch.Steps {
 		step := step // capture loop variable
-		g.Go(func() error {
+		safego.Go(g, func() error {
 			args, err := ResolveArgs(step.Args, execCtx)
 			if err != nil {
 				return fmt.Errorf("tool step %s: failed to resolve args: %w", step.ID, err)
--- a/internal/safego/safego.go
+++ b/internal/safego/safego.go
@@ -1,0 +1,59 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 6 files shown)


solace-aross (commented):
Solid, and the scoping is right. The neutral leaf package dodges the composite->tools import cycle, and it covers every goroutine spawn site: the four errgroup fan-outs plus the one raw go func in main.go, whose only work is ListenAndServe (already covered by the HTTP middleware).

One thing worth a follow-up ticket, not a blocker: the panic-value redaction block is now triplicated (safego.go, recovery/middleware.go, tools/register.go). That redaction is a security invariant, and copy-paste invariants drift. If someone adds %v to one, nothing catches that the others didn't follow. Extract it to one helper. Same theme shows up in #140, so it could be one ticket.

Minor, optional: the recovered-panic error in the composite executor loses the step ID (generic "worker goroutine panicked (%T)"), and it's worth one comment line noting why safego deliberately does not re-raise http.ErrAbortHandler the way the middleware does (no serving frame above a worker goroutine).

solace-awaheed (approved):
Nothing was flagged for the implementation, however, a comment was brought up for the tests: 

> The `safego.Go` tests look good for validating the helper itself, but they don’t quite prove that the actual call sites are wired correctly. Right now they show the helper catches panics in isolation, but not that a panic from the real paths (`structToMap`, `applySelect`, `constructRequestBody`, etc.) gets converted into a request error instead of crashing.
> 
> Maybe we can add a single call-site test as well? Something like an executor step whose work intentionally panics and then verify `Handle` returns an error while the process keeps running. That would help lock in the integration behaviour and protect against future refactors.

Still approving as this is not blocking

amitmorade (approved):
LGTM

PR #140, opened 2026-07-06
fix(sempv1): bound broker error text at capture [SOL-151516]

solace-aross:
## What was wrong
`sempv1.Error.Message` was stored untruncated from the `<parse-error>`, `<permission-error>`, `<limit-error>`, and `<execute-result reason=...>` elements in `parseReply` (`internal/semp/sempv1/envelope.go`). Because the response body is only capped at 16 MiB, a misbehaving or compromised broker could put a multi-MiB string in that text, which then flowed verbatim into `Error.Error()` (logged as the tool-result `detail` field at `manager.go:313`) and through the tool layer's three regex-pass `sanitizeBrokerText` into the agent-facing message (`tools/errors.go:222`). SEMPv2 already truncates the equivalent text at capture (`maxErrorTextLen = 4096`); SEMPv1 did not.

## What the fix does
Adds `truncateErrorText` to `internal/semp/sempv1/errors.go` (rune-safe, 4 KiB cap + `... [truncated]` marker, mirroring the SEMPv2 client) and applies it to the four `Message` capture sites in `parseReply`. Truncating at capture bounds every downstream sink at once.

`Error.Body` is deliberately left untouched: a grep confirms no sink renders it (`Error()` ignores it; the HTTP-error agent message is generic), so it is debug-only transient memory. This is a documented, intentional divergence from SEMPv2, whose `Body` *is* a rendered fallback.

## Tests
- Added `TestTruncateErrorText`: short input unchanged; oversized input cut to cap + marker; multibyte (`€`) input stays valid UTF-8 (rune-boundary backup).
- Added `TestParseReply_TruncatesOversizedMessage`: an oversized `<parse-error>` yields a bounded `Error.Message` ending in the marker.
- Revert-verify: un-wrapping the ParseError assignment made `TestParseReply_TruncatesOversizedMessage` fail (Message = full 16384 bytes); restoring passed. Full sempv1 race suite green.

## Jira
[SOL-151516](https://sol-jira.atlassian.net/browse/SOL-151516)

[SOL-151516]: https://sol-jira.atlassian.net/browse/SOL-151516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/semp/sempv1/envelope.go
+++ b/internal/semp/sempv1/envelope.go
@@ -49,7 +49,7 @@ func parseReply(body []byte) ([]byte, *Error) {
 		return nil, &Error{
 			Kind:       ErrorKindParse,
 			StatusCode: 200,
-			Message:    *r.ParseError,
+			Message:    truncateErrorText(*r.ParseError),
 			Body:       body,
 		}
 	}
@@ -57,7 +57,7 @@ func parseReply(body []byte) ([]byte, *Error) {
 		return nil, &Error{
 			Kind:       ErrorKindPermission,
 			StatusCode: 200,
-			Message:    *r.PermissionError,
+			Message:    truncateErrorText(*r.PermissionError),
 			Body:       body,
 		}
 	}
@@ -65,7 +65,7 @@ func parseReply(body []byte) ([]byte, *Error) {
 		return nil, &Error{
 			Kind:       ErrorKindLimit,
 			StatusCode: 200,
-			Message:    *r.LimitError,
+			Message:    truncateErrorText(*r.LimitError),
 			Body:       body,
 		}
 	}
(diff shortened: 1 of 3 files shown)


solace-aross (commented):
Correct and tested, and a faithful mirror of the SEMPv2 control. Truncation is rune-safe against short/empty input and mid-rune boundaries.

Follow-up, not a blocker: maxErrorTextLen and truncateErrorText are now byte-for-byte copies across sempv1/errors.go and sempv2/client.go. It is a security bound, so if the v2 cap is ever tightened, v1 silently keeps 4096. Hoist to a shared helper. Pairs with the same duplication theme in #139, could be one ticket.

Two minor test notes: the parseReply test exercises only the parse-error site (the other three capture sites are asserted only transitively, a table test would pin all four), and the multibyte assertion checks for the replacement rune where SEMPv2 uses utf8.ValidString, which is the stronger check. Match it for parity.

solace-awaheed (approved):
Nothing flagged on my end, looks good!

PR #141, opened 2026-07-06
fix(config): refuse insecure broker TLS in production unless opted in [SOL-151517]

solace-aross:
## What was wrong
In production (`mcp_client_auth.mode: oauth`) the server enforces `https://` for every broker URL but accepted `insecure_skip_verify: true` with only a `slog.Warn` (`internal/config/config.go`). TLS was "on" while certificate verification was off, so a man-in-the-middle between the MCP server and the broker could present any certificate, terminate TLS, and capture the broker admin credential the server attaches to every SEMP request (`internal/semp/resilience/transport.go:67` wires the flag into `tls.Config`).

## What the fix does
`validate()` now refuses `insecure_skip_verify: true` in production unless a new top-level `allow_insecure_broker_tls: true` opt-in is set — mirroring the existing `allow_remote_unauthenticated` guard for the analogous disabled-mode exposure. The hard error names the broker and the opt-in. When the opt-in is set, the pre-existing startup WARN still fires so the insecure setting stays visible in triage logs. Dev/static modes are unchanged: self-signed brokers remain allowed with no warning.

`ToLower(mode)` normalization now runs at the top of `validate()`, before any broker checks, instead of inline where `mcp_client_auth.mode` was first read. This also fixes a pre-existing bug in `validateBroker` (`config.go:844`): a mixed-case `mode: OAuth` previously skipped the `https://` requirement on broker URLs. `TestLoadConfig_NormalizesModeBeforeBrokerChecks` locks in the ordering.

Documented the new flag in `broker-config.example.yaml` and `docs/configuration.md` (alongside `allow_remote_unauthenticated`), including that it's a single server-wide opt-in rather than per-broker.

Added a CHANGELOG entry — this is a backward-incompatible startup change for anyone currently running `insecure_skip_verify: true` in production.

## Tests
- Added `TestLoadConfig_RejectsInsecureSkipVerifyInProductionModeWithoutOptIn`: production + `insecure_skip_verify` and no opt-in → validation error naming the broker and `allow_insecure_broker_tls`.
- Added `TestLoadConfig_NormalizesModeBeforeBrokerChecks`: mixed-case `mode: OAuth` + `insecure_skip_verify` and no opt-in → still refused (guards the normalization-ordering fix).
- Added `TestLoadConfig_AllowsInsecureSkipVerifyInProductionModeWithOptIn`: same config + `allow_insecure_broker_tls: true` → passes, WARN still emitted.
- `TestLoadConfig_NoWarnOnInsecureSkipVerifyInDevelopmentMode` unchanged and passing (dev mode still allowed).
- Revert-verify: reverting the guard to warn-only made the reject test fail; restoring passed. Full config race suite green.

## Jira
[SOL-151517](https://sol-jira.atlassian.net/browse/SOL-151517)

[SOL-151517]: https://sol-jira.atlassian.net/browse/SOL-151517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solace-aross (commented):
Core logic is right and fail-closed: defaults to false, only fires in production, explicit opt-in required, and a typo'd key fails loudly via unknown-field rejection.

One thing to fix before merge: this is a backward-incompatible startup change (prod + insecure_skip_verify:true used to start with a WARN, now refuses without the opt-in), and there is no CHANGELOG entry. The two analogous tightening changes in [Unreleased] are both marked BREAKING with migration tables. Add an ### Changed BREAKING entry with the one-line migration (add allow_insecure_broker_tls:true or install a trusted cert).

Also worth tightening: the opt-in is a single top-level flag but insecure_skip_verify is per-broker, so setting it to onboard one self-signed broker disables the guard for every broker, including ones added later. It mirrors allow_remote_unauthenticated so the pattern is defensible, but at minimum docs/configuration.md should say the opt-in is global, not per-broker (current wording reads as scoped to the one broker).

helen-zhu-solace (dismissed):
Good, tight fix. Closes a real MITM path and follows the same pattern as `allow_remote_unauthenticated`. The Copilot comment about mode being normalized too late is fixed correctly, with a regression test.

### Approve

Small change, well-tested, docs updated in all three places.

### Suggestions (non-blocking)

**The normalization move fixes more than the PR says.** Moving `ToLower(mode)` to the top of `validate()` also fixes a bug in `validateBroker` at `config.go:844` — before this PR, `mode: OAuth` (mixed case) would have skipped the `https://` requirement for broker URLs too. That's a nice bonus fix, but the PR description only talks about `insecure_skip_verify`. Worth adding a line to the PR body so it's not lost. A second test asserting `mode: OAuth` still rejects `http://` broker URLs would lock this in.

**Global flag, not per-broker.** `allow_insecure_broker_tls` turns off the check for every broker. If someone has one self-signed dev broker mixed with real prod brokers, they have to opt in for all of them. Same shape as `allow_remote_unauthenticated`, so this is consistent — just worth one sentence in the docs saying the flag is server-wide.

**Test name.** `TestLoadConfig_RejectsInsecureSkipVerifyInProductionModeWithMixedCaseMode` is the one that guards the ordering fix, but the name doesn't say that. Something like `..._NormalizesModeBeforeBrokerChecks` makes the intent obvious to whoever breaks it later.

### Nits

- `broker-config.example.yaml:29` — "(disabling broker cert verification exposes …)" reads like the flag causes the exposure. Try "accept the risk that disabling verification exposes …"

Nice work.

amitmorade (dismissed):
LGTM

solace-aross:
Addressed the outstanding feedback:

- Added a `### Changed BREAKING` CHANGELOG entry with the one-line migration (`allow_insecure_broker_tls: true` or a trusted cert).
- `docs/configuration.md` now states the opt-in is server-wide, not per-broker.
- Renamed the mixed-case regression test to `TestLoadConfig_NormalizesModeBeforeBrokerChecks` and called out the mode-normalization bonus fix in the PR description, per @helen-zhu-solace's suggestions.
- Reworded the `broker-config.example.yaml` comment so it reads as accepting a risk rather than the flag causing one.

Rebased on main. CI running now.

amitmorade (approved):
LGTM

helen-zhu-solace (approved):
LGTM too!

PR #142, opened 2026-07-06
fix(resilience): bound the overall SEMP retry chain with a deadline [SOL-151518]

solace-aross:
## What was wrong
`Sender.Do` (`internal/semp/resilience/sender.go`) holds the per-broker semaphore slot across `retryablehttp`'s entire retry chain, but there was no explicit overall deadline on that chain. On the correctly-configured path (`httpClient.Timeout == RequestTimeoutDuration`, as both `sempv1` and `sempv2` clients wire it), the total wall-clock was already bounded — but only *emergently*, by three settings lining up (`Retries+1) × RequestTimeoutDuration + Retries × RetryMaxInterval`) rather than by one explicit, tested invariant.

There's one case where those three settings do **not** bound the chain: `RateLimitLinearJitterBackoff` honors a broker's `Retry-After` header verbatim, uncapped by `RetryMaxInterval`. A broker returning a long `Retry-After` (minutes) could hold a slot for that long per attempt, stalling every other call to that broker once all slots are pinned — that gap is real, not emergent.

## What the fix does
`New()` derives `retryBudget = (RetryMax+1)*RequestTimeout + RetryMax*RetryMaxInterval` — the retry policy's own configured worst case, from existing config (no new knob). `Do` applies it via `context.WithTimeout`. `retryablehttp` honors ctx cancellation during backoff (`client.go:785`), so the slot is released once the budget is spent even if a per-attempt guard is missing or misbehaves.

On the correctly-configured path this is hardening, not a behavior change: it makes an already-implied bound explicit and covered by a test, so it stops depending on three settings staying in sync. On the `Retry-After` path it's a real fix: it caps exposure to a broker announcing an arbitrarily long `Retry-After` instead of letting it park the slot for however long the broker asks.

On success the `cancel` is transferred to `Body.Close` (via a small `cancelOnCloseReadCloser` wrapper) rather than deferred at `Do`'s return — deferring would cancel the request context while the caller is still reading the open response body, aborting the connection instead of pooling it and potentially truncating the body. A zero budget (`RequestTimeout` unset, e.g. some tests) skips the deadline.

## Tests
- Added `TestSender_Do_BoundsOverallRetryChainDeadline`: with **no** per-attempt `httpClient.Timeout` and a slow broker, `Do` returns at ~budget with `context.DeadlineExceeded` and the semaphore slot is released (proves the overall deadline is the effective bound).
- Added `TestSender_Do_RetryBudgetBoundsMultiAttemptChain`: a real `RetryMax=2` chain (3 attempts + backoff, `httpClient.Timeout` wired like production) rather than the single-attempt/zero-backoff case above — elapsed is checked against both the deterministic attempt floor and the budget ceiling.
- Existing `TestClient_TransportPool_ReusesConnections` guards the success-path body/connection behavior — it caught an early `defer cancel()` version and passes with the cancel-on-close design.
- Test helpers now set a realistic `RequestTimeoutDuration` so their budgets are generous.
- Revert-verify: disabling the `WithTimeout` made the new test block ~3s and return no deadline error; restoring passed. Full `internal/semp` race suite green; `go vet` clean.

## Jira
[SOL-151518](https://sol-jira.atlassian.net/browse/SOL-151518)

[SOL-151518]: https://sol-jira.atlassian.net/browse/SOL-151518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/semp/resilience/sender.go
+++ b/internal/semp/resilience/sender.go
@@ -43,9 +43,10 @@ type Sender struct {
 	retryClient   *retryablehttp.Client
 	authenticator auth.Authenticator // for 401 re-auth: delegates recovery to the auth mode
 	rateLimiter   <-chan time.Time
-	rateTicker    *time.Ticker // non-nil when rate limiting enabled; stopped by Close()
-	sem           Semaphore    // bounds in-flight requests; shared per-broker across SEMPv1+v2
-	brokerURL     string       // for logging context
+	rateTicker    *time.Ticker  // non-nil when rate limiting enabled; stopped by Close()
+	sem           Semaphore     // bounds in-flight requests; shared per-broker across SEMPv1+v2
+	brokerURL     string        // for logging context
+	retryBudget   time.Duration // overall deadline for the whole retry chain; 0 disables
 }
 
 // New creates a Sender configured for a specific broker. It sets up retryablehttp
@@ -87,6 +88,20 @@ func New(httpClient *http.Client, sempCfg *config.SEMPConfig, authn auth.Authent
 	retryClient.ErrorHandler = d.errorHandler
 	retryClient.Logger = nil // manual logging in checkRetry
 	d.retryClient = retryClient
+
+	// Overall retry-chain deadline. httpClient.Timeout bounds a single attempt,
+	// but retryablehttp issues up to RetryMax+1 attempts with backoff between
+	// them, and the per-broker semaphore slot (see Do) is held for that entire
+	// span. Without an overall deadline a slow or degraded broker keeps a slot
+	// for the full chain, stalling every other call to that broker. Bound it at
+	// the retry policy's own configured worst case — every attempt at its full
+	// timeout plus every backoff at its cap — so the slot is released once that
+	// budget is spent even if a per-attempt guard is missing or misbehaves. It
+	// never cuts a chain that stays within its configured budget. Zero when
+	// RequestTimeoutDuration is unset (e.g. some tests); Do then skips it.
+	retryMax := *sempCfg.Retries
+	d.retryBudget = time.Duration(retryMax+1)*sempCfg.RequestTimeoutDuration +
+		time.Duration(retryMax)*sempCfg.RetryMaxInterval
 
 	// Per-broker rate limiter: ticker-based interval enforcement.
 	// When interval > 0, each Do() blocks until the ticker fires.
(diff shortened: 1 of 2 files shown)


solace-aross:
CI caught a real regression from the overall-deadline change: `test/integration`'s `TestSenderRetryStateIsolatedUnderConcurrency` constructs a `Sender` directly with `RequestTimeoutDuration` unset (0) and `Retries > 0`. My budget formula then reduced to `retries * RetryMaxInterval` — a backoff-only budget that allots no time to the attempts themselves, so the deadline fired spuriously under concurrent load.

Fixed by anchoring the budget on the per-attempt timeout: it is computed only when `RequestTimeoutDuration > 0`. Production always sets it (`validate()` enforces it), so the deadline is unchanged there; direct-construction tests without a timeout now correctly get no overall deadline. Full `./...` race suite (incl. `test/integration`) is green locally.

solace-aross (commented):
Sound, and leak-free. The deadline is placed after the semaphore slot is acquired so it bounds slot-hold time, composes with any shorter caller deadline, and cancel is handled on every exit path. Off-by-one is right (RetryMax+1 attempts, one full RequestTimeoutDuration left for the final attempt).

Two hardening items, not blockers:

- Document the Retry-After interaction. The configured backoff is RateLimitLinearJitterBackoff, which returns the raw Retry-After unbounded by RetryMaxInterval. So the deadline can cut a chain when a broker returns a long Retry-After, which contradicts the "never cuts a chain within its configured budget" comment. That is actually the sharpest real win here (it caps a degraded broker parking you for minutes), so name it in the comment rather than leaving it implicit.
- The new test runs one attempt and zero backoffs, so it does not exercise the multi-attempt chain the title claims. Add a retries>0 variant (503 loop, small RetryMaxInterval) asserting elapsed is approximately the budget and the slot is released.

solace-wmaryam:
**Not blocking — question on necessity.**

In the shipped config both constructors set `http.Client.Timeout = RequestTimeoutDuration`, so the old accidental bound and the new `retryBudget` work out to the *same* number: `(Retries+1) × RequestTimeoutDuration + Retries × RetryMaxInterval`. So on the correctly-configured path this changes no timing and closes no currently-reachable gap — attempts are already bounded, retries finite, backoff capped.

What it does add: the total slot-hold time was only ever implied by three settings lining up; this makes it one explicit, tested invariant. Nice insurance on a severe-and-silent failure mode (one slow broker starving all calls to it) — but it costs the `cancelOnCloseReadCloser` mechanism, which already caused a bug mid-development (the early `defer cancel()`).

So this is hardening, not a bug fix. Genuinely your call whether the explicit guarantee is worth the complexity or redundant enough to skip. If we keep it, worth tweaking the description to say it's a no-behavior-change hardening rather than closing an existing gap.

solace-aross:
Addressed both hardening items:

- Documented the Retry-After interaction in the retryBudget comment — a broker's uncapped Retry-After can legitimately cut a chain short, and that's the deadline doing its job (capping the exposure), not a bug.
- Added TestSender_Do_RetryBudgetBoundsMultiAttemptChain: a real RetryMax=2 chain (3 attempts + backoff, httpClient.Timeout wired like production) instead of the single-attempt/zero-backoff case the existing deadline test covers. Elapsed is asserted against both the deterministic attempt floor and the budget ceiling; stable across repeated runs and under -race.

solace-aross:
Fair question. Reworded the description: on the correctly-configured path (httpClient.Timeout == RequestTimeoutDuration, which both constructors already set) this is hardening — makes an already-implied bound explicit and tested, doesn't change timing. Keeping it because of the Retry-After case: RateLimitLinearJitterBackoff honors Retry-After verbatim, uncapped by RetryMaxInterval, so that path isn't bounded by the three settings lining up. That's the genuine gap this closes, and I think it's worth the cancelOnCloseReadCloser complexity for it — but if you'd rather drop the general-case hardening and scope this to just the Retry-After cap, say the word.

solace-wmaryam:
Looks good!

solace-wmaryam (approved):

PR #143, opened 2026-07-06
feat(cmd/server): second signal forces immediate shutdown [SOL-151437]

solace-aross:
## What

Once graceful shutdown begins, a **second** SIGINT/SIGTERM now forces an immediate exit instead of being buffered and ignored until the drain window and shutdown timeout elapse.

Resolves [SOL-151437](https://sol-jira.atlassian.net/browse/SOL-151437). Follow-up to the graceful-shutdown work in SOL-151288.

## Why

On shutdown the server flips `/readyz` to 503, waits the drain window (SIGTERM only), then drains in-flight requests up to the shutdown timeout. A second signal was previously dropped, so an impatient second Ctrl-C in local dev appeared to do nothing until the timeout elapsed. In Kubernetes this is bounded by the grace-period SIGKILL, so it is primarily a **local-dev ergonomics** improvement, not a correctness fix.

## How

- Thread a receive-only force-signal channel through `drainAndShutdown` and `gracefulShutdown`. The startup-error path passes `nil`; a nil channel case in a `select` never fires, so that path keeps its exact prior behavior (including the Shutdown-timed-out forced-close fallback).
- **Drain window:** `select` on (timer, second signal); a second signal forces close and returns, skipping the rest of the sequence.
- **Graceful wait:** run `srv.Shutdown` in a goroutine with a buffered result channel (no goroutine leak); `select` on (completion, second signal); a second signal calls `srv.Close`, which unblocks the in-flight `Shutdown`, and we return.
- Shared `forceClose` helper logs one WARN line naming the signal and stating that in-flight connections are dropped, then calls `srv.Close`.
- Bump the `done` signal channel buffer 1→2 so a rapidly delivered second signal is not dropped before the shutdown path arms its `select`.

## Scope

`cmd/server` only. No config, deployment, or API changes. Single-signal path unchanged: SIGTERM still honors the full drain window then drains gracefully; SIGINT still skips the drain. Readiness still flips to 503 before any forced exit.

## Acceptance criteria

- [x] Second signal during the drain window → immediate forced close, rest of drain skipped
- [x] Second signal during the graceful wait → immediate forced close and return
- [x] Single-signal path unchanged (SIGTERM drains, SIGINT skips)
- [x] Forced exit logs a single clear WARN line naming the signal
- [x] No goroutine leak, no double-close panic on any path

## Testing

`make check` passes (build, vet, golangci-lint, race-enabled tests). New tests are deterministic via a `blockingShutdowner` fake whose `Shutdown` blocks until `Close` — no real multi-second sleeps:

- Second signal during drain → immediate `Close`, drain not completed
- Second signal during graceful wait → immediate `Close`, and the `Shutdown` goroutine is asserted to exit (guards the no-leak AC)
- `forceClose` emits exactly one WARN naming the signal (guards the WARN-line AC)
- `Close` called exactly once on every force path (guards the no-double-close AC)

## Review

Reviewed pre-merge through a five-lens principal pass (architect, QA, security, support, SRE). Security clean. Applied the confirmed findings: the WARN line now states connections are dropped (converged support + SRE), and the three missing acceptance-criteria tests were added. Two forward-looking items were considered and deliberately deferred (a distinct return sentinel for forced vs clean shutdown; a deployment.yaml note on duplicate-SIGTERM) as out of scope / YAGNI.

[SOL-151437]: https://sol-jira.atlassian.net/browse/SOL-151437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -295,15 +295,44 @@ func startServer(srv *http.Server, tlsCertFile, tlsKeyFile string) <-chan error 
 // the full shutdownTimeout to drain in-flight requests, exactly as the K8s
 // terminationGracePeriodSeconds calculation in deploy/kubernetes/deployment.yaml
 // assumes (grace = drainDelay + shutdownTimeout + buffer).
-func drainAndShutdown(srv shutdowner, readiness *health.ReadinessState, drainDelay, shutdownTimeout time.Duration) error {
+//
+// forceSig carries a SECOND shutdown signal (SOL-151437). A second SIGINT/SIGTERM
+// arriving while the drain window is sleeping short-circuits the rest of the
+// sequence: forceClose drops in-flight connections and we return immediately,
+// skipping the graceful wait. Readiness has already flipped to 503 above, so a
+// forced exit still leaves /readyz correct. A nil forceSig (the startup-error
+// path, which never drains) simply never fires that select case.
+func drainAndShutdown(srv shutdowner, readiness *health.ReadinessState, drainDelay, shutdownTimeout time.Duration, forceSig <-chan os.Signal) error {
 	readiness.SetShuttingDown()
 	slog.Info("draining before shutdown",
 		slog.String("reason", "signal"),
 		slog.Duration("drain_delay", drainDelay))
 	if drainDelay > 0 {
-		time.Sleep(drainDelay)
+		timer := time.NewTimer(drainDelay)
+		defer timer.Stop()
+		select {
+		case <-timer.C:
+			// Drain window elapsed normally; fall through to the graceful wait.
+		case sig := <-forceSig:
+			forceClose(srv, sig)
+			return nil
+		}
 	}
-	return gracefulShutdown(srv, shutdownTimeout)
+	return gracefulShutdown(srv, shutdownTimeout, forceSig)
+}
+
+// forceClose is the shared "second signal" reaction (SOL-151437): a single WARN
+// line naming the triggering signal, then srv.Close() to drop in-flight
+// connections immediately. Close is the only way to actually drop connections —
+// cancelling Shutdown's context would merely stop the wait and leave them open.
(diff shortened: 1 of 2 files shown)


solace-wmaryam:
Went through the four Copilot comments. Only one looks worth acting on; the other three don't hold up.

  Worth addressing:
  - shutdown_test.go:308 — 30s drain window: Reasonable point. The long window is intentional (the test
  asserts it finishes faster than the window, so the window has to be long), but if the code regresses the
  test would hang for 30s before failing. Shrinking it to a few seconds would keep the check meaningful
  while failing fast.

  Can be dismissed:
  - main.go:396 — "shutdown timed out" message: Shutdown only fails on a timeout or a listener-close
  error; it never returns http.ErrServerClosed (that comes from Serve). The message is accurate, and this
  line isn't new in this PR.
  - main.go:333 — logging Close errors: Close doesn't return http.ErrServerClosed either. It's called
  exactly once per path (asserted by closeCount == 1), so there's no benign error being logged as a
  failure.
  - shutdown_test.go:400 — global logger racing with parallel tests: No race. This test doesn't call
  t.Parallel(), so Go runs it during the serial phase and restores the logger via t.Cleanup before the
  parallel tests in panic_recovery_test.go resume — serial tests always complete before paused parallel
  ones start. It would only race if this test became parallel, which is what the existing "don't make this
  parallel" comment guards against.

solace-aross:
Shrunk the drain window in TestDrainAndShutdown_SecondSignalDuringDrainForcesClose from 30s to 3s, per your comment — it still leaves a wide margin over the microsecond-scale elapsed time the test expects, but a regression now fails in 3s instead of hanging for the full 30s -timeout guard.

solace-wmaryam (approved):

PR #147, opened 2026-07-06
SOL-151574: gate Hop-2 OAuth runtime resources on ENABLE_UNRELEASED_BROKER_OAUTH

amitmorade:
## Summary

Splits config validation from runtime construction for Hop-2 OAuth (MCP server ↔ broker). Config validation continues to run for every YAML shape as before; runtime resources (IdP HTTP client, token exchanger, in-memory client secret) are now built only when the experimental runtime is actually going to be used.

Fixes #(issue number)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The Hop-2 broker-side OAuth runtime is experimental and not yet validated end-to-end against real IdPs. It ships hidden behind the `ENABLE_UNRELEASED_BROKER_OAUTH` env var, and the config validator rejects any broker declaring `auth.mode: oauth` unless that flag is set — so the runtime cannot execute in a normal release.

Before this change, one gap remained. If an operator populated the top-level `broker_oauth:` block but did not switch any broker to `auth.mode: oauth`, config validation passed (it is a valid staging shape, and only emits a WARN), but startup then went ahead and built:

- an `*http.Client` for the IdP,
- a `*tokenexchange.Exchanger` with the client secret held in memory,
- and threaded the exchanger into every broker client in the pool.

None of these were reachable from any request path — the switch on `cfg.Mode` in `newAuthenticator` never hit the `AuthModeOAuth` case — but they existed. This is wasteful in the "no broker uses it" case, and structurally risky in a weekly release cadence where experimental code paths should not construct resources they cannot use.

## Changes Made

- Added `(*ServerConfig).Hop2OAuthActive() bool` in `internal/config/config.go` — the single source of truth for "should Hop-2 OAuth runtime run in this process." True only when all three preconditions hold: `ENABLE_UNRELEASED_BROKER_OAUTH` set truthy, `broker_oauth:` populated, and at least one broker has `auth.mode: oauth`.
- Gated the construction of the token exchanger in `cmd/server/main.go` on `cfg.Hop2OAuthActive()`. The gate lives in `main()` — the wiring layer, where startup decisions belong — rather than inside the constructor.
- `newTokenExchanger` is now a pure builder with a stated precondition; it no longer self-protects.
- Extended the `validate()` doc-comment in `config.go` to state the flag contract explicitly (validation always runs; runtime construction is gated).
- Updated the LIFECYCLE comment on `unreleasedBrokerOAuthEnabled` to describe correct ship-time removal — delete only the flag portion of `Hop2OAuthActive`; the two structural preconditions remain, since building an exchanger for a config that does not consume it would still be waste post-ship.

## Testing

**Test Environment:**
- Go version: (per repo `go.mod`)
- OS: macOS Darwin 24.6.0 (local)
- Broker version (if applicable): N/A — no runtime broker interaction on this code path when flag off

**Tests Performed:**

**Test Coverage:**

- `make test` — all production and internal packages pass. `TestOAuthErrorVisibility` (the Hop-2 integration test) still passes because it sets `ENABLE_UNRELEASED_BROKER_OAUTH=true` AND configures an oauth-mode broker — all three preconditions hold.
- No existing test asserted on the construction of the exchanger in the "provided but not used" scenario, so no tests broke. Existing tests either exercise the config validator (unaffected), or construct exchangers directly for the pool (bypass `newTokenExchanger` entirely), or set the flag and configure oauth mode (still constructs as before).
- Not added: a dedicated unit test for `Hop2OAuthActive`. Willing to add one if reviewers want it — I opted to keep the change minimal since the method is a straight combination of preconditions already covered elsewhere.

## Breaking Changes

Not a breaking change for shipped functionality. The runtime being gated here has never been documented as consumable by operators; the `broker_oauth:` fields are not present in `broker-config.example.yaml` or any operator-facing doc. Behaviour differences:

- Config-only shapes (validation): **unchanged**.
- Flag off, `broker_oauth:` present, no broker uses oauth: **used to build idle resources; now builds nothing.** Observable difference: the `token exchanger created for broker OAuth` INFO log no longer fires in this shape.
- Flag on with all three preconditions met: **unchanged.**

### Code Quality

### Security
- [x] Followed secure logging rules — no changes to logging surface aside from the existing INFO/WARN lines
- [x] Credential-carrying types implement `slog.LogValuer` — unchanged (already the case for `BrokerOAuthConfig.LogValue`)
- [ ] No security vulnerabilities introduced (checked with golangci-lint/gosec) — did not run gosec locally

### Testing
- [x] All tests pass locally: `go test -race ./...` (via `make test`)
- [ ] Edge cases covered by tests — see note above; no new dedicated tests added
- [x] Error paths tested — construction errors already covered by existing tests when flag is on

### Documentation
- [ ] README.md updated (if user-facing changes) — N/A, no user-facing change

### Git & CI
- [ ] All commits are signed off (DCO: `git commit -s`) — commit is not DCO-signed; happy to amend if the project requires it
- [x] Branch is up to date with main (branched off current `main` at `1883393`)
- [x] No merge conflicts
- [ ] CI checks passing (lint, build, test, E2E) — will confirm after PR is created

- N/A

N/A.

## Additional Notes

**Lifecycle plan for ship time.** When the Hop-2 OAuth runtime ships and is ready for release, only the flag portion of `Hop2OAuthActive` needs to be removed. The two structural preconditions (`broker_oauth:` populated AND at least one broker uses `auth.mode: oauth`) are the correct final gate and should remain. The `unreleasedBrokerOAuthEnabled` helper, its constant, the WARN in `applyEnvOverrides`, the `validateBroker` guard, and the `LogOAuthNotSupported` banner branch are all listed in the LIFECYCLE comment on `unreleasedBrokerOAuthEnabled` and should be removed in the same ship-time PR.

**Not addressed in this PR.** The `docs/oauth/…` directory contains internal design documents that reference the OAuth-on-brokers plan; none of them are operator-facing, so this PR doesn't update them. Happy to add a follow-up if reviewers see anything that now contradicts the code.

## Related Issues/PRs

- Related to # (SOL-150070 series — Hop-2 broker OAuth runtime)
- Depends on #
- Blocks #

---

**For Reviewers:**

**Focus Areas**:
1. The doc-comment on `ServerConfig.Hop2OAuthActive` and on `newTokenExchanger` — these encode the contract the change adds; if they read confusingly, that will misdirect the next person modifying this path.
2. Whether the gate should also cover a hypothetical second Hop-2 resource site. Today `newTokenExchanger` is the only one; the maintainer note in the constructor doc-comment says any future constructor must consult `Hop2OAuthActive`. That is convention, not enforcement — is that acceptable, or would you prefer a structural guard (e.g. an unexported "runtime present" token that must be threaded through)?
3. The signature change on `newTokenExchanger` (from `*BrokerOAuthConfig` to `*ServerConfig`). It is a small ergonomic ripple but expands what the constructor can reach. The alternative was to have `main()` pass just what's needed; I chose the `ServerConfig` form so the constructor can call `cfg.Hop2OAuthActive()` locally if the check ever moves back down.

**Questions**:
- Should I add a dedicated unit test for `Hop2OAuthActive` covering the eight (2×2×2) truth-table cells? Kept out for minimality; happy to add if you'd like coverage there.
- Should the commit be DCO-signed? Recent commits on `main` are not signed off; wanted to confirm before amending.


solace-aross (commented):
**Sound change, one comment to fix before merge.** The gate is correct: `Hop2OAuthActive()` faithfully ANDs the three preconditions, `main()` is the right home for the decision, and making `newTokenExchanger` a pure builder is the right call. No correctness or security issue in the executable path. The substance sits in the load-bearing comment this PR exists to add.

Copilot already raised all three of my substantive points, so I won't restate them — here's where I land on each, plus answers to your three focus questions.

### 1. The "FLAG CONTRACT" comment is wrong, and it's the deliverable (blocker)

Copilot is right, and this is the one that matters: a wrong contract comment misdirects the next maintainer harder than no comment would. Lines 1029-1034 claim an operator staging `auth.mode: oauth` without the flag "will still see the SAME validation errors and WARN messages they would see with the flag." The code in the same function says otherwise:

- **`validateBroker` (line 1183):** flag off appends the "not yet supported" error; flag on skips it. A config that *fails* validation flag-off *passes* flag-on.
- **`validate` (line 1048):** flag off logs `LogOAuthNotSupported`; flag on skips that banner and runs `validateHop1Hop2Alignment`, which can append a different error and log `LogHop2WithoutHop1`.

The new comment also contradicts your own adjacent bypass comments at 1043-1047 and 1180-1182, which correctly describe the flag changing exactly these outputs.

The distinction worth encoding: the flag does not change schema/shape validation, but it does gate the temporary not-yet-supported rejection and swap the banner. Suggested reframe:

> The flag does not change schema/shape validation — malformed `broker_oauth:` or `auth.mode: oauth` YAML is rejected identically either way. It does gate the temporary not-yet-supported guard: flag-off, an oauth-mode broker is rejected with an error and the not-supported banner; flag-on, that rejection is bypassed (see the guard in `validateBroker` and the banner site above) and the Hop-1/Hop-2 alignment check runs instead. Runtime construction (IdP client, exchanger, in-memory secret) is separately gated by `Hop2OAuthActive`.

### 2. `newTokenExchanger` "does no I/O at construction" (fix the wording, low priority)

Copilot is right here too. `idpclient.NewHTTPClient` reads `SSL_CERT_FILE` (`os.ReadFile`) and calls `x509.SystemCertPool()` — filesystem/OS I/O. Your intent is clearly "no outbound network calls," so say that: change "None of these do I/O at construction time" to "None of these make an outbound network call at construction time; the first IdP request happens when Exchange() is called."

### 3. Missing unit test for `Hop2OAuthActive` (agree, worth adding)

Add it. It's a small truth table, but it gates whether a client secret lands in memory — exactly the decision point where a cheap test is insurance against a future refactor silently flipping a precondition. You don't need all eight cells; the four that pin each precondition's contribution (all-true then each one flipped) are enough.

### Your three focus questions

1. **Docs read clearly?** The `Hop2OAuthActive` and `newTokenExchanger` doc-comments are good — precondition and lifecycle are explicit. The `validate()` FLAG CONTRACT block is the exception; see above.
2. **Convention vs. structural guard for a future second resource site?** Convention (the doc-comment "any future constructor must consult `Hop2OAuthActive`") is fine for one site behind a flag that's already on a defined removal path. A threaded "runtime present" token buys enforcement you'll delete at ship time — over-engineering for a single call site. Don't build it now.
3. **`newTokenExchanger(*ServerConfig)` vs `*BrokerOAuthConfig`?** Minor, not a blocker, and I'd lean the other way. The gate already lives in `main()` and calls `Hop2OAuthActive()` there, so the constructor doesn't need `ServerConfig` to re-check — it only reads `cfg.BrokerOAuth`. The narrower signature keeps the builder's inputs to exactly what it consumes; "so the check can move back down" is a hypothetical I wouldn't optimize for today. Your call.

solace-aross (commented):
Not approving yet. Checked against the current head (still b235821, unchanged since the last review) — the blocker from the earlier review is still open: the "FLAG CONTRACT" comment in `validate()` (config.go ~1023-1034) still claims operators see identical validation errors/WARNs regardless of the flag. That's contradicted by `validateBroker`'s guard at line 1183 and the banner/alignment-check branch in `validate()` at line 1048, both of which change behavior on `unreleasedBrokerOAuthEnabled()`. Verified by reading both call sites directly.

The rest of the change is sound (gate placement in `main()`, `Hop2OAuthActive()` logic, `newTokenExchanger` as a pure builder) — this is the one thing standing between here and merge. Happy to re-review as soon as the comment is corrected and, ideally, the `Hop2OAuthActive` unit test is added per the earlier feedback.

solace-aross (approved):
All four items addressed. Verified the corrected FLAG CONTRACT comment against the actual branches (validateBroker's guard at ~1170, the banner/alignment-check if/else at ~1035) — it now accurately describes that the flag doesn't touch schema validation but does gate the not-yet-supported rejection and banner. The narrowed `newTokenExchanger(*BrokerOAuthConfig)` signature is the right call — take what you consume. `TestServerConfig_Hop2OAuthActive` proves each precondition is load-bearing, which is exactly the risk (a silent AND-to-OR slip would leak a client secret into memory for a config that shouldn't build one).

CI green, `go test -race ./internal/config/... ./cmd/server/...` clean. Good tightening.

PR #149, opened 2026-07-07
SOL-151506: harden BasicAuthenticator with nil-jar panic

amitmorade:
## Summary

Move the `jar != nil` invariant from `HandleAuthFailure` and the `newAuthenticator` dispatcher into `NewBasicAuthenticator` itself — panic at construction, delete the two now-dead defensive checks. Same pattern as `NewOAuthAuthenticator` in SOL-150800.

Fixes [SOL-151506](https://sol-jira.atlassian.net/browse/SOL-151506)

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue) — the old nil-jar branch masked wiring bugs; the new panic surfaces them

## Motivation and Context

The cookie jar is a **core dependency** of `BasicAuthenticator` — without it, `HandleAuthFailure` cannot clear stale session cookies after a 401, so the retry-on-auth-failure path is broken. A `BasicAuthenticator` constructed with a nil jar is an object that cannot function correctly in production.

Despite that, `NewBasicAuthenticator` accepted `nil` silently and `HandleAuthFailure` returned `retry=false` in that case, hiding the wiring bug instead of surfacing it. The dispatcher `newAuthenticator` in `broker.go` had a duplicate `if jar == nil` check for the same invariant.

In production, `newCookieJar` always creates a real jar for basic auth mode, so the nil path is unreachable — but that's a wiring guarantee, not a type-system one. Centralizing the check in the constructor makes the invariant explicit and lets us remove two now-dead defensive branches.

Rationale for `panic` vs `error`: the jar is built once at startup and set at construction — it can't become nil at runtime. `NewBasicAuthenticator` is internal wiring, never called by external users. Nil jar is a programmer error, so fail-fast is correct. Same reasoning as `NewOAuthAuthenticator`.

## Changes Made

- `internal/semp/auth/basic.go`: `NewBasicAuthenticator` panics on nil jar; `HandleAuthFailure`'s nil guard removed; doc comment updated to describe the new contract.
- `internal/semp/broker.go`: `newAuthenticator` no longer pre-checks `jar == nil` for basic auth — the constructor owns that invariant now.
- **Tests: fixed 4 brittle setups.** Four existing tests passed nil jar as a setup shortcut even though the jar was irrelevant to what they verified. They now pass a real jar / stub, testing what they actually meant to test:
  - `internal/semp/auth/authenticator_test.go`: `TestBasicAuthenticator_AddAuth`, `TestAuthenticator_ConcurrentAddAuth_NoFieldMutation/basic`
  - `internal/semp/resilience/sender_test.go`: `TestSender_RateLimiter_PerBrokerIndependence`
  - `internal/semp/sempv1/client_test.go`: `TestHTTPClient_LogValue_ExcludesCredentials` (basic-auth table row)
- **Tests: added 3 direct unit tests** in a new `internal/semp/auth/basic_test.go`:
  - `TestNewBasicAuthenticator_PanicsOnNilJar` — defends the new panic contract
  - `TestBasicAuthenticator_HandleAuthFailure_ClearsJarAndReturnsTrue` — defends the happy path directly (was only covered transitively via `resilience.Sender`)
  - `TestBasicAuthenticator_HandleAuthFailure_JarClearError` — defends the `jar.Clear()`-fails branch (previously **zero** coverage at any level)

## Testing

**Test Environment:**
- Go version: go1.26.0 darwin/arm64
- OS: macOS (Darwin 24.6.0)
- Broker version: N/A (unit + integration only; no wire-level behavior change)

**Tests Performed:**
- [x] Unit tests added/updated (3 new, 4 fixed)
- [x] Integration tests added/updated (N/A — existing integration tests unchanged and still pass)
- [ ] E2E tests added/updated (N/A — no wire behavior change)
- [ ] Tested with real Solace broker (N/A)

**Test Coverage:**
- New panic contract is defended directly by a `defer/recover` test.
- Both branches of `HandleAuthFailure` (success + `jar.Clear()` error) now have direct unit tests with a `recordingJar` fake, in addition to the pre-existing end-to-end resilience-layer coverage.
- `make check` (build + vet + lint + race unit + integration tests) all green.
- Critical existing tests confirmed intact:
  - `TestBrokerClient_SharedJar_401ClearVisibleAcrossProtocols` (cross-protocol jar sharing)
  - `TestSender_Retry_401_BasicAuth_ClearsCookiesAndRetries` (401 → clear → retry)
  - `TestSenderRetryStateIsolatedUnderConcurrency` (50 goroutines, `-race`: 100/100 hits)

## Breaking Changes

**Impact:** None user-visible. Production wiring goes through `newAuthenticator`, which always supplies a jar for basic auth. Any code that direct-constructed a `BasicAuthenticator` with `nil` jar was already broken (couldn't handle a 401) — such code will now fail fast at construction instead of silently at first 401.

**Migration Guide:** N/A.

### Code Quality
- [x] Self-review completed
- [x] Code is well-commented (WHY, not WHAT)

### Security
- [x] No credentials in code
- [x] Followed secure logging rules (only pre-existing `slog.Warn` retained verbatim; no new logging introduced)

### Testing
- [x] Edge cases covered by tests (both success and jar.Clear-error branches)
- [x] Error paths tested (jar.Clear() error path was previously untested; now covered)

### Documentation
- [x] godoc comments updated on `NewBasicAuthenticator` and `HandleAuthFailure` to describe the new contract
- [ ] README.md updated — N/A (internal wiring change)
- [ ] docs/ updated — N/A (no architecture change)

### Git & CI
- [x] Commits have clear, descriptive messages (split into production change + test change for reviewer clarity)
- [x] No merge conflicts

## Additional Notes

While auditing this surface I noticed `NewOAuthAuthenticator` panics on nil exchanger but has **no test** defending that contract. Filed as follow-up [SOL-151601](https://sol-jira.atlassian.net/browse/SOL-151601) — out of scope here.

## Related Issues/PRs

- Related to SOL-150800 (same encapsulation pattern applied to `NewOAuthAuthenticator`)
- Parent epic: SOL-148417 (Enterprise-Ready Broker MCP: Multi-User Auth & Performance at Scale)
- Follow-up: [SOL-151601](https://sol-jira.atlassian.net/browse/SOL-151601)

---

**For Reviewers:**

**Focus Areas:**
- Whether the panic contract is the right choice vs. returning an error (see rationale under "Motivation and Context")
- Whether the `stubJar` in `authenticator_test.go` and `recordingJar` in `basic_test.go` should be unified or whether the split (no-op stub vs. recording fake) is worth keeping. I kept them separate because they serve different intents.

**Questions:**
- The `jar.Clear()` error branch currently logs `WARN` and returns `retry=false`. That's the pre-existing behavior; I added the test to pin it. If we ever want jar-clear failures to be louder (return an error, surface a metric), that should be a separate ticket — happy to file if desired.

[SOL-151506]: https://sol-jira.atlassian.net/browse/SOL-151506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/semp/auth/basic.go
+++ b/internal/semp/auth/basic.go
@@ -17,10 +17,15 @@ type BasicAuthenticator struct {
 }
 
 // NewBasicAuthenticator returns a BasicAuthenticator that will attach the
-// given credentials to every request via http.Request.SetBasicAuth.
-// jar may be nil — HandleAuthFailure will return retry=false when no jar
-// is available.
+// given credentials to every request via http.Request.SetBasicAuth. The
+// jar is required — HandleAuthFailure clears it to force fresh Basic
+// credentials on the retry, so an authenticator without a jar cannot
+// recover from a 401. Panics if jar is nil — a nil jar is a wiring bug,
+// not a runtime condition.
 func NewBasicAuthenticator(username, password string, jar CookieJarClearer) *BasicAuthenticator {
+	if jar == nil {
+		panic("NewBasicAuthenticator: jar must be non-nil")
+	}
 	return &BasicAuthenticator{username: username, password: password, jar: jar}
 }
 
@@ -34,12 +39,8 @@ func (a *BasicAuthenticator) AddAuth(_ context.Context, req *http.Request) error
 
 // HandleAuthFailure clears stale session cookies so the next request
 // re-sends raw Basic credentials. Returns retry=true on success so the
-// Sender retries the request. Returns retry=false when no jar is
-// available or the clear fails.
+// Sender retries the request. Returns retry=false when the clear fails.
 func (a *BasicAuthenticator) HandleAuthFailure(_ context.Context, _ http.Header) bool {
-	if a.jar == nil {
-		return false
-	}
 	if err := a.jar.Clear(); err != nil {
 		slog.Warn("basic auth: 401 received but failed to clear cookie jar",
 			slog.String("error", err.Error()))
--- a/internal/semp/auth/basic_test.go
+++ b/internal/semp/auth/basic_test.go
@@ -41,6 +41,29 @@ func TestNewBasicAuthenticator_PanicsOnNilJar(t *testing.T) {
 	NewBasicAuthenticator("admin", "s3cret", nil)
 }
 
+// TestNewBasicAuthenticator_TypedNilJar_DoesNotPanic pins the accepted
+// limit of the nil-jar guard: a typed-nil implementation (Go's classic
+// "nil interface vs. nil concrete" gotcha — see
+// https://go.dev/doc/faq#nil_error) produces a non-nil interface value
+// and slips past the == nil check.
+//
+// This is intentional and matches the ecosystem convention: defend at
+// the producer, not the consumer. Production wiring (newCookieJar →
+// newAuthenticator) never emits a typed-nil, so the case is unreachable
+// in production. If we ever adopt reflect-based detection or a linter
+// that upgrades this to a compile-time error, this test will fail — and
+// that failure is the intended signal that we're deliberately changing
+// the contract.
+func TestNewBasicAuthenticator_TypedNilJar_DoesNotPanic(t *testing.T) {
+	var typedNil *recordingJar // nil pointer, but wrapping it in the interface produces a non-nil header
+	defer func() {
+		if r := recover(); r != nil {
+			t.Fatalf("NewBasicAuthenticator with typed-nil jar panicked unexpectedly: %v", r)
+		}
+	}()
+	_ = NewBasicAuthenticator("admin", "s3cret", typedNil)
+}
+
 // TestBasicAuthenticator_HandleAuthFailure_ClearsJarAndReturnsTrue pins
 // the happy-path contract: on a 401, HandleAuthFailure calls jar.Clear()
 // exactly once and returns retry=true so the Sender re-sends with fresh
--- a/internal/semp/auth/basic_test.go
+++ b/internal/semp/auth/basic_test.go
@@ -22,11 +22,20 @@ func (r *recordingJar) Clear() error {
 // a nil jar is a wiring bug and must fail fast at construction, not later
 // at first 401 recovery attempt. Same rationale as
 // NewOAuthAuthenticator's nil-exchanger panic.
+//
+// The panic message is part of the assertion so a future refactor that
+// panics for a different reason (bad username, etc.) can't quietly pass
+// this test while dropping the nil-jar guard.
 func TestNewBasicAuthenticator_PanicsOnNilJar(t *testing.T) {
+	const wantMsg = "NewBasicAuthenticator: jar must be non-nil"
 	defer func() {
 		r := recover()
 		if r == nil {
 			t.Fatal("NewBasicAuthenticator(_, _, nil) did not panic")
+		}
+		got, ok := r.(string)
+		if !ok || got != wantMsg {
+			t.Fatalf("panic value = %#v, want string %q", r, wantMsg)
 		}
 	}()
 	NewBasicAuthenticator("admin", "s3cret", nil)
--- a/internal/semp/auth/basic_test.go
+++ b/internal/semp/auth/basic_test.go
@@ -1,0 +1,68 @@
+package auth
+
+import (
+	"context"
+	"errors"
+	"testing"
+)
+
+// recordingJar records Clear() invocations and can be configured to return
+// an error from Clear() to exercise the failure branch of HandleAuthFailure.
+type recordingJar struct {
+	clearErr   error
+	clearCalls int
+}
+
+func (r *recordingJar) Clear() error {
+	r.clearCalls++
+	return r.clearErr
+}
+
+// TestNewBasicAuthenticator_PanicsOnNilJar pins the constructor contract:
+// a nil jar is a wiring bug and must fail fast at construction, not later
+// at first 401 recovery attempt. Same rationale as
+// NewOAuthAuthenticator's nil-exchanger panic.
+func TestNewBasicAuthenticator_PanicsOnNilJar(t *testing.T) {
+	defer func() {
+		r := recover()
+		if r == nil {
+			t.Fatal("NewBasicAuthenticator(_, _, nil) did not panic")
+		}
+	}()
+	NewBasicAuthenticator("admin", "s3cret", nil)
+}
+
+// TestBasicAuthenticator_HandleAuthFailure_ClearsJarAndReturnsTrue pins
+// the happy-path contract: on a 401, HandleAuthFailure calls jar.Clear()
+// exactly once and returns retry=true so the Sender re-sends with fresh
+// Basic credentials. Covered transitively by resilience.Sender tests, but
+// isolating it here defends the auth-side contract against refactors of
+// the retry loop that might silently drop the Clear() call.
+func TestBasicAuthenticator_HandleAuthFailure_ClearsJarAndReturnsTrue(t *testing.T) {
+	jar := &recordingJar{}
+	a := NewBasicAuthenticator("admin", "s3cret", jar)
+
+	if !a.HandleAuthFailure(context.Background(), nil) {
+		t.Error("HandleAuthFailure returned false, want true on successful clear")
+	}
+	if jar.clearCalls != 1 {
+		t.Errorf("jar.Clear() called %d times, want 1", jar.clearCalls)
+	}
+}
+
+// TestBasicAuthenticator_HandleAuthFailure_JarClearError pins the failure
+// branch: if jar.Clear() fails, HandleAuthFailure returns retry=false so
+// the Sender does not retry with a possibly-inconsistent jar. This branch
+// was previously untested at any level.
+func TestBasicAuthenticator_HandleAuthFailure_JarClearError(t *testing.T) {
+	sentinel := errors.New("clear failed")
+	jar := &recordingJar{clearErr: sentinel}
+	a := NewBasicAuthenticator("admin", "s3cret", jar)
+
+	if a.HandleAuthFailure(context.Background(), nil) {
+		t.Error("HandleAuthFailure returned true, want false when jar.Clear() fails")
+	}
+	if jar.clearCalls != 1 {
+		t.Errorf("jar.Clear() called %d times, want 1", jar.clearCalls)
+	}
+}
--- a/internal/semp/broker.go
+++ b/internal/semp/broker.go
@@ -111,9 +111,6 @@ func newAuthenticator(alias string, brokerCfg *config.BrokerConfig, jar *resilie
 	cfg := brokerCfg.Auth
 	switch cfg.Mode {
 	case config.AuthModeBasic:
-		if jar == nil {
-			return nil, fmt.Errorf("basic auth requires a cookie jar for broker %q", alias)
-		}
 		return auth.NewBasicAuthenticator(cfg.Username, cfg.Password, jar), nil
 	case config.AuthModeBearer:
 		return auth.NewBearerAuthenticator(cfg.Token), nil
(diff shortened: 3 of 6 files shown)


internal/semp/auth/basic.go
```diff
@@ -34,12 +39,8 @@ func (a *BasicAuthenticator) AddAuth(_ context.Context, req *http.Request) error
 
 // HandleAuthFailure clears stale session cookies so the next request
 // re-sends raw Basic credentials. Returns retry=true on success so the
-// Sender retries the request. Returns retry=false when no jar is
-// available or the clear fails.
+// Sender retries the request. Returns retry=false when the clear fails.
 func (a *BasicAuthenticator) HandleAuthFailure(_ context.Context, _ http.Header) bool {
-	if a.jar == nil {
```
amitmorade, line 40: Removed this guard on purpose. Whether `a.jar` can be nil is not this function's problem — it's a construction-time invariant, not a runtime condition. Now that `NewBasicAuthenticator` panics on nil jar, this branch is unreachable, and keeping it would leave a dead defensive check that lies about the contract (implying "callers might get here with a nil jar," which they can't).

Single responsibility: `HandleAuthFailure` clears cookies. Validating the jar exists belongs to the constructor.

internal/semp/resilience/sender_test.go
```diff
@@ -366,9 +366,9 @@ func TestSender_RateLimiter_PerBrokerIndependence(t *testing.T) {
 	}))
 	defer serverB.Close()
 
-	senderA := newTestSender(t, serverA.Client(), auth.NewBasicAuthenticator("admin", "secret", nil), 0)
+	senderA := newTestSender(t, serverA.Client(), auth.NewBasicAuthenticator("admin", "secret", mustNewSafeCookieJar(t)), 0)
```
amitmorade, line 369: This test doesn't use the jar for anything — it verifies rate-limiter isolation between two brokers. The old nil-jar was a setup shortcut that only worked because the pre-hardening constructor accepted it. The new panic contract forces the shortcut out; a real jar is the minimum needed to construct the authenticator, nothing more.

Not a signal that this test is exercising cookie behavior — it isn't. `mustNewSafeCookieJar` is inert here.

internal/semp/broker.go
```diff
@@ -111,9 +111,6 @@ func newAuthenticator(alias string, brokerCfg *config.BrokerConfig, jar *resilie
 	cfg := brokerCfg.Auth
 	switch cfg.Mode {
 	case config.AuthModeBasic:
-		if jar == nil {
```
amitmorade, line 114: Removed the pre-check here on purpose. This dispatcher is a pure switch on auth mode — it shouldn't carry mode-specific invariants that each authenticator's constructor already owns. `NewBasicAuthenticator` now panics on nil jar; duplicating the check here would mean the invariant lives in two places, and the two places could drift.

You might notice we lose the "broker %q" alias in the error message. That's fine: reaching this branch with a nil jar requires `newCookieJar` to have returned nil for `AuthModeBasic`, which it can't (it only returns nil for non-basic modes). If that changes in the future, the constructor's panic will point to a real wiring bug in one place, which is what we want — not two half-baked checks racing each other.

Same pattern OAuth already follows: `newAuthenticator` doesn't pre-check the exchanger; `NewOAuthAuthenticator` panics if it's nil. This aligns basic auth with that shape.

internal/semp/auth/basic.go
```diff
@@ -17,10 +17,15 @@ type BasicAuthenticator struct {
...
+// given credentials to every request via http.Request.SetBasicAuth. The
+// jar is required — HandleAuthFailure clears it to force fresh Basic
+// credentials on the retry, so an authenticator without a jar cannot
+// recover from a 401. Panics if jar is nil — a nil jar is a wiring bug,
+// not a runtime condition.
 func NewBasicAuthenticator(username, password string, jar CookieJarClearer) *BasicAuthenticator {
+	if jar == nil {
+		panic("NewBasicAuthenticator: jar must be non-nil")
+	}
 	return &BasicAuthenticator{username: username, password: password, jar: jar}
```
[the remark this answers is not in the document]
amitmorade: Aware of the typed-nil trap — [Go FAQ covers it](https://go.dev/doc/faq#nil_error). The convention isn't to defend against it at the consumer, it's to make sure the producer never emits a typed-nil interface in the first place.

Our producer is already safe by construction: `newCookieJar` returns `*resilience.SafeCookieJar` (concrete), not `CookieJarClearer` (interface). The only path a typed-nil could reach this constructor is if some future caller explicitly wrote `var j *SafeCookieJar; NewBasicAuthenticator(..., j)`, which isn't a caller pattern production wiring produces.

The `reflect.ValueOf(...).IsNil()` approach is a common suggestion but it's considered a code smell in idiomatic Go — it treats a caller mistake as a runtime concern instead of a wiring concern, and drags `reflect` into a hot-path constructor. `nilness`/`nilaway` linters are the ecosystem-standard tool for catching typed-nil producers at compile-time; not adopting one here, but that's the right layer if we ever need it.

The alternative you mentioned — restructuring to take `*SafeCookieJar` directly and check before interface conversion — would be cleaner if it were free, but `CookieJarClearer` exists specifically to break an import cycle between `auth` and `resilience` (see the doctring on the interface in `auth.go`). Reintroducing that cycle to defend a caller pattern production can't produce isn't a good trade.

Applied the panic-message assertion from your other comment. Not changing this one.

internal/semp/auth/basic.go
```diff
@@ -17,10 +17,15 @@ type BasicAuthenticator struct {
...
-// jar may be nil — HandleAuthFailure will return retry=false when no jar
-// is available.
+// given credentials to every request via http.Request.SetBasicAuth. The
+// jar is required — HandleAuthFailure clears it to force fresh Basic
+// credentials on the retry, so an authenticator without a jar cannot
+// recover from a 401. Panics if jar is nil — a nil jar is a wiring bug,
+// not a runtime condition.
 func NewBasicAuthenticator(username, password string, jar CookieJarClearer) *BasicAuthenticator {
+	if jar == nil {
+		panic("NewBasicAuthenticator: jar must be non-nil")
```
solace-aross, line 27: Minor, non-blocking: this panic carries no broker alias, unlike the dispatcher-level error it replaced (`"basic auth requires a cookie jar for broker %q"`). If this invariant is ever broken by a future wiring change, `withRecovery` only logs `panic_type=string` (never the panic text, per the secure-logging rules), so there'd be no way to tell *which* broker failed from either the panic or the logs.

`NewOAuthAuthenticator` has the identical gap today, so this isn't a new inconsistency this PR introduces — just flagging it since fail-fast invariants are only useful if the failure is diagnosable when they trip. Might be worth a shared follow-up (could ride on SOL-151601) to have `NewBrokerClient` recover-and-rewrap these constructor panics with broker context before they reach the tool-handler recovery layer.

internal/semp/auth/basic_test.go
```diff
@@ -0,0 +1,77 @@
...
+//
+// The panic message is part of the assertion so a future refactor that
+// panics for a different reason (bad username, etc.) can't quietly pass
+// this test while dropping the nil-jar guard.
+func TestNewBasicAuthenticator_PanicsOnNilJar(t *testing.T) {
+	const wantMsg = "NewBasicAuthenticator: jar must be non-nil"
+	defer func() {
+		r := recover()
+		if r == nil {
+			t.Fatal("NewBasicAuthenticator(_, _, nil) did not panic")
```
solace-aross, line 34: Optional: this passes an untyped `nil` literal, which only exercises the true-nil-interface branch of `jar == nil`. A typed nil (`var jar *resilience.SafeCookieJar; NewBasicAuthenticator("u", "p", jar)`) produces a non-nil interface and would skip the panic — the same typed-nil case already discussed with Copilot and settled as unreachable in production.

Not asking for the reflect-based defense you already correctly declined — just noting the test as written can't distinguish "panics on any nil" from "panics only on true nil." A cheap added case with a typed nil would pin down and document the actual (currently unverified) behavior for that path.

solace-aross (dismissed):
Clean, well-scoped hardening change. Verified `go build`, `go vet`, and the affected packages (`go test -race`) pass on this branch; matches the existing `NewOAuthAuthenticator` panic-on-nil-dependency pattern, and the four "brittle test" fixes plus three new direct tests actually exercise what they claim.

Also reviewed the existing thread — the typed-nil-interface question Copilot raised is correctly resolved: I traced `newCookieJar` → `newAuthenticator` myself and confirmed it can never hand a nil/typed-nil jar to `NewBasicAuthenticator` in production wiring, and a panic here (even if it somehow fired) is contained by `withRecovery` — one failed tool call, not a process crash. Nothing blocking.

Two small, non-blocking suggestions inline — neither should hold up the merge.

internal/semp/auth/basic.go
amitmorade: Good catch — the diagnostic gap is real and I'd missed the `withRecovery` redaction interaction. Agreed the fail-fast panic only earns its keep if the failure is diagnosable, and right now it isn't: the operator sees "a broker tool panicked" with no way to identify which broker's wiring is bad.

You're also right that OAuth has the identical gap today, so this isn't a regression introduced by this PR — but it's a coherent story worth fixing symmetrically for both auth modes, not layering more asymmetry on.

Widened [SOL-151601](https://sol-jira.atlassian.net/browse/SOL-151601) to cover it:
- Part 1: symmetric panic-contract tests for `NewOAuthAuthenticator` (mirroring the Basic pair)
- Part 2: broker-context in constructor panics for **both** auth modes, with two candidate designs — a localized string-based one (Design A) and a typed-panic-value + recovery-layer update (Design B). Preference is B because it's the only shape that survives the secure-logging redaction you flagged; Design decision goes in the ticket, not this PR.

Not blocking the merge, per your call. Thanks for tracing this all the way through the recovery layer.

internal/semp/auth/basic_test.go
amitmorade: Fair call — I originally left this as prose-only (in the reply to Copilot on the sibling thread) with the reasoning that pinning "typed-nil does NOT panic" via a test would pin something we don't want to promise. But you're right that the *opposite* framing is stronger: if we ever change our mind and adopt reflect-based detection (or a `nilaway`/`nilness` linter that upgrades this to a compile-time error), a failing test is the correct signal that we're deliberately changing the contract — not a silent behavior drift.

Added `TestNewBasicAuthenticator_TypedNilJar_DoesNotPanic` in commit `2e76319`. Uses a typed-nil `*recordingJar` (already in the test file) to trigger the Go interface-header gotcha, and asserts no panic. Doc-comment references the Go FAQ and explains why the test-failure-on-future-change is the intended signal.

The symmetric coverage for OAuth is folded into the widened [SOL-151601](https://sol-jira.atlassian.net/browse/SOL-151601).

solace-aross (approved):
Re-reviewed after the earlier round of comments. Both points from my prior review are resolved: SOL-151601 now covers diagnosability for both auth modes, and `TestNewBasicAuthenticator_TypedNilJar_DoesNotPanic` pins the accepted typed-nil behavior.

Independently re-traced the typed-nil path through `newCookieJar` -> `newAuthenticator` -> `NewBasicAuthenticator` in broker.go: `newCookieJar` only returns a non-nil `*SafeCookieJar` on the success path for `AuthModeBasic` (the error path returns before `newAuthenticator` runs), so the typed-nil interface conversion discussed on this thread can't occur through production wiring today. Agreed with the resolution.

On the open question about unifying `stubJar` and `recordingJar`: keep them separate. `stubJar` exists only to satisfy the non-nil-jar precondition in tests that don't care about jar behavior; `recordingJar` asserts call counts and injects errors. Merging them would force trivial construction sites to carry unused recording state for no benefit.

Verified on `2e76319`: build, vet, and `go test -race ./internal/semp/...` all green; CI (build/lint/e2e-basic/e2e-oauth/e2e-monitoring/FOSSA) passing. Clean, well-scoped hardening change, consistent with the `NewOAuthAuthenticator` precedent. Approving.

PR #158, opened 2026-07-09
SOL-150455: E2E functional tests for MCP server action tools

solace-lootii:
## Summary

Adds an end-to-end functional test suite (`test/e2e-action/`) for the four action tools — `delete-queue-messages`, `clear-queue-stats`, `disconnect-client`, `clear-client-stats` — driving each over the MCP JSON-RPC wire and verifying the broker state actually changes, on both brokers. Also promotes the `broker-driver` and its shell lifecycle helpers into `test/e2e-common/` now that a second suite depends on them.

Jira: **SOL-150455** (blocked-by SOL-148462; builds on SOL-150725, now merged)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The action tools are SEMPv2 Action PUTs that mutate broker state (clear stats, delete spooled messages, disconnect clients). Until now only unit tests covered them; nothing validated their real end-to-end effect. This suite exercises each tool over the live MCP wire and asserts the state actually changed — plus cross-broker isolation and tool-annotation exposure — on every CI run.

It builds on the `e2e-common` scaffold from SOL-150725. Since the action suite is the first **messaging-based** peer of the monitoring suite, the `broker-driver` (the SMF client that produces spooled messages / connected clients — state SEMP/curl can't create) graduated from monitoring-local to shared.

## Changes Made

- **New `test/e2e-action/` suite** (`.env`, `docker-compose.yml`, `helpers.sh`, `run-all.sh`, `test-action-tools.sh`, `README.md`) — 13 tests: each of the 4 action paths on broker-a and broker-b, plus read-after-write consistency, cross-broker isolation (deleteMsgs + disconnect), and a `tools/list` annotation check. Per-test fixture ownership with entry/exit sweeps.
- **Promoted the broker-driver to shared scaffolding:** moved `broker-driver/` → `test/e2e-common/broker-driver/`, and lifted `build_broker_driver` / `wait_for_pidfile` / `stop_broker_drivers` (glob generalized to `broker-driver-*.pid`) into `e2e-common/lib.sh`. Monitoring now sources them from there — behavior-preserving.
- **Wiring:** new `e2e-action` CI job, `make e2e-action[-up|-down|-all]` targets, `.gitignore` `.env` exception, and the `go vet` broker-driver path updated to `test/e2e-common/broker-driver`.
- **No production Go changes.** Post-action assertions reuse the existing `verify_monitor_object` (its optional jq predicate polls until a field matches), so no new shared helper was needed.

**Assertion fields verified against a real broker — and where they intentionally deviate from the ticket text (which predates design changes), documented in the suite README:**
- `delete-queue-messages` → asserts `liveDepth.currentMsgCount == 0`, **not** cumulative `spooledMsgCount` (which stays at N — per SOL-150260).
- `clear-client-stats` → asserts `dataTxMsgCount == 0`; SEMP client counters are broker-centric, so a *receiving* client bumps `dataTxMsgCount`, not `dataRxMsgCount`.
- `disconnect-client` → asserts a new broker-assigned `clientId`; the SDK auto-reconnects under the same name in ~1s, so "absent from `list-clients`" would flake.
- Annotations → asserts `destructiveHint`; there is no `requiresConfirmation` annotation (confirmation lives in the tool description prose).
- FR-3 "cache invalidation" is reframed as **read-after-write consistency** — the server has no response cache (reads hit the broker live), so there is nothing to invalidate. See the note below.

## Testing

**Test Environment:**
- Go version: 1.25.0
- OS: Linux
- Broker version (if applicable): `solace/solace-pubsub-standard:latest` (two containers, broker-a/broker-b)

**Tests Performed:**
- [ ] Tested locally with `go test -race ./...` *(no production Go changed; suite is bash-driven E2E)*

**Test Coverage:**
- New `e2e-action` suite: **13/13 pass** on both brokers (all 4 action tools, read-after-write, both cross-broker isolation tests, annotations). Clean re-run confirmed (entry sweep) — order-independent, no leaked fixtures.
- Regression: `e2e-monitoring` still **16/16 fixtures + 46/46 tool tests** after the broker-driver move + helper lift.
- `go vet ./...` and `go vet` on the moved broker-driver: clean.

## Breaking Changes

None — test-only. The shared-scaffold change (broker-driver move + helper lift) is behavior-preserving; the monitoring suite was re-run and passes unchanged.

**Impact:** n/a

**Migration Guide:** n/a

## Checklist

- [x] My code follows the project's style and conventions
- [x] I have performed a self-review of my changes
- [x] I have commented non-obvious logic where needed
- [x] I have updated relevant documentation (suite `README.md`)
- [x] My changes generate no new warnings (`go vet` clean)
- [x] New and existing E2E suites pass locally (action + monitoring)
- [x] No production Go code changed (`go vet ./...` clean)


solace-aross (dismissed):
Approving. Test-only, no production Go touched, and the single red check (`e2e-oauth`) is an unrelated CI infra flake: a 504 from github.com while Terraform pulled the keycloak provider, nothing in this diff. A re-run should clear it.

What I verified beyond the diff:

- The broker-driver move is a clean 100%-similarity rename, and the helper lift into `e2e-common/lib.sh` is behavior-preserving. No stale `e2e-monitoring/broker-driver` references remain on the branch, monitoring no longer redefines the lifted helpers, and the updated `go vet` path plus the green `e2e-monitoring` job confirm nothing broke.
- The glob generalization from `broker-driver-f*.pid` to `broker-driver-*.pid` stays safe because `BIN_DIR` is suite-scoped, so one suite can't reap another's drivers.

What stood out as good: every post-action assertion polls rather than reading once, fixtures are per-test with entry and exit sweeps, and the README's "load-bearing broker facts" (live depth vs cumulative `spooledMsgCount`, broker-centric `dataTxMsgCount`, the disconnect/reconnect `clientId` signal) is exactly the context the next person needs to avoid asserting the wrong thing.

Nothing blocking. Nice work.

test/e2e-action/test-action-tools.sh
```diff
@@ -0,0 +1,367 @@
...
+    spawn_action_client "$BROKER_B_SEMP_CONFIG" b "broker-b" "$BROKER_B_URL" "$c" "$q" "$ACTION_TOPIC_DISC" "act-disc-iso" \
+        || { log_fail "isolation(disconnect): client on broker-b failed"; stop_broker_drivers; semp_delete "$BROKER_A_SEMP_CONFIG" "msgVpns/$BROKER_VPN/queues/$q"; semp_delete "$BROKER_B_SEMP_CONFIG" "msgVpns/$BROKER_VPN/queues/$q"; return 1; }
+
+    ca=$(read_client_id "$BROKER_A_URL" "$c"); cb=$(read_client_id "$BROKER_B_URL" "$c")
+    if [ -z "$ca" ] || [ -z "$cb" ]; then
+        log_fail "isolation(disconnect): could not read both clientIds (a='$ca' b='$cb')"; rc=1
+    else
+        call_tool_ok "disconnect-client" \
+            "$(jq -nc --arg c "$c" '{broker:"broker-a",msgVpnName:"default",clientName:$c}')" \
+            "isolation(disconnect): disconnect on broker-a" \
```
amitmorade: The `A && { … } || rc=1` shape reads like `if A then B else C`, but bash actually runs it as *"run A; if A succeeds run B; if the last thing that ran (which could be B) returned non-zero, run C."* So `|| rc=1` fires either when `call_tool_ok` fails **or** when the block's last statement returns non-zero.

It works here because the block ends on `[ "$cb_after" = "$cb" ] || { …; rc=1; }` — both outcomes leave `rc` correct. But the safety comes from the order of assertions inside the block, not from the construct. Reorder them, or add a new final assertion that returns 0 on success, and either a real `call_tool_ok` failure gets swallowed or the test flakes.

The other reason I'd swap it: agents pattern-match on the repo to decide what "our style" is, so any construct we merge tends to spread. Explicit `if/then/else` removes the trap and doesn't leave a shape that only reads correctly when you also check the invariants.

```bash
if call_tool_ok "disconnect-client" "$args" "isolation(disconnect): disconnect on broker-a"; then
    poll_client_reconnected "$BROKER_A_URL" "$c" "$ca" || { log_fail "..."; rc=1; }
    local cb_after; cb_after=$(read_client_id "$BROKER_B_URL" "$c")
    [ "$cb_after" = "$cb" ] || { log_fail "..."; rc=1; }
else
    rc=1
fi
```

Non-blocking.

solace-aross (approved):
Re-reviewing after the head advanced past my earlier (dismissed) approval. The only new author work since then is c55b2f6f, which rewrites the disconnect-isolation test to explicit if/then/else. That is amitmorade's suggestion, applied as written.

Confirmed it's sound: rc starts at 0, each assertion sets it independently, and return $rc no longer depends on the block's last statement, so the construct is robust to reordering assertions. The rest of the range is the main merge, which is clean with no conflict markers in the suite files.

Original review stands: test-only, no production Go, behavior-preserving. Approving.

PR #161, opened 2026-07-10
SOL-151601: Diagnosable constructor-panic contracts for Basic and OAuth authenticators

amitmorade:
## Summary

Makes authenticator constructor panics **diagnosable** — the tool-handler recovery layer can now log the broker alias when a constructor's wiring invariant fires. Also adds the missing symmetric panic-contract tests for `NewOAuthAuthenticator`.

Jira: [SOL-151601](https://sol-jira.atlassian.net/browse/SOL-151601)

## Type of Change

- [x] Refactoring (no functional changes) — dispatcher-guard collapse

## Motivation and Context

Two gaps surfaced by [PR #149](https://github.com/SolaceDev/solace-broker-mcp/pull/149) (SOL-151506) — Andrea Ross raised the diagnostic concern in review:

**Gap 1 — Missing panic-contract test for `NewOAuthAuthenticator`.** SOL-151506 introduced a nil-jar panic in `NewBasicAuthenticator` with tests defending it. `NewOAuthAuthenticator` already had the equivalent nil-exchanger panic (`internal/semp/auth/oauth.go:36-38`) but no test — a future refactor could quietly drop the guard.

**Gap 2 — Constructor panics aren't diagnosable.** Both constructors panicked with a bare string like `"NewBasicAuthenticator: jar must be non-nil"`. When `withRecovery` catches such a panic, secure-logging rules require dropping the panic text (panic values are unaudited, `docs/internal/secure-logging-rules.md`). So the operator saw *"a broker tool panicked"* with no way to identify which broker's config tripped the invariant. The dispatcher's prior error string (`"basic auth requires a cookie jar for broker %q"`) carried the alias — that context regressed when the check moved to the constructor.

## Changes Made

Eight commits, each independently reviewable — the sequence is intentional so `git bisect` stays useful:

1. **`752856b` Add panic-contract tests for NewOAuthAuthenticator** — the two symmetric tests (Part 1 of the ticket): asserts the untyped-nil panic and pins the typed-nil accepted-limit.
2. **`9592ccb` Add WiringError type for authenticator wiring panics** — the typed panic value: `BrokerAlias`, `Reason`. Kept local to `internal/semp/auth` per rule-of-three (2 callers today, both in-package); lift path documented in the type comment.
3. **`b1864ce` Add broker alias parameter to NewBasicAuthenticator** — pure signature threading (11 files, +29/-27). No behavior change; prep for commit 4.
4. **`6461c8c` Panic with WiringError in both authenticator constructors** — both constructors now panic with `auth.WiringError{...}`. The 4 panic-contract tests updated in the same commit to assert the typed shape (type, `BrokerAlias`, `Reason`).
5. **`8a3bd9b` Log broker alias from WiringError panics in tool recovery** — `withRecovery` type-switches on `auth.WiringError` and appends `broker_alias` + `wiring_reason` as pre-vetted structured fields. All other panic types continue to redact per the existing secure-logging rule.
6. **`48c1d1e` Remove redundant nil-exchanger guard from newAuthenticator** — the dispatcher's `if exchanger == nil` block is now dead code; the constructor owns the invariant. Dispatcher goes back to pure auth-mode routing.
7. **`75e6c10` Post-review cleanups from stale-reference sweep** — fixed a stale docstring on `NewBrokerClient` describing the deleted error-return path; softened the cross-reference in HTTP-mux `recovery.HTTPMiddleware` so it no longer claims identical behavior with `withRecovery` after the typed-exception was added; removed the unused `brokerAlias` field from `BasicAuthenticator`'s struct (the constructor still passes it through to the panic, so the operator diagnostic is intact).
8. **`a0cf100` Pass untyped nil at OAuth dispatcher boundary to preserve WiringError** — flagged by @solace-aross: after collapsing the dispatcher guard, `newAuthenticator` was passing the concrete `*tokenexchange.Exchanger` directly into the `tokenExchanger` interface parameter — a typed-nil interface header that silently bypassed the constructor's `== nil` guard. Fixed at the producer boundary (consistent with the "defend at the producer" principle we settled on in PR #149's Basic discussion). Added a defending test at the dispatcher boundary so a future "cleanup" refactor of the pre-conversion check fails loudly.

**Design decision — Design B (typed panic) chosen over Design A (localized string):** The secure-logging contract in `withRecovery` (`internal/tools/register.go:65-70`) drops all panic-value text because panic values are unaudited — a typed panic with named fields is the only shape that survives redaction while keeping the rule intact. Fields are logged because *the field's contract* declares them safe, not because they happen to be strings. See the `WiringError` type doc in `internal/semp/auth/wiring.go`.

## Testing

**Test Environment:**
- Go version: 1.24
- OS: Darwin 24.6.0

**Tests Performed:**
- [x] Integration tests — pre-existing `test/integration/retry_state_isolation_test.go` compiles/runs against the new signature
- [ ] E2E tests — not applicable (no wire-protocol changes)
- [x] Tested locally with `go test -race ./cmd/... ./internal/... ./test/...` — green
- [ ] Tested with real Solace broker — not applicable (panic path only fires on wiring bugs; verified via panic-recovery tests)

**Test Coverage:**

Four constructor-side panic-contract tests, symmetric across Basic and OAuth:
- `TestNewBasicAuthenticator_PanicsOnNilJar` / `TestNewOAuthAuthenticator_PanicsOnNilExchanger` — assert panic value is `auth.WiringError` with the exact `BrokerAlias` + `Reason` fields.
- `TestNewBasicAuthenticator_TypedNilJar_DoesNotPanic` / `TestNewOAuthAuthenticator_TypedNilExchanger_DoesNotPanic` — pin the accepted limit of the `== nil` guard (Go's typed-nil-vs-untyped-nil gotcha). Defend at producer, not consumer.

One recovery-layer test:
- `TestWithRecovery_WiringErrorLogsBrokerAlias` — direct unit test on `withRecovery` (no MCP client/server/transport). Asserts the log record contains `broker_alias`, `wiring_reason`, and `panic_type=auth.WiringError`. The sibling `TestPanicAuditedAsError` continues to defend the non-`WiringError` redaction path.

One dispatcher-boundary test:
- `TestNewBrokerClient_OAuth_NilExchanger_PanicsWithWiringError` — pins the `newAuthenticator` → `NewOAuthAuthenticator` boundary contract. With OAuth mode and a nil `*tokenexchange.Exchanger`, `NewBrokerClient` must panic with `auth.WiringError` carrying the broker alias. Defends against a future refactor that "cleans up" the pre-conversion nil check in `newAuthenticator` as redundant and silently re-introduces the typed-nil trap.

## Breaking Changes

**None user-facing.** All changes are internal to `internal/semp/auth` and `internal/tools`. `NewBasicAuthenticator`'s signature grew by one param (`brokerAlias string`), but it's not part of any external API — only production caller is `internal/semp/newAuthenticator` (one line updated), plus test call sites (all updated in the same commit).

### Code Quality
- [x] Self-review completed
- [x] Complex logic has explanatory comments — see `WiringError` type doc, `withRecovery`'s inline comment on the typed-exception, `newAuthenticator`'s updated docstring on constructor-owned invariants + the interface-conversion hygiene note

### Security
- [x] Followed [secure logging rules](docs/internal/secure-logging-rules.md) — the recovery-layer change is a targeted exception with explicit rationale: `WiringError`'s two fields are declared in-package as safe-to-log (broker alias from config, reason string authored by trusted in-package code). All other panic types continue to redact. Rule not weakened.
- [x] Credential-carrying types implement `slog.LogValuer` — n/a for this change; `WiringError` does not carry credentials

### Testing
- [x] All tests pass locally: `go test -race ./cmd/... ./internal/... ./test/...`
- [x] Edge cases covered by tests (typed-nil accepted limit at consumer, typed-nil trap at producer boundary, non-`WiringError` panics still redact)
- [x] Error paths tested

### Documentation
- [x] README.md updated — not applicable, no user-facing changes
- [x] docs/ updated — **docs sweep performed, no updates required.** The `withRecovery` redaction contract lives in code comments only (not in `secure-logging-rules.md`) — updated inline in commit 5, and the sibling `recovery.HTTPMiddleware` comment was softened in commit 7 to reflect the sibling-layer difference. Three planning-doc references to older constructor signatures were found (SOL-150070 architecture-plan.md, SOL-150070 implementation-decisions.md, hardening/fail-fast-on-exchanger-error.md) — all pre-date SOL-151601 and are historical decision logs; edits would falsify the record. Left as-is.
- [x] godoc comments added/updated — `WiringError` type doc, both constructors' updated docstrings, `newAuthenticator`'s tightened docstring, `withRecovery`'s inline comment on the typed exception
- [ ] CHANGELOG.md — not maintained in this repo
- [x] Examples updated — n/a

### Git & CI
- [ ] All commits are signed off (DCO) — **intentionally omitted per team convention.** Happy to adjust if this repo actually enforces DCO on merge.
- [x] Branch is up to date with main (fast-forwarded before starting)
- [x] No merge conflicts

### Deliberate scope decisions

- **Commit 6 (dispatcher-guard removal) is slightly beyond the letter of the ticket.** The ticket scope lists: "add tests," "choose A or B, apply symmetrically," "update `withRecovery` if B chosen," "add tests for new payload." It does not explicitly say "collapse the dispatcher-level check." I bundled it because leaving the dispatcher guard in place makes Basic and OAuth handle the same class of wiring bug asymmetrically (Basic panics at construction; OAuth returned an error at dispatch) — which contradicts the ticket's stated symmetry goal.
- **`WiringError` has 2 fields, not 3.** The ticket's Design B example included a `Constructor` field. I omitted it — the constructor name is already in the stack trace, so it would duplicate information. Rationale in `wiring.go`'s type doc.
- **`withRecovery` updated, not a new broker-level `recover` in `NewBrokerClient`.** The ticket permits either ("update `withRecovery` **or** add broker-level `recover`"). I picked the simpler option; the tool-handler goroutine is where the panic fires today (broker clients are built lazily on first request per `pool.go:87-110`).

### Follow-up work identified

- **Domain types for `BrokerAlias` and `Password`/`Secret`.** During review of commit 3 I noticed 3 consecutive `string` params in `NewBasicAuthenticator` — arg-order-swap is compiler-silent. A codebase-wide `type BrokerAlias string` and a `type Password` with a `LogValue()` method would tighten this significantly. 50+ existing production references would need to migrate, so it's a real refactor — filed as a follow-up chip locally; will land as its own SOL ticket if the design pans out.

## Related Issues/PRs

- Jira: [SOL-151601](https://sol-jira.atlassian.net/browse/SOL-151601)
- Parent epic: [SOL-148417](https://sol-jira.atlassian.net/browse/SOL-148417)
- Follow-up from: [SOL-151506](https://sol-jira.atlassian.net/browse/SOL-151506) / PR #149

---

**For Reviewers:**

**Focus Areas:**
1. **Commit 5 (`withRecovery` type-switch)** — this is the security-adjacent commit. Please verify that the secure-logging rule remains intact for every non-`WiringError` panic (still logs only `panic_type` + `stack`, no `panic_message`), and that the two new fields on `WiringError`'s branch are genuinely safe (`BrokerAlias` sourced from config, `Reason` authored by trusted in-package code — both mirrored to other config-sourced structured log fields in the codebase).
2. **Commit 6 (dispatcher guard removal) + Commit 8 (typed-nil trap fix)** — read these together. Commit 6 collapsed the dispatcher-level `nil` guard as redundant; commit 8 restores a `nil` check at the dispatcher for a *different* purpose — emitting an untyped-nil interface value so the constructor's guard actually fires (Go's classic nil-interface header gotcha). The two commits together preserve "constructor owns the invariant" while making sure the dispatcher's interface-conversion is truthful.
3. **`WiringError` type location and shape** — kept local to `internal/semp/auth`; only 2 fields. If you'd prefer it live at a shared location (`internal/semp/wiring.go` or similar), the compiler will drive the rename in a follow-up.

### DIFF
--- a/internal/semp/auth/basic.go
+++ b/internal/semp/auth/basic.go
@@ -11,10 +11,9 @@ import (
 // so AddAuth and HandleAuthFailure are safe to call concurrently from
 // any number of goroutines.
 type BasicAuthenticator struct {
-	username    string
-	password    string
-	brokerAlias string
-	jar         CookieJarClearer
+	username string
+	password string
+	jar      CookieJarClearer
 }
 
 // NewBasicAuthenticator returns a BasicAuthenticator that will attach the
@@ -22,14 +21,15 @@ type BasicAuthenticator struct {
 // jar is required — HandleAuthFailure clears it to force fresh Basic
 // credentials on the retry, so an authenticator without a jar cannot
 // recover from a 401. Panics with a WiringError if jar is nil — a nil
-// jar is a wiring bug, not a runtime condition. The brokerAlias flows
-// into the panic payload so the tool-handler recovery layer can log
-// which broker tripped the invariant.
+// jar is a wiring bug, not a runtime condition. brokerAlias flows into
+// the panic payload so the tool-handler recovery layer can log which
+// broker tripped the invariant; it is not stored on the successful path
+// because no runtime code path reads it.
 func NewBasicAuthenticator(username, password, brokerAlias string, jar CookieJarClearer) *BasicAuthenticator {
 	if jar == nil {
 		panic(WiringError{BrokerAlias: brokerAlias, Reason: "NewBasicAuthenticator: jar must be non-nil"})
 	}
-	return &BasicAuthenticator{username: username, password: password, brokerAlias: brokerAlias, jar: jar}
+	return &BasicAuthenticator{username: username, password: password, jar: jar}
 }
 
 // AddAuth attaches an Authorization: Basic header to req. The ctx is
--- a/internal/semp/auth/basic.go
+++ b/internal/semp/auth/basic.go
@@ -11,9 +11,10 @@ import (
 // so AddAuth and HandleAuthFailure are safe to call concurrently from
 // any number of goroutines.
 type BasicAuthenticator struct {
-	username string
-	password string
-	jar      CookieJarClearer
+	username    string
+	password    string
+	brokerAlias string
+	jar         CookieJarClearer
 }
 
 // NewBasicAuthenticator returns a BasicAuthenticator that will attach the
@@ -21,12 +22,13 @@ type BasicAuthenticator struct {
 // jar is required — HandleAuthFailure clears it to force fresh Basic
 // credentials on the retry, so an authenticator without a jar cannot
 // recover from a 401. Panics if jar is nil — a nil jar is a wiring bug,
-// not a runtime condition.
-func NewBasicAuthenticator(username, password string, jar CookieJarClearer) *BasicAuthenticator {
+// not a runtime condition. The brokerAlias is retained for diagnostic
+// use by the wiring-panic payload (see WiringError).
+func NewBasicAuthenticator(username, password, brokerAlias string, jar CookieJarClearer) *BasicAuthenticator {
 	if jar == nil {
 		panic("NewBasicAuthenticator: jar must be non-nil")
 	}
-	return &BasicAuthenticator{username: username, password: password, jar: jar}
+	return &BasicAuthenticator{username: username, password: password, brokerAlias: brokerAlias, jar: jar}
 }
 
 // AddAuth attaches an Authorization: Basic header to req. The ctx is
--- a/internal/semp/auth/basic.go
+++ b/internal/semp/auth/basic.go
@@ -21,12 +21,13 @@ type BasicAuthenticator struct {
 // given credentials to every request via http.Request.SetBasicAuth. The
 // jar is required — HandleAuthFailure clears it to force fresh Basic
 // credentials on the retry, so an authenticator without a jar cannot
-// recover from a 401. Panics if jar is nil — a nil jar is a wiring bug,
-// not a runtime condition. The brokerAlias is retained for diagnostic
-// use by the wiring-panic payload (see WiringError).
+// recover from a 401. Panics with a WiringError if jar is nil — a nil
+// jar is a wiring bug, not a runtime condition. The brokerAlias flows
+// into the panic payload so the tool-handler recovery layer can log
+// which broker tripped the invariant.
 func NewBasicAuthenticator(username, password, brokerAlias string, jar CookieJarClearer) *BasicAuthenticator {
 	if jar == nil {
-		panic("NewBasicAuthenticator: jar must be non-nil")
+		panic(WiringError{BrokerAlias: brokerAlias, Reason: "NewBasicAuthenticator: jar must be non-nil"})
 	}
 	return &BasicAuthenticator{username: username, password: password, brokerAlias: brokerAlias, jar: jar}
 }
--- a/internal/semp/auth/wiring.go
+++ b/internal/semp/auth/wiring.go
@@ -1,0 +1,28 @@
+package auth
+
+// WiringError is the panic value authenticator constructors use when a
+// required dependency is nil. It carries the broker alias so the
+// tool-handler recovery layer (internal/tools.withRecovery) can log it
+// as a pre-vetted structured field.
+//
+// Raw panic strings cannot be logged under the secure-logging rules
+// (docs/internal/secure-logging-rules.md): panic values are unaudited
+// and may carry arbitrary data. A typed panic value with explicitly
+// declared fields is the safe way to route a diagnostic string through
+// recovery — each field's contract makes it safe to log.
+//
+// Kept local to internal/semp/auth: the only callers are the
+// authenticator constructors in this package. If a third package needs
+// the same shape, lift this to a shared location — the compiler will
+// drive the rename.
+type WiringError struct {
+	// BrokerAlias identifies which broker's wiring tripped the panic.
+	// Sourced from the broker config alias, safe to log.
+	BrokerAlias string
+
+	// Reason is the constructor-authored message describing what
+	// invariant was violated (e.g. "NewOAuthAuthenticator: exchanger
+	// must be non-nil"). Authored by trusted code in this package;
+	// safe to log.
+	Reason string
+}
--- a/internal/semp/broker.go
+++ b/internal/semp/broker.go
@@ -117,6 +117,15 @@ func newCookieJar(alias string, mode string) (*resilience.SafeCookieJar, error) 
 // exchanger) live inside each constructor and fail fast via
 // auth.WiringError panics. The dispatcher does not re-validate them;
 // its only error path is an unknown auth mode.
+//
+// The nil-exchanger case for OAuth requires one bit of conversion
+// hygiene: passing a concrete nil *tokenexchange.Exchanger into
+// NewOAuthAuthenticator's tokenExchanger interface parameter would
+// produce a typed-nil interface header ({type: *Exchanger, value: nil})
+// that silently bypasses the constructor's `exchanger == nil` guard
+// (Go's classic nil-interface gotcha — see go.dev/doc/faq#nil_error).
+// Check the concrete pointer here and pass untyped nil so the guard
+// fires and the WiringError panic carries the broker alias.
 func newAuthenticator(alias string, brokerCfg *config.BrokerConfig, jar *resilience.SafeCookieJar, exchanger *tokenexchange.Exchanger) (auth.Authenticator, error) {
 	cfg := brokerCfg.Auth
 	switch cfg.Mode {
@@ -125,6 +134,9 @@ func newAuthenticator(alias string, brokerCfg *config.BrokerConfig, jar *resilie
 	case config.AuthModeBearer:
 		return auth.NewBearerAuthenticator(cfg.Token), nil
 	case config.AuthModeOAuth:
+		if exchanger == nil {
+			return auth.NewOAuthAuthenticator(nil, cfg.Audience, cfg.Scopes, alias), nil
+		}
 		return auth.NewOAuthAuthenticator(exchanger, cfg.Audience, cfg.Scopes, alias), nil
 	default:
 		return nil, fmt.Errorf("unsupported auth mode %q for broker %q", cfg.Mode, alias)
--- a/internal/semp/broker.go
+++ b/internal/semp/broker.go
@@ -46,8 +46,9 @@ type BrokerClient struct {
 // owns its own precondition checks.
 //
 // exchanger is the process-wide token exchanger for OAuth brokers. Pass
-// nil when no broker uses OAuth; newAuthenticator returns an error if an
-// OAuth broker is configured without an exchanger.
+// nil when no broker uses OAuth; NewOAuthAuthenticator panics with an
+// auth.WiringError if an OAuth broker is configured without an exchanger,
+// carrying the broker alias for operator diagnostics.
 func NewBrokerClient(alias string, brokerCfg *config.BrokerConfig, sempCfg *config.SEMPConfig, exchanger *tokenexchange.Exchanger) (*BrokerClient, error) {
 	jar, err := newCookieJar(alias, brokerCfg.Auth.Mode)
 	if err != nil {
--- a/internal/semp/broker.go
+++ b/internal/semp/broker.go
@@ -111,6 +111,11 @@ func newCookieJar(alias string, mode string) (*resilience.SafeCookieJar, error) 
 // auth config. Each mode has its own constructor with mode-specific deps;
 // this switch is the single dispatch point so NewBrokerClient stays
 // focused on wiring clients.
+//
+// Per-mode wiring invariants (e.g. non-nil cookie jar, non-nil token
+// exchanger) live inside each constructor and fail fast via
+// auth.WiringError panics. The dispatcher does not re-validate them;
+// its only error path is an unknown auth mode.
 func newAuthenticator(alias string, brokerCfg *config.BrokerConfig, jar *resilience.SafeCookieJar, exchanger *tokenexchange.Exchanger) (auth.Authenticator, error) {
 	cfg := brokerCfg.Auth
 	switch cfg.Mode {
@@ -119,9 +124,6 @@ func newAuthenticator(alias string, brokerCfg *config.BrokerConfig, jar *resilie
 	case config.AuthModeBearer:
 		return auth.NewBearerAuthenticator(cfg.Token), nil
 	case config.AuthModeOAuth:
-		if exchanger == nil {
-			return nil, fmt.Errorf("oauth auth requires a token exchanger for broker %q", alias)
-		}
 		return auth.NewOAuthAuthenticator(exchanger, cfg.Audience, cfg.Scopes, alias), nil
 	default:
 		return nil, fmt.Errorf("unsupported auth mode %q for broker %q", cfg.Mode, alias)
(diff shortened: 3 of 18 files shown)


internal/semp/broker.go
```diff
@@ -111,17 +111,19 @@ func newCookieJar(alias string, mode string) (*resilience.SafeCookieJar, error)
...
 	case config.AuthModeBasic:
-		return auth.NewBasicAuthenticator(cfg.Username, cfg.Password, jar), nil
+		return auth.NewBasicAuthenticator(cfg.Username, cfg.Password, alias, jar), nil
 	case config.AuthModeBearer:
 		return auth.NewBearerAuthenticator(cfg.Token), nil
 	case config.AuthModeOAuth:
-		if exchanger == nil {
-			return nil, fmt.Errorf("oauth auth requires a token exchanger for broker %q", alias)
-		}
 		return auth.NewOAuthAuthenticator(exchanger, cfg.Audience, cfg.Scopes, alias), nil
```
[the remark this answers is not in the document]
amitmorade: Correct, and thanks for the catch — this is a real regression I introduced when collapsing the dispatcher's `if exchanger == nil` guard in the earlier commit. Before that collapse, the dispatcher's own untyped-nil check caught it before ever reaching the constructor; when I moved sole ownership to the constructor, the interface-conversion typed-nil trap became reachable.

Same interface-header gotcha we discussed on PR #149 for `NewBasicAuthenticator`, but the shape here is different: for Basic, the producer (`newCookieJar`) returns a concrete type and is safe by construction. For OAuth, the producer *is* `newAuthenticator` — it takes a concrete `*tokenexchange.Exchanger` and passes it into an interface parameter. That pointer-to-interface conversion is where the typed-nil emerges.

Applying the same "defend at the producer" principle from PR #149, the fix belongs at the producer boundary — check the concrete pointer, pass untyped nil so the constructor's guard fires. Not adopting the reflect-based consumer defense for the reasons you and I settled on last time.

Fixed in [`a0cf100`](https://github.com/SolaceDev/solace-broker-mcp/commit/a0cf100). Also added `TestNewBrokerClient_OAuth_NilExchanger_PanicsWithWiringError` — pins the dispatcher-boundary contract end-to-end so a future refactor that "cleans up" the pre-conversion nil check as redundant will fail loudly instead of silently re-introducing the trap.

internal/semp/auth/basic.go
```diff
@@ -11,22 +11,25 @@ import (
 // so AddAuth and HandleAuthFailure are safe to call concurrently from
 // any number of goroutines.
 type BasicAuthenticator struct {
-	username string
-	password string
-	jar      CookieJarClearer
+	username    string
+	password    string
+	brokerAlias string
+	jar         CookieJarClearer
```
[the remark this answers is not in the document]
amitmorade: Also removed the unused `brokerAlias` field from `BasicAuthenticator`'s struct in [`75e6c10`](https://github.com/SolaceDev/solace-broker-mcp/commit/75e6c10) — no runtime code path reads it, so it was dead state on the successful-construction path. The panic still carries the alias via the constructor parameter, so the ticket's diagnostic contract is unchanged.

`OAuthAuthenticator` keeps its `brokerAlias` field because it actually uses it at runtime — in `AddAuth`'s `ExchangeInput` and `HandleAuthFailure`'s eviction key. The asymmetry between the two structs reflects real runtime-usage asymmetry, not an oversight.

internal/semp/broker.go
```diff
@@ -111,16 +112,30 @@ func newCookieJar(alias string, mode string) (*resilience.SafeCookieJar, error)
...
 	switch cfg.Mode {
 	case config.AuthModeBasic:
-		return auth.NewBasicAuthenticator(cfg.Username, cfg.Password, jar), nil
+		return auth.NewBasicAuthenticator(cfg.Username, cfg.Password, alias, jar), nil
 	case config.AuthModeBearer:
 		return auth.NewBearerAuthenticator(cfg.Token), nil
 	case config.AuthModeOAuth:
 		if exchanger == nil {
-			return nil, fmt.Errorf("oauth auth requires a token exchanger for broker %q", alias)
+			return auth.NewOAuthAuthenticator(nil, cfg.Audience, cfg.Scopes, alias), nil
```
solace-aross, line 138: This branch changes the OAuth nil-exchanger case from a classified, caller-visible error (`error_type="broker_init_error"`, specific message) to a generic panic that `withRecovery` turns into "contact your administrator." That's a real classification regression, not just symmetry with Basic's existing panic: `Hop2OAuthActive()` gates the real exchanger behind `ENABLE_UNRELEASED_BROKER_OAUTH` independently of whether the broker's `auth.mode: oauth` + `broker_oauth:` config is otherwise valid, so this is reachable in a normal staged-rollout state, not just a wiring bug. Once hit, every tool call against that broker panics and re-runs the full constructor (nothing caches a failed construction), capturing a full stack trace each time with no counter anywhere to distinguish "failing on every call" from "failed once." Any alert/dashboard keyed on `broker_init_error` also goes silent for this case. Worth keeping this branch as a classified error, or making `WiringError` retryable-classified in `classifyBrokerError`, rather than routing a plausible production state through the generic panic path. Separately: `Reason` here just restates the nil-check ("exchanger must be non-nil"); naming the actual cause (`ENABLE_UNRELEASED_BROKER_OAUTH` / missing `broker_oauth:` config) would make it something an on-call engineer can act on.

internal/semp/broker.go
```diff
@@ -111,16 +112,30 @@ func newCookieJar(alias string, mode string) (*resilience.SafeCookieJar, error)
...
+// that silently bypasses the constructor's `exchanger == nil` guard
+// (Go's classic nil-interface gotcha — see go.dev/doc/faq#nil_error).
+// Check the concrete pointer here and pass untyped nil so the guard
+// fires and the WiringError panic carries the broker alias.
 func newAuthenticator(alias string, brokerCfg *config.BrokerConfig, jar *resilience.SafeCookieJar, exchanger *tokenexchange.Exchanger) (auth.Authenticator, error) {
 	cfg := brokerCfg.Auth
 	switch cfg.Mode {
 	case config.AuthModeBasic:
-		return auth.NewBasicAuthenticator(cfg.Username, cfg.Password, jar), nil
+		return auth.NewBasicAuthenticator(cfg.Username, cfg.Password, alias, jar), nil
```
solace-aross, line 133: Same typed-nil trap shape as the OAuth case this PR just fixed: `jar` (a concrete `*resilience.SafeCookieJar`) is boxed directly into the `CookieJarClearer` interface parameter here, so a nil `*SafeCookieJar` would silently bypass `NewBasicAuthenticator`'s `jar == nil` guard, the same way the OAuth bug did before Copilot caught it on this PR. Not exploitable today since `newCookieJar` guarantees non-nil whenever Basic mode reaches here, but that safety is implicit and untested, unlike OAuth, which now has `TestNewBrokerClient_OAuth_NilExchanger_PanicsWithWiringError` as a tripwire. Worth a comment documenting the invariant this relies on, or a matching dispatcher-level test, since this is exactly the class of regression that slipped in unnoticed once already in this PR.

internal/semp/auth/wiring.go
```diff
@@ -0,0 +1,28 @@
...
+// (docs/internal/secure-logging-rules.md): panic values are unaudited
+// and may carry arbitrary data. A typed panic value with explicitly
+// declared fields is the safe way to route a diagnostic string through
+// recovery — each field's contract makes it safe to log.
+//
+// Kept local to internal/semp/auth: the only callers are the
+// authenticator constructors in this package. If a third package needs
+// the same shape, lift this to a shared location — the compiler will
+// drive the rename.
+type WiringError struct {
```
solace-aross, line 18: `WiringError`'s "safe to log" guarantee (both fields bypass panic-value redaction in `withRecovery`) is enforced only by this doc comment, not by the type system: both fields are exported with no constructor, so any future call site can construct one with interpolated or untrusted content and it would still get logged unredacted, purely by matching this concrete type. Not exploitable today since both current call sites use hardcoded literals, but worth closing off now while there's only one caller: unexport the fields and add a small constructor scoped to the two authenticators that need it, so a future contributor has to satisfy the compiler rather than just read the comment.

solace-aross (changes requested):
Good fix for a real diagnosability gap, and the typed-nil regression that crept in mid-PR was caught and fixed cleanly with a solid regression test. Three things worth addressing before merge, left inline.

One more, not tied to a line in this diff: `Hop2OAuthActive()` (`internal/config/config.go`) gates real exchanger construction behind `ENABLE_UNRELEASED_BROKER_OAUTH` independently of whether a broker's oauth config is otherwise valid, so the nil-exchanger panic this PR makes diagnosable is reachable in a normal staged-rollout state, not just a wiring bug. Worth an eager check in `cmd/server/main.go` at boot (WARN or error) when that flag is off but a broker declares `auth.mode: oauth`, so this class of misconfiguration is caught once at startup instead of discovered per call in production.

amitmorade:
Closing this PR — the premise doesn't hold up on a second look.

## Why

The whole PR exists to make an unreachable panic diagnosable. Walking the reachable states:

- `mode: basic` — `newCookieJar` builds the jar and returns an error if it can't. `NewBrokerClient` propagates that error before `newAuthenticator` is ever called. `NewBasicAuthenticator` never sees a nil jar.
- `mode: oauth` — config validation (`validateBrokerConfig` + `validateBrokerOAuthConfig`) rejects `auth.mode: oauth` at startup unless (a) the unreleased-feature flag is set AND (b) `broker_oauth:` is populated. When both hold, `Hop2OAuthActive()` is true and `main.go` builds the exchanger. `NewOAuthAuthenticator` never sees a nil exchanger.

Both constructor panics on `main` today guard states the surrounding wiring already prevents. Making them diagnosable (this PR) doesn't fix a real bug — it just adds infrastructure to log a panic that can't fire.

## Where I think we drifted

@solace-aross — your review comment on #149 was right about the shape: "the constructor should own its invariants, not the dispatcher." That's a good principle. Where I think we went one step further than the code warranted: these constructors aren't a library. They have exactly one caller each, in the same package's dispatcher. And the dispatcher isn't the right invariant owner either — the *config validator* is. It runs before any authenticator exists, rejects the invalid state at startup, and owns the "oauth broker ⇒ exchanger built" (and "basic mode ⇒ jar built successfully") contracts structurally.

Applying library-grade defensive coding to code that isn't a library gave us panic guards defending an unreachable path — and #161 was going to add ~300 lines of `WiringError` + alias threading + recovery type-switch to make those guards diagnosable. Not wrong in principle — just at the wrong altitude.

## What I'm doing instead

New small PR (SOL-151601 refocused). Delete, don't add:

1. Remove the `if exchanger == nil` error return in `newAuthenticator` (dead branch given validator).
2. Remove the `panic` on nil jar in `NewBasicAuthenticator` and the `panic` on nil exchanger in `NewOAuthAuthenticator`.
3. Remove the panic-contract tests defending those guards (from #149 for basic).
4. Add a validator test pinning the two real invariants at their owner:
   - `auth.mode: oauth` + flag off → rejected at startup
   - `auth.mode: oauth` + missing `broker_oauth:` → rejected at startup

Net LOC negative. The invariant is checked once, at startup, in one file.

## Jira

Updating SOL-151601 — the ticket premise ("make the constructor panic diagnosable") is invalidated by the higher-altitude fix.

Happy to keep this open if I'm missing a reachable path I haven't accounted for.

PR #162, opened 2026-07-10
SOL-150692: Enforce TLS posture for the OAuth-mode listener

solace-awaheed:
## Summary

Enforces TLS on the MCP server's own listener in OAuth (production) mode: the server now refuses to start with a plaintext listener unless the operator provides TLS certs or explicitly acknowledges upstream TLS termination via a new tls_terminated_upstream flag.

Fixes # SOL-150692

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

In OAuth mode the server's HTTP listener could silently run plaintext. Config validation only checked that tls_cert_file/tls_key_file were paired, never that TLS was actually on, and cmd/server/main.go falls back to ListenAndServe() when certs are absent. resource_url must be https, but nothing tied that to the listener. Result: a deployment with no TLS-terminating proxy transmitted client bearer tokens and all tool results in cleartext while the config validated as production — with no warning emitted.

TLS termination at an upstream proxy/ingress is a legitimate pattern, so the fix is enforcement with an explicit opt-out, not a hard TLS requirement. This completes the server-listener side of the production-mode transport posture begun in SOL-149665 (which closed the same gap for outbound broker/auth URLs).

## Changes Made

  - Config field: added tls_terminated_upstream (bool, default false) to ServerConfig and the raw YAML struct.
  - Validation: under mcp_client_auth.mode: oauth, no TLS certs and no acknowledgment is now a fatal config error naming both remediation paths.
  - Startup WARN: added banner.LogOAuthPlaintextListener (mirrors the existing LogStaticCleartextExposure pattern) + the OAuthPlaintextListenerAcknowledged() predicate; wired into main.go. Fires when oauth serves plaintext under the acknowledgment, naming the field and logging the bind address.
  - Docs: docs/authentication.md (two deployment patterns — direct TLS vs. terminating proxy/ingress), docs/configuration.md (field table + prose).
  - Examples/manifests: broker-config.example.yaml and deploy/kubernetes/configmap.yaml set tls_terminated_upstream: true (both already model plaintext-behind-upstream-TLS); test/oauth/test-config.yaml and an integration fixture updated to keep passing.
  - CHANGELOG.md: BREAKING entry with migration + upgrade/rollback ordering.

## Testing

**Test Environment:**
- Go version:
- OS:
- Broker version (if applicable):

**Tests Performed:**

**Test Coverage:**
  - OAuth + no certs + no ack → fatal error naming tls_cert_file, tls_key_file, and tls_terminated_upstream
  - OAuth + tls_terminated_upstream: true + no certs → valid; predicate true (WARN path)
  - OAuth + certs → valid; predicate false (TLS listener, unchanged)
  - OAuth + certs + ack both set → valid; certs win, predicate false (no spurious WARN)
  - Non-OAuth (static) + no certs, with the flag set → valid; flag ignored, predicate false
  - Banner emits WARN, names tls_terminated_upstream, logs bind_address
  - Full suite: 1397 passed; golangci-lint clean; /check-logs clean

## Breaking Changes

**Impact:** Existing mcp_client_auth.mode: oauth deployments that run without server-side TLS certs (relying on an upstream proxy/ingress) will fail to start after upgrading until they add the new acknowledgment.

**Migration Guide:**
  - If TLS is terminated upstream: add tls_terminated_upstream: true.
  - To terminate TLS at the server instead: set tls_cert_file and tls_key_file.
  - Ordering: the config parser rejects unknown fields, so add tls_terminated_upstream only after the new binary is deployed; to roll back to a binary that predates the field, remove it from config first.
  - Dev modes (static/disabled) are unaffected — the flag is ignored there.

### Code Quality
- [ ] Code follows the [coding standards](../.github/CONTRIBUTING.md#coding-standards)
- [ ] Self-review completed (checked for typos, logic errors, edge cases)
- [ ] Code is well-commented (explains "why", not "what")
- [ ] Complex logic has explanatory comments
- [ ] No commented-out code or debug statements left in

### Security
- [ ] **No credentials in code** (checked all files, commits, and logs)
- [ ] Followed [secure logging rules](../docs/secure-logging-rules.md)
- [ ] Credential-carrying types implement `slog.LogValuer`
- [ ] No security vulnerabilities introduced (checked with golangci-lint/gosec)

### Testing
- [ ] All tests pass locally: `go test -race ./...`
- [ ] No race conditions detected
- [ ] Edge cases covered by tests
- [ ] Error paths tested
- [ ] Test names clearly describe what is being tested

### Documentation
- [ ] godoc comments added/updated for exported functions

### Git & CI
- [ ] All commits are signed off (DCO: `git commit -s`)
- [ ] Commits have clear, descriptive messages
- [ ] Branch is up to date with main (rebased if needed)
- [ ] No merge conflicts
- [ ] CI checks passing (lint, build, test, E2E)

## Related Issues/PRs

- Related to #
- Depends on #
- Blocks #

---

**For Reviewers:**

**Focus Areas**: 

**Questions**:


internal/config/config.go
```diff
@@ -995,6 +1005,21 @@ func validate(cfg *ServerConfig) error {
...
 		}
+		// Listener transport: OAuth mode is production, so a plaintext listener
+		// must be an explicit choice. With no server-side TLS and no upstream-
+		// termination acknowledgment the listener would carry client bearer
+		// tokens and tool results in cleartext while validating as production.
+		// resource_url being https:// does not imply the listener is encrypted.
+		// TLSCertFile == "" is the "TLS is off" signal (same convention as
+		// StaticTokenExposedCleartext); a half-configured cert/key pair is caught
+		// independently by the pairing check below.
+		if cfg.TLSCertFile == "" && !cfg.TLSTerminatedUpstream {
```
solace-aross, line 1026: This file trims every other operator-supplied required string before comparing to `""` (see the `dev_token` check, and `applyDefaults`'s handling of `ListenAddress`), specifically because a whitespace-only value from an unresolved `${VAR}` substitution or a secret mounted with a trailing newline resolves to a non-empty string. `TLSCertFile` isn't trimmed here or in `OAuthPlaintextListenerAcknowledged()` below, so a whitespace-only cert path satisfies this check with no acknowledgment required and no config error, then fails later inside `ListenAndServeTLS` instead of clearly at `LoadConfig`. That reopens the exact silent-plaintext-in-production scenario this PR closes. Worth trimming in both places for consistency with the rest of the file.

internal/config/config.go
```diff
@@ -56,6 +56,14 @@ type ServerConfig struct {
 	TLSCertFile   string              // path to TLS certificate file (optional, enables HTTPS)
 	TLSKeyFile    string              // path to TLS private key file (optional, requires TLSCertFile)
 
+	// TLSTerminatedUpstream acknowledges that TLS is terminated by an upstream
+	// proxy or ingress, so the server may serve a plaintext listener under
+	// mcp_client_auth.mode: oauth. Without it, OAuth mode with no cert/key is a
+	// fatal config error (the listener would transmit client bearer tokens and
+	// tool results in cleartext while validating as production). Honored only in
+	// OAuth mode; ignored in the dev modes. See OAuthPlaintextListenerAcknowledged.
+	TLSTerminatedUpstream bool
```
solace-aross, line 65: `tls_terminated_upstream` only unlocks `http.ListenAndServe()`; it doesn't touch `ListenAddress`, which still defaults to all interfaces under oauth mode (`applyDefaults`). Both example configs that set this flag (`broker-config.example.yaml`, `deploy/kubernetes/configmap.yaml`) leave `listen_address` unset, so as shipped they put the bearer-token-carrying plaintext listener on `0.0.0.0`. The flag's entire safety premise is that only the upstream proxy can reach that port, but nothing in the code defaults to or enforces a restricted bind. Worth defaulting to loopback when the flag is set and `listen_address` is unset (mirroring the existing disabled-mode default), or adding a second WARN specifically for the wildcard-bind + acknowledgment combination.

solace-aross (changes requested):
Solid fix for a real gap: oauth mode could previously start fully plaintext with no validation error at all. Two things worth addressing before merge, both concrete and small. Not restating the CHANGELOG migration-order contradiction since Copilot already flagged it.

solace-aross (dismissed):
Approving. Both round-1 items are resolved and verified in the current head, and the fix holds up under a fresh read.

**Resolved:**
- Whitespace-only cert path: `applyDefaults` now trims `tls_cert_file`/`tls_key_file`, so an unresolved `${VAR}` or a secret with a trailing newline reads as "no TLS" and fails cleanly at `LoadConfig`. Covered by `TestLoadConfig_OAuthWhitespaceCertPath_TreatedAsNoTLS`.
- Bind scope: `ListensOnAllInterfaces()` plus the escalated WARN, and both example configs now document that the flag permits plaintext but does not restrict the bind.

Also spot-checked the surrounding logic: the cert-without-key case can't slip through to plaintext (the XOR pairing check at `config.go:1023` errors independently), and `ListensOnAllInterfaces()` is safe against a `host:port` value because `listen_address` is validated to a bare IP or `localhost`. Test matrix covers the full ack/cert/precedence grid.

One non-blocking note, your call, not a merge blocker:

The escalated wildcard-bind WARN fires on the recommended Kubernetes deployment. A pod must bind `0.0.0.0` for the Service to reach it, so `deploy/kubernetes/configmap.yaml` trips the loudest banner on every startup even though the traffic is ClusterIP-scoped and safe. The process can't distinguish a locked-down pod netns from a LAN-exposed bare-metal host, so the escalation reads the same in both. The risk is alert fatigue: the scary variant becomes background noise exactly where it should carry signal. Worth considering whether to keep the base WARN for wildcard-in-oauth and reserve escalation for a condition the config can actually tell apart. Fine to defer or close as documented.

solace-aross (approved):
Approving. The one change since my last pass is the wildcard-bind escalation drop I'd flagged, and it's the right call.

`LogOAuthPlaintextListener` no longer escalates on a wildcard bind. The base plaintext WARN still fires on every acknowledged plaintext start and logs `bind_address`, and the bind-scope risk is now documented in banner.go, docs/authentication.md, and the example configs instead. That removes the alert-fatigue trap — the loud variant no longer fires hardest on the recommended Kubernetes deployment — without weakening the gate: plaintext is still opt-in behind `tls_terminated_upstream`, and the whitespace-cert-path trim that closes the silent-plaintext hole is untouched by the merge.

Docs now match the code: authentication.md states `https://` is enforced on both `issuer` and `resource_url`, that direct server TLS takes precedence when both cert and flag are set, and the CHANGELOG migration order works against the strict-field parser (deploy binary + config together, remove the field before rolling back). All the round-1 doc notes are closed. CI green including e2e-oauth.

PR #163, opened 2026-07-10
SOL-151741: Stop implying a health verdict in get-broker-status example queries

solace-aross:
## Summary

- `get-broker-status` already reports raw state, not a health verdict — its tool description says so explicitly. The example queries in the docs hadn't caught up, and still modeled "Is prod-broker healthy?" as the trigger phrase.
- Replaces that with status-oriented phrasing in `docs/tools-reference.md` and `docs/examples.md`, consistent with the disclaimer already in `internal/tools/sempv1/brokerstatus/handler.go`.

## Why

Support (Alexandra Masse) flagged this as a customer-trust risk: our own examples inviting the LLM to answer "is it healthy?" reintroduces the exact liability we removed when `get-broker-health` was renamed to `get-broker-status`. Full context in [SOL-151741](https://sol-jira.atlassian.net/browse/SOL-151741).

A related but separate issue — `get-vpn-health` has the same naming problem and needs its own decision on whether to rename — is tracked in [SOL-151742](https://sol-jira.atlassian.net/browse/SOL-151742), intentionally not bundled into this PR.

## Test plan

- [x] `grep -rn "healthy" docs/tools-reference.md docs/examples.md` returns no matches
- [x] `make check` passes
- [x] Docs-only change; no behavioral/code changes

[SOL-151741]: https://sol-jira.atlassian.net/browse/SOL-151741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-151742]: https://sol-jira.atlassian.net/browse/SOL-151742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/tools-reference.md
+++ b/docs/tools-reference.md
@@ -219,7 +219,7 @@ useful "has replication been flaky recently?" signal.
 
 ### list-vpns
 
-List Message VPNs with enabled state, connection count, and basic health.
+List Message VPNs with enabled state, connection count, and basic status.
 
 **Parameters:**
 
--- a/docs/tools-reference.md
+++ b/docs/tools-reference.md
@@ -160,7 +160,7 @@ appliances add a `hardwareDetails` section. Field shape is documented in
 { "broker": "prod-broker" }
 ```
 
-**Example request:** "Is prod-broker healthy? When did it last restart?"
+**Example request:** "What's prod-broker's current status? When did it last restart?"
 
 ### get-redundancy-status
 
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -101,7 +101,7 @@ parameters. Representative queries and the shape they return:
 | You ask | Tool invoked | Returns (shape) |
 |---|---|---|
 | "What brokers are configured?" | `list-brokers` | `{ "brokers": ["prod-broker", "dev-broker"] }` |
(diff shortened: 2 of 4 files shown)


solace-wmaryam (dismissed):
LGTM.

docs/tools-reference.md
amitmorade: is "basic health" here intentional? Same category as the example queries this PR is fixing — it's a verdict word in a user-facing tool summary, and `list-vpns` doesn't return a health field (line 231 is raw state).

solace-aross: Good catch, and actually worse than the pattern this PR is fixing: the listVpns postprocess handler doesn't compute anything called health, just raw counts (disabled/down/standby/zero-connection). Swapped "health" for "status" in 19491a0 — tools.yaml, tools-reference.md, and user-guide.md.

get-vpn-health has the same issue but bigger blast radius (tool rename, not just wording), so leaving that to SOL-151742 — already has an open PR (#165).

amitmorade (approved):
LGTM

PR #164, opened 2026-07-10
SOL-151601: Remove unreachable nil-guards in authenticator constructors

amitmorade:
## Summary

Removes two dead nil-guards that duplicated a stronger invariant already owned by config validation. Net −35 LOC across 4 files. Behavior unchanged.

Jira: [SOL-151601](https://sol-jira.atlassian.net/browse/SOL-151601)

## Type of Change

## Motivation and Context

Follow-up to [#161](https://github.com/SolaceDev/solace-broker-mcp/pull/161) (closed). #161 tried to make constructor panics *diagnosable* by adding a `WiringError` type + alias threading + a recovery-layer type-switch. On a second look, the panics guard states the surrounding wiring already prevents — the correct fix is to delete the guards, not to make them nicer.

**The reachable states, walked:**

- `mode: basic` — `newCookieJar` builds the jar and returns an error if it can't. `NewBrokerClient` propagates that error at `broker.go:54` *before* `newAuthenticator` is ever called. `NewBasicAuthenticator` never sees a nil jar.
- `mode: oauth` — config validation (`validateBroker` + `validateBrokerOAuthConfig` in `internal/config/config.go`) rejects `auth.mode: oauth` at startup unless (a) the unreleased-feature flag is set AND (b) the `broker_oauth:` block is populated. When both hold, `Hop2OAuthActive()` is true and `cmd/server/main.go` builds the exchanger. `NewOAuthAuthenticator` never sees a nil exchanger.

The `if exchanger == nil` return in `newAuthenticator` and both constructor panics are therefore unreachable. The invariant lives — and belongs — at the config validator, which runs at startup with full context, produces an actionable error, and prevents the whole invalid state from existing. Runtime guards downstream of that check are noise.

## Changes Made

Four commits:

1. **`6878312` Remove unreachable nil-exchanger checks in OAuth wiring**
   - `internal/semp/broker.go` — collapse the OAuth branch of `newAuthenticator` to a single-line return (no more `if exchanger == nil` error return).
   - `internal/semp/auth/oauth.go` — remove the `panic("NewOAuthAuthenticator: exchanger must be non-nil")`. Docstring updated to point at the invariant's actual owner (config validation).

2. **`6914387` Remove unreachable nil-jar check in NewBasicAuthenticator**
   - `internal/semp/auth/basic.go` — remove the `panic("NewBasicAuthenticator: jar must be non-nil")`. Docstring updated to point at the caller chain (`newCookieJar` + `NewBrokerClient`'s error propagation).
   - `internal/semp/auth/basic_test.go` — remove `TestNewBasicAuthenticator_PanicsOnNilJar` and `TestNewBasicAuthenticator_TypedNilJar_DoesNotPanic`. They pinned a runtime check that no longer exists.

3. **`570a3a7` Fix stale and phantom identifiers in updated docstrings** (from independent-review feedback)
   - `internal/semp/broker.go` — `NewBrokerClient`'s docstring still described `newAuthenticator` as returning an error for a nil exchanger; that path was deleted in commit 1. Rewritten to name the actual invariant owners.
   - `internal/semp/auth/oauth.go` — docstring referenced `validateBrokerConfig`, which doesn't exist; the actual function is `validateBroker`. Fixed. Also added a "failure mode if the invariant is violated" line so on-call has a lead if the upstream chain ever breaks.

4. **`96bfd49` Rewrite stale newAuthenticator docstring in NewBrokerClient** (from @solace-awaheed's review)
   - `internal/semp/broker.go` — the "newAuthenticator is a pure dispatcher; each Authenticator constructor owns its own precondition checks" line was true before this PR and false after (none of the three constructors owns a precondition check anymore). It also asserted the exact mental model this PR argues against. Rewritten to describe what `newAuthenticator` actually does now and to name the real owners of each precondition invariant. Missed in commit 3.

## Testing

**Test Environment:**
- Go version: 1.24
- OS: Darwin 24.6.0

**Tests Performed:**
- [x] Unit tests updated (removed two panic-contract tests that defended deleted code)
- [x] Integration tests — no change needed; race-enabled run of `./test/integration/...` green
- [ ] E2E tests — not applicable (no wire-protocol changes)
- [x] Tested locally with `go test -race ./internal/... ./cmd/... ./test/integration/...` — green

**Test Coverage:**

The invariants these guards defended are already exhaustively tested at their real owner (config validator):

- `TestLoadConfig_BrokerOAuth` (`config_test.go:2527`) — schema-valid oauth-mode broker rejected at load with `"auth.mode \"oauth\" is recognized but not yet supported"`.
- Same test, case at line 2579 — `auth.mode: oauth` + missing `broker_oauth:` block → rejected with `"broker_oauth block is required when any broker uses auth.mode"`.
- `TestServerConfig_Hop2OAuthActive` (`config_test.go:3554`) — all three preconditions pinned as load-bearing.

Every basic-auth test exercises the wiring path (`NewBrokerClient` → `newCookieJar` → `newAuthenticator` → `NewBasicAuthenticator` with a real jar). The "constructor never sees nil" invariant is defended by every one of those tests continuing to pass.

## Breaking Changes

**None.** All changes are internal to `internal/semp/auth` and `internal/semp`. No signature changes. No exported behavior changes. Configs that would previously have hit the deleted guards are still rejected at startup by config validation — the earlier, better rejection layer.

## Scope — what this PR does not touch

Deliberately narrow. The principle here is "remove guards whose invariant is owned upstream and has never fired." The codebase has other constructor-side nil-guards that look similar at first glance but are out of scope:

- `internal/semp/resilience/sender.go:67-73` panics on nil `authn` and nil `sem`. The `sem` guard is not defensive scaffolding — its docstring names SOL-150116 explicitly, and it prevents a specific regression (a Sender-private semaphore fallback would silently allow 2× the configured per-broker concurrency cap). That check has incident-history justification and stays. Deleting only the `authn` half while leaving the `sem` half creates asymmetry inside the same constructor with no clear rationale, so we leave both.
- `internal/semp/sempv1/client.go` and `internal/semp/sempv2/client.go` have similar `authn`/`sem` guards, forwarded to `resilience.New`. Same rationale applies.

The line, stated: **we remove guards whose invariant is owned upstream and has never fired. We keep guards tied to a specific past incident, because they're documentation as much as check.**

### Code Quality
- [x] Self-review completed
- [x] Code is well-commented (constructor docstrings updated to name the higher-altitude invariant owner)

### Security
- [x] Followed [secure logging rules](docs/internal/secure-logging-rules.md)
- [x] Credential-carrying types implement `slog.LogValuer` — n/a for this change

### Testing
- [x] Race-enabled tests pass everywhere: `go test -race ./internal/... ./cmd/... ./test/integration/...`
- [x] Edge cases covered by tests (config-validator tests defend the real invariant)

### Documentation
- [x] README.md — no user-facing changes
- [x] docs/ — no updates required
- [x] godoc comments updated — both authenticator constructors and `NewBrokerClient` now name the invariant's actual owner in their docstrings
- [ ] CHANGELOG.md — not maintained in this repo
- [x] Examples updated — n/a

### Git & CI
- [ ] All commits signed off (DCO) — omitted per team convention; happy to adjust if enforced
- [x] No merge conflicts

### Relationship to #161

#161 is closed in favor of this PR. Its stated goal ("make the constructor panic diagnosable") was addressing symptoms — the correct fix is to remove the unreachable panic itself. Details in the closing comment on #161.

### The principle in one line

> Enforce an invariant once, at the highest layer that has the information to enforce it. Don't duplicate checks downstream — they either drift or become dead code.

Config validator has the information ("does this YAML declare a broker that requires an exchanger, and are the exchanger's preconditions met?") and enforces it at startup. Downstream code trusts that check. When the feature flag is deleted at ship time, precondition (a) goes away and (b) remains — no other code changes.

## Related Issues/PRs

- Jira: [SOL-151601](https://sol-jira.atlassian.net/browse/SOL-151601) (scope reframed)
- Supersedes: [#161](https://github.com/SolaceDev/solace-broker-mcp/pull/161) (closed)
- Follow-up from: [#149](https://github.com/SolaceDev/solace-broker-mcp/pull/149) / SOL-151506

---

**For Reviewers:**

**Focus Areas:**
1. **Confirm the reachability analysis.** Both nil-check deletions rest on the claim that config validation + the caller chain prevent the guarded state. If you can construct a real reachable path that hits `NewBasicAuthenticator` with `jar == nil` or `NewOAuthAuthenticator` with `exchanger == nil`, that's the finding.
2. **Docstring accuracy.** The updated constructor docstrings and `NewBrokerClient`'s docstring name specific validator functions and caller-chain functions as the invariant owners. If any of those names drift (rename, move), the docs go stale — worth a check.
3. **Scope boundary.** The "Scope — what this PR does not touch" section names `resilience.New` and the sempv1/v2 client constructors as deliberately left alone. If you disagree with that line, that's a discussion worth having.

**Questions:**
- Is deleting the panic-contract tests in `basic_test.go` the right call, or would a reviewer prefer to keep them as documentation of the historical contract?


internal/semp/broker.go
solace-awaheed: This line is now stale after this PR. It reads "each Authenticator constructor owns its own precondition checks," but this change removes those checks from NewBasicAuthenticator and NewOAuthAuthenticator, and NewBearerAuthenticator never had one — so none of the three constructors owns a precondition check anymore. It also asserts the exact mental model the PR argues against (enforce upstream, not in the constructor), and it sits in the same docstring commit 570a3a7 was fixing for accuracy

solace-awaheed (dismissed):
LGTM, only thing flagged on my end was a comment update

internal/semp/broker.go
amitmorade: Good catch — you're right on both counts.

The line asserted the pre-PR mental model ("constructor owns the precondition") which contradicts what this PR argues for. It also lived in the same docstring the previous commit was supposed to be cleaning up for accuracy — I fixed two other stale references there and missed this one.

Pushed [`96bfd49`](https://github.com/SolaceDev/solace-broker-mcp/pull/164/commits/96bfd49) to rewrite the paragraph. It now states what `newAuthenticator` actually does (mode → constructor mapping with pass-through) and names the real owners of each precondition (`newCookieJar` + this function's error propagation for the jar; the config validators + `cmd/server/main.go` for the exchanger). Points readers at the per-constructor docstrings for the full contract.

Thanks for the close read.

solace-aross (approved):
Verified the reachability claim rather than taking it on faith: NewSafeCookieJar never returns (nil, nil), the config validator genuinely rejects an oauth broker unless Hop2OAuthActive()'s preconditions hold, main.go builds the exchanger (or exits) before the pool is constructed, and NewBrokerClient has exactly one production call site (pool.go:110). Both nil states really are unreachable today.

Also checked whether removing the returned-error check in newAuthenticator's oauth branch trades away real diagnostics for a future regression, since that one wasn't a panic before, it was a clean Go error. It doesn't: classifyBrokerError wraps it as broker_init_error, which isn't one of the audited error types in manager.go's logging gate, so only the Go type was ever logged, never the message text. A future nil-pointer panic would actually surface with a full stack trace, so there's no diagnosability loss here.

One non-blocking observation: the "Scope" section's line between keeping and removing guards (incident-backed vs. not) doesn't quite cover internal/semp/resilience/sender.go's authn nil-check the same way it covers sem, since authn has no SOL-150116-style silent-failure story of its own. Not something this PR needs to fix, just worth a comment there sometime so a future pass doesn't assume the two guards share one rationale.

Good tests, clean commit sequence, and the closed-#161 context makes the design call clear.

amitmorade:
> Verified the reachability claim rather than taking it on faith: NewSafeCookieJar never returns (nil, nil), the config validator genuinely rejects an oauth broker unless Hop2OAuthActive()'s preconditions hold, main.go builds the exchanger (or exits) before the pool is constructed, and NewBrokerClient has exactly one production call site (pool.go:110). Both nil states really are unreachable today.
> 
> Also checked whether removing the returned-error check in newAuthenticator's oauth branch trades away real diagnostics for a future regression, since that one wasn't a panic before, it was a clean Go error. It doesn't: classifyBrokerError wraps it as broker_init_error, which isn't one of the audited error types in manager.go's logging gate, so only the Go type was ever logged, never the message text. A future nil-pointer panic would actually surface with a full stack trace, so there's no diagnosability loss here.
> 
> One non-blocking observation: the "Scope" section's line between keeping and removing guards (incident-backed vs. not) doesn't quite cover internal/semp/resilience/sender.go's authn nil-check the same way it covers sem, since authn has no SOL-150116-style silent-failure story of its own. Not something this PR needs to fix, just worth a comment there sometime so a future pass doesn't assume the two guards share one rationale.
> 
> Good tests, clean commit sequence, and the closed-#161 context makes the design call clear.

Good point — thanks for calling that out. The "Scope" wording lumps two guards under one rationale (SOL-150116), but that story really only applies to `sem`.

The honest split:
- `sem` guard: kept for the incident history.
- `authn` guard: same shape as the ones deleted here, no independent incident story of its own.

I left both because splitting them inside the same constructor felt like unmotivated asymmetry, but I think you're right that it's a stylistic call more than a substantive one. If someone does a future pass and removes the `authn` check on the same principle, that would be reasonable.

I'll add a small note on `sender.go` clarifying the split so nobody assumes both guards share the same rationale. Not for this PR — just tracking.

Appreciate the careful review.

PR #167, opened 2026-07-13
SOL-151646: Rework write action tools to use embedded SEMPv2 API

solace-wmaryam:
## Summary

Migrates the four SEMPv2 write-action tools (delete-queue-messages, clear-queue-stats, disconnect-client, clear-client-stats) from bespoke Go handlers to composite YAML definitions, aligning them with every other write tool already served by the composite framework.

Fixes #[SOL-151646](https://sol-jira.atlassian.net/browse/SOL-151646)

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The composite framework already backs every other write tool in tools.yaml (create/update/delete VPN, queue, topic endpoint, RDP), executor-tested against action/* operations, but the four action tools still hand-rolled their own sempv2.Operation, JSON schemas, and ToolHandler boilerplate because the SEMPv2 action OpenAPI spec was dropped in SOL-148431 and never re-added. Re-embedding the action spec lets these tools live in YAML alongside their siblings and removes two Go packages of duplicated infrastructure.

## Changes Made

- Embedded semp-v2-swagger-private-action-extended.json (load 10.26.0.8615, same source as the config spec) and wired __private_action__ into validSpecTypes + privateToPublicSpecType in internal/semp/sempv2/operation.go.
- Added four composite YAML tool definitions (delete-queue-messages, clear-queue-stats, disconnect-client, clear-client-stats) to internal/composite/definitions/tools.yaml using action/do* operations and strategy: collect. Descriptions copied verbatim from the retired native handlers.
- Deleted internal/tools/sempv2/queueactions/ and internal/tools/sempv2/clientactions/ and removed registerSEMPv2Tools (with its imports) from cmd/server/main.go.
- Fixed a stale docstring reference in internal/tools/queuemetrics/handler.go.
- Updated docs/internal/architecture.md (embedded-spec row) and docs/tools-reference.md.

## Testing

**Test Environment:**
- Go version: go1.26.3 darwin/arm64
- OS: macOS 26.5.1
- Broker version (if applicable): 10.26.0.8615 (spec source)

**Tests Performed:**

**Test Coverage:**
- New loader tests for each of the four tools (parses YAML, asserts annotations + operation name).
- New executor tests that assert each tool issues the expected PUT /SEMP/v2/__private_action__/... call with both path params filled in, synthesizes an empty request body, and collects under the step ID.
- TestParseSpecs_SpecTypeDerivation extended to assert the action spec loads under the action/ prefix.
- E2E suite from SOL-150455 (test/e2e-action/test-action-tools.sh) exercises live-broker semantics: clearStats resets counters, deleteMsgs drains live depth, disconnect terminates the session. All assertions read broker state via SEMP rather than parsing the tool response body, so they carry over unchanged.

## Breaking Changes

**Impact:**
The four tools' response envelope changes from the strict { status: "ok", msgVpnName, queueName|clientName } (with additionalProperties: false) to the composite step-keyed envelope every other write tool already returns:

  { "deleteMsgs": { "data": {}, "meta": { "responseCode": 200, "request": { ... } } } }

Success is now signalled by meta.responseCode == 200. The path params (msgVpnName, queueName/clientName) are no longer echoed in the response body — SEMPv2 action endpoints don't return them, and the caller supplied them.

**Migration Guide:**
Clients that parsed the strict envelope should switch to reading meta.responseCode (or rely on the MCP isError protocol field, which is unchanged). Nothing else about the tool contract — names, input schema, annotations, gating behind enable_write_tools, or destructive-confirmation prompts — changes.

## Additional Notes

**Guardrail dropped:** The retired native handlers set `minLength: 1` on their `msgVpnName`/`queueName`/`clientName` params and rejected empty strings at schema-validation time. Composite `ParameterDef` has no equivalent field, so an empty string now passes validation and templates into a double-slash SEMP path (`.../msgVpns//queues/foo/deleteMsgs`), which the broker rejects with a 4xx. Not a correctness or safety regression — the request fails either way — but the error surfaces one round trip later and with a broker-side message rather than a schema-validator message. Consistent with the rest of the composite write family, which already lacks this guard.

**Follow-up:** Worth a separate ticket to cross-reference composite `ParameterDef` against the operations catalog and auto-emit `minLength: 1` on parameters that map to SEMP path params. Would restore the schema-level rejection uniformly across every composite tool — not just the four migrated here.

[SOL-151646]: https://sol-jira.atlassian.net/browse/SOL-151646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solace-aross (approved):
Approving. Clean, well-scoped refactor that removes two packages of hand-rolled boilerplate and lands the action tools alongside their composite siblings. The PR description names the breaking change and migration path clearly. Build, vet, and the composite/sempv2/tools test packages pass on the PR head.

I checked the two things most at risk in this move, and both hold:

- Write-gating survives. `isWriteTool` keys off `!ReadOnly` and the loader maps `readOnly: false` through to `Metadata`, so all four tools still register only under `enable_write_tools` (`register.go:143`).
- The destructive WARNING still fires. `manager.go:181` logs off `Annotations.Destructive`, which the loader now populates for `delete-queue-messages` and `disconnect-client`.

One thing to decide, low severity, not blocking:

The `minLength: 1` guard on the path params is dropped. The old native handlers enforced `minLength: 1` plus explicit empty-string checks on `msgVpnName`/`queueName`/`clientName` ("a malformed request is never sent"). The composite `ParameterDef` has no constraint field, so an empty `queueName` now passes validation and templates into a double-slash SEMP path. Worst case is a broker 4xx, not an unintended destructive action, so it is minor. But it is a documented guardrail removed on destructive tools without a mention in the notes.

Two options: accept it, since every other composite write tool already lacks `minLength`, and add a line to the notes so it is a conscious call; or, if empty-segment protection matters for the destructive tools, add a constraints field to `ParameterDef` so the fix lives in the framework rather than per-tool Go. My lean is the former. Your call.

Copilot already covered the `recorded[0]` length-guard nits, so I won't restate them.

solace-wmaryam:
In response to Copilot's comments, I don't believe these concerns are worth fixing. I looked at whether an empty slice is actually reachable in these tests, and it doesn't look like this can happen - the tool has exactly one step declared in the source, and any executor error is caught by an earlier if err != nil { t.Fatalf(...) } before we ever reach the index. Not adding a guard for a case that isn't reachable.

solace-wmaryam:
@solace-aross thanks — good catch. Updated the PR body's Additional Notes to call the guardrail drop out on its own line rather than folding it into the follow-up suggestion. Agreed on the lean: accept in this PR, file the `ParameterDef` constraint-field extension separately so it applies uniformly across every composite tool.

PR #169, opened 2026-07-13
SOL-151816: remove per-broker OAuth scopes from Hop 2 token exchange

amitmorade:
## Summary

Removes the per-broker `brokers.<alias>.auth.scopes` field end-to-end from the Hop-2 OAuth (RFC 8693) token-exchange path, before the runtime ships. The exchange request now omits the RFC 6749 §3.3 `scope` parameter, so the IdP grants its per-client / per-user default scopes based on the subject token.

Related: [SOL-151816](https://sol-jira.atlassian.net/browse/SOL-151816) (the follow-up designing the successor mechanism), [SOL-150799](https://sol-jira.atlassian.net/browse/SOL-150799) (the T5 sub-task that introduced the field).

## Type of Change

- [x] Refactoring (removes an unshipped config surface; no user-visible behavior change)

## Motivation and Context

Scope entitlement is a property of the **user** (carried by the subject token), not of the **broker**. A single static per-broker list has two failure modes and no third setting that works:

- **Overscoped for read-only users.** On a broker configured with `scopes: [semp:admin]`, a read-only user's exchange requests `semp:admin` with their subject token. Strict IdPs (Entra always; Keycloak depending on client-scope config) reject this with `invalid_scope` — the read-only user gets 401s on every tool, including reads.
- **Underscoped for admins.** Set to the least-privilege value, admin operations fail broker-side because the exchanged token doesn't carry the entitlements the user has.

The Hop-2 OAuth runtime is gated behind `ENABLE_UNRELEASED_BROKER_OAUTH` and was never activated in the wild with the field populated, so removing this now costs no user-visible behavior. It prevents a wrong-axis knob from calcifying into a public contract.

## Changes Made

**Production code (feature removed):**

- `internal/config/config.go` — dropped `AuthConfig.Scopes []string` and the `broker.Auth.Scopes` whitespace-validation loop in `validateBroker`.
- `internal/tokenexchange/types.go` — dropped `ExchangeInput.Scopes`, reworded the package doc.
- `internal/tokenexchange/request.go` — dropped `setScopes` and its call in `buildIdPRequest`. Added a doc comment explaining the intentional omission of the `scope` parameter.
- `internal/semp/auth/oauth.go` — dropped the `scopes` field, dropped the `scopes []string` param from `NewOAuthAuthenticator`, dropped `Scopes: a.scopes` in `AddAuth`.
- `internal/semp/broker.go` — sole production call site updated to the new 3-arg constructor.

**Tests:**

- Deleted the three now-dead scope tests in `internal/tokenexchange/request_test.go` and replaced with a single `TestBuildIdPRequest_NoScopeParameter` that pins the "IdP defaults" behavior.
- `internal/tokenexchange/exchange_test.go` — removed the `Scopes:` input from `TestExchange_RequestBodyContainsExpectedFields` and added an assertion that the `scope` form field is absent on the wire.
- `internal/semp/auth/oauth_test.go` — updated all six `NewOAuthAuthenticator` call sites for the new signature and dropped the "scopes reach exchanger" assertion.
- `internal/config/config_test.go` — replaced the whitespace-validation subtest with a **strict-decoder** subtest that pins `KnownFields(true)` rejecting any config that still declares `auth.scopes`. Stripped `scopes:` from the schema-valid fixture and reworded a stale comment.

**Docs:**

- `CHANGELOG.md` — reworded the `Added` entry for the `broker_oauth:` block (audience-only), removed a comparison clause that referenced the removed feature, and added a new `### Removed` bullet pointing at SOL-151816.

## Testing

**Test Environment:**
- Go: repo default (via `go.mod`)
- OS: macOS 25.5.0
- Broker version: n/a (no runtime behavior change)

**Tests Performed:**

**Test Coverage:**
- New `TestBuildIdPRequest_NoScopeParameter` pins: the request omits `scope` — the IdP applies defaults.
- New strict-decoder subtest pins: any config that still declares `auth.scopes` fails at startup with `field scopes not found in type config.AuthConfig`. This makes it structurally hard to silently reintroduce the field.
- Existing signature-only test edits keep coverage of `NewOAuthAuthenticator`, `Exchange`, and `AddAuth` intact.
- `make check` ran clean **on this diff**. Two pre-existing `staticcheck SA5011` findings in `internal/tools/sempv1/redundancy/handler_test.go:112,116` predate this branch (April 2026, `19d5e581`, unrelated code) — flagging so reviewers aren't confused if their local `make check` output shows the same two hits.

## Breaking Changes

**Impact:** None user-visible. `AuthConfig.Scopes` was schema-shipped in v0.5.0 but the OAuth runtime that would have consumed it is gated behind `ENABLE_UNRELEASED_BROKER_OAUTH` and has never activated in the wild with this field populated.

**Migration Guide:** Any config that still declares `brokers.<alias>.auth.scopes` will now fail at startup with a strict-decoder error naming the removed field. Remove the key.

### Code Quality
- [x] Follows coding standards
- [x] Self-review completed

### Security
- [x] No credentials in code
- [x] Followed secure logging rules; `/check-logs` clean on the changed files
- [x] Credential-carrying types unchanged; `AuthConfig` still implements `slog.LogValuer`
- [x] No new gosec/golangci-lint findings introduced by this diff

### Testing
- [x] `go test -race ./...` (via `make check`) — passing on this diff
- [x] Edge cases covered: strict-decoder rejection subtest, "no scope field on the wire" assertion in both request-builder and end-to-end tests

### Documentation
- [x] CHANGELOG.md updated (`Added` reworded, new `Removed` bullet)

### Git & CI
- [ ] All commits signed off (DCO)
- [x] Commit message is descriptive
- [x] Branch fresh off `main` (base commit `59e2468`)

## Additional Notes

- The token-exchange dedup / cache key is unchanged today (`sha256(subjectToken, brokerAlias)`). If the successor mechanism designed under SOL-151816 makes scopes vary per call for the same subject token, that key **must** be extended or the cache will serve wrong-scoped tokens. Called out explicitly in the CHANGELOG and in SOL-151816's acceptance criteria.
- A stay-list was actively verified during cleanup — `AudienceParamScope` (Entra wire-format for *audience*, not the removed feature), Hop-1 principal `Scopes`, `scopes_supported` metadata, `dev_token_scopes`, Keycloak IdP client-scope provisioning, and historical design docs under `docs/oauth/token-exchange-SOL-150070/` — all untouched.

## Related Issues/PRs

- Follow-up: [SOL-151816](https://sol-jira.atlassian.net/browse/SOL-151816) — Redesign per-user scope handling for Hop 2 token exchange.
- Predecessor: [SOL-150799](https://sol-jira.atlassian.net/browse/SOL-150799) — the T5 sub-task that ported the Exchanger and consumed `scopes` on the wire.
- Parent story: [SOL-150070](https://sol-jira.atlassian.net/browse/SOL-150070).

---

**For Reviewers:**

**Focus Areas**:
1. Confirm the stay-list is right — nothing on the Hop-1 or `audience_parameter_name` side got touched by accident.
2. The strict-decoder subtest in `config_test.go` — is that the right way to prevent silent re-introduction of `auth.scopes`, or would you prefer a differently-scoped guard?
3. The single-assertion replacement test for scope-absence — enough coverage, or should there be a positive dual to cover more of the request-builder wiring?

**Questions**: none blocking.

[SOL-151816]: https://sol-jira.atlassian.net/browse/SOL-151816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-150799]: https://sol-jira.atlassian.net/browse/SOL-150799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-151816]: https://sol-jira.atlassian.net/browse/SOL-151816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solace-aross (approved):
Approve. Clean deletion-only removal of an unshipped surface, and the reasoning holds: scope entitlement is a user property carried on the subject token, not a per-broker constant.

I verified the three load-bearing claims rather than taking them on faith:

- Strict decoding is real. `config.go` sets `dec.KnownFields(true)`, and it recurses into `AuthConfig`, so the guard subtest's asserted error (`field scopes not found in type config.AuthConfig`) is genuine, not decorative. Nice way to make the field structurally hard to reintroduce.
- Stay-list is accurate. The only surviving `Scopes` references are Hop-1 principal scopes and `scopes_supported` metadata. Nothing on the `audience_parameter_name` path moved.
- `dedup_key.go`, referenced in three comments, exists. The forward-looking note about per-user scopes joining the dedup key points at a real file.

On your focus questions: the strict-decoder subtest is the right guard, and scope-absence is pinned at two layers (request builder plus end-to-end form body), which is enough. Deleting the duplicate/blank-element edge tests was correct since they pinned behavior that no longer exists.

Two minor points, both description/changelog only, neither blocking:

1. The description says `auth.scopes` was "schema-shipped in v0.5.0". There is no v0.5.0. The whole `broker_oauth:` block, scopes included, is under `[Unreleased]`; latest release is `0.1.0`. Worth correcting, and it strengthens the safety argument: the field never shipped in any release.

2. Since `scopes` only ever existed under `[Unreleased]`, the `### Removed` bullet is optional. Editing the `Added` entry in place (which you did) already handles it; the extra bullet makes the eventual release notes read "added the OAuth block, removed a sub-field of it" in one version. Your call whether the loud signal for anyone on `main` is worth that. I lean toward dropping it.

The Copilot comment is already resolved in the current diff.

amitmorade:
> Approve. Clean deletion-only removal of an unshipped surface, and the reasoning holds: scope entitlement is a user property carried on the subject token, not a per-broker constant.
> 
> I verified the three load-bearing claims rather than taking them on faith:
> 
> * Strict decoding is real. `config.go` sets `dec.KnownFields(true)`, and it recurses into `AuthConfig`, so the guard subtest's asserted error (`field scopes not found in type config.AuthConfig`) is genuine, not decorative. Nice way to make the field structurally hard to reintroduce.
> * Stay-list is accurate. The only surviving `Scopes` references are Hop-1 principal scopes and `scopes_supported` metadata. Nothing on the `audience_parameter_name` path moved.
> * `dedup_key.go`, referenced in three comments, exists. The forward-looking note about per-user scopes joining the dedup key points at a real file.
> 
> On your focus questions: the strict-decoder subtest is the right guard, and scope-absence is pinned at two layers (request builder plus end-to-end form body), which is enough. Deleting the duplicate/blank-element edge tests was correct since they pinned behavior that no longer exists.
> 
> Two minor points, both description/changelog only, neither blocking:
> 
> 1. The description says `auth.scopes` was "schema-shipped in v0.5.0". There is no v0.5.0. The whole `broker_oauth:` block, scopes included, is under `[Unreleased]`; latest release is `0.1.0`. Worth correcting, and it strengthens the safety argument: the field never shipped in any release.
> 2. Since `scopes` only ever existed under `[Unreleased]`, the `### Removed` bullet is optional. Editing the `Added` entry in place (which you did) already handles it; the extra bullet makes the eventual release notes read "added the OAuth block, removed a sub-field of it" in one version. Your call whether the loud signal for anyone on `main` is worth that. I lean toward dropping it.
> 
> The Copilot comment is already resolved in the current diff.

Thanks for the review, and good catch on the version confusion — but I think the real issue is that our CHANGELOG hasn't been kept consistent.

On the "schema-shipped in v0.5.0" claim: `git tag` shows v0.5.0 tagged 2026-07-09 (commit `fa903d2`), and `git show v0.5.0:internal/config/config.go` has `Scopes []string` at line 381. So the field did ship in the v0.5.0 binary — my agent referred to that. What likely made this look off is that our CHANGELOG has drifted from reality: we have four real releases (v0.2 through v0.5, May → July) with no version blocks — everything from those three months is still under `[Unreleased]`. Reading the CHANGELOG alone gives "latest release is v0.1.0", which contradicts what actually shipped.

Two follow-ups I'm taking:
1. Update the project-wide `CLAUDE.md` so keeping the CHANGELOG in sync with PR changes becomes default behaviour for anyone working in this repo through Claude Code — not something the model has to guess at.
2. Filed [SOL-151818](https://sol-jira.atlassian.net/browse/SOL-151818) under the GA Quality Plan epic to roll the current `[Unreleased]` content into proper `[0.2.0]`/`[0.3.0]`/`[0.4.0]`/`[0.5.0]` blocks by walking the tag dates.

Given the above, I'll keep the `Added` rewording and the `Removed` bullet as-is — `scopes` did ship in v0.5.0 binaries, so the operator signal is real, even if the section it's landing in is stale for other reasons.

[SOL-151818]: https://sol-jira.atlassian.net/browse/SOL-151818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #170, opened 2026-07-14
SOL-151624: OAuth 401 retry re-authenticates in-flight request (PrepareRetry hook)

solace-wmaryam:
## Summary

Re-authenticate in-flight requests on OAuth 401 by wiring a `PrepareRetry` hook that re-invokes `AddAuth` before the retry attempt.

Fixes [SOL-151624](https://sol-jira.atlassian.net/browse/SOL-151624)

## Type of Change

## Motivation and Context

Before this change, an OAuth 401 from the broker terminated the call immediately with no retry. `HandleAuthFailure` evicted the cached token (benefiting the *next* request), but returned `false` because retrying would replay the same stale `Authorization` header — a guaranteed second 401.

This wires the missing `PrepareRetry` hook so the retry carries a fresh token.

## Changes Made

- Add `needsReauth` field to `retryState` in `retry.go`
- Set `needsReauth = true` in the 401 branch when `HandleAuthFailure` returns `true`
- Add `prepareRetry` method to `Sender` that re-invokes `AddAuth` when `needsReauth` is set
- Wire `PrepareRetry` hook in `New()`
- Flip `OAuthAuthenticator.HandleAuthFailure` to return `true` after cache eviction
- Update docstrings to reflect new behavior
- Remove stale followup comment from `oauth_cache_wiring_test.go`

## Testing

**Test Environment:**
- Go version: 1.24
- OS: macOS Darwin 25.5.0

**Tests Performed:**

**Test Coverage:**
- `TestSender_Retry_401_OAuth_ReAuthsOnRetry` — 401 triggers retry with a different `Authorization` header
- `TestSender_Retry_429_SkipsReAuth` — 429 does NOT re-invoke `AddAuth` (guards against unnecessary IdP round-trips)
- Existing Basic/Bearer 401 tests still pass unchanged

## Breaking Changes

None. Behavior change only affects OAuth mode on 401, which previously failed immediately.

### Code Quality
- [x] Self-review completed
- [x] Code is well-commented

### Security
- [x] No credentials in code

### Testing
- [x] Error paths tested

### Documentation

### Git & CI

## Related Issues/PRs

- Related to [PR #146](https://github.com/SolaceDev/solace-broker-mcp/pull/146) (OAuth token exchange cache)
- Parent epic: SOL-148417

---

**For Reviewers:**

**Focus Areas**: The `prepareRetry` hook wiring and the guardrail that prevents re-auth on 429/503 retries.

**Questions**: None.

[SOL-151624]: https://sol-jira.atlassian.net/browse/SOL-151624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solace-awaheed (commented):
The same Co-pilot suggestion was caught on my end as well, everything else looks good!

internal/semp/resilience/sender.go
```diff
@@ -238,6 +239,17 @@ func (b *cancelOnCloseReadCloser) Close() error {
 	return err
 }
 
+// prepareRetry re-runs AddAuth when the previous attempt failed with 401.
+// Skipped on 429/503/5xx retries to avoid unnecessary IdP round-trips.
+func (d *Sender) prepareRetry(req *http.Request) error {
+	state := getRetryState(req.Context())
+	if state.needsReauth {
+		state.needsReauth = false
+		return d.authenticator.AddAuth(req.Context(), req)
```
solace-aross, line 257: If the IdP is down when this re-runs `AddAuth`, retryablehttp aborts with `prepareErr` and `errorHandler` returns `RetriesExhaustedError{StatusCode: 0, Err: ...}`. Pre-PR the same persistent 401 surfaced as `StatusCode: 401`. And unlike the `resp != nil` path, this branch emits no `slog.Error` and drops `OperationIDKey`, so a caller keying on `StatusCode == 401` misclassifies it and the resilience layer goes quiet.

Worth a test where the second `AddAuth` errors, plus a log here or in the `errorHandler` `err != nil` branch naming broker and operation.

internal/semp/resilience/retry.go
```diff
@@ -106,6 +107,7 @@ func (d *Sender) checkRetry(ctx context.Context, resp *http.Response, err error)
 			state.auth401Retried = true
 			retry := d.authenticator.HandleAuthFailure(ctx, resp.Header)
 			if retry {
+				state.needsReauth = true // signal PrepareRetry to re-run AddAuth
```
solace-aross: The second-401 case (bad credentials, refresh did not help) is the realistic bad path, and it is exactly what the `auth401Retried` cap protects. No test covers it today, so a regression that re-set `needsReauth` on the second 401 would ship unnoticed.

Suggest a test: server 401s twice then 200s, asserting exactly 2 requests, 2 `AddAuth` calls, and `StatusCode: 401`. While there, the re-auth path is only tested over GET; a `WithRetrySafe` POST variant would cover the SEMPv1 `<show>` production path.

internal/semp/auth/oauth.go
```diff
@@ -78,21 +78,9 @@ func (a *OAuthAuthenticator) AddAuth(ctx context.Context, req *http.Request) err
...
-// AddAuth between attempts (retryablehttp replays the same *http.Request
-// with the stale Authorization header). Returning true would guarantee
-// another 401 plus one wasted round-trip and backoff — strictly worse
-// than pre-cache behavior. SOL-151624 tracks wiring a PrepareRetry hook
-// that re-runs AddAuth on retry; once that lands, this returns true.
-//
-// Basic/Bearer are unaffected — their AddAuth sets a static header, so
-// replaying the request on retry carries the same (still-correct) value.
-// Only OAuth needs the PrepareRetry hook to make in-flight recovery work.
+// HandleAuthFailure evicts the cached token and returns true to trigger an
```
solace-aross: Good to see the contract spelled out here. `HandleAuthFailure` returning `true` now guarantees the resilience layer re-invokes this authenticator's `AddAuth` on the in-flight request. That obligation lives only in this comment, not on the `Authenticator` interface in `auth.go`, which is the boundary the resilience layer actually couples to. Worth lifting a line onto the interface doc so a future auth mode knows returning `true` triggers `AddAuth` re-invocation.

internal/semp/resilience/retry.go
```diff
@@ -106,6 +107,7 @@ func (d *Sender) checkRetry(ctx context.Context, resp *http.Response, err error)
...
 				slog.Warn("retrying: 401 received, auth handler recovered",
```
solace-aross: "recovered" fires before recovery is actually attempted. The token fetch happens later in `prepareRetry`'s `AddAuth`, which can fail. When it does, the log reads "recovered" right before the failure, pointing the operator away from the IdP. Consider "401 received, evicted cached token, retrying with fresh token" and reserve any "recovered" claim for after a successful retry.

solace-aross (dismissed):
Approving. The fix is correct and well-scoped.

I traced the retryablehttp loop end to end: `checkRetry` sets `needsReauth`, then after backoff `PrepareRetry` re-runs `AddAuth` on the shared request and overwrites the stale `Authorization` before the retry. The flag rides the per-request `retryState` pointer, so it propagates across attempts and the one-shot clear is correct. `auth401Retried` caps re-auth at one attempt, and 429/503 correctly skip it. The happy-path test genuinely fails without the wiring.

None of the notes below block the merge. A few inline; two more here because they touch unchanged lines.

**retryBudget invariant** (`sender.go` ~109-127). The docstring promises one full `RequestTimeoutDuration` remains for the final attempt. The re-auth exchange now runs inside the chain on a detached context (`exchange.go:65`), so the semaphore slot can be held up to `exchangeTimeout` (10s) beyond the computed budget. Immaterial at defaults, ~30% under short tunings like `retries: 1, request_timeout: 2s`. Worth folding `exchangeTimeout` into the formula, or a one-line caveat so the invariant stays honest.

**Persistent-authz 401 and IdP load** (question). The retry fixes the expiry race. When the 401 is authz rejection (wrong audience/scope, broker distrusts the IdP, clock skew), the re-exchanged token carries the same claims and 401s again, after evict + backoff + a fresh exchange. On a broker-wide invalidation, concurrent 401s from different subjects become synchronous exchanges at the same instant (singleflight only collapses same-subject). Bounded to ~2x by `auth401Retried`, so not a blocker. Was the slow or degraded IdP case weighed? A line in the runbook noting the expected IdP-latency signature would help on-call.

Minor: no CHANGELOG or architecture entry for an operator-visible change; `architecture.md:451` still describes the Sender as plain "auth-failure re-auth."

Clean test coverage on the happy path and the 429 guard.

amitmorade:
Clean, well-scoped change. The `PrepareRetry` hook is the right mechanism and the isolation model is solid (per-request `retryState` on context, single-goroutine retry loop). No blockers.

### Medium

**`needsReauth` flag is set by the resilience layer, not the authenticator**

`checkRetry` sets `needsReauth = true` whenever `HandleAuthFailure` returns `true`. This means Basic auth 401 retries also go through `prepareRetry` → `AddAuth`, which re-sets the same static `Authorization: Basic ...` header. Harmless today (idempotent, no network call), but it's the resilience layer deciding "retry = re-auth" when that's really the authenticator's concern.

Two options, your call:
- (a) Leave as-is and note that `AddAuth` on Basic retry is a no-op — simplest, works fine
- (b) Enrich the signal from `HandleAuthFailure` (e.g., return a struct with `Retry bool` + `ReAuth bool`) so the authenticator owns the decision — cleaner abstraction, bigger change

Not blocking — just flagging the abstraction boundary. If we ever add an auth mode where re-running `AddAuth` on retry has side effects, this assumption breaks. (+1 to Andrea's comment about lifting this contract onto the `Authenticator` interface doc.)

### What's good

- Minimal blast radius — 4 production lines changed, rest is docs/tests
- `needsReauth = false` reset in `prepareRetry` prevents double re-auth on subsequent non-401 retries
- Negative test (`429_SkipsReAuth`) guards against unnecessary IdP round-trips
- Stale comment cleanup in `oauth_cache_wiring_test.go`

solace-wmaryam:
@solace-aross — thanks for the thorough pass. Addressed the rest of your notes:

- **errorHandler silent on re-auth failure**: the `err != nil` path now logs `slog.Error` with broker + operation, and carries the last observed `401` through `RetriesExhaustedError.StatusCode` instead of `0`. Added a test that fails the retry's `AddAuth` (IdP-down) and asserts both the surfaced error and the log.
- **Missing coverage**: added the double-`401` cap test (exactly 2 requests / 2 `AddAuth` calls, returns `401`) and a `POST` + `WithRetrySafe` re-auth variant for the SEMPv1 `<show>` path.
- **retryBudget invariant**: added a docstring caveat noting the re-auth exchange runs on a detached context bounded by its own exchange timeout, so a re-auth retry can hold the slot up to that beyond the budget.
- **Docs**: CHANGELOG entry + a design-decisions row in architecture.md.

On the **persistent-authz / slow-IdP signature** — agreed it's worth capturing for on-call. Where would you like it to live: a new `docs/internal/` runbook, or somewhere existing?

internal/semp/resilience/sender.go
```diff
@@ -238,6 +239,17 @@ func (b *cancelOnCloseReadCloser) Close() error {
 	return err
 }
 
+// prepareRetry re-runs AddAuth when the previous attempt failed with 401.
```
amitmorade: The docstring says "when the previous attempt failed with 401" — but the actual trigger is `state.needsReauth`, which comes from `AuthFailureResult.ReAuth`. A 401 where `ReAuth=false` (Basic auth) won't trigger this. Now that the authenticator owns the re-auth decision via the struct, the comment should reflect that.

Suggestion:
```go
// prepareRetry re-runs AddAuth when the authenticator signalled ReAuth
// on the preceding 401. Skipped for 429/503/5xx retries and for auth
// modes whose recovery does not require fresh credentials (e.g. Basic).

solace-wmaryam: Thanks for catching that! Will update.

amitmorade (dismissed):
LGTM

solace-aross (approved):
Approving. Cleaner than the version I looked at before, and every round-2 note — mine, Copilot's, and amitmorade's — is resolved.

The change I like most: the re-auth decision now lives in the authenticator via `AuthFailureResult{Retry, ReAuth}`, so the resilience layer no longer hardcodes "OAuth needs re-auth." The obligation I'd asked to lift onto the interface is now a typed field documented on the boundary itself, which is stronger than a comment. Basic returns Retry without ReAuth (static header), Bearer declines, OAuth returns both — each authenticator owns what only it knows.

Re-traced the path end to end: `checkRetry` sets `needsReauth` only when `Retry` is true, `prepareRetry` re-runs `AddAuth` once and clears the flag, and `auth401Retried` still caps re-auth at one attempt, so a persistently-rejected credential surfaces as a 401 instead of looping. The IdP-down path now carries the last observed 401 and logs broker + operation (my earlier gap), and the retryBudget / exchange-timeout interaction is documented where it belongs.

Test coverage is thorough: the cap, the 429 skip, the re-auth-AddAuth-failure surfacing, and the `WithRetrySafe` POST path for the SEMPv1 <show> production route. CI green including -race and e2e-oauth. Nice work.

amitmorade (approved):
LGTM

PR #178, opened 2026-07-15
SOL-151875: Add ToolAuthorizationConfig schema and validation (T1)

amitmorade:
## Summary

Adds the `tool_authorization` config block to `MCPClientAuthConfig` — schema,
defaults, validation, observability (`slog.LogValuer`), and 16 contract tests.
This is **Ticket T1** of the claims-based tool authorization feature (SOL-151440).

All validation is gated behind `ENABLE_TOOL_AUTHORIZATION=true` so existing
deployments are completely unaffected. The YAML block is parsed but ignored
when the flag is unset.

## Type of Change

## Motivation and Context

SOL-151440 introduces claims-based tool authorization (RBAC) for the MCP
server. T1 lays the config foundation — schema parsing, invariant enforcement,
and sensible defaults — so that later tickets (T2–T6: enforcement middleware,
access-level resolution, tool filtering) can build on a validated config
without re-litigating field semantics.

The feature flag (`ENABLE_TOOL_AUTHORIZATION`) follows the same pattern as
`ENABLE_UNRELEASED_BROKER_OAUTH` and will be removed when the full feature
ships (T2–T6 complete).

### Config schema (`internal/config/config.go`)

- **`ToolAuthorizationConfig` struct** with presence-sentinel fields:
  - `Enabled *bool` — pointer so omission is distinguishable from `false`
  - `GroupsClaimName *string` — claim name in the JWT; defaults to `"groups"`
  - `AccessLevelGroups map[string][]string` — maps group names to allowed tool lists

- **`ToolAuthorization *ToolAuthorizationConfig`** field on `MCPClientAuthConfig`
  (pointer so an absent YAML block is `nil`, not zero-value)

- **`LogValue()` method** — implements `slog.LogValuer` with:
  - 3-way `enabled` string: `"true"` / `"false"` / `"unset"`
  - Verbatim `groups_claim_name` (not sensitive)
  - `access_level_groups` intentionally omitted (can contain group names)

- **`applyToolAuthorizationDefaults()`** — synthesizes defaults:
  - Creates an empty `ToolAuthorizationConfig{}` when the block is omitted
    under `mode: oauth`, so the validator always sees a non-nil pointer
  - Defaults `groups_claim_name` to `"groups"` when the field is omitted

- **`validateToolAuthorization()`** — enforces invariants:
  - `tool_authorization` block is only legal under `mode: oauth`
  - `enabled` must be set explicitly in oauth mode (no implicit default)
  - `groups_claim_name` cannot be empty or whitespace-only when set
  - `enabled: true` requires non-empty `access_level_groups`
  - Group names in `access_level_groups` cannot be empty/whitespace

- **Feature flag**: `ENABLE_TOOL_AUTHORIZATION` env var gates both functions.
  When unset/false, the YAML block is parsed by Go but all validation and
  defaulting is skipped — zero behavioral change for existing deployments.

### Contract tests (`internal/config/tool_authorization_test.go`)

16 tests covering every documented behavior, all driven through YAML → `LoadConfig`:

| Test | Behavior |
|------|----------|
| OmittedBlockSynthesizedInOAuthMode | Omitted block synthesized, validator requires enabled |
| EnabledFalseAloneIsLegal | enabled: false alone is a valid opt-out |
| EnabledFalseWithPopulatedFieldsLoads | enabled: false with other fields loads correctly |
| OmittedGroupsClaimNameDefaultsToGroups | Omitted groups_claim_name defaults to "groups" |
| BlankGroupsClaimNameRejected | Empty/whitespace groups_claim_name rejected (3 subtests) |
| EnabledTrueEmptyAccessLevelGroupsRejected | enabled: true with no groups → error |
| CaseSensitiveGroupNamesPreserved | Case-sensitive group names preserved |
| WhitespaceInGroupNamePreservedVerbatim | Trailing whitespace in group name kept verbatim |
| EmptyAndWhitespaceOnlyGroupNamesRejected | Empty/whitespace group names rejected (3 subtests) |
| LogValueEnabledRendering | LogValue renders enabled as "true"/"false" (2 subtests) |
| LogValueGroupsClaimNameVerbatim | LogValue renders groups_claim_name verbatim |
| LogValueOmitsAccessLevelGroups | LogValue omits access_level_groups |
| NonOAuthModeWithBlockRejects | tool_authorization under non-oauth mode → error (2 subtests) |
| OAuthModeEnabledOmittedRejects | Omitted enabled in oauth mode → error |

## Example Config

```yaml
port: 9091
tls_terminated_upstream: true
mcp_client_auth:
  mode: oauth
  issuer: "https://idp.example.com/realms/solace"
  audience: "solace-mcp-server"
  resource_url: "https://mcp.example.com/mcp"
  tool_authorization:
    enabled: true
    groups_claim_name: "groups"      # optional — "groups" is the default
    access_level_groups:
      mcp-readers:
        - list-queues
        - get-broker-status
        - get-vpn-health
      mcp-viewers:
        - list-queues
        - get-broker-status
      mcp-operators:
        - list-queues
        - get-broker-status
        - create-queue
        - update-queue
      mcp-admins:
        - list-queues
        - get-broker-status
        - create-queue
        - update-queue
        - delete-queue
```

With `enabled: false` (authorization parsed but not enforced):

```yaml
mcp_client_auth:
  mode: oauth
  # ... issuer, audience, resource_url ...
  tool_authorization:
    enabled: false
```

**Feature flag**: Set `ENABLE_TOOL_AUTHORIZATION=true` to activate validation.
Without it, the `tool_authorization` block is silently ignored.

## Testing

**Test Environment:**
- Go version: 1.24
- OS: macOS (Darwin 25.5.0)

**Tests Performed:**

**Test Coverage:**
- 16 contract tests covering all documented behaviors
- All tests use YAML fixtures through LoadConfig (no direct struct construction)
- Each test enables the feature flag via `t.Setenv` for isolation
- Existing oauth test suites pass without modification (feature flag off by default)
- E2E suites unaffected (flag is off, so no validation changes)

## Breaking Changes

None. The feature flag ensures zero behavioral change for existing deployments.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code
- [x] `access_level_groups` intentionally omitted from LogValue (contains group names)

### Testing
- [x] Error paths tested

### Git & CI
- [x] No merge conflicts

## Adjacent Changes

Refactors `envBool` from `internal/config/observability.go` to
`internal/config/config.go` and adds a `context` string parameter so warning
messages identify which env-var subsystem (observability, tool authorization,
broker OAuth) produced a bad value. Existing callers updated to pass their
subsystem name. Mechanical change, no behavior change.

## Development-Time Gating

This PR lands behind `ENABLE_TOOL_AUTHORIZATION` env var while the remaining
tool-authorization tickets are still being merged. Config parsing runs
unconditionally; only the effects (validation errors, `LogValue` rendering,
`applyDefaults` synthesis) are gated. This is a development-time seam that
will be removed when the enforcement wrapper lands. Without it, any PR merged
in the interim that adds `tool_authorization: {enabled: true}` to a config
would silently be inert. Reviewers should not treat this env var as product
surface — it does not appear in customer docs.

## Additional Notes

- The `ToolAuthorizationConfig` field names mirror the Solace broker's OAuth
  profile (`accessLevelGroupsClaimName`, `accessLevelGroups`) for cross-product
  consistency.
- The presence-sentinel pattern (`*bool`, `*string`) follows the principle that
  config-side values = reject nonsense; the operator must explicitly opt in or out.
- This PR is T1 only — enforcement middleware (T2), access-level resolution (T3),
  tool filtering (T4–T6) follow in subsequent PRs.

## Related Issues/PRs

- Part of SOL-151440 (Claims-Based Tool Authorization)
- Ticket: SOL-151875

---

**For Reviewers:**

**Focus Areas:**
- Invariant logic in `validateToolAuthorization()` — are the error messages actionable?
- `LogValue()` redaction posture — is omitting `access_level_groups` the right call?
- Feature flag lifecycle — the constant, function, and two gate sites are documented as one linked lifecycle to delete together.

**Questions:**
- Should `groups_claim_name` also be redacted in `LogValue()`? Currently emitted verbatim since it's a claim key name, not a value.

**Open Questions:**
- **Should we support a wildcard identifier (e.g. `"*"`) that means "grant all tools"?** The current design requires the admin to explicitly list every tool name per group. This is deliberate — if a future release adds a new tool, that tool is *not* silently granted to any group. The admin must review the new tool and consciously add it to the appropriate groups. A wildcard like `"*"` would invert this: new tools would be auto-granted on upgrade without the admin even knowing, which is a security concern. On the other hand, requiring explicit enumeration means more config maintenance on every release that ships new tools. Do we accept that trade-off, or do we want a wildcard with a logged warning on startup?


internal/config/config.go
```diff
@@ -1552,6 +1658,24 @@ func applyEnvOverrides(cfg *ServerConfig) error {
...
+// toolAuthorizationFeatureEnabled reports whether ENABLE_TOOL_AUTHORIZATION
+// is set truthy.
+//
+// LIFECYCLE: this function, its constant, and the two gate sites
+// (applyToolAuthorizationDefaults, validateToolAuthorization) are one linked
+// lifecycle. When the full tool-authorization feature ships, delete this
+// function and its constant, and make the gated calls unconditional.
+func toolAuthorizationFeatureEnabled() bool {
+	return envBool(envEnableToolAuthorization, false)
+}
```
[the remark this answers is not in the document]
amitmorade: Good catch — already addressed in e2f0e14. Rather than duplicating the parse logic per caller (which would become copy-paste debt as more feature flags appear), I moved `envBool` out of `observability.go` into `config.go` and added a `context` parameter so each caller names its own subsystem in the warning:

\```go
func envBool(name string, def bool, context string) bool {
    // ...
    slog.Warn(fmt.Sprintf("ignoring unparseable %s flag; using default", context),
        "var", name, "value", v, "default", def)
    // ...
}
\```

Callers now read:
- `envBool(envEnableToolAuthorization, false, "tool authorization")`
- `envBool(envEnableUnreleasedBrokerOAuth, false, "broker OAuth")`
- `envBool(envObsCorrelationIDEnabled, true, "observability")`

One helper, subsystem-specific warnings, no duplication.

solace-lootii (dismissed):
Approving but just a couple of non-blocking notes. 

`ToolAuthorizationConfig.LogValue()` has a value receiver, but it's stored as a nilable ***ToolAuthorizationConfig** so calling it on nil panics (slog recovers gracefully and logs a garbled string instead of crashing, so not a live risk today). 
Also, this is not new to this PR:  `BrokerOAuthConfig` has the identical shape and also isn't logged anywhere yet but  maybe worth a nil-guard or pointer receiver before T2 wires this into a config-dump log.

Everything else LGTM !

amitmorade:
> Approving but just a couple of non-blocking notes.
> 
> `ToolAuthorizationConfig.LogValue()` has a value receiver, but it's stored as a nilable ***ToolAuthorizationConfig** so calling it on nil panics (slog recovers gracefully and logs a garbled string instead of crashing, so not a live risk today). Also, this is not new to this PR: `BrokerOAuthConfig` has the identical shape and also isn't logged anywhere yet but maybe worth a nil-guard or pointer receiver before T2 wires this into a config-dump log.
> 
> Everything else LGTM !

Good catch — fixed in 434f011. Switched to a pointer receiver with a nil guard:

```go
func (t *ToolAuthorizationConfig) LogValue() slog.Value {
    if t == nil {
        return slog.GroupValue()
    }
    // ...
}
```

In non-oauth modes `ToolAuthorization` is `nil` by design (the concept doesn't apply), so the method needs to handle that. Returns an empty group on nil — keeps the return type consistent (always a group, never a magic string). When a future PR wires up config-dump logging, it can decide whether to skip the field entirely for non-oauth modes or let the empty group through.

solace-lootii (dismissed):

internal/config/tool_authorization_test.go
```diff
@@ -0,0 +1,506 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
```
solace-awaheed, line 1: All 14 tests enable the feature flag, so the flag-off path isn't covered. Could we add one test with the flag unset and tool_authorization: {} under mode: oauth that verifies the config loads and GroupsClaimName remains nil? That would catch regressions where the toolAuthorizationFeatureEnabled() guard is accidentally removed.

amitmorade: Good call — added in bec5178. Placed the test in `config_test.go` alongside the existing `ENABLE_UNRELEASED_BROKER_OAUTH` feature flag tests since it's guarding the feature flag gate rather than a tool authorization behavior — felt like a more natural home following the existing pattern.

The test loads `tool_authorization: {}` under `mode: oauth` with the flag unset and asserts `err == nil` — that's the user-facing behavior we're guarding. If someone accidentally removes the `toolAuthorizationFeatureEnabled()` gate, this config would fail with `"enabled must be set explicitly"` and the test catches it.

Kept it to just the error check rather than inspecting struct fields — the user doesn't care about `GroupsClaimName` being nil, they care that the server starts.

solace-awaheed (approved):
Thanks for the update, LGTM!

solace-aross (commented):
Reviewed post-merge, so this is for the record and for T2.

Solid PR. The three prior findings were all handled well: the `envBool` context param, the pointer receiver + nil guard on `LogValue`, and the flag-off test. The gating design (parse-but-ignore behind the flag), the presence-sentinel pointers, and omitting `access_level_groups` from `LogValue` are all the right calls, and the contract tests are thorough.

One minor thing for T2 to sweep up (godoc only, no behavior): when `envBool` moved into config.go, it landed directly under the `applyEnvOverrides` doc comment with no blank line between them (config.go:1624-1648). godoc now attaches "applyEnvOverrides checks for environment variable overrides..." to `func envBool`, and `func applyEnvOverrides` has lost its own doc comment. A blank line plus a one-line `envBool` comment fixes it. Since T2 removes the feature-flag gate and touches this file anyway, easy to fold in there.

PR #185, opened 2026-07-17
SOL-151829: Automate the OAuth token-exchange E2E suite (Keycloak + two brokers + MCP server)

solace-lootii:
## Summary

Automates the OAuth token-exchange E2E suite — Hop 1 (agent → MCP server JWT validation) and Hop 2 (MCP server → broker RFC 8693 token exchange) — against a real Keycloak IdP and two real Solace brokers, and retires the older Hop-1-only `test/oauth/` suite it supersedes.

Fixes # SOL-151829

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The manual OAuth token-exchange flow (Keycloak + two brokers + MCP server) was fully documented and tooled for local use, but never ran in CI — regressions in JWT validation, token exchange, cache invalidation, or group/audience mapping could only surface via manual reproduction. The existing `e2e-oauth` CI job (`test/oauth/`) only covered Hop 1 (`client_credentials` grant, single broker, no token exchange, no group/audience mapping) and predates the token-exchange cache work entirely — it never talks to a broker at all. This closes that gap with a suite that exercises the full path end to end.

## Changes Made

- Added `test/e2e-oauth/`: Keycloak (TLS, realm import) + two Solace brokers with OAuth profiles configured, driving 6 scenarios over the real MCP JSON-RPC wire with JWTs minted from a live Keycloak (no mocks, no static dev token): admin path, cache hit, cache invalidation on 401, insufficient permission, audience isolation, wrong-audience rejection.
- Added `make e2e-oauth-up` / `e2e-oauth` / `e2e-oauth-down` / `e2e-oauth-all` targets, matching the sibling suite conventions.
- Replaced the `e2e-oauth` CI job in `build-and-test.yml` (was Terraform + `test/oauth/`) to run the new suite; dropped the `hashicorp/setup-terraform` step.
- Deleted `test/oauth/` in the same PR (Terraform config, old realm/client setup, old test script) — its Hop-1 coverage is now just the first step of the new suite's admin-path scenario, so nothing is lost by retiring it.
- Updated `docs/authentication.md` and `test/integration/README.md` to remove stale references to the deleted suite.
- Removed the now-unused Terraform block from `.gitignore`.

## Testing

**Test Environment:**
- Go version: 1.25.0 (per `go.mod`)
- OS: Linux
- Broker version: `solace/solace-pubsub-standard:latest`; Keycloak `26.2.5`

**Tests Performed:**
- [x] E2E tests added/updated

**Test Coverage:**
- All 6 scenarios run against real containers (not mocked): admin path succeeds; 2 consecutive calls by the same user produce exactly 1 token-exchange event; poisoning then restoring a broker's required audience proves cache eviction + refresh; a readonly user's write-tool call is denied with the alias-tagged authorization message; a tool call against the second broker proves per-broker audience selection; a deliberately-misconfigured wrong-audience broker alias proves the broker rejects a well-formed-but-wrong-audience token.
- No `.go` files changed in this diff; `go test -race ./...` and `golangci-lint run` both run clean as a regression check.
- Full local timing: `make e2e-oauth-all` end to end (certs, bring-up, configure, build, run, teardown) in ~15–70s depending on image-cache warmth, against the acceptance criterion's 3-minute budget.

## Breaking Changes

**Impact:** None to shipped server behavior. `test/oauth/` (an internal CI test suite, not a public interface) is removed; nothing external depends on it.

**Migration Guide:** N/A

### Code Quality

### Security
- [x] No credentials in code — the one secret checked in (`realm-export.json`'s `mcp-server-client` secret) is a stable, test-realm-only value for a throwaway local/CI Keycloak instance, the same convention already used by every sibling e2e suite's static dev token in `.env`
- [ ] Followed secure logging rules — N/A, no `.go` files changed
- [ ] Credential-carrying types implement `slog.LogValuer` — N/A, no Go types added
- [x] No security vulnerabilities introduced (`golangci-lint run`: 0 issues)

### Testing
- [x] Error paths tested

### Documentation
- [ ] godoc comments added/updated for exported functions — N/A, no Go functions added

### Git & CI
- [ ] All commits are signed off (DCO: `git commit -s`) — **not yet done, needs an amend before merge if this repo enforces DCO**
- [ ] Branch is up to date with main (rebased if needed) — rebased once during development; `main` has moved further since with no file-level overlap (verified via `git merge-tree`), but worth a final rebase before merge
- [x] No merge conflicts
- [ ] CI checks passing (lint, build, test, E2E) — pending this PR's own CI run

- N/A — not a breaking change

N/A

## Additional Notes

Two real gaps were found in the reference manual-setup tooling only by running a genuinely fresh stack, not discoverable from docs: (1) a fresh Solace broker has its SEMP-TLS port enabled but no certificate installed, so a cert install step (`configure-oauth-profiles.sh`) was added; (2) Keycloak's standard token-exchange grant needs an actual registered client behind the second broker's audience, separate from the custom-audience protocol mapper — added to this suite's realm export. Both are noted in `test/e2e-oauth/README.md` for future reference.

## Related Issues/PRs

- Related to # SOL-148417 (parent epic — Enterprise-Ready Broker MCP)
- Depends on # 170, # 173 (both already merged — this suite exercises behavior from both)


test/e2e-oauth/test-oauth-scenarios.sh
```diff
@@ -0,0 +1,152 @@
...
+    token=$(mint_token "$TEST_ADMIN_USER" "$TEST_USER_PASSWORD") || return 1
+    args=$(jq -nc --arg b "$BROKER_A_ALIAS" '{broker:$b,maxResults:500}')
+
+    before=$(count_token_exchanges)
+    mcp_call_tool_as "$token" "list-vpns" "$args" >/dev/null || { log_fail "first call failed"; return 1; }
+    mcp_call_tool_as "$token" "list-vpns" "$args" >/dev/null || { log_fail "second call failed"; return 1; }
+    after=$(count_token_exchanges)
+    delta=$((after - before))
+
+    if [ "$delta" -ne 1 ]; then
```
solace-aross, line 68: This asserts exactly one exchange from a single immediate read of the Keycloak log, and nothing waits for that log line to arrive. The `TOKEN_EXCHANGE` event travels JVM stdout, container stdout, then docker's json-file driver before `count_token_exchanges` can see it via `docker logs`. On a loaded runner the line from call 1 can still be in flight when `after` is sampled, so `delta` comes back 0 and the test fails on a perfectly healthy cache. Exact-equality with no retry is the most fragile point in the suite.

Suggest polling: after call 1, wait (bounded, ~5s) for the count to rise by one, then make call 2, then assert the count is stable at +1. Keeps the exactly-once check without racing the flush.

The cache logic itself is right (process-wide key on subject token + broker alias, so the two sessions share the entry). Only the observation is racy.

test/e2e-oauth/test-oauth-scenarios.sh
```diff
@@ -0,0 +1,152 @@
...
+    # for the real audience) fails broker-side validation on next use.
+    upsert_profile "$BROKER_A_SEMP_PORT" "poisoned-audience-temp" || { log_fail "failed to poison profile"; return 1; }
+
+    # Exercise the poisoned state — outcome intentionally not asserted.
+    mcp_call_tool_as "$token" "list-vpns" "$args" >/dev/null 2>&1 || true
+
+    # Restore.
+    upsert_profile "$BROKER_A_SEMP_PORT" "$BROKER_A_AUDIENCE" || { log_fail "failed to restore profile"; return 1; }
+
+    if ! mcp_call_tool_as "$token" "list-vpns" "$args" >/dev/null; then
```
solace-aross: You check success once here, no retry. `upsert_profile`'s read-after-write loop only confirms the SEMP config store reports the restored audience, not that the broker's OAuth enforcement has reloaded it. If enforcement is briefly still on `poisoned-audience-temp` when this call lands, it rejects and the test fails on healthy code. Wrap this in a short bounded retry, same shape as the profile loop.

test/e2e-oauth/run-all.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Redundant with the Makefile's own pre-flight wait (safety net for standalone
+# invocation — matches e2e-common/start-server.sh's own internal re-wait).
+check_build_deps
+wait_for_all_brokers 60
+wait_for_keycloak 60
+
+build_server
+
+CONFIG_FILE=$(mktemp /tmp/e2e-oauth-config-XXXXXX.yaml)
```
solace-aross: `mktemp /tmp/e2e-oauth-config-XXXXXX.yaml` only substitutes the `X`s on GNU mktemp. BSD mktemp (macOS) requires the `X`s to be last; with `.yaml` after them it creates a literal `...-XXXXXX.yaml` with no randomization (confirmed on macOS 26: exit 0, no substitution). Two local runs then collide on one predictable path, and the leftover literal file lingers. The README points local macOS devs at `make e2e-oauth`, so this will bite. Use `mktemp "${TMPDIR:-/tmp}/e2e-oauth-config-XXXXXX"` and append `.yaml`, or `mktemp -t`.

test/e2e-oauth/helpers.sh
```diff
@@ -0,0 +1,431 @@
...
+# browser-based Authorization Code + PKCE flow.
+mint_token() {
+    local username="$1" password="$2"
+    local response token
+    response=$(curl -sf --cacert "$KEYCLOAK_CERT" -X POST "$KEYCLOAK_TOKEN_ENDPOINT" \
+        -H "Content-Type: application/x-www-form-urlencoded" \
+        -d "grant_type=password" \
+        -d "client_id=${HOP1_CLIENT_ID}" \
+        -d "username=${username}" \
+        -d "password=${password}")
```
solace-aross: `mint_token` puts the password on the curl command line via `-d password=...`, which is visible in `ps`/`/proc` to other users on the host. That's the exact pattern `semp_curl` was built to avoid with `curl -K -`. It's only the test password for a public client, so low, but it breaks the off-argv convention the rest of the suite follows. Feeding the form body on stdin (`--config -` / `--data-urlencode @-`) keeps it consistent.

solace-aross (changes requested):
Strong, genuinely end-to-end, and unusually well-documented. The header comments explaining why the helpers weren't folded into the shared lib, and the two gaps you found by running a fresh stack (broker cert install, second registered client), are exactly the kind of thing that saves the next person hours. No shipping code changes, so blast radius is CI-only.

My comments are all about test reliability, not design. I'm marking request-changes on two of them because they break healthy code, not because the approach is wrong:

- The cache-hit test will intermittently red CI (see `test-oauth-scenarios.sh`).
- The `mktemp` template deterministically breaks `make e2e-oauth` on macOS, which the README points local devs at.

Everything else is comment-level. Happy to re-review quickly once the two are addressed.

On the two Copilot comments, I agree with both:
- Realm signing keys in `realm-export.json`: worth stripping the `KeyProvider` block. The suite mints everything at runtime and depends on no pre-signed material, so Keycloak generates fresh keys on import at zero test cost, and a token-forgery key leaves the repo. The risk is realm reuse, not this disposable run.
- `-u admin:admin` in the failure-snapshot step: route it through `semp_curl` like the rest of the suite. Those snapshot curls also hardcode `8102`/`8104` instead of reading `.env`, so the switch fixes the argv exposure and the port coupling in one move.

One nit I'm not raising inline: `set -e` is suspended inside every scenario function (they run as an `if` condition), so a future unguarded command won't fail the test. The current scenarios are all explicitly guarded, so no action needed, just worth knowing.

bczoma (approved):
Looks good to me

solace-aross (approved):
Re-reviewed at 421af71. Every item from the last round is addressed, so this is good to merge.

Both blocking findings:
- macOS mktemp: fixed with the get-random-name-then-add-.yaml shape, and the cleanup trap tracks the final path.
- Cache-hit flake: the bounded poll for the exchange count to rise before the second call closes the race while keeping the exactly-once assertion. Right fix.

The rest:
- Audience restore now has the bounded retry, so the enforcement-reload gap is absorbed.
- mint_token feeds the password via --data-urlencode password@-, off argv.
- The CI failure-snapshot uses semp_curl now, so no credentials on argv and no hardcoded ports.
- The KeyProvider signing keys are stripped from realm-export.json. The one remaining secret is the mcp-server-client client secret, which the suite needs and which grants nothing outside the disposable local realm. No concern.

CI is green across all the E2E suites. Nice work running a fresh stack to surface the cert-install and second-client gaps.

PR #186, opened 2026-07-17
SOL-151978: Enforce tool authorization at request time via a wrapper composed inside withRecovery

amitmorade:
## Summary

Adds the request-time enforcement layer for claims-based tool authorization. An OAuth-authenticated caller's `groups` claim is now checked against a compiled `*authz.Policy` before each tool call. Denials return a uniform tool-level error result; the tool never runs. `list-brokers` is structurally exempt so every authenticated caller keeps discovery. When tool authorization is off, dispatch is byte-identical to the pre-RBAC path.

Part of [SOL-151440](https://sol-jira.atlassian.net/browse/SOL-151440) (Claims-Based Tool Authorization), ticket [SOL-151978](https://sol-jira.atlassian.net/browse/SOL-151978).

## Type of Change

## Motivation and Context

Prior to this PR, any authenticated OAuth caller could invoke any tool the server exposed — including destructive ones (`delete-queue`, `disconnect-client`, etc.). The parent feature SOL-151440 closes that gap by introducing group-based authorization derived from an OIDC `groups` claim. T1 landed the config schema, T2 landed the leaf `internal/authz` decision engine, T3 landed the middleware that extracts the groups claim onto `TokenInfo.Extra`. This PR wires those pieces into request dispatch: a wrapper reads the groups off the request, asks the compiled policy for a decision, and either passes through to the tool handler or returns a uniform "not authorized" tool-level error result.

**How it works at request time:**

```
Client sends tool call over MCP
    ↓
withRecovery (outer — panics, correlation ID)
    ↓
withAuthorization  ← reads req.Extra.TokenInfo.Extra["groups"]
    │             ← calls policy.Authorize(groups, toolName)
    │             ← emits INFO audit line (allowed/denied/denied_missing_claim)
    ↓
tool handler runs OR uniform deny result returned
```

**How it works at startup:**

```
buildToolPolicy(cfg)
    │
    ├─ gate off → returns (nil, nil)
    └─ gate on  → returns (compiled Policy, nil)
    ↓
fail-closed guard: gate on AND policy nil → slog.Error + os.Exit(1)
    ↓
RegisterWithServer(..., policy)
    ├─ policy == nil → skips withAuthorization; dispatch byte-identical to pre-RBAC
    └─ policy != nil → composes withAuthorization inside withRecovery on every tool
    ↓
RegisterListBrokers(server, pool)  ← never receives policy; structurally exempt
    ↓
ValidatePolicyToolNames(cfg, mgr)  ← catches YAML typos; fatal on unknown tools
    ↓
Startup INFO announcement naming the RBAC posture on both enabled and disabled paths
```

## Design rationale — three-layer separation of concerns

The extraction / plumbing / interpretation split across the three components in this feature is a load-bearing discipline. Recording it here explicitly so a future contributor understands which layer owns which failure-handling responsibility, and doesn't drift one layer's job into another.

Three distinct jobs, three distinct owners:

1. **Resolver** (`ResolveGroups` in `internal/auth/claims.go`) — structural detector. Given a decoded claims map and a claim name, it either produces a usable `[]string` or signals it can't. It owns *only that signal*. It does not know why any caller cares, does not shape any diagnostic, and does not decide what to do next.

2. **Middleware** (`createOIDCTokenVerifier` in `internal/auth/middleware.go`) — claim-plumbing carrier. Runs the detector, decides whether the outcome is worth recording on the request (structural anomalies get a WARN), and attaches the outcome to `TokenInfo.Extra` for the next layer to read. It does not interpret. It does not shape user-facing responses. It does not decide policy.

3. **Enforcement wrapper** (`withAuthorization` in `internal/tools/authorization.go`) — policy interpreter. It reads the plumbing off the request, applies the compiled policy (`Policy.Authorize`), and shapes the two observable outcomes: the audit event on the server log side and the caller-facing response on the wire side (uniform "not authorized" string, no configuration metadata).

Consequences worth defending against drift:

- **The resolver's signal is not diagnostic surface.** Its one-bit answer is consumed by machine (the middleware's `if !ok` branch); if a future contributor is tempted to enrich its message with the configured claim name or similar, that's a layering violation — claim-name context is the wrapper's diagnostic input, not the resolver's output.
- **The middleware does not shape user-facing behavior.** If a future contributor is tempted to make the middleware return 401 on `!ok`, that's a layering violation — the token *is* authenticated (verifier already ran), just not carrying the groups claim. Token validity and authorization are different concerns.
- **The wrapper owns the "shout."** The loud, actionable diagnostic (which claim was missing, which correlation ID to grep, which subject) is emitted by the wrapper's audit event, not by the resolver or middleware.

**Test this layering by asking:** would a change to the caller-facing wording touch the resolver? No — it touches the wrapper. Would a change to the resolver's coalescing rules touch the wrapper? No — only the resolver's edge-case tests. Would a change to which structured fields appear in the audit event touch the middleware? No — only the wrapper's audit call site. Three orthogonal axes of change.

## Design Decisions

**Why the wrapper composes INSIDE `withRecovery`, not outside:**
The composition order is `withRecovery(name, withAuthorization(policy, name, next))`. If reversed, three concrete failures appear. First, a panic anywhere inside the authorization wrapper (e.g. a future bug in the audit-emission code) would escape uncaught and crash the goroutine — the SDK has no recover of its own. Second, deny results would not get the correlation ID stamp `withRecovery` applies, so a caller receiving a "not authorized" message would have no correlation ID to hand to an admin for the log lookup. Third, the audit event fires before `next` runs on the allow path; if `withAuthorization` sat outside, an allow followed by a tool-handler panic would produce an "allowed" audit line *and* a recovered error, and reviewers might interpret the pairing as a successful allow followed by a separate failure rather than one request that never completed. Current nesting means allow/deny/panic all get uniform correlation and containment.

**Why `RegisterWithServer` uses `policy != nil` as the "RBAC on" signal instead of a separate boolean:**
The two facts (RBAC enabled, compiled policy exists) are perfectly coupled by the invariant `buildToolPolicy` enforces (gate enabled ⟺ policy non-nil). Encoding one bit of information as one value gives zero representable invalid states. The alternative I initially proposed — splitting into `enableAuthorization bool + policy *authz.Policy` — creates four states, two invalid (`{true, nil}` and `{false, non-nil}`), which converts a documentation problem into a synchronization problem: every caller now has to keep the two arguments consistent, including test scaffolding. The stdlib does the same thing (`http.Server.ErrorLog`, `tls.Config.VerifyPeerCertificate`) — nil-for-optional is the standard idiom when the single check-point is well-documented, which this one is.

**Why `buildToolPolicy` was extracted from `main` even though it's five lines:**
The invariant "gate enabled ⟺ policy non-nil" is the contract `RegisterWithServer` relies on to interpret its nil-policy path as "RBAC off." If that invariant were established inline in `main`, it would be an implicit convention scattered across startup wiring. Extracting to `buildToolPolicy` makes it a postcondition of a named function that tests can pin (`tool_policy_test.go` asserts both directions and the non-oauth-mode edge case). The pattern matches `buildRootHandler` and `buildMCPEndpoint` elsewhere in `main`, which the codebase adopted for exactly this reason — the assembled shape gets tested against real code rather than a hand-rebuilt copy in the test file.

**Why there's a fail-closed guard immediately after `buildToolPolicy` even though it's structurally unreachable today:**
```go
if config.ToolAuthorizationEnabled(cfg) && policy == nil {
    slog.Error("tool authorization is enabled in config but no policy was compiled — refusing to start rather than silently bypass authorization")
    os.Exit(1)
}
```
Under the current `buildToolPolicy` implementation the guard cannot fire — the function returns non-nil exactly when the gate is on. The guard defends against a future refactor that accidentally breaks the postcondition (e.g. someone adds an early-return branch in `buildToolPolicy` that returns `(nil, nil)` under a new condition). If that ever happens without the tests catching it, this guard converts a silent fail-open (every OAuth request bypasses authorization) into a loud, immediate startup crash naming the failure. SOL-149989 is the prior incident where a fail-open on the auth path went undetected for a period; this guard is proportionate defense-in-depth against that class.

**Why `authzDeniedMessage` and `authzMissingClaimMessage` are two constants with identical values in v1:**
The caller-facing distinction between "your groups don't grant this tool" and "your token has no groups claim at all" would be information disclosure — an unauthenticated-authorization caller learning that the server knows about a groups claim tells them the RBAC feature is enabled, and telling them their token is structurally incomplete is admin-configuration metadata they can't act on. Both branches return the same uniform string, so a caller cannot distinguish which fired. The audit log on the server side does distinguish (via the `decision` field: `denied` vs `denied_missing_claim`) so operators can tell the two failure classes apart. The two constants stay as separate identifiers so a future evolution can amend one without touching the other; `TestAuthzMessages_IdenticalForV1` locks the v1 identity constraint against accidental divergence.

**Why the deny result shape mirrors `withRecovery`'s panic branch:**
```go
&mcp.CallToolResult{
    StructuredContent: map[string]any{"error": msg, "retryable": false},
    Content:           []mcp.Content{&mcp.TextContent{Text: msg}},
    IsError:           true,
}
```
Same fields, same values, same `retryable: false` marker. A reviewer looking at any tool-level error result in this codebase sees one shape across authorization, panic recovery, and validation errors. `retryable: false` tells the agent client not to loop on the same denied call — retries won't change the outcome. Returning `(*CallToolResult, nil)` rather than `(nil, error)` means it's a tool-level error the agent surfaces to its user, not a JSON-RPC protocol error that would kill the session.

**Why `list-brokers` is exempt structurally (via a separate registration function) rather than as a runtime check inside the wrapper:**
`RegisterListBrokers` takes no `policy` argument and never composes `withAuthorization`. The exemption is expressed at the registration API's shape, meaning a YAML typo, a policy misconfiguration, or a bug in the wrapper cannot disable discovery. Every authenticated caller — regardless of group membership or claim structure — can list brokers and see which aliases the server knows about, which is the minimum discovery primitive the MCP contract requires. If the exemption lived inside `withAuthorization` as a name check (e.g. `if toolName == "list-brokers" { return next(...) }`), a refactor that split or renamed the wrapper could accidentally drop the check. Test `TestRegisterListBrokers_NeverComposesWithAuthorization` locks the current shape.

**Why `ValidatePolicyToolNames` collects unknown-tool errors via `errors.Join` with one row per unknown tool:**
Admin authors YAML in group-first form. A single typo (e.g. `delete-que-messages` instead of `delete-queue-messages`) referenced across three groups is *one* bug the admin needs to fix, not three. The function collects a map from unknown tool name to the set of referencing groups, sorts both dimensions alphabetically, and emits one row per unknown tool with the group list joined:
```
unknown tool "delete-que-messages" (referenced by groups: Contractors, Ops, Support)
```
Sorted output means the error is byte-identical across startup restarts — no dependency on Go's randomized map iteration order — so operators can diff startup logs across environments reliably. `list-brokers` grants are inert (the tool is exempt) but not errors; they surface as WARN so an admin who added `list-brokers` to a group intending some effect learns their grant has none.

**Why the audit event fires at INFO for every outcome in this PR (allow/deny/missing) rather than splitting levels:**
The audit-level split (deny at WARN for alerting, allow at INFO) requires refining the audit field schema (bounded matched-groups slice, correlation-ID field naming, structured decision codes) that this PR intentionally does not touch. Splitting audit refinement into its own ticket keeps this PR focused on the enforcement wiring and lets reviewers see one concern at a time. This PR's audit line is a v1 placeholder: INFO level always, minimal field set (identity, tool, decision string). A follow-up ticket refines. Tests assert on distinguishability of decision values rather than the literal schema so the follow-up doesn't churn these tests.

## Changes Made

- **`internal/tools/authorization.go`** (new):
  - `withAuthorization(policy, toolName, next) mcp.ToolHandler` — the enforcement wrapper. Reads groups via `requestGroups`, calls `policy.Authorize`, emits the decision audit line, dispatches to `next` on allow or returns a uniform tool-level error on deny/missing-claim.
  - `authzDeniedMessage` and `authzMissingClaimMessage` constants — identical strings for v1; separate identifiers for future evolution.
  - `authzErrorResult(message)` — builds the uniform deny/missing `*CallToolResult` (IsError:true, retryable:false, StructuredContent + TextContent).
  - `ValidatePolicyToolNames(cfg, mgr)` — startup check that every tool name in `AccessLevelGroups` resolves to a registered handler or is `list-brokers`. Alphabetized, deduped, one row per unknown tool via `errors.Join`. WARN on list-brokers grants (inert but noticeable).
  - `listBrokersToolName` constant — names the exemption target so a grep finds every touchpoint.

- **`internal/tools/register.go`**:
  - `RegisterWithServer` gains a `policy *authz.Policy` parameter.
  - Tool-registration loop composes `withRecovery(name, withAuthorization(policy, name, closure))` when policy is non-nil; keeps the pre-RBAC `withRecovery(name, closure)` shape when nil.
  - `RegisterListBrokers` is unchanged — signature deliberately does not accept a policy so exemption is enforced at compile time.
  - Godoc comments refreshed to reflect the new composition and the exemption.

- **`cmd/server/main.go`**:
  - New `buildToolPolicy(cfg) (*authz.Policy, error)` — single testable decision point returning `(nil, nil)` when gate is off, `(policy, nil)` when gate is on and compilation succeeds, `(nil, err)` on compilation failure.
  - Composition-time gate wires the flow: `buildToolPolicy` → fail-closed guard → `RegisterWithServer(..., policy)` → `RegisterListBrokers` → `ValidatePolicyToolNames` (when gate is on).
  - Startup INFO announcement naming the RBAC posture on both enabled and disabled paths, with reason on disabled (auth mode vs feature flag).
  - `slog.Error` + `os.Exit(1)` on any startup failure — partial RBAC state must never reach the request path.

- **`internal/tools/authorization_test.go`** (new): 12 wrapper unit tests.
- **`internal/tools/authorization_validate_test.go`** (new): 8 startup-validator tests.
- **`internal/tools/register_test.go`**: 3 composition-site integration tests appended (drive real client-server calls via `mcp.NewInMemoryTransports`).
- **`cmd/server/tool_policy_test.go`** (new): 3 `buildToolPolicy` postcondition tests.

## Testing

**Test Environment:**
- Go version: 1.26.0
- OS: darwin/arm64

**Tests Performed:**
- [x] Unit tests added
- [x] Integration-style composition-site tests added
- [x] `go test -race ./internal/tools/...` and `go test -race ./cmd/server/...` pass
- [x] Full `make check` passes

**Test Coverage — `withAuthorization` wrapper (12 tests):**

Allow path:
- `next` handler is called exactly once and its result is returned pointer-identical
- INFO `"tool authorization"` audit line fires with `tool` name and a `decision` value distinguishable from deny/missing values

Deny path (caller has groups, but none grant the tool):
- Returns `(*CallToolResult, nil)` — tool-level error, not JSON-RPC error
- `IsError: true` on the result
- `StructuredContent` carries `error = authzDeniedMessage` and `retryable = false`
- `Content` contains one `*TextContent` whose `Text` equals `authzDeniedMessage`
- `next` is NOT called
- INFO audit line fires with `decision` value distinguishable from allow and missing-claim

Missing-claim path (caller's token has no groups key in `TokenInfo.Extra`):
- Same result shape as deny, but `error = authzMissingClaimMessage`
- `next` is NOT called
- INFO audit line fires with `decision` distinguishable from plain deny (operator can tell IdP misconfig from grant miss)

Edge cases:
- Nil `req.Extra` → treated as missing-claim (test scaffolding safety)
- Nil `req.Extra.TokenInfo` → treated as missing-claim (disabled-mode safety)
- `authzDeniedMessage == authzMissingClaimMessage` — v1 identity constraint locked; also asserts both are non-empty
- Nil `*authz.Policy` passed to wrapper → panics on first call (precondition violation is loud, not silent — outer `withRecovery` catches it)

**Test Coverage — `ValidatePolicyToolNames` (8 tests):**

Happy path:
- All configured tool names resolve → `err == nil` AND zero log output (silent no-op)

Error return shape:
- One unknown tool referenced by one group → single error row naming both identifiers
- Same unknown tool referenced across three groups → one deduped row, group names alphabetized
- Three distinct unknown tools → three rows, alphabetized by tool name (deterministic across restarts)
- Unknown tool AND `list-brokers` grant on the same config → both surface independently (error return + WARN log)

`list-brokers` handling:
- Grant of `list-brokers` alone → `err == nil`, exactly one WARN log line naming the referencing group
- Case-mismatched `List-Brokers` → treated as unknown tool (unknown-tool error, NO list-brokers WARN); prevents silent case-folding drift on the exempt name

Empty inputs:
- Nil or empty `AccessLevelGroups` → `err == nil`, zero log output

**Test Coverage — composition-site wiring (3 tests, full MCP SDK loop):**

- Nil policy → dispatch is byte-identical to pre-RBAC: existing `"tool invoked"` audit line still fires, no `"tool authorization"` line appears
- Non-nil policy → wrapper is composed and consulted: caller receives `IsError:true` (missing-claim branch under no-middleware test scaffold), exactly one `"tool authorization"` audit line fires, decision value is consistent with the deny result (doesn't contain "allow")
- `RegisterListBrokers` under a policy-enabled server where the policy grants nothing for `list-brokers` → call succeeds (`IsError:false`), no `"tool authorization"` line fires (structural exemption verified end-to-end)

**Test Coverage — `buildToolPolicy` postcondition (3 tests):**

- Feature flag unset → `(nil, nil)` regardless of RBAC block contents in cfg (gate-off is total)
- Feature flag set + valid RBAC block + oauth mode → non-nil `*Policy`, `err == nil`
- Feature flag set + non-oauth mode → `(nil, nil)` (gate still off; tool authorization requires OAuth for identity)

**Test discipline notes:**
- No test asserts on the literal `decision` string schema. Assertions are "does not contain 'allow'" or "differs from 'denied'" so the audit-refinement follow-up ticket can rename fields without churning these tests.
- No test uses mocks for `*authz.Policy`. Real policies are constructed via `authz.NewPolicy`, so drift between wrapper and policy surfaces at test time.
- Composition-site tests drive real SDK transport (`mcp.NewInMemoryTransports`), not mocked wiring — regressions in the SDK's `RequestExtra` population would be caught.

## Independent review passes

Ran two independent principal-architect review passes on the branch — an architecture/design lens and a security lens — before opening this PR.

**Architecture review** (design decisions, layering, dependency direction, testability): overall verdict **Approve as-is**. All eight assessed decisions returned "Solid — no change needed" except the fail-closed guard, marked "Marginal — worth a follow-up comment but not blocking." No structural changes recommended.

**Security review** (SECURE framework — fail-open detection, silent bypass, information disclosure, panic containment): overall verdict **Ship with follow-ups**. No Critical or High findings. Three Low/Informational items surfaced for future tickets, none blocking:
- Add wrapper-level test for `[]string{}` empty-but-present groups edge case (belt-and-suspenders; behavior is correct today)
- Add `authz.Authorize` case-mismatch test for group names and document case-sensitivity in operator guide (documentation gap, not vulnerability)
- Consider whether the nil-policy panic path (unreachable in production due to startup guard) should be moved to a `defer`-based audit for zero-gap coverage (code hygiene)

Both reviewers praised: the three-layer separation actually honored in code (not just docs), the `policy != nil` convention + fail-closed guard, the leaf-package split of `authz`, the composition order, and the uniform-message + separate-constant pattern with `TestAuthzMessages_IdenticalForV1` as the drift guard.

## Breaking Changes

None. `RegisterWithServer` gained a new parameter, but this function is called from exactly one call site (`cmd/server/main.go`) which this PR also updates. No external consumers.

## What is deliberately not in this PR

- **Audit event refinement.** Level split (deny at WARN), bounded matched-groups slice, full field schema — separate follow-up ticket.
- **E2E OAuth flow tests.** Separate follow-up ticket owns the real-broker OAuth end-to-end coverage.
- **Admin-facing documentation for the RBAC config block.** Separate follow-up ticket.
- **`RegisterWithServer` → `RegisterAuthProtected` rename.** Was originally in scope, then descoped mid-implementation to keep this PR focused on the security-sensitive wrapper wiring rather than a codebase-wide naming refactor.

### Code Quality
- [x] Self-review completed
- [x] Comment scan for workflow-internal labels — three found and removed in a follow-up commit on this branch

### Security
- [x] Followed secure logging rules — `/check-logs` scan passed
- [x] `authzDeniedMessage` and `authzMissingClaimMessage` are identical in v1; locked by test
- [x] `authz.Policy` and `authz.Decision` have `LogValue` implementations (from T2) that emit counts only, never group/tool names
- [x] Fail-closed guard defends the "gate on ⟹ policy non-nil" invariant against future refactor
- [x] `list-brokers` exemption is structural (registration API shape), not a runtime check that can be bypassed
- [x] Independent principal-level security review pass completed — no Critical/High findings

### Testing
- [x] Edge cases covered (nil TokenInfo, empty groups, unknown tools, empty tool names, case-mismatched exempt name)
- [x] Security-critical invariants explicitly locked: "next NOT called on deny", "next NOT called on missing-claim", "nil policy panics (no silent bypass)"
- [x] Test names describe observable behavior, not implementation

### Documentation
- [x] godoc comments added for all exported symbols in `authorization.go`
- [x] `RegisterWithServer` godoc updated to describe the new `policy` parameter and nil-vs-non-nil semantics
- [x] `buildToolPolicy` godoc names the invariant it enforces and the failure classes it protects against

### Git & CI
- [x] Commits have clear, descriptive messages (four commits: implementation, invariant refactor, tests, comment hygiene)
- [x] No merge conflicts

## Related Issues/PRs

- Part of epic [SOL-148417](https://sol-jira.atlassian.net/browse/SOL-148417) (Claims-Based Tool Authorization)
- Depends on:
  - #178 — T1: config schema + validation
  - #180 — T2: `internal/authz` decision engine
  - #183 — T3: middleware groups extraction
- Blocks: follow-up audit-refinement ticket, follow-up E2E OAuth ticket, follow-up admin-docs ticket

---

**For Reviewers:**

**Focus Areas:**
- **Fail-closed posture.** `buildToolPolicy` postcondition + the guard immediately after + the wrapper's nil-policy panic form a three-layer defense against silent RBAC bypass. If a bug in any of them makes an OAuth request quietly skip authorization, that's the disaster case this PR must prevent.
- **`list-brokers` exemption.** Verified by `TestRegisterListBrokers_NeverComposesWithAuthorization`. Is there a code path where a `list-brokers` request could still hit the wrapper?
- **Caller-side information disclosure.** Deny and missing-claim must be indistinguishable to the caller. Any observable difference (timing, error shape, extra fields, message wording) between the two branches would leak the claim-structure state to an unauthenticated-authorization caller.
- **Composition order.** `withAuthorization` inside `withRecovery` — reversing this breaks panic containment and correlation-ID stamping. Test-locked, but the ordering is the kind of thing that could regress in a well-meaning refactor.

**Questions:**
- The audit event's `decision` string values (`"allowed"` / `"denied"` / `"denied_missing_claim"`) are v1 placeholder. Tests assert on distinguishability, not literals. Is there a preferred schema convention I should match for the follow-up ticket, or is that up to the audit-refinement work?
- The `list-brokers` WARN in `ValidatePolicyToolNames` is one WARN per config-load, not one per grant (dedup on the exempt tool name). Is that the right calibration, or would per-grant WARNs be more actionable for a busy admin?

## Note on the DCO checklist item

Per team convention I do not sign commits with `-s`. Leaving that checkbox unchecked deliberately; happy to discuss if the DCO enforcement bot flags it.

[SOL-151440]: https://sol-jira.atlassian.net/browse/SOL-151440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-151978]: https://sol-jira.atlassian.net/browse/SOL-151978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-148417]: https://sol-jira.atlassian.net/browse/SOL-148417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -470,29 +470,17 @@ func newTokenExchanger(oauthCfg *config.BrokerOAuthConfig) (*tokenexchange.Excha
 	return exchanger, nil
 }
 
-// buildToolPolicy is the single, testable decision point for whether the
-// server should run with tool authorization and, if so, which compiled
-// Policy to use. Extracted from main so the invariant "gate enabled ⟺
-// returned policy is non-nil" is a postcondition of a named function
-// rather than a convention scattered across startup wiring — same
-// technique used for buildRootHandler and buildMCPEndpoint so the
-// assembled shape is asserted against real code rather than a hand-
-// rebuilt copy.
+// buildToolPolicy is the single, tested decision point for whether the server
+// runs with tool authorization and which compiled Policy to use. Extracted so
+// the invariant "gate enabled ⟺ returned policy is non-nil" is a
+// postcondition of a named function rather than folklore scattered across
+// startup.
 //
-// Return contract:
-//
-//   - Gate off (config.ToolAuthorizationEnabled(cfg) == false): returns
-//     (nil, nil). RegisterWithServer receives nil and skips the wrapper.
-//   - Gate on, NewPolicy succeeds: returns (policy, nil). RegisterWithServer
-//     composes withAuthorization on every registered tool.
-//   - Gate on, NewPolicy fails: returns (nil, err). Caller must fail
-//     startup — a partial RBAC state must never reach the request path.
-//
-// A non-nil policy is never returned when the gate is off. main's
-// fail-closed guard immediately after this call asserts the mirror
-// invariant (gate on ⟹ policy non-nil) so any future refactor that
-// breaks this function's contract aborts startup rather than silently
-// bypasses authorization on every OAuth request.
+// Returns (nil, nil) when the gate is off, (policy, nil) on success,
+// (nil, err) on compilation failure — callers must fail startup on the last.
+// Main's fail-closed guard immediately after this call asserts the mirror
+// direction so any future refactor that breaks the postcondition aborts
+// startup rather than silently bypassing authorization.
 func buildToolPolicy(cfg *config.ServerConfig) (*authz.Policy, error) {
 	if !config.ToolAuthorizationEnabled(cfg) {
 		return nil, nil
@@ -685,14 +673,10 @@ func main() {
 	registerMixedTools(mgr)
 
 	// Compile the tool-authorization policy. buildToolPolicy owns the
-	// gate-and-compile decision so the "enabled ⟺ policy non-nil"
-	// invariant is a named, tested postcondition of one function rather
-	// than a convention scattered across main. RegisterWithServer still
-	// receives a *authz.Policy that is nil iff RBAC is off; the fail-
-	// closed check below guards against a future refactor breaking the
-	// invariant (e.g. gate flipped on but constructor accidentally
-	// skipped) — silent fail-open on a security switch is exactly the
-	// SOL-149989 failure class this feature exists to prevent.
+	// gate-and-compile decision so the "enabled ⟺ policy non-nil" invariant
+	// is a tested postcondition of one function. The guard below defends
+	// against a future refactor breaking that postcondition — silent
+	// fail-open on a security switch is the SOL-149989 failure class.
 	policy, err := buildToolPolicy(cfg)
 	if err != nil {
 		slog.Error("tool authorization startup failed", slog.String("error", err.Error()))
@@ -707,23 +691,15 @@ func main() {
 	slog.Info("tool registration complete",
 		slog.Bool("enable_write_tools", cfg.EnableWriteTools))
 
-	// list-brokers is a discovery tool registered directly on the MCP
-	// server (it doesn't need broker resolution or the ToolManager pipeline).
-	// It also does not compose withAuthorization — the exemption is
-	// structural: RegisterListBrokers takes no policy argument and every
-	// authenticated caller keeps discovery regardless of the RBAC state.
+	// list-brokers is registered directly (no broker resolution needed) and
+	// takes no policy — the RBAC exemption is expressed structurally at this
+	// API surface.
 	tools.RegisterListBrokers(server, pool)
 
-	// Validate every configured tool name against the full registered
-	// handler set + list-brokers, now that both registrations have
-	// populated it. A typo in the admin's YAML would silently fail to
-	// grant what the admin thinks — an admin intending to grant
-	// delete-queue-messages who writes delete-que-messages would see
-	// the grant simply not take effect at request time, with no signal
-	// anywhere that the config is wrong. Catching that at startup is
-	// fatal by design: refuse to accept any request under a config
-	// whose grants do not mean what the admin thinks they mean.
-	// Skipped when RBAC is off.
+	// Validate every configured tool name now that both registrations have
+	// populated mgr. An admin typo would silently produce a grant that never
+	// takes effect at request time; catching it at startup is fatal by
+	// design. Skipped when RBAC is off.
 	if policy != nil {
 		if err := tools.ValidatePolicyToolNames(*cfg.MCPClientAuth.ToolAuthorization, mgr); err != nil {
 			slog.Error("tool authorization startup failed", slog.String("error", err.Error()))
@@ -732,13 +708,10 @@ func main() {
 		slog.Info("tool authorization is enabled",
 			slog.Any("policy", policy))
 	} else {
-		// Symmetric announcement: name the reason so a restart's startup
-		// log unambiguously records the RBAC posture. The config
-		// validator rejects a tool_authorization block outside oauth
-		// mode and rejects an oauth-mode block that omits the enabled
-		// key, so at this point exactly two disabled-paths are
-		// possible: auth mode is not oauth (block absent by contract),
-		// or auth mode is oauth and enabled is explicitly false.
+		// Announce the RBAC posture with the reason so startup logs are
+		// unambiguous. Config validator rejects a block outside oauth mode
+		// and requires enabled to be set in oauth mode, so exactly two
+		// disabled paths remain: non-oauth mode, or oauth with enabled:false.
 		if cfg.MCPClientAuth.Mode == config.AuthModeOAuth {
 			slog.Info("tool authorization is disabled (enabled=false in config)")
 		} else {
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -470,6 +470,36 @@ func newTokenExchanger(oauthCfg *config.BrokerOAuthConfig) (*tokenexchange.Excha
 	return exchanger, nil
 }
 
+// buildToolPolicy is the single, testable decision point for whether the
+// server should run with tool authorization and, if so, which compiled
+// Policy to use. Extracted from main so the invariant "gate enabled ⟺
+// returned policy is non-nil" is a postcondition of a named function
+// rather than a convention scattered across startup wiring — same
+// technique used for buildRootHandler and buildMCPEndpoint so the
+// assembled shape is asserted against real code rather than a hand-
+// rebuilt copy.
+//
+// Return contract:
+//
+//   - Gate off (config.ToolAuthorizationEnabled(cfg) == false): returns
+//     (nil, nil). RegisterWithServer receives nil and skips the wrapper.
+//   - Gate on, NewPolicy succeeds: returns (policy, nil). RegisterWithServer
+//     composes withAuthorization on every registered tool.
+//   - Gate on, NewPolicy fails: returns (nil, err). Caller must fail
+//     startup — a partial RBAC state must never reach the request path.
+//
+// A non-nil policy is never returned when the gate is off. main's
+// fail-closed guard immediately after this call asserts the mirror
+// invariant (gate on ⟹ policy non-nil) so any future refactor that
+// breaks this function's contract aborts startup rather than silently
+// bypasses authorization on every OAuth request.
+func buildToolPolicy(cfg *config.ServerConfig) (*authz.Policy, error) {
+	if !config.ToolAuthorizationEnabled(cfg) {
+		return nil, nil
+	}
+	return authz.NewPolicy(*cfg.MCPClientAuth.ToolAuthorization)
+}
+
 func main() {
 	if len(os.Args) == 2 && (os.Args[1] == "-version" || os.Args[1] == "--version") {
 		fmt.Println(version.Version())
@@ -654,18 +684,23 @@ func main() {
 	registerSEMPv1Tools(mgr)
 	registerMixedTools(mgr)
 
-	// Compile the tool-authorization policy when the composition-time gate
-	// says RBAC is on. NewPolicy runs BEFORE RegisterWithServer because the
-	// wrapper it composes needs the compiled *Policy. When the gate is off,
-	// policy stays nil and RegisterWithServer skips the wrapper — dispatch
-	// is byte-identical to the pre-RBAC path.
-	var policy *authz.Policy
-	if config.ToolAuthorizationEnabled(cfg) {
-		policy, err = authz.NewPolicy(*cfg.MCPClientAuth.ToolAuthorization)
-		if err != nil {
-			slog.Error("tool authorization startup failed", slog.String("error", err.Error()))
-			os.Exit(1)
-		}
+	// Compile the tool-authorization policy. buildToolPolicy owns the
+	// gate-and-compile decision so the "enabled ⟺ policy non-nil"
+	// invariant is a named, tested postcondition of one function rather
+	// than a convention scattered across main. RegisterWithServer still
+	// receives a *authz.Policy that is nil iff RBAC is off; the fail-
+	// closed check below guards against a future refactor breaking the
+	// invariant (e.g. gate flipped on but constructor accidentally
+	// skipped) — silent fail-open on a security switch is exactly the
+	// SOL-149989 failure class this feature exists to prevent.
+	policy, err := buildToolPolicy(cfg)
+	if err != nil {
+		slog.Error("tool authorization startup failed", slog.String("error", err.Error()))
+		os.Exit(1)
+	}
+	if config.ToolAuthorizationEnabled(cfg) && policy == nil {
+		slog.Error("tool authorization is enabled in config but no policy was compiled — refusing to start rather than silently bypass authorization")
+		os.Exit(1)
 	}
 
 	tools.RegisterWithServer(mgr, server, pool, cfg.EnableWriteTools, policy)
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -717,8 +717,13 @@ func main() {
 	// Validate every configured tool name against the full registered
 	// handler set + list-brokers, now that both registrations have
 	// populated it. A typo in the admin's YAML would silently fail to
-	// grant what the admin thinks; catching it at startup is fatal by
-	// design (spec goal 7). Skipped when RBAC is off.
+	// grant what the admin thinks — an admin intending to grant
+	// delete-queue-messages who writes delete-que-messages would see
+	// the grant simply not take effect at request time, with no signal
+	// anywhere that the config is wrong. Catching that at startup is
+	// fatal by design: refuse to accept any request under a config
+	// whose grants do not mean what the admin thinks they mean.
+	// Skipped when RBAC is off.
 	if policy != nil {
 		if err := tools.ValidatePolicyToolNames(*cfg.MCPClientAuth.ToolAuthorization, mgr); err != nil {
 			slog.Error("tool authorization startup failed", slog.String("error", err.Error()))
@@ -728,10 +733,12 @@ func main() {
 			slog.Any("policy", policy))
 	} else {
 		// Symmetric announcement: name the reason so a restart's startup
-		// log unambiguously records the RBAC posture. When the auth mode
-		// is not oauth the block cannot be present (config validator I1);
-		// when the mode is oauth the block must set enabled explicitly
-		// (I3), so the only remaining false path is enabled: false.
+		// log unambiguously records the RBAC posture. The config
+		// validator rejects a tool_authorization block outside oauth
+		// mode and rejects an oauth-mode block that omits the enabled
+		// key, so at this point exactly two disabled-paths are
+		// possible: auth mode is not oauth (block absent by contract),
+		// or auth mode is oauth and enabled is explicitly false.
 		if cfg.MCPClientAuth.Mode == config.AuthModeOAuth {
 			slog.Info("tool authorization is disabled (enabled=false in config)")
 		} else {
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/SolaceDev/solace-broker-mcp/internal/auth"
+	"github.com/SolaceDev/solace-broker-mcp/internal/authz"
 	"github.com/SolaceDev/solace-broker-mcp/internal/banner"
 	"github.com/SolaceDev/solace-broker-mcp/internal/composite"
 	"github.com/SolaceDev/solace-broker-mcp/internal/composite/definitions"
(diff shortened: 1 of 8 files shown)


solace-aross (approved):
Approving. Careful work, and the security posture is right: every path fails closed. I traced the ones that matter.

- Missing or nil groups claim denies. requestGroups returns present=false on nil Extra/TokenInfo or an absent claim, and the wrapper returns the missing-claim result before Authorize runs.
- No group match and unknown tool both return Decision{} (deny).
- Nil policy is guarded three ways: composition only wraps when policy is non-nil, main.go refuses to start if enabled-but-nil, and the wrapper's own precondition panics into withRecovery.
- list-brokers exemption is structural and tested, and the pre-RBAC path stays byte-identical when disabled.

Test coverage is genuinely thorough: allow, both deny classes, nil TokenInfo, nil policy, the exemption, startup typo validation, and decision-matches-outcome. CI is green across build, lint, and all e2e including oauth.

On the Copilot comment about the nil-policy contract: the doc says the wrapper panics, but only the groups-present branch does; the missing-claim branch returns a deny instead. Worth aligning the comment, but it isn't a hole. Both nil-policy paths fail closed, so there's no bypass, only a wording overreach in the doc.

One nit, not blocking: requestGroups still carries //nolint:unused with a 'next ticket' note. This PR is that ticket, so the directive and the comment are now stale.

bczoma:
## Review — SOL-151978 (tool authorization enforcement)

Reviewed the enforcement wiring end-to-end. The security-critical logic is sound: fail-closed at every boundary (startup postcondition + guard + wrapper), the three-layer separation holds in code, deny/missing-claim are indistinguishable to the caller, and `LogValue` keeps group/tool names out of the logs. Test coverage of the changed behavior is thorough and non-brittle. 👍

Before merge I'd like the user-facing docs this change makes **wrong right now** addressed, plus the still-open nil-policy contract comment. The rest are minor or fold into the acknowledged audit-refinement follow-up. Nothing here is a defect in the enforcement path itself.

---

### 🟡 1. `README.md:93` — "only enforced control" is now false
`enable_write_tools` is no longer the only enforced control. With `mcp_client_auth.mode: oauth` and tool authorization enabled, the caller's `groups` policy is a second enforced control — broader, since it can deny read-only tools too. Suggest:

> **Confirmation is not enforced.** In `oauth` mode with tool authorization enabled, a per-caller `groups` policy additionally gates every tool (except `list-brokers`); otherwise `enable_write_tools` is the only server-side enforced control. `destructiveHint` and the confirmation text in tool descriptions are hints, not enforced by the MCP protocol — whether the user is actually prompted depends on the client and the model.

### 🟡 2. `internal/tools/authorization.go:53` — nil-policy contract holds on only one branch (re: Copilot's comment)
Confirmed, and worth closing out. The `!present` (missing-claim) branch returns before `policy.Authorize` is dereferenced, so nil-policy + missing-groups never panics — only nil-policy + groups-present does, which is the sole shape `TestWithAuthorization_NilPolicy_Panics` covers. Fail-*closed* and unreachable in production, so not a defect, but the stated precondition is honored on only one input. Cheapest fix that makes the doc true:

```go
if policy == nil {
    panic("withAuthorization: nil policy (composition-site invariant violated)")
}
```

Otherwise scope the doc to "the Authorize branch panics; missing-claim fails closed." I'd prefer the guard — it also survives a future reorder of the branches.

### 🟡 3. `CHANGELOG.md` — missing `### Added` entry
No entry for this user-observable change, breaking the per-SOL convention (SOL-150796, SOL-149606). Suggested `### Added` bullet:

> - Request-time tool authorization for `oauth` client-auth mode: when enabled, an OAuth caller's `groups` claim is checked against a compiled policy before each tool call, and tools the caller's groups don't grant are denied with a uniform tool-level error. `list-brokers` is always permitted so every authenticated caller keeps broker discovery. Startup emits an INFO line naming the RBAC posture, and an unknown tool name in the authorization config is fatal at startup. No behavior change when tool authorization is disabled. Tracked under SOL-151978.

### 🟢 4. `docs/tools-reference.md` — user-facing authz story incomplete
The overview presents `enable_write_tools` as the whole authorization model; with RBAC live it's partial.
- `:8-12` / `:73-87` — note that in `oauth` mode with `tool_authorization` on, a caller can be denied any tool (read-only included) their `groups` don't grant, independent of `enable_write_tools`.
- `:20` — "Every tool **except `list-brokers`**" is framed only as the `broker`-param exemption; `list-brokers` is now also the tool-authorization exemption. One clause avoids conflating them.
- Same "log-only" staleness applies to `README.md:135` — the `groups` claim is now consumed for a decision, not just logged.

### 🟢 5. `internal/tools/authorization.go:72` — audit lines omit the correlation ID
All three audit lines carry `tool`/`decision`/identity but not the correlation ID — though it's in `ctx` here and `withRecovery` already stamps it onto the caller-facing deny result. So the server-side decision and the caller's denial can't be joined by ID, the correlation an RBAC audit trail most needs. Comment already flags this as a v1 placeholder, so it can fold into the audit-refinement follow-up:

```go
slog.String("correlation_id", correlation.From(ctx)),
```

amitmorade:
Thank you @bczoma and @solace-aross for the feedback. These are important suggestions and I will make sure to factor those in into the next PR

PR #189, opened 2026-07-17
SOL-150757: Add unit tests for secret-handling hygiene (credential redaction beyond logs)

solace-wmaryam:
## Summary

Adds unit tests for three credential-redaction paths that were previously untested (`newSlogHandler` `ReplaceAttr`, `AuthConfig.LogValue`, and the startup banner). Sentinel-string assertions make a leak a deterministic substring hit rather than a guess at what a secret looks like.

Fixes [SOL-150757](https://sol-jira.atlassian.net/browse/SOL-150757).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

Primary change is test-only; the small refactor lifts an anonymous `ReplaceAttr` closure into a named `redactSecretAttr` function so the redaction logic is directly testable. No behavior change.

## Motivation and Context

The Broker MCP Quality Plan review (SOL-147161 §3.3) flagged that credential redaction is implemented (`slog.LogValuer` on credential-bearing types, `ReplaceAttr` filter on the JSON handler, URL-userinfo sanitization) but three specific paths have no test coverage — a regression could silently leak secrets. This PR closes those gaps.

## Changes Made

- **`cmd/server/main.go`** — lifted the `ReplaceAttr` closure in `newSlogHandler` into a package-level `redactSecretAttr` function and moved `redactedKeys` to a package var. `newSlogHandler` signature and behavior unchanged.
- **`cmd/server/redact_test.go`** (new) —
  - `TestRedactSecretAttr`: table test asserting `[REDACTED]` on the seven redacted keys, case-insensitivity (`PASSWORD`, `Api_Key`), and substring matching (`db_password`, `x-authorization-header`, `access_token_expiry`); passthrough for non-matching keys.
  - `TestNewSlogHandler_RedactsInRenderedOutput`: end-to-end assertion that the rendered JSON contains `[REDACTED]` and never the sentinel value.
- **`cmd/server/startup_banner_test.go`** (new) — `TestStartupBanner_NoCredentialLeak`: mirrors `main.go`'s startup banner sequence across `disabled` / `static_cleartext` / `oauth_plaintext_upstream` modes with sentinel credentials planted on `ServerConfig.MCPClientAuth.DevToken` and a broker `AuthConfig`; asserts no sentinel appears in stderr.
- **`internal/config/config_test.go`** — `TestAuthConfig_LogValue_RedactsCredentials`: asserts `Username` / `Password` / `Token` sentinels never appear in slog output across `basic` / `bearer` / `oauth` modes; asserts `mode` is exposed.

Note: the ticket description points at `internal/auth/banner_test.go`, but the banner package moved to `internal/banner/` a while back. The banner emitters only accept primitive `mode` / `issuer` / `bindAddr` — a banner-package-only test can't meaningfully assert "with credentials in the config", so the startup-banner assertion lives at the call site (`cmd/server/startup_banner_test.go`).

## Testing

**Test Environment:**
- Go version: 1.24 (matches project baseline)
- OS: macOS
- Broker version: N/A (unit-only, no broker required)

**Tests Performed:**

**Test Coverage:**
- Every entry in `redactedKeys` (password, token, secret, authorization, credential, api_key, private_key) — exact match.
- Case-insensitivity (uppercase and mixed-case keys).
- Substring matching (prefixed and suffixed keys).
- Passthrough for non-credential keys (username, url, mode).
- Rendered JSON validity of the redacted output.
- `AuthConfig.LogValue` across all three modes.
- Startup banner across all three mode paths, including the follow-up `LogStaticCleartextExposure` and `LogOAuthPlaintextListener` emissions.

## Breaking Changes

None.

### Code Quality

### Security
- [x] **No credentials in code** (all secret-shaped strings are `SENTINEL_*` test literals)
- [x] Credential-carrying types implement `slog.LogValuer` (unchanged; tests pin the behavior)

### Testing
- [x] Edge cases covered by tests (case-insensitivity, substring, passthrough)
- [x] Error paths tested (leak assertion is the error path)

### Documentation
- [ ] README.md updated (n/a)
- [ ] docs/ updated (n/a — the referenced `secure-logging-rules.md` remains authoritative)
- [x] godoc comments added/updated for the new `redactSecretAttr` / `redactedKeys`
- [ ] CHANGELOG.md updated (n/a)
- [ ] Examples updated (n/a)

### Git & CI
- [ ] All commits are signed off (DCO)
- [x] No merge conflicts
- [ ] CI checks passing (pending on this push)

[SOL-150757]: https://sol-jira.atlassian.net/browse/SOL-150757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/server/startup_banner_test.go
```diff
@@ -0,0 +1,111 @@
...
+					Username: sentBasicUser,
+					Password: sentBasicPass,
+					Token:    sentBearerTok,
+				},
+			}
+
+			out := captureStderr(t, func() {
+				// Install a handler that writes to (the captured) os.Stderr so
+				// slog output flows through the pipe.
+				slog.SetDefault(slog.New(newSlogHandler(slog.LevelInfo)))
```
solace-aross, line 138: `slog.SetDefault` here is never restored. `newSlogHandler` binds to `os.Stderr` at construction, and `captureStderr` has already swapped `os.Stderr` for the pipe, so the default logger ends up holding a handler that writes to `w`. When the test ends, `captureStderr` restores `os.Stderr` and closes `w`, but the default logger still points at the closed pipe, so any later test in this package that uses `slog.Default()` writes to a dead fd.

`shutdown_test.go:399-401` already has the pattern:

```go
prev := slog.Default()
slog.SetDefault(slog.New(newSlogHandler(slog.LevelInfo)))
t.Cleanup(func() { slog.SetDefault(prev) })
```

Save the previous default before the swap and restore it in a Cleanup.

cmd/server/startup_banner_test.go
```diff
@@ -0,0 +1,111 @@
...
+					Audience:    "mcp",
+					ResourceURL: "https://mcp.example.com/mcp",
+					DevToken:    sentDevTok,
+				},
+				TLSTerminatedUpstream: tc.tlsTermUp,
+			}
+			// The broker Auth block also holds credentials; keep sentinels here
+			// even though the banner call sites don't reference brokers today.
+			// Guards against a future refactor piping broker auth through.
+			_ = config.BrokerConfig{
```
solace-aross: This `BrokerConfig` is assigned to `_` and never passed anywhere, and `cfg` only feeds `Mode`/`Issuer`/`BindAddress()` into the banner calls. So none of the four sentinels can reach stderr regardless of what the banner does, and the leak loop at the bottom passes vacuously.

The comment says this guards against a future refactor piping broker auth through, but the test mirrors main()'s sequence by hand instead of calling main(), so a real change in main() wouldn't be caught here either.

Two honest options: drop the sentinel machinery and keep this as a smoke test that the three emission paths run and emit only mode/issuer/bind_address; or route the banner calls through a shared helper that main() also uses, and feed a credential-bearing value through it so the assertion has something real to catch.

solace-aross (changes requested):
The three tests that do real work (redactSecretAttr, the rendered-output check, and AuthConfig.LogValue) are solid, and the sentinel-substring approach is the right call for making a leak a deterministic hit.

Requesting changes on one issue in the startup-banner test that leaves shared state broken, and flagging a second point on that same test worth fixing while you're in there.

- SetDefault is never restored, so the global default logger is left writing to a closed pipe after this test runs. shutdown_test.go:399-401 already has the save/restore idiom. (blocking)
- The four sentinels never reach the banner call path, so those leak assertions pass no matter what. Details inline.

Happy to re-review quickly once the first one is addressed.

amitmorade (commented):
Reviewed the PR against the ticket (SOL-150757) and the surrounding code. Framing this as a test-only PR guarding an already-shipped security control — the bar is "do the tests actually assert what they claim," not "does the code work."

**The three tests that do real work — `TestRedactSecretAttr`, `TestNewSlogHandler_RedactsInRenderedOutput`, and `TestAuthConfig_LogValue_RedactsCredentials` — are solid.** Sentinel-substring approach is the right technique, coverage of the seven `redactedKeys` + case-insensitivity + substring matching is thorough, and the three `AuthConfig` modes are all exercised. The `redactSecretAttr` / `redactedKeys` refactor is genuinely zero-behavior-change and the smallest thing that makes the filter directly testable. Nicely scoped.

**The two blocking findings on `startup_banner_test.go` are already covered by @solace-aross, and both are the right calls** — I won't repeat the details, just adding my agreement:

1. `slog.SetDefault` save/restore — the pattern at `cmd/server/shutdown_test.go:373-376` is exactly what's needed here.
2. The sentinel assertions are vacuous against the current call path. I want to underline the meta-point: the test's *name* (`NoCredentialLeak`) makes a claim its *implementation* can't verify, because none of the four sentinels are ever passed into any function that writes to stderr in this test. On a security-relevant path that mismatch matters more than usual — a passing test named "NoCredentialLeak" reads as "we've proven no credential can leak here," when what we've actually proven is "these three primitive-taking banner functions ran without panicking." aross's two remediation options (reframe as an emission smoke test, or wire through a shared helper `main()` also uses) both work; the reframe is the lighter fix.

Nothing new to add beyond what's already been flagged. Happy to re-review once the banner test is addressed.

solace-aross (approved):
Both findings from the last round are addressed correctly, so this is good to go.

- slog.SetDefault save/restore is now in place via t.Cleanup, and captureStderr restores os.Stderr with a defer. Shared state is clean after the test.
- The vacuous-sentinel problem is fixed the right way: sentinels are planted in a real loaded config, assertContains proves they reached cfg, and the test drives logStartupBanners, the same helper main() now calls. That's the wire-through-a-shared-helper option, which is the stronger of the two I offered.

The main.go extraction is behavior-preserving and CI is green across all the E2E suites.

Two non-blocking notes, take or leave:
- Copilot's two comments at the current head (SetDefault at line 142, os.Stderr in captureStderr at line 130) are stale. The current code already restores in both places, so they can be resolved.
- The extraction dropped the two rationale comments that explained why static-cleartext and oauth-plaintext warn. logStartupBanners would be a natural home for a one-liner on each, since that 'why' is security-relevant. Optional.

amitmorade (approved):
LGTM as well, thank you!

PR #190, opened 2026-07-17
SOL-152039: Refine tool-authorization audit event and relocate the claim sanitizer

amitmorade:
## Summary

Refines the placeholder audit event that landed with the tool-authorization
enforcement wrapper into a production-quality audit surface, and relocates
the shared claim-sanitizer utility from `internal/tools/identity.go` to a
new `internal/observability/logging/sanitize` package. Feature ticket:
[SOL-152039](https://sol-jira.atlassian.net/browse/SOL-152039). Parent
epic: SOL-148417 (Claims-Based Tool Authorization).

## Type of Change

- [x] Refactoring (no functional changes) — the sanitizer relocation

## Motivation and Context

The tool-authorization enforcement wrapper landed in a previous ticket
with a placeholder audit event: INFO for every outcome, minimal field
set. That was a deliberate scaffold — the enforcement mechanics needed
to land first so the audit surface could be refined against a working
wrapper. This PR is the refinement: proper level split, decision +
decision_reason two-field sub-axis, bounded matched-groups slice with
truncation flag, and a sanitized expected_claim on missing-claim events.

The sanitizer relocation makes the utility available to the audit event
without forcing `internal/tools/authorization.go` to reach into
`internal/tools/identity.go`'s unexported surface. Pure behavior-
preserving rename — same 256-byte cap on rune boundary, same Cc/Cf/Zl/Zp
strip set. Two-commit structure keeps the relocation reviewable in
isolation from the security-relevant schema change.

## Changes Made

- New package `internal/observability/logging/sanitize` exports
  `Claim(s string) string`, `MaxLen = 256`,
  `VerifierBugSentinel = "<verifier-bug>"`. Algorithm is byte-for-byte
  identical to the pre-move `sanitizeClaim`.
- `internal/tools/identity.go` drops the unexported symbols and routes
  callers through the new package.
- `internal/tools/authorization.go` refines the three audit-event call
  sites (allow / deny / missing-claim). New `configuredGroupsClaimName`
  parameter on `withAuthorization`; new `boundMatchedGroups` helper
  applying the bound of 32 and per-element sanitization.
- `internal/tools/register.go` and `cmd/server/main.go` plumb the
  configured claim name through to the wrapper.
- New test file `internal/tools/authorization_audit_schema_test.go` —
  six tests pinning the refined schema, including load-bearing negative
  assertions on `caller_groups` absence.
- Three additive tests in `internal/observability/logging/sanitize/sanitize_test.go`
  covering inclusive-cap boundary, prefix truncation identity, and
  combining-mark pass-through (Mn exclusion).

## Design rationale — audit-log disclosure discipline

The deny event carries `matched_groups=[]` and `matched_groups_total=0`
but deliberately does NOT carry the caller's actual extracted groups
list. This is a load-bearing separation-of-duties decision that a
future refactor could inadvertently regress; the negative assertion in
`TestAuditEvent_Deny_EmitsWarnWithNotPermittedAndNoCallerGroups` exists
to fail loudly if it does.

In real organizations, MCP server operators (platform/SRE) and IdP
administrators (identity team) are frequently different roles with
different access surfaces. Logging the caller's complete extracted
group membership on every deny would give MCP operators a shadow user
→ group map derived from audit trails — knowledge the identity team
may have deliberately kept internal. This crosses a boundary the IdP
enforces by policy.

Group names themselves can be sensitive. Some organizations use names
that reveal role, seniority, or team affiliation (VP-Direct-Reports,
Legal-Confidential-Reviewers, M&A-Advisory). Emitting these on every
denial would leak organizational structure to whoever tails MCP logs.

There is a deliberate asymmetry with `matched_groups` on allow.
`matched_groups` reveals only which admin-configured groups matched a
specific tool grant — that disclosure is audit-mandatory (the "under
what authority" question compliance auditors ask). `caller_groups` on
deny would reveal the user's complete membership set, most of it
unrelated to the denied tool. Different disclosure category, different
discipline. The audit contract answers "why was this denied?" with
"the tool grants list has none of the caller's groups," which is a
complete audit fact even without enumerating the caller's groups.

If beta feedback shows operators genuinely need `caller_groups` on
deny (regulated-customer scenarios where operator = IdP admin, for
example), a config knob `audit_include_caller_groups` with default
`false` is a backward-compatible add. Default-off preserves the
discipline; opt-in per deployment lets high-trust deployments override.

## Design rationale — audit-event schema and level split

The event carries three matched-groups fields (`matched_groups`,
`matched_groups_total`, `matched_groups_truncated`) rather than the
two the interface design originally proposed. The extra boolean is
not schema noise: operators tailing logs by eye need truncation as an
immediate visual signal. Comparing `len(matched_groups)` against
`matched_groups_total` works for SIEM parsers but not for a human
scanning terminal output; the boolean makes the state legible at a
glance. ~15 extra bytes per line; negligible.

The bound of 32 on `matched_groups` is tuned for log-line readability,
not compliance. 32 group names at ~40 bytes each is ~1.3 KB — under
every SIEM per-field cap and comfortable in both structured JSON and
text-handler console output. Realistic multi-group memberships fit
without truncation; truncation firing is almost certainly a broken
IdP emission or hostile input. When it does fire,
`matched_groups_total` records the true count so nothing is silently
dropped.

The decision axis is a two-field split (`decision`, `decision_reason`),
not a single three-value enum. `decision` carries the primary axis
(allowed vs. denied) that SIEM queries filter on cleanly.
`decision_reason` distinguishes the two non-allow sub-cases
(`not_permitted` vs. `missing_claim`) without flattening them onto the
allow axis. A single enum would blur the "notable but non-error" deny
category with the "configuration event" missing-claim category —
different operator responses, different alerting rules.

Level split is INFO for allow, WARN for both deny outcomes.
Missing-claim is deliberately NOT ERROR: "misconfiguration fails
loudly" in the spec refers to visibility (structured field, distinctive
`decision_reason`), not severity. ERROR would put missing-claim events
on the same footing as system failures (panics, exit-1); an
IdP-misconfigured deployment could flood ERROR-level dashboards and
pagers designed to alert on actual failures. WARN with the reason
field lets operators distinguish missing-claim from deny cleanly
without conflating configuration with system errors.

## Testing

**Test Environment:**
- Go version: 1.25
- OS: darwin arm64

**Tests Performed:**
- [x] Unit tests added/updated — 3 sanitize boundary tests + 6 audit
      schema tests, all in the T5-touched packages.
- [x] Tested locally with `go test -race ./...` — clean.
- [ ] Tested with real Solace broker — not applicable (no broker
      interaction in this change).

**Test Coverage:**
- Inclusive cap at MaxLen; prefix truncation identity at MaxLen+1;
  combining marks survive (Mn exclusion boundary from the outside).
- Every field on every outcome (allow INFO, deny WARN with
  `not_permitted`, missing-claim WARN with `missing_claim` and
  `expected_claim`).
- Load-bearing negative assertions: `caller_groups` absent on deny;
  `matched_groups*` fields absent on missing-claim; `decision_reason`
  and `expected_claim` absent on allow.
- Sanitize applied to `expected_claim` (configured claim name) and to
  every element of `matched_groups` (multi-element test catches
  "sanitize only first" refactor bugs).
- Bounded matched-groups: 16 (under), 32 (inclusive cap), 33 (one
  over) — table-driven, asserts set-membership and dedup rather than
  sequence.
- Correlation ID propagation empirically verified via raw log-line
  dump: `correlation_id` stamped by the correlation slog handler
  reading it from ctx; wrapper never adds it manually.

## Breaking Changes

None. The refined audit event is a new field schema on log records;
downstream SIEM parsers that read the placeholder event will continue
to parse — the tool/decision/identity fields are unchanged, only new
fields appear on the record.

### Code Quality
- [x] Code follows the coding standards.
- [x] Self-review completed.
- [x] Comments explain "why", not "what".
- [x] No commented-out code or debug statements.

### Security
- [x] No credentials in code.
- [x] Secure-logging rules followed — every admin/IdP-controlled
      string reaching a log record passes through `sanitize.Claim`.
- [x] `caller_groups` deliberately omitted from deny events (see
      design rationale above).
- [x] gosec / staticcheck clean via `make check`.

### Testing
- [x] `make check` passes at each commit boundary independently
      (verified via detached-HEAD checkout of the sanitizer-relocation
      commit).
- [x] `go test -race ./internal/tools/... ./internal/observability/...`
      clean.
- [x] Negative assertions carry regression-warning messages so
      future contributors see the contract they would violate.

### Documentation
- [x] Godoc comments on new exported symbols (`sanitize.Claim`,
      `MaxLen`, `VerifierBugSentinel`).
- [x] Design rationale documented in this PR description (not just in
      internal design docs) so the caller_groups decision is visible
      in PR history.

### Git & CI
- [x] Commits have clear, descriptive messages.
- [x] Two-commit-plus-followups structure keeps the sanitizer
      relocation reviewable in isolation from the security-relevant
      audit refinement.
- [ ] Signed off (DCO) — not signed per project convention on this
      repo.

## Related Issues/PRs

- Jira: [SOL-152039](https://sol-jira.atlassian.net/browse/SOL-152039)
- Parent epic: SOL-148417

---

**For Reviewers:**

**Focus areas:**
1. `internal/tools/authorization.go` audit-event call sites — is the
   field ordering / naming what you'd want to see in a SIEM query?
2. The `caller_groups` absence decision — captured in the "audit-log
   disclosure discipline" section above. If you have regulated-customer
   experience where operator = IdP admin, please push back explicitly
   so we can weigh in on the future config knob.
3. `TestAuditEvent_MatchedGroupsBounding` — the bound is set at 32.
   Sanity check: is 32 the right number for your operational context?
4. Commit structure — the sanitizer relocation is pure refactor; the
   audit refinement is the security-relevant change. Reviewing them
   as separate commits, not as one squashed PR, is the intended
   workflow.

[SOL-152039]: https://sol-jira.atlassian.net/browse/SOL-152039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SOL-152039]: https://sol-jira.atlassian.net/browse/SOL-152039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -687,7 +687,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	tools.RegisterWithServer(mgr, server, pool, cfg.EnableWriteTools, policy)
+	// Pull the configured groups claim name for the missing-claim audit
+	// event. applyDefaults guarantees GroupsClaimName is non-nil whenever
+	// the tool_authorization block is non-nil, so this dereference is safe
+	// under the same gate that produced a non-nil policy. When RBAC is
+	// disabled the string is unused by RegisterWithServer.
+	var groupsClaimName string
+	if policy != nil {
+		groupsClaimName = *cfg.MCPClientAuth.ToolAuthorization.GroupsClaimName
+	}
+	tools.RegisterWithServer(mgr, server, pool, cfg.EnableWriteTools, policy, groupsClaimName)
 	slog.Info("tool registration complete",
 		slog.Bool("enable_write_tools", cfg.EnableWriteTools))
 
--- a/internal/observability/logging/sanitize/sanitize.go
+++ b/internal/observability/logging/sanitize/sanitize.go
@@ -1,0 +1,92 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package sanitize centralizes audit-log field hygiene. Admin- or
+// IdP-controlled strings pass through Claim before they reach an slog
+// record, so a hostile or buggy source cannot inject forged log lines
+// (CWE-117) or spoof audit identities via bidi controls (CWE-1007).
+//
+// This package is a leaf: it has no dependencies outside the standard
+// library and imports no other repo packages. Both the tool-authorization
+// audit event and the identity-audit log line depend on it, so hosting
+// the utility here — under the observability tree, alongside correlation —
+// keeps every caller import-cycle-free.
+package sanitize
+
+import (
+	"strings"
+	"unicode"
+	"unicode/utf8"
+)
+
+// MaxLen is the byte cap applied by Claim. A real IdP-issued sub/jti is
+// 20–50 chars; the cap defends against a hostile or buggy issuer shipping
+// a multi-megabyte claim into audit logs.
+const MaxLen = 256
+
+// VerifierBugSentinel is the audit sentinel emitted by callers when their
+// own verifier stashed a non-string value where a string was expected.
+// Distinct from an "absent" sentinel so log consumers can tell "IdP did
+// not issue this claim" (normal) from "our code put garbage in Extra"
+// (alarm).
+const VerifierBugSentinel = "<verifier-bug>"
+
+// Claim defends against log-injection (CWE-117) and audit-spoofing
+// (CWE-1007) by stripping non-graphic Unicode and capping the result at
+// MaxLen bytes. The four stripped categories are:
+//
+//   - Cc — Control characters. Includes the ASCII C0 block (NUL, TAB, LF,
+//     CR, ESC, …), DEL, and the C1 block (U+0080–U+009F). This is the
+//     CWE-117 surface: LF and CR can break log-line framing.
+//   - Cf — Format characters. Includes bidi/RTL overrides (U+202A–U+202E,
+//     U+2066–U+2069), zero-width joiners, BOM, soft hyphen, language tags.
+//     Bidi controls are the CWE-1007 surface: a sub like
+//     "alice<RLO>nimda" renders as "aliceadmin" in any bidi-aware UI,
+//     misattributing the action to the wrong user in SIEMs/terminals.
+//   - Zl / Zp — Line and Paragraph separator (U+2028, U+2029). Some
+//     renderers honor these as line breaks; they're CWE-117-adjacent.
+//
+// JSON-handler escaping already neutralizes CR/LF for the JSON sink, but
+// field-level sanitization protects re-emission through error messages,
+// panic strings, custom handlers, and metric labels.
+//
+// Allocation and CPU are bounded by MaxLen: the working buffer is sized at
+// min(len(s), MaxLen) and the loop breaks as soon as another rune would
+// exceed the cap. A malicious or buggy IdP shipping a multi-megabyte claim
+// cannot force proportional work here.
+//
+// Truncation lands on a rune boundary — the projected length is checked
+// BEFORE writing, so the output is always valid UTF-8 even if the cap
+// falls inside what would have been a multibyte codepoint.
+func Claim(s string) string {
+	b := strings.Builder{}
+	b.Grow(min(len(s), MaxLen))
+
+	for _, r := range s {
+		// Single filter, complete spec — see godoc above. Cc covers both
+		// ASCII (C0/DEL) and C1 controls, so no separate fast-path is
+		// needed.
+		if unicode.In(r, unicode.Cc, unicode.Cf, unicode.Zl, unicode.Zp) {
+			continue
+		}
+		// Stop before writing a rune that would push us past the cap. This
+		// is both the DoS defense (bounds total work) and the UTF-8 safety
+		// guarantee (never emits half a multibyte codepoint).
+		if b.Len()+utf8.RuneLen(r) > MaxLen {
+			break
+		}
+		b.WriteRune(r)
+	}
+	return b.String()
+}
--- a/internal/observability/logging/sanitize/sanitize_test.go
+++ b/internal/observability/logging/sanitize/sanitize_test.go
@@ -1,0 +1,218 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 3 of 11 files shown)


amitmorade:
following up on what I said I'd address from PR https://github.com/SolaceDev/solace-broker-mcp/pull/186:

**#1 (Andrea's nit): stale `//nolint:unused` on `requestGroups`** — dropped in `bab53f7`. The directive and the "next ticket" note are both gone; that ticket is this one.

**#2 (Balazs): nil-policy contract holds on only one branch** — closed in `e137507`. Added an explicit top-of-closure guard so nil-policy panics uniformly regardless of which downstream branch a request would have taken. Without it, nil + missing-claim returned first and silently masked the invariant violation — exactly the gap you called out. Also survives a future branch reorder, as you noted. Companion test `TestWithAuthorization_NilPolicy_PanicsOnMissingClaimBranchToo` pins the new panic path so the guard can't be silently removed.

**#5 (Balazs): correlation ID on audit lines** — this one is done, but not the way the comment suggests. The correlation ID does land on every `"tool authorization"` record via the correlation slog handler decorator (`internal/observability/correlation/sloghandler.go`) — it reads the ID from ctx at emit time and stamps `correlation_id` onto the record. The wrapper deliberately does NOT add `correlation_id` at the call site because doing so would double-stamp the field (the handler doesn't dedup by key). Empirically verified: my three schema tests seed `correlation.With(ctx, "corr-…")` and assert `rec["correlation_id"] == "corr-…"` on all three outcomes. So the "server-side decision and caller's denial can't be joined by ID" concern isn't the current state — they share the ID via the handler. The wrapper's docstring on `withAuthorization` was updated to spell this out (the pre-refinement placeholder comment was stale).

**#1 / #3 / #4 (docs — README.md, CHANGELOG.md, docs/tools-reference.md):** deliberately holding these off. The feature is still behind a config gate and hasn't been exercised end-to-end against the OAuth harness yet. I want to test more thoroughly early next week before the flag comes off — if anything shifts (message wording, config-knob shape, missing-claim ergonomics, edge cases in the auth flow), I'd rather update the customer-facing surface once with the version that actually ships than land partial docs I'd have to amend. I'll file a follow-up ticket to track the CHANGELOG entry + README fix + `docs/tools-reference.md` sweep so it doesn't fall off the radar when the flag flips.

solace-aross (approved):
Approving. The audit refinement is careful work, and the design rationale in the description carries most of the reviewer's load.

What I checked:
- Sanitizer relocation is a true behavior-preserving move: same 256-byte rune-boundary cap, same Cc/Cf/Zl/Zp strip set, leaf package with no repo imports. The stale `//nolint:unused` on `requestGroups` is gone as well.
- Enforcement paths: the nil-policy guard is now uniform across all three branches, `matched_groups` is bounded and per-element sanitized, and deny/missing-claim both emit WARN with the `decision`/`decision_reason` split.
- Build, lint, and the E2E suites are green. `e2e-monitoring` was still running as I write this; nothing in this change touches the monitoring surface, so I'm not holding on it.

On your two invited focus areas:
- `caller_groups` on deny: agreed on omitting it. The separation-of-duties argument holds, and `matched_groups_total=0` answers the audit question without enumerating the caller's membership. A default-off `audit_include_caller_groups` knob is the right escape hatch if a regulated customer ever needs it, and there's no reason to build it now.
- Bound of 32: fine. It's a readability bound with `matched_groups_total` preserving the true count on overflow, so nothing is silently dropped. I wouldn't tune it without a real membership distribution to point at.

Clean split between the relocation and the schema commits.

PR #191, opened 2026-07-20
Reject empty or whitespace-only tool names in access_level_groups

amitmorade:
## Summary

Config validation for `mcp_client_auth.tool_authorization.access_level_groups` previously did not check tool-name **values** inside each group's list, so mistakes like `Ops: ["", "list-queues"]` or `Ops: ["list-queues "]` silently compiled into inert grants. This adds a single-invariant validator at config load: **a tool name must be non-empty and equal to its trimmed form** — one rule that rejects empty, whitespace-only, and surrounding-whitespace values at startup with a clear, indexed error.

Ticket: [SOL-151914](https://sol-jira.atlassian.net/browse/SOL-151914)

## Type of Change

## Motivation and Context

A config like:

```yaml
tool_authorization:
  enabled: true
  access_level_groups:
    Ops: ["", "list-queues"]
```

passed validation, reached `NewPolicy`, and compiled `""` as a valid tool entry. `Authorize(["Ops"], "")` then returned allowed. There is **no runtime exploit** — the MCP dispatch layer never routes an empty tool name — so this is a fail-fast / defense-in-depth fix, not a security patch.

The same class of soft failure covers **surrounding whitespace** too: `"list-queues "` (bad template substitution, editor whitespace, copy-paste artifact) compiled as a distinct-from-`"list-queues"` policy entry, silently making the operator's grant unreachable. Under `enabled: false` (draft-policy workflow) this wouldn't even be caught by the later registry check.

The real value: catching operator mistakes — dropped values, trailing commas, bad template substitutions, editor whitespace — that today are silently accepted as inert entries instead of surfacing as errors at startup.

## Decisions and Rationale

Three shape choices deserve to be called out; all were deliberate.

### Decision 1 — Reject, do not trim

An empty, whitespace-only, or padded tool name causes startup to **fail** with a clear error. It is **not** silently trimmed or normalized.

**Why:** the whole point of this fix is to surface operator mistakes. YAML config is authored, not user input — someone typed it, and a trailing space in `"list-queues "` is almost certainly a mistake (copy-paste artifact, editor whitespace, bad template substitution), not intent. Silent normalization would:

- Hide the exact mistake we want to expose.
- Create two sources of truth — the value on disk differs from the value in memory, so an operator debugging *"why doesn't my ACL match?"* greps the YAML, sees the string they wrote, and can't reproduce what the server actually loaded.
- Punish the next reader of the file (does `"list-queues "` matter here? mystery).

Config validation is a one-time cost at startup. Failing loud once is cheaper than debugging silent trimming at 2am.

### Decision 2 — One rule, not several

The check is `tool != "" && tool == strings.TrimSpace(tool)`. **One invariant, stated positively:** *"non-empty and equal to its trimmed form."*

**Why:** consistency beats cleverness. *"Empty = error, whitespace-only = error, leading/trailing space = silently fixed"* is three rules. *"Non-empty, no surrounding whitespace"* is one rule. One rule is easier to document, test, and reason about, and the error message can state the invariant directly rather than enumerate every violation shape.

Error message states the rule (*"must be non-empty and have no leading or trailing whitespace"*) rather than describing the violation (*"is empty or whitespace-only"*). The rule wording is stable across all four failure modes.

**Boundary:** internal whitespace (`"list queues"`) is intentionally NOT rejected by this validator. Such a name is a syntactically clean string that isn't a real tool — that's the registry check's job (`ValidatePolicyToolNames`). See the follow-up note at the bottom.

### Decision 3 — Collapse multiple offenders per group into one line

When a group has more than one bad tool name, one error line per group with all offending indices joined (`[0, 2]`), rather than one line per offender.

**Why:** the outer group-name check `break`s after the first empty group because map iteration is nondeterministic — reporting "index 0" vs "index 3" of a map traversal would be flaky. That reasoning does not apply to slices: tool names live in a stable-order slice, so joining every offending index for a given group stays deterministic. The offending value is not echoed back in the message (operators open the YAML to see what they wrote), so collapsing several near-identical lines into one loses no signal.

Consequence: grammar switches between singular and plural (`tool name ... [0]` ↔ `tool names ... [0, 2]`) so the message reads naturally in both cases. Every group with offenders is still reported — accumulation across groups is preserved via `errors.Join`.

## What the operator sees

**Case A — single unclean tool name.** Config:

```yaml
access_level_groups:
  Ops:
    - ""
    - list-queues
```

Startup log:

```
validating config: mcp_client_auth.tool_authorization.access_level_groups: tool name at "Ops" [0] must be non-empty and have no leading or trailing whitespace
```

The empty string is invisible inside the message text itself, so the index (`[0]`) is what locates the mistake. The space between `"Ops"` and `[0]` is intentional readability polish so the eye doesn't run the group name and index together.

**Case B — mixed offenders across multiple groups.** Config:

```yaml
access_level_groups:
  Ops:
    - ""
    - list-queues
    - "list-queues "     # trailing space
  Admin:
    - get-broker-status
    - " get-broker-status"  # leading space
```

Startup log:

```
validating config: mcp_client_auth.tool_authorization.access_level_groups: tool names at "Ops" [0, 2] must be non-empty and have no leading or trailing whitespace
mcp_client_auth.tool_authorization.access_level_groups: tool name at "Admin" [1] must be non-empty and have no leading or trailing whitespace
```

Both groups reported. `Ops` has two offenders (empty at `[0]`, padded at `[2]`) → plural, indices joined into one line. `Admin` has one offender → singular. Startup aborts; operator fixes all three positions from a single restart cycle.

## Changes Made

**Commit 1 — Reject empty or whitespace-only tool names** (`abcd9a2`)
- `internal/config/config.go` — new validation loop in `validateToolAuthorization` mirroring the existing group-name check's style. Rejected empty and whitespace-only tool names.
- `internal/config/tool_authorization_test.go` — table-driven test for the empty/whitespace cases + a multi-offender collapse test.

**Commit 2 — Reject surrounding whitespace** (`3935c0a`)
- `internal/config/config.go` — extended the rule to a single invariant: `tool != "" && tool == strings.TrimSpace(tool)`. Now covers empty, whitespace-only, **and** surrounding whitespace. Error wording shifts from describing the violation to stating the invariant.
- `internal/config/tool_authorization_test.go` — table extended with trailing-space and leading-space rejection cases, plus a control for internal whitespace being accepted (locks in the surrounding-only boundary for future readers). Multi-offender test updated for the extended rule.

**Commit 3 — Merge main** (`3730391`) to keep the branch current.

## Explicitly out of scope

- **Registry-based typo detection** for tool names (`list-queuse` → unknown tool). `internal/tools/authorization.go:164` already provides `ValidatePolicyToolNames` for that.
- **Internal whitespace** in tool names (`"list queues"`). Syntactically a clean string; the fact that it isn't a real tool is a registry-check concern.
- **Deduplication within a group.** Handled elsewhere.
- **Sibling `[]string` yaml lists.** Grep confirmed `AccessLevelGroups map[string][]string` (`internal/config/config.go:342`) is the only `[]string` YAML list field in the config package. No sibling deny/allow/name-list types needed the same treatment.

## Known follow-up (separate ticket, not this PR)

`ValidatePolicyToolNames` is gated on `policy != nil`, which is only true when `enabled: true`. Under `enabled: false` (draft-policy workflow) it does not run, so a typo like `"list-queuse"` or an internal-whitespace name like `"list queues"` slips through until the operator flips `enabled: true`. The config-load-time structural checks in this PR run regardless of `enabled`, but the registry-dependent typo check does not — that's a real cross-layer inconsistency worth its own ticket. **Not fixed here** because it's a behavior change for existing operators running with draft policies and deserves its own design review.

## Testing

**Test Environment:**
- Go version: 1.24
- OS: macOS (darwin/arm64)
- Broker version: N/A (config-only change)

**Tests Performed:**
- [x] Tested locally with `go test -race ./internal/config/...`
- [x] `go vet ./internal/config/...` clean
- [x] `/check-logs` scan clean (no logging violations)

**Test Coverage:**
- Empty string, single space, tab, mixed whitespace, trailing space, leading space — all rejected with indexed error.
- Valid tool names accepted.
- Internal whitespace (`"list queues"`) accepted at this layer — locks in the surrounding-only boundary.
- Multi-offender per group with singular/plural rendering across two groups (empty + padded in one group, leading-space in another).
- Existing group-name empty-check tests unaffected.

## Breaking Changes

None. Any config that was **previously accepted** as a real, clean tool name remains accepted. Any config that was **previously silently compiled with an inert entry** (empty, whitespace-only, padded) now fails startup with a clear error — that is the intended fail-fast behavior.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code

### Testing
- [x] All new/existing tests pass locally: `go test -race ./internal/config/...`
- [x] Edge cases covered (empty, whitespace-only, tab, mixed whitespace, leading space, trailing space, internal whitespace boundary, multi-offender per group)
- [x] Error paths tested

### Documentation
- [x] godoc / inline comments added where non-obvious ("why" only)

### Git & CI
- [x] Branch is up to date with main (merged, not rebased)
- [x] No merge conflicts

## For Reviewers

**Focus areas:**
1. Decision 1 (reject vs trim) — do you agree fail-fast is the right posture over silent normalization?
2. Decision 2 (one invariant, stated positively) — is the *"must be non-empty and have no leading or trailing whitespace"* wording clear, or would you prefer the violation-listing form?
3. Decision 3 (collapse per group) — is the per-group collapse the right operator UX given the value isn't echoed in the message?
4. The internal-whitespace boundary — the check accepts `"list queues"` because that's a registry-check concern; noted as a follow-up for making `ValidatePolicyToolNames` run under `enabled: false`. Comfortable with that split?

[SOL-151914]: https://sol-jira.atlassian.net/browse/SOL-151914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solace-aross (dismissed):
Clean, and the PR body does the reasoning for me. Confirmed the new loop sits inside the `ta != nil` block after the group-name check, and strconv/strings/fmt are already imported. Ranging a nil map is safe, per-group indices are deterministic on stable slice order, and cross-group line ordering being nondeterministic is harmless since errors accumulate.

Fail-fast over silent trim is the right call and matches the existing validator convention. The tests assert the exact rendered messages (singular/plural, joined indices), so they'd fail if the behavior regressed. The group-name break vs tool-name accumulate asymmetry is deliberate and you pre-empted it.

On your three focus questions: agree on reject-not-trim, agree the per-group collapse is the right operator UX, and the message format including the space before the index reads fine.

internal/config/config.go
```diff
@@ -1226,6 +1226,29 @@ func validateToolAuthorization(cfg *ServerConfig) []error {
...
 		}
+
+		// Empty or whitespace-only tool name inside a group's list is a config
+		// coherence error. Indices are reported because an empty string is
+		// invisible in an error message otherwise; when a group has multiple
+		// offenders, they collapse into one line per group.
+		for groupName, tools := range ta.AccessLevelGroups {
+			var badIdx []string
+			for i, tool := range tools {
+				if strings.TrimSpace(tool) == "" {
```
solace-awaheed: TrimSpace(tool) == "" only catches entries that are completely blank. A tool name with extra whitespace (e.g. "list-queues " from a bad template substitution still gets through here.

Normally ValidatePolicyToolNames catches that, but it only runs when policy != nil. If the block is present with enabled: false, this check runs but ValidatePolicyToolNames doesn't, so tool names with leading/trailing whitespace slip through unnoticed in that case.

Edit: This PR is only going for empty and whitespace-only so not blocking.

solace-wmaryam:
LGTM as well!

solace-awaheed (approved):
LGTM!

PR #192, opened 2026-07-20
SOL-151876: list-vpns: derive zeroConnectionCount from msgVpnClients (drop #client invariant) 

helen-zhu-solace:
## Summary

Replace list-vpns' `zeroConnectionCount` derivation with a version-independent per-VPN probe, and add a fan-out step type to the composite framework as the first primitive that lets a step iterate a prior step's rows.

Fixes SOL-151876

## Type of Change

## Motivation and Context

`zeroConnectionCount` was previously derived by counting VPNs where `msgVpnConnections <= 1`. That predicate rides on an empirical broker invariant — the reserved `#client` shows up as exactly one connection per enabled+up VPN — that isn't part of the SEMP contract. A tripwire in `verify-fixtures.sh` asserted `msgVpnConnections == 1` on a fixture VPN to catch drift.

New derivation: for each enabled VPN, probe `getMsgVpnClients` filtered server-side by `clientUsername != #*` (the `#` prefix is Solace's documented reserved-name contract). If any row comes back, the VPN has a real client; otherwise it counts as zero-conn. Version-independent, and the tripwire is no longer load-bearing.

The composite framework had no primitive for issuing a per-row follow-up call to a prior step. This PR adds one — `forEach` on `Step` — as the general mechanism, with `list-vpns` its first consumer. Other list-* tools (e.g. `get-rdp-status`) can adopt it later.

## Changes Made

**Framework (fan-out step type):**
- `Step` gains `ForEach`, `ForEachIf`, `ForEachKey`, `Concurrency` (`internal/composite/definition.go`).
- `fetchFanOut` iterates the parent step's `data[]` under a bounded errgroup (default 8, cap 32), keyed by `row[ForEachKey]`. `ForEachIf` skips falsy rows and tallies under `skipped`; fail-fast on the first per-row error. `.Item` binds the current row for per-iteration templating (`internal/composite/executor.go`).
- Loader validation: prior-step reference, no `parallel`, key-in-parent-select, concurrency range, config-smell rejection (`internal/composite/loader.go`).

**list-vpns rewire:**
- New `real-clients` step in `tools.yaml`: `forEach: vpns`, `forEachIf: {{.Item.enabled}}`, `where: clientUsername!=#*`, `count=1`, `select: [clientName]`. Boolean check — no pagination per VPN.
- Handler reads `real-clients.byKey[vpnName].data` and counts empties as zero-conn. `RequiredFieldsPerStep` split across both steps; drops the `msgVpnConnections` requirement.

**E2E:**
- Removed the `msgVpnConnections == 1` tripwire from `verify-fixtures.sh`.
- Rewrote the summary-recompute assertion in `test-monitoring-tools.sh` to cross-reference `real-clients.byKey` instead of the old `<=1` predicate. Switched to a bespoke `assert_json_field` because the recompute needs top-level context that `assert_recompute_count` strips.
- Updated the fixture comment in `helpers.sh`.

## Testing

**Test Environment:**
- Go version: matches `go.mod`
- OS: Linux
- Broker version: Solace PubSub+ 10.26.0.8586 (e2e-monitoring docker fixtures)

**Tests Performed:**
- [ ] Integration tests added/updated (if applicable) — n/a
- [x] E2E tests added/updated
- [x] Tested with real Solace broker (local e2e-monitoring docker; verified filter behaviour end-to-end)

**Test Coverage:**
- `executor_fanout_test.go` — 9 cases: basic iteration, keyed result, `ForEachIf` skip and non-bool error, empty parent, fail-fast, concurrency cap (barrier-client), `.Item` templating, missing parent step, missing forEachKey on row.
- `loader_fanout_test.go` — 8 cases across the validation rules (forward ref, `parallel` combo, missing forEachKey, key not in parent select, concurrency out-of-range, fan-out fields without forEach).
- `list_vpns_test.go` — rewritten to cover the new fan-out shape; added missing-byKey-entry test; expanded `ValidateTool` cases for both step IDs.

## Breaking Changes

None. `list-vpns`'s response gains a top-level `real-clients` block alongside the existing `vpns` block; `summary` fields are unchanged.

### Code Quality

### Security
- [x] Followed secure logging rules (fan-out per-row errors include VPN name — non-sensitive)
- [x] Credential-carrying types implement `slog.LogValuer` — n/a, no credential types touched
- [x] No security vulnerabilities introduced (`golangci-lint` clean)

### Testing
- [x] All tests pass locally: `go test -race ./...` (via `make check`)
- [x] Edge cases covered by tests (empty parent, missing key, non-bool predicate, concurrency cap)
- [x] Error paths tested

### Documentation
- [ ] README.md updated (if user-facing changes) — no user-facing surface change
- [ ] docs/ updated — architecture unchanged at doc level; may follow up in `docs/internal/architecture.md` if fan-out gets a second consumer
- [x] godoc comments added/updated for exported symbols (Step fan-out fields, `ExecuteContext.Item`, `fetchFanOut`, `validateFanOut`, updated `ListVpns` header)
- [ ] CHANGELOG.md updated (if applicable) — n/a
- [ ] Examples updated (if applicable) — n/a

### Git & CI
- [x] No merge conflicts 

## Additional Notes

- Filter semantics verified live before implementation: on a VPN with one `#client-username`, `where=clientUsername==%23*` returns 1 row and `!=%23*` returns 0 — proving `!=` respects the `*` wildcard.
- Concurrency default set to 8 (capped at 32 in the loader).
- Follow-ups explicitly not in this PR: adopting fan-out in `get-rdp-status` (SOL-151343), best-effort per-iteration errors, adopting per-iteration pagination in another consumer (`FollowPages` already works on fan-out iterations).

## Related Issues/PRs

- Fixes SOL-151876
- Supersedes the `<=1` predicate landed in SOL-151771
---

**For Reviewers:**

**Focus Areas:** `fetchFanOut` in `internal/composite/executor.go` (concurrency scaffold + per-iteration template context); YAML `where: "clientUsername!=#*"` (double-URL-encoding was a bug I hit — the SEMP client encodes the value, so the YAML must contain a raw `#`); new recompute jq in `test-monitoring-tools.sh` (needed a bespoke form because `assert_recompute_count` strips top-level context).

**Questions:** Concurrency default of 8 OK, or would you prefer 4 for cautious rollout? Worth landing `forEachIf` in this PR vs. follow-up (it's small and useful but not strictly required)?


solace-aross (dismissed):
Approving. Clean, and the fan-out primitive is well-scoped for its first consumer.

I verified the four earlier Copilot points are all resolved at this head:
- Self-reference in `validateFanOut` now rejected explicitly (+ test).
- `fetchFanOut` errors on non-list parent `data` instead of silently iterating zero rows (+ test).
- Concurrency slot acquired in the dispatch loop, not per-goroutine; duplicate `forEachKey` values are a hard error (+ tests).
- `list-vpns` description rewritten to the `clientUsername != #*` probe semantics.

Also checked the concurrency model (fail-fast errgroup, bounded semaphore, gCtx abort in the dispatch select), the `count=1` boolean-probe shape, and the truncation semantics carrying over to the visible page. The test suite earns its green. CI and e2e-monitoring pass.

One non-blocking note: `forEachIf: {{.Item.enabled}}` probes every enabled VPN, but the handler only consults `byKey` on the `up` branch. Enabled+down and enabled+standby VPNs get a getMsgVpnClients call whose result is never read, which is wasted work and extra fail-fast surface. Tightening the predicate to enabled AND state==up would match exactly what's consumed. The yaml comment reasons about disabled VPNs but not enabled-but-not-up ones. Doesn't affect the count, so leaving it to your judgment.

solace-aross (approved):
Re-reviewing at head (`abce7f9d` / merge `38d53b3`). My earlier approval was dismissed by the new commit; the delta since is a single change plus a `main` merge.

The new commit tightens `real-clients` to `forEachIf: {{and .Item.enabled (eq .Item.state "up")}}`, which is exactly the enabled-AND-up predicate I raised. It now matches what the handler consumes: `state` is `select`ed on the parent `vpns` step, and `list_vpns.go` only calls `hasRealClient` under `case "up"`. Enabled+down and enabled+standby VPNs no longer get a `getMsgVpnClients` call whose result is discarded, which removes the wasted work and the extra fail-fast surface. The yaml comment now reasons about the enabled-but-not-up rows too.

Verified `e2e-monitoring` and the full check suite are green at this head, so the recompute assertion still agrees with the narrower probe set.

Nothing else changed in the PR. Re-approving.

PR #197, opened 2026-07-21
SOL-152102: F-series LLM eval: fix fabrication check false positives and discards timing flake

helen-zhu-solace:
## Summary

Harden the LLM eval suite (`test/e2e-llm`) against three sources of false-negative CI failures: fabrication-check false positives on adjacent entity names, timing races against the F7 discards fixture, and a silent-exit bug in the ground-truth runner. Also broadens a couple of overly narrow substring gates and drops a redundant turn-2 followup.

Fixes SOL-152102

## Type of Change

## Motivation and Context

Several F-series scenarios were flaking or producing misleading FAILs:

1. **Fabrication check over-flagged legitimate context.** The old check diffed the answer's entity set against `ground_truth.jq`'s narrow projection (only discarding queues, only unhealthy RDPs). When the model correctly named an adjacent seeded entity for context (a healthy sibling `test-rdp`, an adjacent `test-queue-*`), it was flagged as fabrication.
2. **`read-list-queue-discards` raced the fixture.** `topOffenderQueues[]` is only populated after F7 has actually discarded messages; if the scenario ran before warm-up, the truth set was empty and the assertion degenerated silently.
3. **Silent-exit bug in `run-scenario.sh`.** When the LLM guessed the wrong VPN and produced a non-JSON error tool_result, `expected_set+=$(jq … 2>/dev/null)$'\n'` inside a `while` body returned jq's exit code 5, and `set -euo pipefail` killed the script mid-assertion. No `[PASS]`/`[FAIL]` line printed; run-all just recorded a bare FAIL. Reproduced ~4/10 runs of `f3-subscriptions`.
4. **Overly narrow substring gates** on `f1-broker-status` (missed valid `AD-Active` / `synced` phrasings) and `a2-deletemsgs-say-yes` followup (missed valid `purged` / `dropped` phrasings).

## Changes Made

- **`run-scenario.sh`**
  - Fabrication check now substring-matches against the raw concatenated `tool_result` content (fabrication universe), while `gt_jq` still produces the must-appear set (omission check). Two-directional check, two different universes.
  - Added `|| true` on the per-iteration `jq -r "$gt_jq"` call so a non-JSON tool_result (error payload from a wrong-VPN lookup) doesn't trip `set -e` inside the loop body.
  - Header docblock updated to describe the new fabrication semantics.
- **`README.md`** — reworded `ground_truth.jq` / `ground_truth.answer_regex` section to match the new two-universe check.
- **`read-list-queue-discards.json`** — added `setup.cmd` that polls the three F7 discard counters via SEMPv2 until each is non-zero (30 s timeout) before turn 1.
- **`read-list-rdps.json`** — doc note that `test-rdp` context mentions won't trip the fabrication check now.
- **`f1-broker-status.json`** — added `"active"`, `"synced"` to `required_substrings_any_of`.
- **`a2-deletemsgs-say-yes.json`** — added `"purged"`, `"dropped"` to followup `required_substrings_any_of`.
- **`b1-select-clear-the-queue.json`** — removed the redundant turn-2 `"no"` followup; `a3-delete-queue-say-no` already covers the decline-after-prompt path. Doc updated to match.

## Testing

**Test Environment:**
- Go version: 1.25
- OS: Linux
- Broker version: 10.26.0

**Tests Performed:**
- [x] E2E tests added/updated (`test/e2e-llm` suite)
- [x] Tested with real Solace broker (dockerized broker-a/broker-b)

**Test Coverage:**
- Re-ran `test/e2e-llm` full suite (35 scenarios) against local brokers.
- Isolated the `f3-subscriptions` silent-exit with `bash -x` + temp debug logging and traced the trip to the compound assignment on the jq-fails path; `|| true` fix confirms the loop now completes deterministically on wrong-VPN sequences.
- Verified `read-list-queue-discards` setup polling exits promptly once F7 warm-up completes and errors clearly if counters never accumulate.
- Broadened substring gates verified against sample answers observed during earlier flake investigation.
- Verified changed scenarios with `test/e2e-llm/run-flake-check.sh` 

## Breaking Changes

None. Scenario schema and CLI are unchanged.

## Additional Notes

- LLM sampling is non-deterministic; scenarios where the model reasonably splits across paraphrases (`operational` vs `active`) will still occasionally trip narrow gates. Follow-up worth considering: per-scenario retries in `run-all.sh` (e.g. up to 3 attempts) so a genuinely broken scenario still fails all attempts, while single-sample flakes are absorbed. Not in this PR.
- The fabrication universe is now the raw `tool_result` string, so a scenario whose `answer_regex` is very loose (e.g. matches many similarly-named entities) could pass through fabrications that appear elsewhere in unrelated tool output. Existing scenarios all constrain the regex tightly enough for this not to matter.

## Related Issues/PRs

- Related to SOL-151828 (prior E2E LLM tooling coverage work)

## Additional Info
- Follow up: After testing with Github GHA, we might want to incorporate a retry system where we retry failed scenarios 2-3 times to see if its A) a flakey test or B) an actual issues


test/e2e-llm/run-scenario.sh
```diff
@@ -453,33 +461,57 @@ run_assertions() {
...
+            local raw
+            raw=$(jq -r --arg id "$id" '
+                select(.type=="user") | .message.content[]?
+                | select(.type=="tool_result" and .tool_use_id==$id)
+                | .content
+            ' "$run_file")
+            haystack+="$raw"$'\n'
+            expected_set+=$(jq -r "$gt_jq" <<<"$raw" 2>/dev/null || true)$'\n'
+        done <<<"$expected_ids"
+        expected_set=$(echo "$expected_set" | grep -v '^$' | sort -u)
```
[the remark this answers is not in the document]
helen-zhu-solace: Resolved in [2ee3a6c](https://github.com/SolaceDev/solace-broker-mcp/pull/197/commits/2ee3a6cfccc03e34f60fcfe96689da1e21ceb362)

solace-wmaryam:
Nice work on this; each flake is documented with its root cause and repro rate, which made reviewing straightforward. The core `set -e` / `pipefail` fix is correct, and the refactor of how ground-truth is extracted is functionally equivalent to before. Approving after two small notes, neither is a blocker.

**1. The relaxed fabrication check may now be too forgiving.**
The new check passes an entity if its name appears anywhere in the raw tool_result text. Because some names are fragments of others (`test-rdp` is contained inside `test-rdp-failing`), the assistant could name a non-existent `test-rdp` and it would never be flagged. This is a reasonable trade to kill the false failures for now, so I would not block on it, but it is worth a follow-up ticket. Cheapest fix: match on the whole quoted value (search for `"test-rdp"` rather than `test-rdp`), which closes the fragment hole with a one-line change and no per-scenario config.

Also minor: this semantic change affects all five scenarios that use `ground_truth.jq` (`f1-list-vpns`, `f3-subscriptions`, and `read-list-brokers` in addition to the two you documented). Behavior is uniform so no action needed, just flagging that the doc updates were not exhaustive.

**2. The `b1` doc note points at the wrong sibling test.**
The `_doc` justifies removing the turn-2 followup with "A2 already covers confirm-gate honoring," but A2 is the say-yes path. The decline-after-prompt case is actually covered by `a3-delete-queue-say-no`. The removal itself is fine and genuinely redundant, but consider pointing the note at `a3` so the next reader does not go looking in A2.

test/e2e-llm/scenarios/f1-broker-status.json
```diff
@@ -2,6 +2,6 @@
   "_doc": "F1 positive: broker-level prompt with no VPN/queue/client name should route to get-broker-status (not list-vpns, not get-vpn-status). Broker is up, fixtures are healthy, so the model must report operational state — catches a model that hedges or invents capacity issues from a clean snapshot.",
   "prompt": "What's the overall status of $BROKER?",
   "expected_tool": "mcp__solace-broker__get-broker-status",
-  "required_substrings_any_of": ["operational", "running", "healthy", "is up"],
+  "required_substrings_any_of": ["operational", "running", "healthy", "is up", "active", "synced"],
```
solace-aross: Gates match with `grep -qi` and are unanchored (run-scenario.sh:419), so `active` also matches inside `inactive` and `synced` inside `unsynced` / `not synced`. Neither appears in this scenario's `forbidden_substrings`, so a model that misreports the healthy broker as "inactive" or "not synced" now passes the check f1 exists to catch (a model that hedges or invents trouble from a clean snapshot). Suggest anchoring the added phrasings (`is active`, `in sync`, or `AD-Active`) or adding `inactive` / `unsynced` / `not synced` to `forbidden_substrings`.

test/e2e-llm/scenarios/read-list-queue-discards.json
```diff
@@ -1,7 +1,8 @@
...
   "prompt": "Are any queues in the default VPN dropping messages on $BROKER?",
   "expected_tool": "mcp__solace-broker__list-queue-discards",
+  "setup": {"cmd": "for _ in $(seq 1 30); do [ \"$(semp_curl --max-time 5 -sf \"$BROKER_URL/SEMP/v2/__private_monitor__/msgVpns/$BROKER_VPN/queues/test-queue-discards-spool\" | jq -r '.data.maxMsgSpoolUsageExceededDiscardedMsgCount // 0')\" -gt 0 ] && [ \"$(semp_curl --max-time 5 -sf \"$BROKER_URL/SEMP/v2/__private_monitor__/msgVpns/$BROKER_VPN/queues/test-queue-discards-ttl\" | jq -r '((.data.maxTtlExpiredDiscardedMsgCount // 0) + (.data.maxTtlExpiredToDmqMsgCount // 0) + (.data.maxTtlExpiredToDmqFailedMsgCount // 0))')\" -gt 0 ] && [ \"$(semp_curl --max-time 5 -sf \"$BROKER_URL/SEMP/v2/__private_monitor__/msgVpns/$BROKER_VPN/queues/test-queue-lowprio-congestion\" | jq -r '.data.lowPriorityMsgCongestionDiscardedMsgCount // 0')\" -gt 0 ] && exit 0; sleep 1; done; echo 'setup: F7 discard counters did not accumulate after 30s — check that setup-fixtures.sh ran' >&2; exit 1"},
```
solace-aross: `semp_curl` is a bare `curl -K -` with no `--connect-timeout` or `--retry` (lib.sh:116), and setup runs with no `timeout` wrapper (run-scenario.sh:232). So this loop is bounded by iterations, not wall clock. Against a slow or wedged broker each of the 30 iterations can spend up to 3x5s in curl plus the 1s sleep, roughly 8 minutes before `exit 1`, not the 30s the doc states. That turns a degraded broker into a multi-minute per-run stall, which cuts against the flake-reduction goal. Suggest a wall-clock bound (`end=$((SECONDS+30)); while [ "$SECONDS" -lt "$end" ]; do ...`) or wrapping setup in `timeout 45`, plus `--connect-timeout` so an unreachable host fails fast. Minor: on a failed curl the substitution is empty and `[ "" -gt 0 ]` prints `integer expression expected` each iteration; defaulting to 0 on curl failure quiets it.

test/e2e-llm/run-scenario.sh
```diff
@@ -453,33 +461,57 @@ run_assertions() {
...
         log_info "$label expected set: $(echo "$expected_set" | tr '\n' ' ')"
         log_info "$label answer set:   $(echo "$answer_set" | tr '\n' ' ')"
         missing=$(comm -23 <(echo "$expected_set") <(echo "$answer_set"))
-        extra=$(comm -13 <(echo "$expected_set") <(echo "$answer_set"))
         [ -n "$missing" ] && fail "$label: answer omitted entities: $(echo "$missing" | tr '\n' ' ')"
-        [ -n "$extra" ]   && fail "$label: answer fabricated entities: $(echo "$extra" | tr '\n' ' ')"
+        extra=""
+        while IFS= read -r name; do
+            [ -z "$name" ] && continue
+            grep -qF -- "$name" <<<"$haystack" || extra+="$name "
```
solace-aross: The raw-tool_result haystack with `grep -qF` is the right instinct for the adjacent-entity case, but it also passes a fabricated name that is a substring of a real one. Concrete case in this repo: `read-list-queue-discards` uses `answer_regex: test-queue-[a-zA-Z0-9_-]+`, so a model naming `test-queue-discards` (not a real fixture) matches inside `test-queue-discards-spool` and is not flagged. The Additional Notes justify the raw universe as bounded by tight regexes, but that scenario's regex is not tight enough for this prefix case. Consider matching answer names against an extracted token set (same regex, or a `fabrication_jq`, applied over the haystack) with a whole-token match rather than a raw-blob substring. Keeps the sibling-entity tolerance and restores token-level precision. Non-blocking if you would rather tighten the regex per scenario instead.

solace-aross (changes requested):
Solid flake fix. The silent-exit root cause (the missing `|| true` on the per-iteration jq) is correctly diagnosed, the two-universe split is the right shape, and fixing the discards race with a setup poll beats loosening the assertion.

Holding on a few tightenings before merge, all cheap. This is an eval harness, so a false PASS that hides a model regression is the failure mode I weight most, and two of these changes open one.

**Blocking**
- f1 `active` / `synced` gate matches inside `inactive` / `unsynced` (inline).
- discards `setup.cmd` has no wall-clock bound; worst case is minutes, not the 30s the doc claims (inline).

**Worth tightening, your call**
- fabrication check can pass a name that is a prefix of a real entity (inline).

**Non-blocking, flagging so they are not lost**
- The discards `setup.cmd` inlines the F7 readiness contract (queue names, SEMP fields, the TTL three-field sum) that `test/e2e-monitoring/verify-fixtures.sh` already owns, across the boundary this suite sources from. A named `wait_for_discard_fixtures` in the shared helpers would single-source it and match the existing `setup.cmd` convention (b1/a2 both call a named function).
- On a fabrication FAIL the log shows the name but not the haystack it was tested against, and `RUN_FILE` is deleted on exit. Logging the names present in the tool_result (same regex over the haystack) would make a failure reproducible without a rerun of a non-deterministic eval.
- The setup timeout message blames `setup-fixtures.sh` for every failure mode including an unreachable broker, and does not say which of the three counters is still zero.

Nice work overall.

helen-zhu-solace:
The first non blocking comment left from @solace-aross, Jira task has been opened for it: https://sol-jira.atlassian.net/browse/SOL-152176

test/e2e-llm/scenarios/f1-broker-status.json
helen-zhu-solace: Resolved in [dac8170](https://github.com/SolaceDev/solace-broker-mcp/pull/197/commits/dac81701aea4321374acf6694ea9492b66a2f28b)

test/e2e-llm/run-scenario.sh
helen-zhu-solace: Resolved in [dac8170](https://github.com/SolaceDev/solace-broker-mcp/pull/197/commits/dac81701aea4321374acf6694ea9492b66a2f28b)

test/e2e-llm/scenarios/read-list-queue-discards.json
helen-zhu-solace: Resolved in [dac8170](https://github.com/SolaceDev/solace-broker-mcp/pull/197/commits/dac81701aea4321374acf6694ea9492b66a2f28b)

solace-wmaryam:
Looks good overall! There's still the minor comment regarding the doc note left to address but that one's not a big deal.

solace-wmaryam (dismissed):

helen-zhu-solace:
@solace-wmaryam _doc comment has been addressed in [9629f87](https://github.com/SolaceDev/solace-broker-mcp/pull/197/commits/9629f8756e244af2409d118a3e79efd2c1eb5461), thanks!

solace-wmaryam (approved):

solace-aross (approved):
Approving; this clears my earlier changes-requested. All three items I flagged are resolved, and cleanly:

- f1: the substring trap is closed. Gate moved to AD-Active / in sync, and inactive, unsynced, not synced, not active are now forbidden, so a degraded broker can't pass on a substring of the healthy token.
- discards setup: a real SECONDS-based wall-clock deadline with --connect-timeout 3, bounded to ~30s instead of the old worst-case minutes, and the failure line now reports which counter is still zero.
- fabrication check: whole-token set diff via comm against the gt_regex-extracted haystack, so a prefix like test-queue-discards no longer slips past test-queue-discards-spool.

You also picked up two of the non-blocking notes: the haystack is logged on a fabrication FAIL, and the timeout message no longer pins every failure on setup-fixtures.sh. CI is green across all e2e suites. Good to merge.

PR #200, opened 2026-07-22
SOL-152203: Remove unused pointers to SEMP schema from write tool des…

solace-wmaryam:
…criptions

## Summary

The create/update tool descriptions do not currently enumerate all the object attributes that can be configured and instead mention referring to the corresponding SEMP schema to see the full list. Since the end-user is likely to be an agent/LLM, the pointer to the SEMP schema provides no useful information as the LLM does not know where to access this data. This PR refines the tool descriptions by removing these lines from the write tools. 

Fixes #https://sol-jira.atlassian.net/browse/SOL-152203

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Changes Made

- Remove the 'See the SEMP schema line' from the create/update tools for message-vpns, queues, topic-endpoints, and rdps

## Testing

Ran create queue prompts 5 times using Sonnet 4.6 with and without the SEMP schema line in the tool description and didn't see any alarming differences. The most noticeable change was that the LLM provided less suggestions to the user on its own accord, which I would consider a good thing.

## Breaking Changes

N/A


solace-awaheed (dismissed):
LGTM

solace-aross (dismissed):
Approving. Removing the SEMP schema pointer is the right call: an LLM consumer can't act on it, and each description still ends in "etc.", so the "there's more" signal survives. Nothing actionable is lost.

Confirmed these were the only such pointers in the file, and nothing in the Go tests or golden files asserts on the removed text, so nothing breaks.

Non-blocking thought for later: if we ever want agents to discover the full attribute set, the answer is a discovery mechanism (a resource or a describe-schema tool), not prose in the description. Worth a ticket if it comes up, not this PR.

solace-awaheed (approved):
LGTM

PR #201, opened 2026-07-22
fix(semp): reject empty and dot-segment SEMP path parameters [SOL-152208]

solace-aross:
## What

Rejects empty and dot-segment (`.`, `..`) values for SEMPv2 path parameters in `buildURL`, before any HTTP request is issued.

## Why

`buildURL` (`internal/semp/sempv2/client.go`) substituted path-parameter values with `url.PathEscape`, which escapes `/` but not `.` or `..`, and did no emptiness check. A value of `""`, `"."`, or `".."` could produce an empty or dot segment. On a broker or fronting proxy that normalizes dot-segments, that can collapse a targeted request onto an unintended (e.g. parent) path. This matters most for the destructive SEMP operations (`config/deleteMsgVpn`, `config/deleteMsgVpnQueue`, `action/doMsgVpnQueueDeleteMsgs`), and `tool_authorization` is off by default.

Because `PathEscape` makes each path-parameter value exactly one URL segment, `""`, `"."`, and `".."` are the complete set of dangerous values. The guard lives at the single choke point (`buildURL`), so it covers read and destructive operations uniformly.

## Changes

- `internal/semp/sempv2/client.go` — guard in `buildURL` rejecting `""`/`"."`/`".."` path-parameter values with a clear, parameter-named error.
- `internal/semp/sempv2/client_test.go` — `TestClient_Execute_RejectsUnsafePathParam`: the three unsafe values are rejected before any HTTP request (handler not called) with an error naming the parameter; a legitimate dotted value (`my.queue.v2`) is unaffected.

## Testing

- `go test ./internal/semp/sempv2/` passes; full `make check` clean.
- Revert-verified: the test fails with the guard removed, passes with it restored.

## Non-goals

Out of scope for this defense-in-depth fix (all resolve to harmless single literal segments under standard single-decode normalization): double-decoding proxies (`%252e`), whitespace/unicode dot lookalikes, and `nil` → `"<nil>"`.

Ticket: https://sol-jira.atlassian.net/browse/SOL-152208


solace-lootii (approved):
LGTM!

PR #203, opened 2026-07-22
SOL-152124: Add list-bridges and get-bridge-status monitoring tools

solace-lootii:
## Summary
- Adds `list-bridges` and `get-bridge-status` MCP tools for Bridge monitoring, following the existing `list-rdps`/`get-rdp-status` list-then-drill-down pattern
- `list-bridges` returns enabled state, inbound/outbound connection state, and last inbound failure reason per bridge in a VPN (default 100, max 500 via `maxResults`), with a summary block (`downCount`, `disabledCount`, `byInboundFailureReason`)
- `get-bridge-status` returns additional per-bridge diagnostic detail (connection establisher, failure category, client name) for one bridge, identified by `msgVpnName` + `bridgeName` + `bridgeVirtualRouter` — bridges are the only object in this server keyed by two names rather than one
- Docs updated (README, tools-reference, user-guide, CHANGELOG); read-only tool count now 19 (was 17)

## Known gap / follow-up
E2E fixture coverage in `test/e2e-monitoring` is intentionally deferred to a separate follow-up ticket — a real bridge fixture needs two brokers actively bridging to each other, which needs verification against a live broker rather than guessed at.

## Test plan
- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` — all 31 packages pass
- [x] New unit tests: postprocess handler (12 cases covering counts, outbound-only failures, not-applicable direction handling, truncation, empty input, errors, missing/nil fields), executor-level pagination/truncation + compound path-param substitution, YAML loader parsing for both tools
- [x] Self-reviewed diff for description accuracy and test clarity before pushing
- [ ] Live-broker E2E bridge fixtures (follow-up ticket)


internal/composite/postprocess/handlers/list_bridges.go
```diff
@@ -0,0 +1,132 @@
...
+			continue
+		}
+		if !enabled {
+			disabled++
+			continue
+		}
+		unhealthy := !bridgeInboundUpStates[inboundState] || !bridgeOutboundUpStates[outboundState]
+		if unhealthy {
+			down++
+			if inboundFailureReason != "" {
```
solace-aross: Confirmed, and blocking. When only outbound is down (inbound in an up state), this still buckets a non-empty `inboundFailureReason`, which SEMP can leave stale after an inbound recovery. That contradicts this handler's own contract ("active inbound failures only ... recovered bridges with stale reasons are excluded") and the tool description.

Gate the bucket on the inbound direction being the unhealthy one:

```go
if !bridgeInboundUpStates[inboundState] && inboundFailureReason != "" {
    byReason[inboundFailureReason]++
}
```

The test gap that hid this: `TestListBridges_OutboundOnlyFailure` uses an empty `inboundFailureReason`, so it passes either way. Add a case with inbound up, outbound down, and a non-empty (stale) `inboundFailureReason`, asserting `byInboundFailureReason` stays empty.

solace-aross (changes requested):
Strong PR overall. The list-then-drill-down shape mirrors `list-rdps` cleanly, the two-name bridge identifier is handled correctly end to end, and the robustness tests (skip-on-bad-row, truncation, empty) are thorough.

**One blocking item**, in the postprocess handler: `byInboundFailureReason` buckets a stale inbound reason when only the outbound direction is down. That contradicts the "active inbound failures only" contract stated in both the tool description and the handler comment, and it can misdirect an operator. Inline below, with the one-line fix and the test case that would have caught it.

Not blocking, worth fixing while you're here: the user-guide still reads "13 write tools (32 total)" while the README and tools-reference say 16 write / 35 total. Copilot flagged the same line. Reconciling it keeps the count story consistent across the docs this PR already updates.

Deferring the live-broker E2E bridge fixtures to a follow-up is the right call given they need two brokers actively bridging.

solace-aross:
Took a second, deeper pass through the composite postprocess and executor paths. The one blocking item stands: the inbound-reason bucketing in `ListBridges`. Everything else I chased is clean; recording it here so you don't have to re-derive it.

Cleared:
- **`get-bridge-status` selecting via an `args` string instead of the structured `select:` list.** Works as written. `applySelect` skips the join when a step has no structured `select:`, so the literal arg survives untouched, and the loader permits `args.select` on its own (it rejects only setting both). SEMPv2 accepts the spaced form, so the query on the wire matches the structured path. Fine for a `collect` tool. One forward note, not for this PR: if `get-bridge-status` ever moves to a `postProcess` strategy, switch it to the structured `select:` list so startup `ValidateTool` can cross-check the handler's fields.
- **The "max 500" claim in the docs and descriptions.** Correct. `resolveMaxResults` caps at 500 globally, so the bridge tools inherit it without declaring anything.

Once the bucketing is gated on inbound state, with a test covering inbound-up plus outbound-down plus a stale reason, this is an approve.

solace-aross (approved):
Approving. The inbound-reason fix is exactly right, and `TestListBridges_OutboundOnlyFailureWithStaleInboundReason` closes the gap that hid it: inbound up, outbound down, stale reason, asserting the bucket stays empty.

One non-blocking cleanup, since these commits reconciled the write-tool counts. Two prose lines still say "nine Config-API management tools", which now contradicts the corrected headers (12 management, 16 write):
- `docs/user-guide.md:166`
- `docs/tools-reference.md:93`

Both predate this PR. Fix here or in a quick follow-up; not gating the merge.

PR #205, opened 2026-07-22
SOL-151600: Process-level circuit breaker for IdP token exchange

amitmorade:
## Summary

Adds a process-level circuit breaker in front of the RFC 8693 token exchange to the IdP, so a sustained IdP outage fails fast instead of driving every broker's requests into the full retry/timeout budget. Configurable under `broker_oauth.circuit_breaker`, with production-safe defaults and an escape hatch.

Tracked under SOL-151600. (Jira, not a GitHub issue.)

## Type of Change

Not breaking: the whole Hop-2 broker-OAuth runtime is still gated behind `ENABLE_UNRELEASED_BROKER_OAUTH` and unactivated in the wild, and the `circuit_breaker` block is optional with a safe default (breaker on).

## Motivation and Context

The token-exchange layer already retries transient IdP failures (SOL-151520), but nothing bounds the *aggregate* load when the IdP is genuinely down. Every broker's request independently walks the full retry chain (~19s worst case) before failing. A shared circuit breaker detects sustained IdP unhealth and short-circuits, protecting both the MCP server and the IdP from a failure storm.

## Changes Made

Built in dependency order across commits:

- **Preserve failure cause through retry exhaustion.** The token-exchange layer collapsed 429, 5xx, network, and body-read failures into one `ErrExchangeTransport` sentinel, and the retries-exhausted rewrap severed even that. Added a `FailureClass` field on `ExchangeError`, carried through the rewrap, so the breaker can tell an outage (5xx/network — count) from a rate limit (429 — exclude). This is the load-bearing fix the breaker depends on.
- **Breaker classification.** Three pure, side-effect-free functions map an exchange outcome to gobreaker's success/failure/excluded verdicts. Server owns the policy: transient network/5xx/body-read failures and a whole retry chain that hangs until the deadline all count; 429, endpoint misconfiguration (bad TLS/hostname/DNS-not-found), cancellations, and local errors are excluded; a rejection counts as success (the IdP answered).
- **Config + defaults + validation**, then **wiring** the `sony/gobreaker/v2` breaker into `Exchange` — inside the singleflight function, so one logical exchange = one breaker outcome regardless of HTTP attempts, and a collapsed burst counts once. Open-state rejections map to a new `ErrExchangeCircuitOpen` sentinel (transient class, distinct from retries-exhausted).
- **YAML plumbing** under `broker_oauth.circuit_breaker` (nested there because the breaker only protects OAuth token exchange). `enabled: false` is the escape hatch; every field optional with a safe default.
- **Upper bounds** on the numeric config, single-sourced as exported constants shared by the config-layer and runtime validators, with operator errors that name the field, state the range, and echo the bad value. Named `IdP*` (the breaker protects the IdP, not a broker).
- **CHANGELOG** entry.

## Defaults, Bounds, and Their Rationale

Failure classification is server-owned (operators cannot reconfigure it); the values below are the tunable surface. Each default is anchored to what mainstream breakers ship (Hystrix, Resilience4j, Polly, Envoy outlier detection, sony/gobreaker), then adjusted for this deployment's shape: **one shared IdP, one process-wide breaker, token-exchange volumes** (lower than a general RPC path).

| Field | Default | Lower bound | Upper bound | Rationale |
|---|---|---|---|---|
| `failure_rate_window` | `30s` | `10ms` | `1h` | Long enough to ride out a blip, short enough that a resolved outage stops influencing within ~30s. Deliberately longer than mainstream defaults (Hystrix 10s, Polly 30s, Envoy 10s) because token-exchange traffic is lower-volume. Floor is `10ms` because the breaker derives its bucket period as `window/10` (must stay non-zero). |
| `minimum_requests` | `10` | `1` | `1,000,000` | Sample floor before the rate rule may trip, so a 2-of-2 blip can't open it. Lower than Hystrix's 20 / Resilience4j's 100 on purpose — those protect high-volume RPC; token exchange sees less traffic, and the consecutive rule covers the low-traffic outage case. |
| `failure_rate_threshold_percent` | `50` | `>0` | `100` | Half of counted (non-excluded) ops failing is a clear outage signal. Mirrors Resilience4j's and Polly's failure-rate defaults exactly. |
| `consecutive_failure_threshold` | `5` | `0` (disables) | `1,000` | Fast-trips a full outage regardless of sample size — the low-traffic path. Matches gobreaker's and Envoy's consecutive-5xx default of 5. `0` explicitly disables the rule. |
| `open_state_duration` | `30s` | `>0` | `1h` | Time open before probing recovery. Between Hystrix's 5s and Resilience4j/gobreaker's 60s. A multi-replica fleet may raise it so replicas don't all probe at once. |
| `half_open_probe_requests` | `2` | `1` | `10` | Probes admitted in half-open; 2 requires two consecutive successes to close (stronger than a single trial) without bursting a fragile IdP. Between the common "1 trial call" and Resilience4j's 10. |

**On the bounds specifically:**
- **Lower bounds** reject values that would break the mechanism (a zero/negative window, a zero sample floor, a non-positive open duration).
- **Upper bounds are guardrails, not tuned ranges.** Only `half_open_probe_requests ≤ 10` is a domain-meaningful limit (10 = Resilience4j's default; more would flood a recovering IdP). The rest (`1h`, `1,000,000`, `1,000`) exist only to reject absurd/typo values — real deployments run far below them. The validation error deliberately does *not* steer operators toward good values (that belongs in operator docs); it only fires when a value is out of the sane range, and when it does it names the field, states the range, and echoes the offending value.
- **`consecutive_failure_threshold` is deliberately NOT coupled to `minimum_requests`.** The two trip rules are independent (a run of failures vs. a rate over a sample); tying them would break the low-traffic case the consecutive rule exists for.

Defaults are exposed via a `DefaultCircuitBreakerConfig()` **function** (not a package var) so each caller gets a fresh, unmutable copy. The ceiling constants are single-sourced (`config.Max/MinIdP*`) and shared by both the config-layer YAML validator and the runtime validator, so the two can't drift.

### YAML shape

**The breaker is on by default.** It requires no configuration action: whenever the Hop-2 OAuth runtime is active (feature flag set, `broker_oauth` present, and a broker on `auth.mode: oauth`), the breaker is constructed with the shipped defaults automatically. The `circuit_breaker` block only exists to *tune* or *disable* it — omitting the block, or any individual field, keeps the safe defaults. (It does nothing in a deployment that isn't using broker-OAuth, since there's no token exchange to protect.) Every field below is shown with its default:

```yaml
broker_oauth:
  # ... existing IdP coordinates (idp_token_endpoint, mcp_server_client_id, ...) ...
  circuit_breaker:
    enabled: true                            # false disables the breaker (retries still run); logs a WARN
    failure_rate_window: 30s                 # rolling window the failure rate is computed over        [10ms .. 1h]
    minimum_requests: 10                      # min classified ops in the window before the rate rule can trip [1 .. 1_000_000]
    failure_rate_threshold_percent: 50        # failure % (of counted ops) that opens the breaker       (0 .. 100]
    consecutive_failure_threshold: 5          # consecutive failures that trip immediately; 0 disables  [0 .. 1000]
    open_state_duration: 30s                  # how long the breaker stays open before probing recovery (0 .. 1h]
    half_open_probe_requests: 2               # probes admitted while testing recovery                  [1 .. 10]
```

Not operator-configurable (server-owned): the failure classification itself, the derived rolling-bucket granularity (`window/10`), and the breaker name. Out-of-bounds values are rejected at config load with an error naming the field, the valid range, and the offending value.

## Design Decisions Made During Implementation

These are the structural judgment calls made while building this, each with the short rationale we settled on:

- **Breaker wraps *inside* the singleflight function, not around it.** One logical exchange must record exactly one breaker outcome; singleflight already collapses a concurrent burst of identical calls into one real IdP call, so wrapping outside would record N outcomes for one call and inflate the counters. Inside, only the burst winner reaches the breaker.
- **`FailureClass` is a field on `ExchangeError`, not a set of new sentinels.** The coarse `ErrExchangeTransport` sentinel must stay coarse (AgentMessage and the `retryable` flag treat all transport causes identically). A new orthogonal field carries the sub-cause (network / 5xx / 429 / body-read) — and is copied through the retries-exhausted rewrap, which replaces the sentinel and would otherwise sever the cause.
- **Endpoint misconfiguration is excluded from the breaker, distinguished from transient network faults.** `situations.md` says an untrusted/expired TLS cert, a hostname mismatch, or a DNS name-not-found must *not* count — they are permanent operator faults, not IdP outages, and counting one would trip the shared breaker for every tenant with no chance to heal. Go's stdlib surfaces these as `errors.As`-able types (`x509.UnknownAuthorityError`/`CertificateInvalidError`/`HostnameError`, `net.DNSError{IsNotFound}`), so `doExchange` classifies them into a new `FailureClassConfig` (excluded). A DNS *timeout* (`net.DNSError{IsTimeout}`) is deliberately left as a transient network fault and still counts. This was a spec/impl contradiction caught in review — the code previously counted all transport errors — now resolved in the spec's favor.
- **New `ErrExchangeCircuitOpen` sentinel, not reusing `ErrExchangeRetriesExhausted`.** "The breaker refused to try" is a different fact from "we tried and gave up." Both are transient-class for the agent (same back-off advice), but distinct in logs/metrics for diagnosis.
- **A timed-out retry chain is classified `Network`, so it counts as a breaker failure.** When the whole chain hangs until the chain-deadline fires, the rewrap produces `ErrExchangeRetriesExhausted` with *no* underlying `*ExchangeError` — leaving `FailureClassNone`. Because the breaker classifies transport outcomes solely by `FailureClass`, that would fall through as a *success* and dilute the failure rate during exactly the sustained IdP slowness the breaker exists to catch. Fixed on the `FailureClass` axis (the deadline path now sets `FailureClassNetwork` — a hung chain *is* "no usable response received") rather than by adding a sentinel check to the breaker, keeping it single-axis. Caught by Copilot review.
- **Two config types, one per layer.** `config.IdPCircuitBreakerConfig` (YAML-tagged schema, owns parsing + validation messages) is translated by `FromConfig` into `tokenexchange.CircuitBreakerConfig` (runtime, `time.Duration` fields). Same schema→runtime split already used for `grant_type` and `audience_parameter_name`.
- **Schema fields are pointers (`*uint32`, `*time.Duration`, …).** So the translator can distinguish "operator omitted this field → take the default" from "operator explicitly set the zero value → validate/reject it." A non-pointer zero can't tell those apart.
- **`enabled: false` maps to a nil breaker, not a disabled-flag on the breaker.** The escape hatch turns the breaker off entirely (retries still run); representing "off" as the absence of a breaker keeps the hot path free of an `if enabled` check and is why `enabled` isn't on the runtime struct.
- **Ceiling/floor constants live in `config`, reused by `tokenexchange`.** The dependency runs `tokenexchange → config` (never the reverse), so the single source of truth must sit in `config`; putting it in `tokenexchange` would force an import cycle. Exported as `config.Max/MinIdP*`.
- **Named `IdP*`, not `Broker*`.** The breaker protects the IdP token endpoint. Its sibling types are `Broker*` (a subsystem-naming convention), but "broker circuit breaker" reads as "breaker for the broker," which is backwards — it breaks on IdP health.
- **Validation runs at config load, in both validators.** A bad value fails at startup (inside `validateBrokerOAuthConfig`, joined with all other config errors) rather than surfacing later at Exchanger construction. The runtime `Validate()` is the final gate for callers that build the config directly.
- **`OnStateChange` logs at WARN with `slog.Warn` (no context).** It's the one breaker callback allowed to do work — it runs off the lock the classification callbacks hold, so logging there is safe; the other callbacks stay side-effect-free. No context is available in gobreaker's callback signature, and a state transition is process-global, not request-scoped.
- **Circuit-open logs a structured `breaker_state="open"`, not a `failure_class`.** A fast-failed (open-breaker) call makes no IdP request, so it carries no `failure_class` or HTTP status — leaving it identifiable only by substring-matching the free-text `detail`. `LogAttrs` emits `breaker_state="open"` when the sentinel is `ErrExchangeCircuitOpen`, giving operators a structured field to filter/alert on. `failure_class` is deliberately *not* reused: it classifies a transport failure that occurred, and here none did.

## Testing

**Test Environment:**
- Go version: 1.25
- OS: darwin

**Tests Performed:**
- [x] Tested locally with `go test -race ./...` (via `make check` — build, vet, golangci-lint, race tests all green)
- [ ] Tested with real Solace broker (unit/integration only; the Hop-2 runtime is flag-gated and not yet exercised against a live IdP here)

**Test Coverage:**
- One-logical-operation-counts-once (a 5xx chain that retries 3× records one breaker failure, not three).
- 429 excluded from the failure rate; rejection counts as success.
- Open-state fast-fail without touching the IdP; open-state error is `ErrExchangeCircuitOpen`, transient-classified, enriched.
- Escape hatch (`enabled: false`) disables the breaker while retries still run.
- **Concurrency** (shared process-wide singleton): distinct-key concurrent failures record exactly N (counter integrity, start-barrier + `-race`); a concurrent stampede against an already-open breaker is rejected cleanly with no raw-library-error leak and no IdP leak.
- Config validation: lower and upper bounds, presence semantics, and that the operator error carries field path + valid range + offending value.
- Config-fault exclusion: each transport error type maps to the right class (incl. the DNS timeout-counts vs. name-not-found-excluded split), and end-to-end an untrusted-TLS `httptest` IdP leaves the breaker closed across repeated attempts even when set to trip on a single counted failure.
- Logging: `breaker_state="open"` is emitted on circuit-open errors (and survives enrichment) but omitted otherwise, and such errors carry no `failure_class`.

## Breaking Changes

None.

### Code Quality
- [x] Self-review completed
- [x] Code is well-commented (explains "why")

### Security
- [x] No credentials in code
- [x] Followed secure logging rules (ran `/check-logs`; breaker logs emit only fixed names/states, no request content)

### Testing
- [x] All tests pass locally: `make check`
- [x] Edge cases and error paths covered

### Documentation
- [x] godoc comments added/updated for exported symbols
- [x] CHANGELOG.md updated
- [ ] Operator config reference (follow-up — see Additional Notes)

### Git & CI
- [ ] DCO signoff — intentionally omitted per repo maintainer's standing convention; happy to add if required
- [ ] CI checks passing — draft; will confirm once CI runs

## Additional Notes

**Draft** for early review of the approach before finalizing.

Two deliberate design calls worth a reviewer's eye:
- The breaker sits **inside** the singleflight function so stampede-collapsed calls record one outcome, not N. Placing it outside would inflate the counters.
- The config upper bounds are **guardrails, not tuned ranges** — they reject absurd/typo values (a multi-hour window, a billion minimum-requests) but do not steer operators toward good values (real windows are 10-30s). Guiding operators to sensible values belongs in operator docs, not the validator. That operator-config doc (incl. the "one logical operation" definition and the open-duration vs ~19s chain-deadline interaction) is the main follow-up.

## Related Issues/PRs

- Builds on the token-exchange retry loop (SOL-151520).

---

**For Reviewers:**

**Focus Areas**: (1) the singleflight-inside placement and its concurrency implications; (2) the `FailureClass` carry-through the retries-exhausted rewrap; (3) whether the guardrail bounds and their framing are right.


internal/tokenexchange/defaults.go
helen-zhu-solace: Small nit - stale reference: `internal/tokenexchange/defaults.go:108–112` — says the override parameter is a seam for "a future YAML … override support later," but this PR ships that YAML surface. Reword to describe the current mechanism.

internal/tokenexchange/circuitbreaker_exchange_test.go
```diff
@@ -0,0 +1,521 @@
...
+)
+
+// fastTripBreakerConfig trips after a single consecutive failure and stays
+// open long enough that a follow-up call in the same test is rejected without
+// racing the recovery timer. Small numbers keep the tests fast and
+// deterministic.
+func fastTripBreakerConfig() CircuitBreakerConfig {
+	c := DefaultCircuitBreakerConfig()
+	c.ConsecutiveFailureThreshold = 1
+	c.OpenStateDuration = time.Hour
```
solace-aross, line 39: Blocking: the recovery half of the state machine has no coverage. This sets `open_state_duration` to 1h specifically to avoid the recovery timer, and every other trip test does the same, so open to half-open to closed, half-open back to open on a failed probe, and the `half_open_probe_requests` semantics (the default of 2 needing two consecutive successes to close) are never exercised. Recovery is the defining behavior of a breaker. A test with a short open duration driving open to half-open to closed and half-open to open, asserting two probes are required, would close this.

internal/tokenexchange/circuitbreaker_exchange_test.go
```diff
@@ -0,0 +1,521 @@
...
+	//   - consecutive rule disabled (0),
+	//   - rate rule starved (minimum requests unreachable),
+	//   - threshold pinned high too, so no default can surprise the test.
+	// And pin the counting window far longer than the test runs: if the
+	// rolling window rotated mid-test, old buckets would age out and Counts()
+	// could read < N with no bug in the breaker — a latent flake. 1h guarantees
+	// a single window for the whole run.
+	cfg := DefaultCircuitBreakerConfig()
+	cfg.ConsecutiveFailureThreshold = 0
+	cfg.MinimumRequests = 1_000_000 // the max allowed; far above N so the rate rule never trips
```
solace-aross, line 381: The failure-rate trip rule is never exercised: `minimum_requests` is pinned to 1,000,000 here to isolate the consecutive rule, and `newReadyToTrip` has no direct unit test. At production volume the rate rule is the primary trip mechanism. A table test over `gobreaker.Counts` (e.g. 9 evaluated -> no trip; 10 at 50% -> trip; a window of exclusions -> no trip) covers it cheaply.

internal/tokenexchange/circuitbreaker.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+
+	"github.com/sony/gobreaker/v2"
+)
+
+// breakerName labels the breaker in logs and metrics. Internal only — there is
```
solace-aross: This says the breaker is labeled "in logs and metrics", but no metric is emitted anywhere in the package and gobreaker/v2 emits none on its own. Either drop "and metrics" or wire a breaker-state metric. See the review body for the operability follow-up.

internal/tokenexchange/circuitbreaker.go
```diff
@@ -0,0 +1,157 @@
...
+			return false
+		}
+
+		failureRate := float64(counts.TotalFailures) / float64(evaluated) * 100
+		return failureRate >= cfg.FailureRateThresholdPercent
+	}
+}
+
+// logBreakerStateChange is the one breaker callback allowed to do work: it is
+// not on the counter-classification path, so it runs outside the lock the
```
solace-aross: Not accurate: in gobreaker v2.4.0, `OnStateChange` fires from `afterRequest` while `cb.mutex` is held, so `logBreakerStateChange` runs under the lock, not outside it. Harmless today since it only logs, but the comment invites a future maintainer to add real work here believing it is off-lock. Correct the comment and keep this callback to cheap logging.

internal/tokenexchange/circuitbreaker.go
```diff
@@ -0,0 +1,157 @@
...
+// The success type is *Token because that is what Exchange's protected section
+// returns. BucketPeriod is derived from the window (never operator-set) so the
+// failure rate is computed over ~10 rolling buckets.
+func newTokenExchangeCircuitBreaker(cfg CircuitBreakerConfig) *gobreaker.CircuitBreaker[*Token] {
+	return gobreaker.NewCircuitBreaker[*Token](gobreaker.Settings{
+		Name:          breakerName,
+		Interval:      cfg.FailureRateWindow,
+		BucketPeriod:  cfg.FailureRateWindow / 10,
+		Timeout:       cfg.OpenStateDuration,
+		MaxRequests:   cfg.HalfOpenProbeRequests,
```
solace-aross, line 40: Half-open behavior worth confirming: gobreaker gates half-open admission on `validRequests = Requests - TotalExclusions`. An excluded outcome (a 429) refunds its own probe slot, so a recovering-but-throttling IdP admits unbounded sequential probes here, which partly defeats the probe budget this knob exists for. Defensible under the "429 = IdP up" policy, and instantaneous concurrency is still capped, but please confirm it is intended and note it.

solace-aross (changes requested):
Reviewed across four lenses: security, concurrency, classification/config, and architecture. The implementation is verified sound. Classification holds end to end, the singleflight-collapse-counts-once and open-state-rejection paths are correct against gobreaker v2.4.0, config bounds are exact, the escape hatch is right, and the "hung chain reads as success" bug from the earlier pass is genuinely fixed. Security is clean, CI is green including e2e-oauth. Nicely done on a tricky primitive.

Requesting changes on test coverage and comment accuracy, not on correctness. The recovery half of the state machine (open to half-open to closed) and the failure-rate trip rule have no behavioral coverage, and for a circuit breaker that recovery behavior is the part a test most needs to pin, especially given the half-open probe-budget subtlety flagged inline. Two code comments are also factually wrong. Details inline.

Non-blocking, worth a follow-up ticket: there is no alertable metric to separate "breaker open" from "IdP down" for a failure mode that fast-fails the whole fleet for 30s. The log signal is well-designed but only supports a query, not a threshold alert. The feature is still flag-gated, so tracking it is fine.

`types.go`'s package doc is stale: it still says the Exchanger does not cache, does not retry, and classifies into three sentinels, all three now false. It sits just outside this diff so I could not anchor it inline, but the file is edited here, so worth fixing in the same pass.

If the team would rather land now, at minimum I would want the two comment fixes and the half-open question resolved here, with the recovery and rate-rule tests tracked and closed before `ENABLE_UNRELEASED_BROKER_OAUTH` is flipped.

solace-aross:
Follow-up after a deeper pass on the breaker's trip mechanics. Staying with changes-requested, and the reason is now sharper than "recovery is untested."

### One new finding, and it strengthens the test-coverage ask

The consecutive-failure rule is window-bound, so it will not trip at the low-traffic regime its own docs say it exists for.

`BucketPeriod` is set to `window/10`, so gobreaker rolls the counter, and `subtract()` decays `ConsecutiveFailures` as buckets age out (`counter.go:127-196`). A single deadline-bound failure can take the full ~19s chain budget. So when counted failures arrive more than ~`window/5` (≈6s) apart, at most ~2 consecutive failures ever coexist: the 5-consecutive rule never fires, and the rate rule stays below `minimum_requests`. Net effect at low concurrency with slow/hanging failures: the breaker never opens. A hard connection-refused IdP fails fast and still trips, so the gap is specifically the slow-IdP regime.

It fails open, so it is not dangerous, but `circuitbreaker.go:48-52` and the "fast-trips at low traffic" framing describe the opposite of what the library does. This is exactly why the recovery and rate-rule tests I flagged matter: only a time-controlled trip/recovery test surfaces it. Either track consecutive failures outside gobreaker, or correct the docs and state the regime gap plainly.

### Resolved since my last pass, no action

The chain-deadline path now sets `FailureClassNetwork` (`exchange.go:259-261`) and `TestClassifyRetryOutcome_DeadlinePathClassifiesAsNetwork` pins it. Copilot's repeated "counts as success" flag is stale; no need to chase it.

### Small fixes to fold into the doc pass already requested

- `errors.go:30-32`: the `Sentinel` list omits `ErrExchangeCircuitOpen`, which this PR adds and routes through `ExchangeError`.
- `from_config.go:29-30`: "the only file in the package that imports internal/config" is no longer true; `circuitbreaker_config.go` imports it too.
- `defaults.go:83-84`: "a fleet may raise it so replicas don't all probe at once" does not hold. `open_state_duration` maps to gobreaker's `Timeout`, which has no jitter, so a uniform larger value shifts the recovery herd rather than de-correlating it. Add per-replica jitter or drop the note.
- On the off-lock comment (`circuitbreaker.go:70`): since `OnStateChange` fires under `cb.mutex`, the fix is not only the wording. The `slog.Warn` serializes all breaker traffic while it runs. Consider dispatching the log off the callback.

The approach is right and the hard parts hold: classification end to end, singleflight-inside so a burst counts once, the concurrency tests, and the config-fault exclusion that keeps one tenant's typo from tripping the shared breaker. The gate is the trip and recovery behavior, plus the docs that describe it.

amitmorade:
Thanks for the thorough review @solace-aross — all addressed across two commits.

**Tests (b97b6fa):**
- Recovery state machine: open → half-open → closed (two consecutive probe successes required, one asserted insufficient), half-open → open on a failed probe, and over-budget half-open calls rejected with `ErrExchangeCircuitOpen` without touching the IdP. gobreaker has no injectable clock, so these use a real 250ms open window — but the open → half-open transition is lazy (evaluated inside `State()`) and half-open never self-expires, so polling can't miss it; slow CI delays, never flakes. Green under `-race` at `-count=10`.
- Rate rule: `TestNewReadyToTrip`, 11-case table over `gobreaker.Counts` — sample floor, 50% boundary (`>=`), exclusions neither fill the floor nor dilute the rate, and consecutive-rule-before-floor ordering.
- Half-open 429: confirmed intended ("429 = IdP is up"; probes are traffic-driven with concurrency still capped) and pinned by `TestBreaker_HalfOpenExcludedProbeRefundsSlot`.

**Docs (35e134d):**
- Consecutive-rule regime gap: you're right — the counter is window-bound and decays with the buckets, so it only trips fast-failing outages. Stated plainly at `newReadyToTrip` and corrected everywhere that claimed "fast-trips at low traffic" (defaults, field docs, validator, CHANGELOG). Regime fails open, so the mechanism fix is out of this PR.
- Comment fixes: `OnStateChange` runs *under* `cb.mutex` (kept synchronous — one rare WARN, and async would lose transition ordering); dropped "and metrics"; sentinel list now includes `ErrExchangeCircuitOpen`; `from_config.go` importer claim removed; jitter note corrected; stale `types.go` package doc rewritten.

**Follow-ups filed** (to close before the flag flips):
- [SOL-152284](https://sol-jira.atlassian.net/browse/SOL-152284) — alertable breaker-state metric
- [SOL-152285](https://sol-jira.atlassian.net/browse/SOL-152285) — honor `Retry-After` on 429
- [SOL-152286](https://sol-jira.atlassian.net/browse/SOL-152286) — consecutive-failure tracking outside gobreaker (the regime-gap mechanism fix)

[SOL-152284]: https://sol-jira.atlassian.net/browse/SOL-152284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

internal/tokenexchange/defaults.go
amitmorade: Good catch that this reads stale — reworded in 88b51fe. One clarification though: this seam is actually still future, not shipped. The comment is on `ComputeChainDeadline`, whose override would come from a `broker_oauth.retry.chain_deadline` key — and the retry knobs have no YAML surface (this PR ships `broker_oauth.circuit_breaker`, not `broker_oauth.retry`; no production caller passes a positive override, only tests via `Params.ChainDeadline`). The old wording made it easy to conflate the two now that a sibling YAML block exists in the same area, so it now names both explicitly.

solace-aross (approved):
Everything I gated on is resolved, and cleanly.

Recovery now has real behavioral coverage: open to half-open to closed with the two-consecutive-probe requirement asserted at each step, half-open to open on a failed probe, over-budget rejection, and the excluded-probe slot refund that pins the half-open 429 subtlety I flagged. TestNewReadyToTrip covers the rate rule and the consecutive-before-floor ordering.

The consecutive-rule regime gap is stated plainly at newReadyToTrip and echoed in the field docs and validator, with the mechanism fix tracked in SOL-152286. All the comment fixes landed: the sentinel list now includes ErrExchangeCircuitOpen, the from_config importer claim is gone, the types.go package doc is rewritten, the jitter note is corrected, and OnStateChange is documented as running under cb.mutex. Agree with keeping that callback synchronous for one rare WARN.

CI is green including e2e-oauth. The three follow-ups (alertable metric, Retry-After, consecutive tracking outside gobreaker) are the right things to close before the flag flips. Approving.

PR #206, opened 2026-07-22
SOL-152231: Add E2E test coverage for list-bridges and get-bridge-status

solace-lootii:
## Summary
- Adds live-broker E2E coverage for `list-bridges`/`get-bridge-status` (SOL-152124) to `test/e2e-monitoring`, as the deferred follow-up from that PR
- New F8 fixture: three bridges per broker, each pointed at the *other* broker (the only fixture in this suite that's inherently cross-broker) — `test-bridge` (healthy), `test-bridge-failing` (enabled, unreachable remote), `test-bridge-disabled` (`enabled=false`)
- New Tool 14/15 tests in `test-monitoring-tools.sh`: `list-bridges` (presence, pagination, summary aggregation) and `get-bridge-status` (healthy + failing, exercising the compound `bridgeName`+`bridgeVirtualRouter` identifier against a real SEMP round-trip for the first time)
- New SEMP-direct fixture check in `verify-fixtures.sh`, matching every other fixture's dual coverage (state check + tool check)
- `README.md` documents several load-bearing lab findings from building this against real brokers (SEMP 2.46): `outboundState` only converges to `"ready"` once both sides' bridges exist; `inboundFailureReason` does NOT reliably populate for connection-level failures (unlike RDPs' `lastFailureReason`) — tested three ways; deleting a bridge cascades its `remoteMsgVpns` sub-resource automatically (unlike RDPs)

## Test plan
- [x] All three modified shell scripts pass `bash -n` syntax check
- [x] Ran the full suite live against real Dockerized brokers (not just syntax-checked): **20/20 fixture-verification tests, 66/66 MCP tool-level tests pass**, including all new bridge tests on both broker-a and broker-b
- [x] Re-ran the full suite again after rebasing onto `main` (post SOL-152124 merge, which included an unrelated SEMP schema version bump to 10.26.2 and a new composite-engine "fanout" capability) — still 20/20 and 66/66 green
- [x] Verified upstream's changes to these same files (the `list-vpns.zeroConnectionCount` rework, SOL-151876) don't semantically overlap with the bridge fixtures/tests


test/e2e-monitoring/test-monitoring-tools.sh
```diff
@@ -1002,6 +1002,129 @@ test_get_broker_status() {
...
+    assert_json_field "$content" '.bridges.data | length' "1" \
+        "list-bridges [$broker]: maxResults=1 must return exactly 1 bridge" || return 1
+    assert_json_field "$content" '.bridges.truncated' "true" \
+        "list-bridges [$broker]: maxResults=1 must flag truncated=true" || return 1
+    # The uncapped call returns every bridge (3: test-bridge, test-bridge-failing,
+    # test-bridge-disabled), untruncated.
+    response=$(mcp_call_tool "list-bridges" \
+        "$(jq -nc --arg b "$broker" '{broker:$b,msgVpnName:"default"}')") || return 1
+    content=$(extract_content "$response")
+    assert_json_field "$content" '(.bridges.data | length) >= 3' "true" \
```
helen-zhu-solace: The test asserts length >= 3, but F8 creates exactly 3 bridges — so it should be == 3. The >= version silently passes if a pagination bug duplicates a bridge and returns 4. Tightening to == closes that loophole in this test directly instead of relying on the separate uniqueness check to catch it.

solace-aross (dismissed):
Approving. Careful work, and the F8 lab findings are the valuable part. Confirming that a bridge's `inboundFailureReason` stays empty for connection-level failures (unlike RDPs) is exactly what justifies keying the down assertions on `inboundState`/`downCount` instead of the reason map, and it validates the handler's admin-disabled exclusion against real broker behavior rather than unit-test assumptions. The first live confirmation that the compound `bridgeName`+`bridgeVirtualRouter` path resolves closes a real gap. CI is green end to end, including e2e-monitoring.

Two of Copilot's flags are already handled at head (compound-key dedup, and excluding `not-applicable` from the down predicate). The other two are low and non-blocking, fold in here or skip:
- verify-fixtures.sh checks `inboundState` for test-bridge-failing but not `enabled=true`; the tool test does, so the fixture-layer check is just weaker than its sibling.
- README's cleanup-order line reads queues, bridges, VPNs, but `cleanup_fixtures` tears the VPNs down before bridges.

One forward watch, not a change request: cross-broker `outboundState` convergence is the most likely slow-runner flake source in this suite (F4 has bitten us there). The 30s poll looks fine; flagging so it's on the radar if CI ever goes intermittent.

test/e2e-monitoring/helpers.sh
```diff
@@ -126,6 +126,102 @@ cleanup_multi_queue_on() {
...
+
+# Polls all three bridges to their converged terminal state. Call only after
+# create_bridges_on has run for BOTH brokers — test-bridge's outboundState
+# depends on the peer's reciprocal bridge existing (lab-verified: a one-sided
+# bridge reports outboundState "not-applicable" until both sides are up).
+verify_bridges_on() {
+    local broker_url="$1"
+    local label="$2"
+    log_info "Verifying bridge fixtures visible on $label ..."
+    verify_monitor_object "$broker_url" "$label" "msgVpns/$BROKER_VPN/bridges/test-bridge,auto" \
```
solace-wmaryam, line 249: **`verify_bridges_on` will hard-abort the whole suite if a bridge is slow to converge.**

`verify_monitor_object` is documented to return non-zero on timeout so callers can *proceed anyway* ("a transient monitor lag doesn't block the scaffold" — lib.sh:461), and the optional-poll sites honor that with `|| true`. The three calls here are bare statements under `set -euo pipefail`, so a timeout makes `verify_bridges_on` → `create_fixtures` abort setup entirely rather than letting the per-tool tests report a clean failure.

This matches the existing pattern for required fixtures (`create_empty_enabled_vpn_on`, F5), so it may be intentional. But bridges are the one fixture that depends on a two-sided, cross-broker handshake — the most timeout-prone case in the suite, and the exact convergence @solace-aross flagged as the likely flake source. Worth making the choice explicit: either add `|| true` so a slow handshake degrades to a test failure instead of a setup abort, or add a one-line comment noting the abort is deliberate for this fixture.

test/e2e-monitoring/test-monitoring-tools.sh
```diff
@@ -1002,6 +1002,129 @@ test_get_broker_status() {
...
+test_get_bridge_status_failing() {
+    local broker="$1"
+    local response content
+    response=$(mcp_call_tool "get-bridge-status" \
+        "$(jq -nc --arg b "$broker" '{broker:$b,msgVpnName:"default",bridgeName:"test-bridge-failing",bridgeVirtualRouter:"auto"}')") || return 1
+    content=$(extract_content "$response")
+    assert_json_field "$content" '.bridgeStatus.data.enabled' "true" \
+        "get-bridge-status [$broker]: test-bridge-failing must be enabled (down despite being enabled)" || return 1
+    assert_json_field "$content" \
+        '.bridgeStatus.data.inboundState != "ready-in-sync" and .bridgeStatus.data.inboundState != "ready-subscribing" and .bridgeStatus.data.inboundState != "not-applicable"' "true" \
```
solace-wmaryam: **The "not-healthy" predicate is duplicated verbatim in three places and can silently drift from the server's classification.**

`inboundState != "ready-in-sync" and != "ready-subscribing" and != "not-applicable"` appears here, in `verify-fixtures.sh` (verify_bridge_state), and in `helpers.sh` (verify_bridges_on); the healthy `ready-in-sync`/`ready` pair is likewise repeated. This list mirrors `bridgeInboundUpStates` / `bridgeOutboundUpStates` in `list_bridges.go` — the code these tests are meant to guard.

If that server-side up-states list ever changes, all three copies have to be updated in lockstep or the tests quietly diverge from the code they validate (and could pass while the handler is wrong). Cross-layer duplication between the SEMP-direct and MCP-tool paths is partly justified, but pulling the states into a shared shell constant would remove the drift risk. Non-blocking.

solace-wmaryam:
Left two comments, both non-blocking!

solace-aross (dismissed):
Everything from the last round is resolved, cleanly. The README cleanup-order line now matches the actual teardown (queues, VPNs, bridges), test-bridge-failing asserts enabled=true at the fixture layer to match its tool-test sibling, and the bridge health-state predicate is extracted to one shared source mirroring list_bridges.go instead of three hand-copied copies.

Tightening the summary guards from >= 1 to exact 3/1/1 is a real improvement: an overcounting regression (the healthy bridge also landing in downCount) would have slipped past the old guard. Full suite green including e2e-monitoring on the current head. Approving.

solace-aross (approved):
The ready-subscribing relaxation is correct. The SEMP monitor spec lists both ready-subscribing and ready-in-sync as up states ("the bridge is up and is attracting traffic"), with everything else explicitly a down state — so accepting it aligns the assertion with the broker's own classification and closes a real flake window when a fixture bridge is caught mid-subscription-add. It's also the proper complement of the existing unhealthy predicate, so the two read consistently.

Verified against the current head: the delta since my last approval is this one commit plus the #207/#208 main merges, no other suite-file change. Full suite green including e2e-monitoring. Approving.

PR #208, opened 2026-07-23
SOL-152207: Surface SEMP schema version in unknown-attribute rejections

solace-wmaryam:
## Summary

Surface the embedded SEMP schema version in unknown-attribute rejections so readers can distinguish a typo from a broker-newer-than-server mismatch.

Fixes SOL-152207

## Type of Change

## Motivation and Context

When a write tool receives a body attribute the embedded schema doesn't recognize, the server rejects it client-side with a message that frames the only cause as a naming mistake. That's usually correct, but it hides the one legitimate case: the target broker is newer than the spec the server was built against and the attribute genuinely exists. Operators seeing the old message had no way to tell which case they were in.

## Changes Made

- Added `SchemaVersion` field to `sempv2.Operation`, populated from the spec's `info.version` (e.g. `10.26.2.9715`) in `ParseSpecs`.
- Rewrote the rejection in `constructRequestBody` to name both possible causes and interpolate the version:
  `'X' is not a known attribute of operation 'Y' (embedded SEMP schema: broker v10.26.2.9715). Check for the correct name — or, if 'X' is a newly released broker attribute, try a newer version of the MCP server`
- Added test assertions for the new field and error message content.
- CHANGELOG entry under `[Unreleased] › Changed`.

## Testing

**Test Environment:**
- Go version: 1.24
- OS: Linux (Fedora 41)

**Tests Performed:**

**Test Coverage:**
- `TestParseSpecs_BodyFields` — asserts `SchemaVersion` is populated from embedded specs.
- `TestExecute_UpdateMessageVPN_RejectsUnknownBodyField` — asserts error message contains the version string (`broker v10.26.2.9715`).

## Breaking Changes

N/A — the error message text changed but this is informational, not a contract.


internal/composite/executor.go
```diff
@@ -703,10 +703,19 @@ func (ce *CompositeExecutor) constructRequestBody(op *sempv2.Operation, args, pa
...
+	//
+	// A field can be unknown for two reasons: a wrong/typo'd name (the common
+	// case), or a genuinely new attribute on a broker that is newer than the
+	// embedded schema. The message names both and reports the schema version so
+	// the reader can tell which case they're in.
 	if op.BodyFields != nil {
 		for field := range body {
 			if !op.BodyFields[field] {
-				return nil, fmt.Errorf("request body field %q is not a known attribute of operation %q; tool-only params must be declared as path/query/header params, not spread into the body", field, op.ID)
+				schema := "embedded SEMP schema"
```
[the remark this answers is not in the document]
amitmorade: Worth Looking at it, not a real bug here today

helen-zhu-solace (dismissed):
Nothing flagged on my side - LGTM!

internal/semp/sempv2/operation.go
```diff
@@ -15,13 +15,14 @@ import (
...
-	Description string         // human-readable description from the spec
-	BodyFields map[string]bool // BodyFields is the set of attribute names the operation's request body accepts, resolved from the body parameter's
-	                           // schema. It is nil when the operation has no body or its schema couldn't be resolved to a concrete property set.
+	ID          string          // operationId from OpenAPI spec (e.g., "getMsgVpnQueue")
+	Method      string          // HTTP method (GET, POST, PUT, PATCH, DELETE)
+	Path        string          // path template including basePath (e.g., "/SEMP/v2/monitor/msgVpns/{msgVpnName}/queues/{queueName}")
+	Parameters  []Parameter     // path, query, and body parameters
+	Description string          // human-readable description from the spec
+	BodyFields  map[string]bool // BodyFields is the set of attribute names the operation's request body accepts, resolved from the body parameter's
+	// schema. It is nil when the operation has no body or its schema couldn't be resolved to a concrete property set.
```
solace-awaheed: Not blocking but leaving the comment in line doesn't pass on the go linter

internal/composite/executor.go
```diff
@@ -703,10 +703,19 @@ func (ce *CompositeExecutor) constructRequestBody(op *sempv2.Operation, args, pa
...
+				if op.SchemaVersion != "" {
+					schema += ": broker v" + op.SchemaVersion
+				}
+				return nil, fmt.Errorf("request body field %q is not a known attribute of operation %q (%s). Check for the correct name — or, if %q is a newly released broker attribute, try a newer version of the MCP server. If %q is a tool-only parameter, declare it as a path/query/header parameter rather than including it in the request body", field, op.ID, schema, field, field)
```
amitmorade: This is pre-existing behavior, not something this PR introduced — but the PR is what tips it over, so worth raising here.

Today `constructRequestBody`'s rejection has one output serving two audiences at once: the agent/caller (who needs a short, actionable "field X is wrong, here's what to try") and the operator (who'd want a durable, searchable record of *why*). There's currently no server-side log line for this rejection at all — I checked, and neither `constructRequestBody` nor its callers (`runSingle`, `fetchFanOut`) emit anything via `slog`. Compare with the pagination-cap case a bit further down in this same file, which does get an operator-facing `slog.Warn` in addition to what's returned to the caller.

That gap was tolerable when the message was a single sentence. This PR adds the schema version and a second causal branch (typo vs. broker-newer-than-schema), which is exactly the kind of detail an operator would want to correlate across many calls — e.g. "we're seeing this same attribute rejected repeatedly, our embedded schema is stale, time to ship a server update" is a fleet-level signal that only accumulates in a log, not in any single request's return value. So the message is carrying real operator-relevant weight now, with nowhere else for that weight to land.

Would it make sense to split this: keep the caller-facing error terse (field, operation, "check the name or try a newer server version"), and add a `slog.Debug`/`slog.Info` at the rejection site with the fuller context (schema version, operation ID, field) for operators? Not asking to block on it, since it predates this PR, but this change is a natural point to make the split since it's the one adding the extra weight.

internal/semp/sempv2/operation.go
solace-wmaryam: will fix!

amitmorade (approved):
LGTM all feedback is addressed, thank you! CI is also green

solace-aross (approved):
Clean. Moving the schema version into a slog.Debug line while keeping the user-facing message terse is the right call: operators get the correlation signal without the error string growing unbounded, and the message now names all three causes (typo, tool-only param, broker newer than the embedded schema). The struct-comment split resolves the earlier doc-comment concern, and the debug line carries only field name, operation, and version, no values.

Verified the executor.go and operation.go changes are identical to the pre-rebase head I reviewed, and CI is green including e2e. One minor non-blocking note: the PR description still describes the old version-in-message wording, worth syncing it to what shipped. Approving.

PR #209, opened 2026-07-23
SOL-152122: Add list-to-detail cross-references for list-vpns, list-queues, list-clients

solace-lootii:
## Summary
- Adds the one-line "for full detail, use X" pointer that `list-rdps`/`list-bridges` already had to their three siblings: `list-vpns` → `get-vpn-status`, `list-queues` → `get-queue-metrics`, `list-clients` → `get-client-details`
- Confirmed `list-rdps`/`list-bridges` already correct — used as the wording/tone template, left unchanged
- Verified what's *actually* exclusive to each detail tool by diffing `select` fields (not guessed) before writing the pointer text: `get-vpn-status` adds AMQP + TLS-variant service status; `get-queue-metrics` adds the authoritative SEMPv1 live depth plus discard/redelivery counters; `get-client-details` adds discard counts, byte/message rates, and software version
- Mirrored the same three pointers into `docs/tools-reference.md` for consistency (it already had them for `list-rdps`/`list-bridges`); left `docs/user-guide.md` untouched since its narrative table has no cross-references anywhere, not even for `list-rdps`, so adding them there would create a new inconsistency rather than fix one
- Found and fixed an unrelated pre-existing bug while in the area: `docs/tools-reference.md`'s `list-queues` line said "with depth," but `list-queues` returns `spooledMsgCount` — a cumulative lifetime counter the codebase's own conventions (and `list-queues`' own tool description) explicitly say must never be called "depth." Reworded to match the correct phrasing already used in `README.md`/`docs/user-guide.md`
- New regression test (`internal/composite/loader_cross_reference_test.go`) loads the real embedded `tools.yaml` and asserts all 5 list tools mention their detail-tool counterpart by name, so a future edit can't silently drop the pointer
- `CHANGELOG.md` entry added per this repo's own `CLAUDE.md` rule (any `tools.yaml` change needs one)

## Test plan
- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` — all 31 packages pass, including the new cross-reference test (5 sub-tests, one per list tool)
- [x] Description-only + doc-only + one new test file — no executor/postprocess/schema changes, nothing that touches runtime behavior


solace-aross (approved):
Clean, and I like the rigor. You diffed the actual select fields to write accurate pointers rather than guessing, and the regression test loads the real embedded tools.yaml so a future edit can't silently drop a pointer. Catching the list-queues "depth" mislabel while you were in the area is the right call, and leaving user-guide.md untouched (no cross-refs anywhere) is the correct consistency judgment.

One real nit, non-blocking: Copilot's byte/message fold. The description is a folded scalar, so the line break after "byte/" renders as a space ("byte/ message rates"). Worth tightening since this PR is specifically about description quality. Approving.

PR #210, opened 2026-07-23
SOL-152125: Add payload-discipline guardrails (size test, marshal benchmark, drift test)

solace-lootii:
## Summary
Three measurement/regression-guard artifacts refining SOL-149494's payload-discipline half. No user-facing behaviour change — pure engineering hygiene against future re-bloat.

- **`internal/composite/loader_realfile_validate_test.go`** — fast unit test loading the real embedded `tools.yaml` and asserting `ValidatePostProcess` passes, so a select/field drift is caught by `go test ./internal/composite/...` alone, without a full server boot or E2E run. (Reuses the exact pattern from SOL-152122's cross-reference test.)
- **`cmd/server/toollist_budget_test.go`** — replicates `main()`'s real tool-registration pipeline (calls the same unexported `registerSEMPv1Tools`/`registerMixedTools` helpers directly, so it can't drift from what production actually registers), drives a genuine `tools/list` round-trip over an in-memory MCP transport, and asserts token budgets for both `enable_write_tools` states. Measured baseline (already grown since the ticket was scoped, now that bridges exist): 19 tools/~21.1KB/~5.3k tokens read-only (budget 15k), 35 tools/~40.6KB/~10.2k tokens with write (budget 25k) — ~3x headroom by design, a static ceiling meant to occasionally require a deliberate bump as the catalog legitimately grows, not a tight cap.
- **`internal/composite/postprocess/handlers/marshal_bench_test.go`** — `go test -bench` covering `json.Marshal` cost at N=100/500 for all 5 list tools (the ticket predates `list-bridges`; included it too rather than leave an obvious gap). Built with each tool's full `select:` field set, not the minimal fixtures the correctness tests use. Sub-5ms even at N=500 for every tool.
- Added a `make bench` target and a CI step so the benchmarks actually execute (Go silently skips `Benchmark*` funcs without `-bench` — without this they'd exist in the repo but never run). Informational only, per the ticket's instruction not to gate on wall-clock — the actual regression gate is the byte/token threshold in `TestToolListBudget`.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `golangci-lint run` clean
- [x] `go test -race ./...` (matches CI) — 31 packages, 0 failures
- [x] `make bench` runs successfully repo-wide and produces the expected per-tool timing output
- [x] Re-verified everything after merging in `main`'s subsequent changes (an unrelated `executor.go` update) — no regressions


solace-aross (dismissed):
Solid engineering hygiene, and the design choices are the right ones. Driving a real tools/list round-trip through the same unexported registerSEMPv1Tools/registerMixedTools helpers main() uses means the budget test can't drift from what the server actually exposes, and following pagination to completion future-proofs it against a later change that starts paginating. The framing in the test comment (regression guard with ~3x headroom, bump deliberately as the catalog grows, not a tight cap) is exactly where that intent needs to live.

One cross-PR heads-up, not a blocker: #211 (SOL-151994) also adds a real-embedded-tools.yaml ValidatePostProcess check, so loader_realfile_validate_test.go overlaps it. Worth a quick sync with Wajiha on which to keep once both land. CI is green. Approving.

solace-aross (approved):
All three follow-ups landed cleanly. Ceiling division on the token estimate is the right call for a guardrail (round up, never under-report at the boundary), and the comment says why. Deleting loader_realfile_validate_test.go is the correct dedup now that #211 merged: its TestLoadTools_EmbeddedDefinitions runs ValidatePostProcess against the real embedded tools.yaml, so the coverage is preserved, not lost. And continue-on-error on the bench step backs the "informational" claim with an actual guarantee rather than a comment.

Verified #211's embedded check is present on this branch and CI is green end to end. Approving.

PR #211, opened 2026-07-23
SOL-151994: Add unit test validating embedded composite tool definitions

solace-wmaryam:
## Summary

Add a unit test that validates the real embedded `tools.yaml` via `LoadTools(definitions.FS)`, closing the gap where composite tools were only validated at server startup.

Fixes [SOL-151994](https://sol-jira.atlassian.net/browse/SOL-151994)

## Type of Change

## Motivation and Context

The embedded `tools.yaml` (~30 composite tools) was only validated at server startup. Existing loader tests use hand-written `fstest.MapFS` mocks, so a semantic regression in a tool not exercised by e2e could pass CI undetected.

## Changes Made

- Add `internal/composite/loader_embedded_test.go` with independent subtests:
  - `count`: Guards against silent tool drops (floor of 30)
  - `naming`: Unique kebab-case names (no uppercase, no underscores)
  - `postprocess`: Validates postProcess tools against registered handlers via `ValidatePostProcess`
  - `operations`: Every step operation exists in the embedded SEMPv2 specs
  - `spot/list-vpns`: Fan-out struct decoding (forEach/forEachKey fields)

## Testing

**Test Environment:**
- Go version: 1.26.3
- OS: darwin/arm64
- Broker version (if applicable): N/A (no broker required)

**Tests Performed:**

**Test Coverage:**
- Loads real `definitions.FS` and validates all 30 composite tools
- Cross-checks operations against embedded OpenAPI specs
- Validates postProcess handler registration

## Breaking Changes

N/A

## Additional Notes

**Out of scope (for future consideration):** Select-field validation against response schema — would require extending `Operation` struct with `ResponseFields`. The ticket noted this as "where feasible"; operation-existence already catches the higher-risk class of bugs (misspelled/removed operations).

[SOL-151994]: https://sol-jira.atlassian.net/browse/SOL-151994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solace-aross (dismissed):
Good gap to close. The loader validates tool structure, fan-out references, and post-process registration at startup, but the existing loader tests run against fstest.MapFS mocks — so this is the first test exercising those invariants against the real embedded tools.yaml. The operation-existence cross-check against the embedded SEMPv2 specs is the highest-value assertion here; it catches a misspelled or removed operation that no mock would.

Copilot's two nits are both fair and both optional: the naming check covers the two conventions that actually bite (uppercase, underscore) but not full kebab-case, and the fan-out spot-check keys off the first ForEach step rather than a stable step id. Neither blocks. Approving.

bczoma (changes requested):
Some comments from me, would be good to address at least the first two

AC #5 unmet — the ticket's named target get-replication-status is not asserted; list-vpns was substituted
internal/composite/loader_embedded_test.go (whole file) — flagged by good:pr + good:architecture + good:code
The ticket explicitly requires a focused assertion on get-replication-status (step id replication, operation monitor/getMsgVpn, expected select field list). That tool is the entire motivation — it's the composite tool no e2e suite exercises, so a semantic regression in it passes lint/build/unit/e2e today. The PR's operations subtest only checks that the operation string exists; it never validates the selected fields, so dropping/renaming a replication* field still passes silently. Substituting list-vpns closes the ticket in appearance while leaving the named gap open.
Fix: add a spot/get-replication-status subtest asserting step id replication, operation monitor/getMsgVpn, and the expected select fields. Note this tool declares fields via a templated args["select"] (a comma-joined string), not the structured Select []string — so the assertion must parse step.Args["select"]. Keep list-vpns as an additional fan-out check if desired.

Count guard is a floor (< 30), not the exact constant AC #2 asks for
loader_embedded_test.go:44 — flagged by good:code (good:architecture, good:pr concur at lower severity)
There are exactly 30 tools today, so the test sits on its floor. A silent addition (accidental duplicate, or a merge re-adding a removed tool) goes completely uncaught — and a rename-with-replacement (remove one, add one → still 30) also slips through. AC #2 says the count must "match an expected constant," which is a stronger guarantee.
Fix: assert exact equality (len(tools) != wantToolCount) with a comment telling maintainers to bump it deliberately.

AC #4 (required fields / ≥1 step) is only covered transitively
loader_embedded_test.go — flagged by good:pr; contextualized by good:code
LoadTools → validateTool already enforces name/description/≥1-step/step-id/operation, and the test fails on any LoadTools error — so this is effectively covered. But it's implicit; the AC asked for an explicit per-tool assertion, and a reader can't see the guarantee.
Fix: optional — add a short subtest iterating all tools asserting len(tool.Steps) >= 1, or note in the PR that AC #4 is satisfied via validateTool.

solace-wmaryam:
@solace-aross @bczoma comments have been addressed!

bczoma (approved):
Looks good!

PR #215, opened 2026-07-27
Migrate OAuth TokenCache backend from Otter v1 to Otter v2

amitmorade:
## Summary

Swaps the OAuth `TokenCache` backend from `github.com/maypok86/otter` v1.2.4 to `github.com/maypok86/otter/v2` v2.3.0, plus targeted cleanup of dead code the v1 API surfaced. Internal Go API only — on-the-wire MCP schemas, operator config, and the `TokenCache` interface method set are unchanged. Hop-2 OAuth runtime remains gated behind `ENABLE_UNRELEASED_BROKER_OAUTH` and unactivated in the wild.

Tracked under [SOL-152397](https://sol-jira.atlassian.net/browse/SOL-152397).

## Commit-by-commit story

Seven commits, each independently compiling and passing its tests.

1. **Add tests pinning TokenCache validation order, MaxTTL clock asymmetry, Close return, concurrent Delete, and status Level mapping** — five new tests against v1 that pin caller-observable behaviour. These are the equivalence gate for the swap: same tests pass before and after.
2. **Drop test assertions on GetMissExpired and PutDroppedFull** — surgical removal of three assertions on enum values slated for deletion. Alignment step; no behaviour change.
3. **Remove unreachable GetMissExpired and PutDroppedFull statuses** — deletes the enum values from the interface AND their producing branches in the v1 implementation. Both were unreachable: `GetMissExpired` because v1's own expiration already returned `(nil, false)` before the wrapper's double-check could observe it; `PutDroppedFull` because v1's admission only rejects at `cost > MaxSize/10` with the default cost function, requiring `MaxSize < 10`.
4. **Swap TokenCache backend from Otter v1 to Otter v2** — the migration itself. Wrapper rewritten against v2's API (`otter.New(&Options{...})`, `GetIfPresent`, `Set` without per-call TTL, `Invalidate`, `StopAllGoroutines`). Per-entry TTL now supplied via `ExpiryWritingFunc` at construction. `go mod tidy` transitively bumps `testify` to v1.11.1 and drops two v1-only transitive deps (`gammazero/deque`, `dolthub/maphash`).
5. **Rename GetMissAbsent to GetMiss for semantic honesty** — the surviving non-hit status covers both "backend reports no entry" and "wall-clock double-check rejected an entry"; neither `Absent` nor `Expired` alone was honest.
6. **Drop unreachable wall-clock double-check in Get** — the wrapper's wall-clock freshness check after a `GetIfPresent` hit cannot fire in v2. v2's `GetIfPresent` runs a nanosecond-precise `HasExpired` internally, and the `ExpiryCalculator` installs each entry with a deadline at or before `ExpiresAt`. Traced through both cases before removing.
7. **Add CHANGELOG entry** — one bullet under `[Unreleased] > Changed`.

## Why v2

- v1 is on maintenance-only; the active line, security hardening, and feature work are on v2. Migrating before Hop-2 activates is the cheapest time.
- v2's adaptive W-TinyLFU replaces v1's S3-FIFO — same or better hit rate across all workloads per the maintainer.
- v2 has nanosecond expiration precision (v1: ~1s), closing the window the wrapper needed a wall-clock double-check to defend against.
- v2 adds HashDoS protection.

## Why the wrapper stayed small

Only one file imports Otter directly (`internal/oauth/cache/otter.go`), and only one package uses the wrapper (`internal/tokenexchange/`) via the `TokenCache` interface. The library swap is a single-file change; the Exchanger is untouched.

## Test plan

- [x] `go test -race ./internal/oauth/cache/` — passes.
- [x] `go test -race ./internal/tokenexchange/` — passes.
- [x] `go test -race ./...` — passes.
- [x] `go build ./...` and `go vet ./...` — clean.
- [x] `go.mod` shows only `github.com/maypok86/otter/v2 v2.3.0` (v1 gone).
- [x] No file outside `internal/oauth/cache/` imports the library.
- [x] `internal/tokenexchange/exchange.go` is unchanged.
- [ ] Manual E2E via `dev/oauth-token-exchange/` — attempted but blocked by an unrelated realm-export bug in `origin/amorade/dev-oauth-manual-e2e` (`solace-broker-second` referenced in role definitions but not defined as a client). Not this PR's problem to fix; flagged for the e2e-branch owner. Unit + race + build gates are the acceptance criteria for this PR.

## Follow-ups explicitly out of scope

- Adopting v2's built-in `Loader` singleflight to replace the Exchanger's own `singleflight.Group`.
- Adopting v2's `RefreshCalculator` for proactive async token refresh.
- Adopting v2's `stats.Recorder` for pull-based cache metrics.
- Adopting v2's `OnDeletion` typed deletion event listener.
- Fixing the pre-existing double-subtraction of `DefaultTokenExpirySkew` in `internal/tokenexchange/response.go:139`.
- Exposing `MaxSize`, `ClockSkew`, or `MaxTTL` as operator-configurable knobs.

[SOL-152397]: https://sol-jira.atlassian.net/browse/SOL-152397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amitmorade:
Closing in favour of a cleanup pass — will land as a new PR shortly with stale comments and Copilot-flagged items fixed. Same branch, same commits, plus one additional cleanup commit.

PR #218, opened 2026-07-27
docs/tools: standardize on 'status' vocabulary, not 'health'

solace-aross:
## Summary

Standardizes the monitoring tools' wording on **status** (raw operational state: up/down, counts) rather than **health** (a healthy/unhealthy verdict these tools deliberately do not make). Follows the `get-broker-health`→`get-broker-status` (SOL-150707) and `get-vpn-health`→`get-vpn-status` (SOL-151742) precedent, now applied to the newer tools and their docs.

Description/comment text only. **No tool names, output field names, parameters, step keys, or behavior change.**

## Changed

| Area | Change |
|---|---|
| Kafka receiver/sender tool descriptions (`tools.yaml`) | "topic/queue-binding **health**" → "**status**"; "…bindings are actually **healthy**" → "actually **up**" (matches the `…UpCount` field vocabulary) |
| Docs describing those tools (`README.md`, `docs/user-guide.md`, `docs/tools-reference.md`) | Same wording aligned |
| `get-redundancy-status` | "assess HA **health**" → "assess HA **status**" |
| `get-discard-stats` | "broker-level **health checks**" → "**status checks**" |
| `redundancy/response.go` comment | "mate-link **health** flags" → "**status** flags" |
| `CHANGELOG.md` `[Unreleased]` Kafka entry | Same wording, so release notes match the shipped descriptions |

## Deliberately kept

- The "**not a health verdict** / whether X is healthy depends on deployment intent" framing in `get-vpn-status` and `get-broker-status` — it justifies the `-status` naming.
- Real feature names: the HTTP `/health` endpoint, `--health` CLI flag, and Docker health-check references.

## Verification

- `go build`, `go vet`, and affected package tests (`internal/tools/...`, `internal/composite/...`) pass.
- Reviewed by a skeptical Opus pass (`/good-developer` + `/understand-explain`): **GO** — meaning preserved, no behavior change, string concatenation intact. Its one completeness catch (the `[Unreleased]` CHANGELOG wording) is fixed here.


solace-awaheed (approved):
LGTM

bczoma:
For consideration:

internal/composite/postprocess/handlers/list_bridges.go — this file wasn't swept, and since the PR does rewrite one internal comment (redundancy/response.go:32, "mate-link health flags" → "status flags"), it's the inconsistency a reader will notice: bridges are the direct sibling of the Kafka tools being cleaned up here.

Nine lines, all internal — four doc comments (:33 "treated as healthy rather than down", :65 "either direction is unhealthy", :73 "specifically is unhealthy", :76 "inbound healthy, outbound not") and the locals inboundUnhealthy/outboundUnhealthy at :115-120. Suggest up/down rather than "status", since the file already names its tables bridgeInboundUpStates/bridgeOutboundUpStates and its output field downCount — so inboundDown/outboundDown reads straight into if inboundDown || outboundDown { down++ }. Both are function-local, so no test churn.

Alternative if you'd rather keep the diff purely tool-facing: drop the redundancy/response.go comment change instead, and the PR's "description/comment text only" claim lands cleanly at the tool-surface boundary.

Either way, worth adding brokerstatus/handler.go:85 to "Deliberately kept" — "asks about a broker's status or health — whether it is healthy, slow, restarted" is trigger-matching on user vocabulary, not the verdict framing the existing bullet covers.

PR #221, opened 2026-07-27
Honor IdP Retry-After on exhausted 429 chains

amitmorade:
## Summary

Closes a gap left in the token-exchange circuit breaker (SOL-151600): 429 is excluded from breaker failure counting, but gobreaker refunds the half-open probe slot for excluded outcomes, so a still-throttling IdP could admit unlimited sequential probes with no pacing between them. Adds a shared, process-wide backoff gate that honors the IdP's own `Retry-After` on an exhausted 429 chain.

Fixes SOL-152285

### Closed breaker: multiple concurrent callers hitting a throttling IdP

```mermaid
sequenceDiagram
    participant A as Caller A (broker-1)
    participant B as Caller B (broker-2)
    participant C as Caller C (broker-3)
    participant G as Shared Gate
    participant Br as Circuit Breaker (closed)
    participant IdP

    Note over G: gatedUntil = 0 (not gated)

    A->>G: Exchange()
    G->>Br: gate open, proceed
    Br->>IdP: attempt 1
    IdP-->>Br: 429 Retry-After: 30
    Br->>IdP: attempt 2 (intra-chain backoff, honors 30s uncapped)
    IdP-->>Br: 429 Retry-After: 30
    Br->>IdP: attempt 3 (retries exhausted)
    IdP-->>Br: 429 Retry-After: 30
    Br-->>A: ErrExchangeRetriesExhausted (429 excluded, breaker stays closed)
    A->>G: raiseGate(clamp(30s))
    Note over G: gatedUntil = now + 30s

    B->>G: Exchange()
    G-->>B: ErrExchangeRateLimited (fails fast, no IdP call, no breaker touch)

    C->>G: Exchange()
    G-->>C: ErrExchangeRateLimited (fails fast, no IdP call, no breaker touch)

    Note over G: 30s elapses

    B->>G: Exchange() (retry)
    G->>Br: gate open, proceed
    Br->>IdP: attempt 1
    IdP-->>Br: 200 OK
    Br-->>B: token
```

Only Caller A pays the cost of discovering the throttle (one full retry chain). B and C are protected the moment A's chain exhausts — they never touch the IdP, and the breaker's own counters are never involved.

### Half-open: the scenario this PR actually fixes

```mermaid
sequenceDiagram
    participant P1 as Probe 1
    participant P2 as Probe 2
    participant G as Shared Gate
    participant Br as Circuit Breaker (half-open, budget=2)
    participant IdP

    Note over Br: Breaker just transitioned open → half-open

    P1->>G: Exchange()
    G->>Br: gate open, proceed
    Br->>IdP: probe attempt(s), exhausts as 429
    IdP-->>Br: 429 Retry-After: 30 (final attempt)
    Br-->>P1: ErrExchangeRetriesExhausted
    Note over Br: 429 excluded -> gobreaker REFUNDS the probe slot<br/>(budget still 2, unspent) - this is the pre-existing gap
    P1->>G: raiseGate(30s)
    Note over G: gatedUntil = now + 30s

    P2->>G: Exchange() (would-be probe, moments later)
    G-->>P2: ErrExchangeRateLimited
    Note over Br: Breaker never even consulted -<br/>no probe attempt made, budget untouched either way

    Note over G: 30s elapses, IdP now healthy

    P2->>G: Exchange() (retry)
    G->>Br: gate open, proceed
    Br->>IdP: probe 1
    IdP-->>Br: 200 OK
    Br-->>P2: token
    Note over Br: 1 of 2 consecutive successes - still half-open

    P2->>G: Exchange() (next probe)
    G->>Br: gate open, proceed
    Br->>IdP: probe 2
    IdP-->>Br: 200 OK
    Br-->>P2: token
    Note over Br: 2 of 2 - breaker closes
```

Without this gate, gobreaker's refund-on-exclude behavior means a still-throttling IdP could let probe attempts fire back-to-back indefinitely (excluded → refunded → immediately eligible again). The gate doesn't touch that refund logic at all — it just stops the *next* attempt from happening until the IdP's own `Retry-After` window passes, regardless of how many probe slots are technically still "available."

## Design Decisions & Rationale

This PR carries several deliberate tradeoffs that aren't obvious from the diff alone. Full detail (including dead ends) is in `docs/superpowers/plans/2026-07-27-sol-152285-retry-after-implementation-plan.md`; the short version:

**Why a gate, not a change to breaker semantics.** Researched how gobreaker, Polly, resilience4j, Envoy, and Hystrix each handle "an outcome that shouldn't count as breaker failure, but still carries a backpressure signal." Nobody ships this combination out of the box — Hystrix had the equivalent feature requested twice (adding a Retry-After-driven timer to half-open) and declined both times. The one real precedent is the AWS SDK's retry engine, which keeps its failure-based retry quota and its server-directed delay (`x-amz-retry-after`) as two fully separate mechanisms — a timing gate layered in front of, not inside, the existing policy. That's the shape adopted here: the gate is checked *before* `breaker.Execute` and a gated call never reaches breaker bookkeeping at all, so `isBreakerExcluded`/`isBreakerSuccess` and every existing breaker test are untouched.

**Why the gate is process-wide, not per-broker/per-key.** The circuit breaker it sits in front of already guards one shared IdP for the whole process. A 429 is a statement about the IdP's capacity, not about whichever broker's exchange happened to trigger it — so broker A's throttling correctly paces back broker B's unrelated exchange against the same IdP. Rejected a per-dedup-key gate because it would under-protect: concurrent *distinct* callers (different brokers/tenants hitting the same IdP) are exactly the case this needs to cover, and singleflight already collapses identical concurrent callers before they'd reach the gate anyway.

**Why the gate is raised only on chain *exhaustion*, not on the first 429 seen.** Original design leaned toward raising the gate as soon as any attempt in a chain saw a 429. Reconsidered: a single 429 might resolve on the very next retry, so gating off one early data point would preemptively block every other caller before the triggering caller even finished acting on it. Moved the trigger to `classifyRetryOutcome`'s exhaustion path instead — only a chain where *every* attempt was rate-limited and the retry budget ran out is strong enough evidence to justify blocking everyone else.

**Why a new `ErrExchangeRateLimited` sentinel instead of reusing `ErrExchangeCircuitOpen`.** Both mean "rejected without calling the IdP," but for different reasons — one means the breaker judged the IdP unhealthy from failures, the other means the IdP is healthy and asked us to wait. Conflating them would make it impossible for an operator (or alerting keyed on `errors.Is`) to tell "IdP is down" apart from "IdP is fine, just throttling, working as intended." Distinct `LogAttrs` marker (`gate` vs `breaker_state`); same `AgentMessage` phrasing since both are shared-IdP-wide, not broker-specific.

**Why the operator-facing cap follows the existing `CircuitBreakerConfig` pattern (shipped default → YAML override → sanity ceiling) rather than a hardcoded constant.** The operator is the one who actually knows their IdP's throttling behavior; hardcoding it would be the one tunable in this subsystem that doesn't follow the established shape, and would take a real decision away from the person best positioned to make it. Note this is a **ceiling only** — it can shrink what the IdP asked for, never invent a wait the IdP didn't ask for. An absent/unparseable `Retry-After` is a no-op regardless of the cap (see below); there is no fallback-duration knob in this PR.

**A real bug caught during config-surface implementation, not before.** `IdPRetryAfterConfig.MaxHonoredDuration` initially allowed an explicit `0`, following the same "pointer distinguishes omitted-from-set" pattern used elsewhere. That silently breaks here: the *runtime* field it feeds (`Params.MaxHonoredRetryAfter`, a plain `time.Duration`) already uses zero to mean "use the shipped default" — so an operator writing `max_honored_duration: 0` intending "honor nothing" would get the opposite of what they configured, with no error. Fixed by rejecting an explicit zero at validation, with the reasoning captured on both the config type and the validator.

**Absent/unparseable `Retry-After` is a no-op, not a guess.** If the IdP gives us nothing usable, the gate simply isn't raised — matches the acceptance criteria ("preserve current behavior") rather than inventing a fallback backoff duration nobody asked for. Logged distinctly (absent vs. malformed) since the second case is a more interesting signal about the IdP's own behavior. A configurable fallback-duration knob for IdPs that throttle hard but never send the header was raised as a legitimate, non-redundant gap (both this gate *and* the breaker's 429-exclusion leave that IdP unprotected) — deliberately left out of scope here since it inverts "only ever act on what the IdP told us," and noted as a candidate follow-up ticket.

**Two bugs caught by automated PR review, both fixed.** Copilot's review flagged: (1) `parseRetryAfter` didn't trim optional surrounding whitespace (RFC 9110 §5.6.3 OWS), so a header like `"120 "` would be wrongly treated as unparseable and the gate would silently never engage — fixed by trimming a local copy before parsing while keeping the logged `raw` value untrimmed for faithful diagnostics. (2) The gate-set/clamped log lines fired even when nothing was actually raised — either because the delay floored to zero, or because a concurrent chain's longer wait already won the CAS race — misrepresenting what actually happened. Fixed by having `raiseGate` report back whether it genuinely changed the gate, and only logging "set"/"clamped" when it did. Both fixes have dedicated regression tests confirmed to fail against the pre-fix code.

## Changes Made

- `internal/tokenexchange/response.go`: `parseRetryAfter` parses both RFC 9110 §10.2.3 forms (delta-seconds, HTTP-date), trims optional surrounding whitespace before parsing, and floors negative/past values to zero rather than propagating a negative duration.
- `internal/tokenexchange/retry_after_gate.go` (new): a shared `gatedUntil` atomic on `Exchanger`, checked before the circuit breaker on every call and raised only when a 429 chain fully exhausts; `raiseGate` reports whether it actually changed the gate so logging never misrepresents a no-op.
- New `ErrExchangeRateLimited` sentinel (see rationale above).
- Three mutually-exclusive WARN log lines per exhausted 429 chain: gate set, gate set but clamped to the configured cap, or gate not set because the IdP sent no usable `Retry-After` (distinguishing absent from malformed) — only emitted when the gate was genuinely affected.
- New optional `broker_oauth.retry_after.max_honored_duration` config block (`internal/config/idp_retry_after.go`), overlaying a 60s shipped default.
- Zero changes to the breaker's 429-exclusion classification (`isBreakerExcluded`/`isBreakerSuccess`) or the intra-chain retry pacing that already honors `Retry-After` uncapped (SOL-151520).
- Full design rationale, alternatives considered, and implementation-time findings (including both post-review fixes) are in `docs/superpowers/plans/2026-07-27-sol-152285-retry-after-implementation-plan.md`.

## Testing

**Test Environment:**
- Go version: per `go.mod`
- OS: macOS (dev), CI matrix for the rest
- Broker version (if applicable): n/a — no live-broker dependency, this is IdP-side only

**Tests Performed:**

**Test Coverage:**
- Retry-After parsing: delta-seconds, HTTP-date, absent, unparseable, clock-skew/past-date flooring, multiple headers, surrounding whitespace on both forms.
- Gate mechanics: blocks-then-allows across expiry, no-op on non-positive delay, concurrent max-wins CAS, only exhausted+rate-limited chains raise it, no misleading log on a zero-delay or CAS-losing raise.
- Breaker integration: gate check runs unconditionally on breaker state (nil/open/half-open), a gated call touches zero breaker bookkeeping, and a new end-to-end half-open test using a *retrying* client actually exercises the ticket's motivating scenario (the existing pinned `TestBreaker_HalfOpenExcludedProbeRefundsSlot` needed no change — it uses a non-retrying client that structurally can't reach the exhausted-429 path, confirmed by running it rather than assumed).
- Config: YAML round-trip, omitted-block defaults, out-of-bounds/zero rejection, end-to-end plumbing from YAML into the constructed `Exchanger`'s clamp behavior.
- Logging: all three operator-facing states, including the raw-value length cap on a malformed header, and both post-review logging-correctness fixes.

## Breaking Changes

None.

### Code Quality

### Security
- [x] Followed secure logging rules (raw `Retry-After` values are length-capped before logging)
- [x] Credential-carrying types implement `slog.LogValuer` (unchanged by this PR)
- [x] No security vulnerabilities introduced (checked with golangci-lint)

### Testing
- [x] Error paths tested

### Documentation
- [ ] README.md updated (if user-facing changes) — n/a, gated behind `ENABLE_UNRELEASED_BROKER_OAUTH`
- [x] docs/ updated (if architecture/design changes) — implementation plan doc added

### Git & CI
- [ ] All commits are signed off (DCO: `git commit -s`) — not applied per this repo's standing convention for this contributor
- [x] No merge conflicts
- [ ] CI checks passing (lint, build, test, E2E) — pending CI run on this PR

## Additional Notes

Rides entirely under the `broker_oauth` runtime, which stays gated behind `ENABLE_UNRELEASED_BROKER_OAUTH` and is unactivated in the wild — no existing deployment is affected.

**For Reviewers:**

**Focus Areas**: the exhaustion-only gating condition in `raiseGateOnExhaustedRateLimit` (why an intermediate 429 in the same retry chain must not raise the gate), and the deliberate zero-value rejection in `validateIdPRetryAfter` (§9.5 of the plan doc has the full reasoning if useful).

**Questions**: none open.


bczoma (approved):
Looks good, no significant findings.

PR #225, opened 2026-07-27
Remove ENABLE_UNRELEASED_BROKER_OAUTH flag; broker OAuth is fully supported

amitmorade:
## Summary

Removes the `ENABLE_UNRELEASED_BROKER_OAUTH` testing-only flag and its three linked guards, so broker `auth.mode: oauth` (RFC 8693 token exchange, "Hop 2") becomes a fully supported broker authentication mode alongside `basic`/`bearer`. Confirmed no production deployment sets this env var — it was undocumented in operator-facing docs by design and existed only to unblock manual E2E testing while the runtime (retries, circuit breaker, `Retry-After` handling, token caching) was built out across prior tickets.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The flag was a temporary bypass introduced alongside the Hop-2 OAuth runtime (SOL-150799/800/801) so the runtime could be built, tested end-to-end, and hardened (SOL-151520 retries, SOL-151600 circuit breaker, SOL-152285 retry-after, SOL-152397 cache swap) before general availability — all still gated behind the same flag. With that work shipped and no customer running this mode without the flag, this PR flips broker OAuth from "rejected at startup unless you opt into the unreleased bypass" to "fully validated and supported," and documents it for operators for the first time.

## Changes Made

**Flag removal:**
- `internal/config/config.go`: removed the flag check from `Hop2OAuthActive()` (two structural preconditions remain — `broker_oauth:` set and at least one oauth-mode broker); removed the temporary "OAuth not supported" guard from `validate()` so the permanent Hop1/Hop2 alignment check now runs unconditionally; removed the per-broker rejection in `validateBroker`; removed the startup WARN, the env var constant, and `unreleasedBrokerOAuthEnabled()`.
- `internal/banner/banner.go`: removed `LogOAuthNotSupported` — `LogHop2WithoutHop1` is now the sole, unconditional operator banner for a Hop1/Hop2 mismatch.
- `internal/config/config_test.go`, `test/e2e-oauth/helpers.sh`, `test/integration/oauth_error_visibility_test.go`: removed/updated tests and dropped the now-unnecessary flag exports.
- `cmd/server/main.go`, `README.md`: fixed a stale flag-lifecycle comment and a README line that still said broker auth was basic/bearer only.

**Reviewer-driven fix (Andrea, post-review):** `internal/config`'s `validAudienceParams` accepted `audience`, `scope`, and `resource`, and all three passed `validate()` cleanly — but `scope`/`resource` then failed later at server startup inside `tokenexchange.resolveAudienceParam` with "schema-accepted but not yet implemented". That gap predated this PR but was unreachable before it (oauth mode was rejected outright); removing the flag made it reachable at exactly the moment operators are told how to configure the mode. Fixed by narrowing `validAudienceParams` to `audience` only, so `scope`/`resource` are now rejected at config load, joined with every other `broker_oauth` error, instead of surfacing later at Hop-2 runtime construction. Also folded `resolveAudienceParam`'s now-unreachable `scope`/`resource` cases into its generic `default` branch (removing `AudienceParamScope`/`AudienceParamResource`, which lost their last consumer), added test coverage for the config-load rejection, and added a CHANGELOG bullet — this also affects operators who staged a `broker_oauth:` block in advance of switching any broker to oauth mode, since that block validates independent of whether it's in use yet.

**Docs (first real operator documentation for this feature):**
- `docs/authentication.md`: new "Step 2b: Configure broker OAuth (Hop 2)" walkthrough — YAML sample, field table, IdP client-registration guidance, the Hop1-dependency error text, dedicated Grant type / Audience parameter name subsections, and a server→broker flow diagram (cache hit vs. cache miss) showing exactly when the token cache is read and written.
- `docs/configuration.md`: fixed the `auth.mode` reference row (previously said only `basic`/`bearer`); added a full **Broker OAuth (Hop 2)** reference section — every field, the `circuit_breaker`/`retry_after` sub-blocks with real shipped defaults, and explicit dependency/error-condition call-outs.
- `docs/user-guide.md`, `broker-config.example.yaml`: updated prerequisites/troubleshooting and added a commented `broker_oauth:` template, matching the file's existing patterns.
- `CHANGELOG.md`: rewrote `[Unreleased]` bullets that described themselves as gated/unactivated to describe the shipped, active behavior; added bullets for the new capability and the `audience_parameter_name` narrowing.

## Testing

**Test Environment:**
- Go version: per `go.mod`
- OS: macOS (dev), Linux (CI)
- Broker version (if applicable): N/A — config/validation layer only, no SEMP wire changes

**Tests Performed:**
- [ ] E2E tests added/updated (if applicable) — no new E2E scenarios; `test/e2e-oauth` continues to pass with the flag export removed

**Test Coverage:**
- `internal/config`: full config-validation suite, including rewritten `TestLoadConfig_BrokerOAuth` cases, `TestLoadConfig_Hop1Hop2Alignment`, `TestServerConfig_Hop2OAuthActive`, and new cases confirming `scope`/`resource` are rejected at config load with the exact error text.
- `internal/tokenexchange`: `TestFromConfig_UnknownAudienceParamReturnsError` now covers `scope`/`resource` alongside any other unrecognized value.
- `test/integration`: `TestOAuthErrorVisibility`, `TestOAuthCache_TwoRequestsSameBearerHitIdPOnce` pass without the env var.
- `make check` (build, vet, lint, race-enabled tests) passes clean on the merged branch.

## Breaking Changes

Not breaking in the conventional sense — broker `auth.mode: oauth` configs that previously failed at startup with "not yet supported" now load and run, and existing `basic`/`bearer` configs are unaffected. One narrow behavior change: a `broker_oauth:` block staged in advance (no broker on `auth.mode: oauth` yet) with `audience_parameter_name: scope` or `resource` previously loaded cleanly and now fails at config load — documented in the CHANGELOG with a migration note.

### Code Quality

### Security
- [x] No credentials in code
- [x] Credential-carrying types implement `slog.LogValuer` (unchanged by this PR)

### Testing
- [x] Error paths tested

### Documentation

### Git & CI
- [ ] All commits are signed off (DCO) — not applied per team convention on this branch
- [x] Branch is up to date with main (rebased if needed) — merged with `origin/main`
- [x] No merge conflicts

## Additional Notes

Andrea approved with one finding (audience_parameter_name gap above, now fixed). bczoma's two findings (stale flag comment in main.go, README not updated for the third auth mode) are also fixed on the branch.

Follow-up not in scope here: `docs/authentication.md`/`broker-config.example.yaml`'s existing `tool_authorization` example under `mcp_client_auth.mode: oauth` is missing an explicit `enabled` value, which independently fails `LoadConfig` on the example file as shipped — pre-existing, unrelated to this change, not touched here.

Tracked in [SOL-152442](https://sol-jira.atlassian.net/browse/SOL-152442).

[SOL-152442]: https://sol-jira.atlassian.net/browse/SOL-152442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bczoma:
🟡 cmd/server/main.go:633-640 — stale flag comment at the construction site (good:code + good:architecture, independently; confirmed by challenger)

The comment still reads "three preconditions must all hold: the unreleased-feature flag is set, broker_oauth: is populated, AND at least one broker uses auth.mode: oauth". Hop2OAuthActive() now has two preconditions and no flag. main.go is not in the PR's changed-file set, so this was missed. This is the canonical description of the activation gate at the one place that builds the runtime — the seam a reader goes to when asking "is broker OAuth live in this process?"
Fix: same edit already made to the Hop2OAuthActive docstring; drop the "lifecycle notes for ship time" pointer.

🟡 README.md:148 and README.md:74 — README not updated for the third auth mode (docs)

Line 148 still says auth.mode — basic or bearer, and the architecture diagram at line 74 labels the server→broker leg basic / bearer. Both now contradict docs/configuration.md:79 (basic, bearer, or oauth) and broker-config.example.yaml:160 within the same commit. The README is the front door and the first config reference most operators read; it silently hides the feature this PR exists to ship.
Fix: "basic, bearer, or oauth" with a pointer to docs/authentication.md#step-2b-configure-broker-oauth-hop-2 (matching the docs/user-guide.md:32 wording this PR already added); relabel the diagram leg.

bczoma (changes requested):
see above comment

solace-aross (commented):
Nice piece of work. The removal follows the lifecycle comments you left behind exactly, the alignment invariant is now load-bearing in the right place, and the docs are a real step up from nothing. I can see bczoma's two points are already fixed on the branch, so this is additive to those.

One thing I would want settled before we call this mode fully supported.

**`audience_parameter_name` accepts values that cannot work.** `validAudienceParams` in `internal/config/config.go` allows `audience`, `scope`, and `resource`, and all three pass `validate()` cleanly. Two of them then fail at server startup, in `internal/tokenexchange/from_config.go:153` and `:155`, with "schema-accepted but not yet implemented".

That gap predates this PR, but it was unreachable: `auth.mode: oauth` was rejected outright, so nobody got as far as the audience parameter. This PR makes it reachable, and it does so at the same moment we start telling operators how to configure the mode. The two unimplemented values are exactly the ones an Entra operator (`scope`) or an RFC 8707 IdP (`resource`) reaches for, going by our own comments on those constants.

You call it out in the docs, which is the right instinct. My concern is where it fails rather than whether it is disclosed. A config that passes validation and then kills the process at startup is a worse trip than one rejected in the joined config error next to everything else. Every new test case uses `audience_parameter_name: audience`, so nothing exercises the other two paths either.

Suggestion: reject `scope` and `resource` in config validation with the same message the runtime uses. Small change, moves the failure to where operators already look, and the schema keeps documenting the intent. Alternatively narrow the allowlist to `audience` until the others land.

Not blocking from me, and happy to be told it belongs in a follow-up.

---

**One bot comment to ignore.** The thread on `CHANGELOG.md` claims the Unreleased block contradicts itself by still describing Hop 2 as gated behind `ENABLE_UNRELEASED_BROKER_OAUTH`. It does not. That line is 91, inside the `[0.5.0]` block, and it is accurate history: at 0.5.0 the runtime genuinely was gated behind the flag. Editing it would rewrite a shipped release note to say something untrue. Worth resolving that one rather than acting on it.

bczoma (dismissed):
Requested changes addressed, recommend fixing Copilot's finding in `internal/banner/banner.go`

amitmorade:
Fixed the audience_parameter_name gap: narrowed `validAudienceParams` to `audience` only, so `scope`/`resource` are now rejected at config load (joined with every other `broker_oauth` error) instead of surfacing later at Hop-2 runtime construction. Added test coverage for both rejected values and updated the docs to match the new failure point.

Also fixed `LogHop2WithoutHop1`'s doc comment, which said `validateHop1Hop2Alignment` is called "whenever any broker uses auth.mode: oauth" — contradicting the same sentence's "unconditionally" (the call is unconditional; the oauth-mode-broker check happens inside the function, not at the call site). Same fix applied to `validateHop1Hop2Alignment`'s own doc comment, which had the identical inaccuracy.

Pushed in 2d9ae2e and a6e820d.

CHANGELOG.md
```diff
@@ -53,11 +54,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Per-broker `brokers.<alias>.auth.scopes` field, removed before its first release with the field populated. Rationale: scope entitlement is a property of the user (carried on the subject token), not of the broker — a single static per-broker list either hard-fails less-privileged users (strict IdPs reject `invalid_scope`) or under-scopes admins. The exchange request now omits the RFC 6749 §3.3 `scope` parameter, so the IdP grants its per-client / per-user default scopes. Config decoding is strict (`KnownFields(true)`), so any config still declaring `auth.scopes` fails loudly at startup — this is intentional and covered by a test. Follow-up SOL-151816 will design a per-user scope mechanism; if per-call scopes ever vary for the same subject token, they must also join the token-exchange dedup key (`internal/tokenexchange/dedup_key.go`).
```
solace-aross, line 61: Add a sibling bullet here for the `audience_parameter_name` narrowing.

`[0.5.0]` (line 94) advertised the field as "allowlisted to `audience` | `scope` | `resource`", and `validateBrokerOAuthConfig` checks every `broker_oauth` field whenever the block is non-nil, whether or not any broker is on `auth.mode: oauth`. So the staging pattern that function's own doc comment invites (rule 2, "the block may be staged in advance of switching brokers to oauth mode") is exactly the path that breaks. I verified it: `broker_oauth` with `audience_parameter_name: scope` plus a single `basic`-mode broker returns `nil` from `LoadConfig` on `main` and fails at this head. Config decoding being strict makes it a hard startup failure, not a warning.

Suggested:

```
- `broker_oauth.audience_parameter_name` no longer accepts `scope` or `resource`. Both were
  schema-accepted in 0.5.0 but never implemented at the wire-construction layer, so a config
  that passed load would have failed later at Hop-2 runtime construction. The allowlist is now
  `audience` only, and the rejection joins the other `broker_oauth` errors at config load.
  Migration: a staged `broker_oauth:` block using `scope` or `resource` must switch to
  `audience`; broker OAuth is not yet usable against an IdP that requires either style.
```

Your call on `Changed` versus `Removed`. `Removed` reads better next to the `auth.scopes` bullet, since it is an allowlist member going away.

internal/config/config.go
```diff
@@ -218,23 +218,26 @@ var validGrantTypes = []string{
...
-// accepted by the schema. Membership here does not imply V1 runtime support
-// for that wire format — see the decisions doc.
+// accepted at config load. Only AudienceParamAudience — schema and runtime
+// support must land together: a value that validates here but isn't
+// implemented would move the failure from config load (joined with every
+// other broker_oauth error) to server startup once the Hop-2 runtime is
+// actually constructed, which only happens when Hop 1 is oauth AND
+// broker_oauth: is set AND a broker uses auth.mode: oauth — a worse and
+// later place for an operator to discover a typo-shaped mistake.
 var validAudienceParams = []string{
```
solace-aross, line 233: This comment is now the accurate account of the decision, but the branch it justifies still contradicts it. `resolveAudienceParam` in `internal/tokenexchange/from_config.go:152-155` returns:

```
tokenexchange: audience_parameter_name %q is schema-accepted but not yet implemented
```

for `scope` and `resource`, and after this PR neither is schema-accepted. The two tests pinning that string (`from_config_test.go:159` and `:175`) carry the same framing in their names.

Two ways out. Reword the message to drop the schema claim, or fold both cases into the existing `default` branch: at that layer `scope` is now indistinguishable from any other unaccepted value, and `TestFromConfig_UnknownAudienceParamReturnsError` already covers it. I lean toward folding, because the defensive branch was only ever earning its keep while the schema was wider than the runtime.

Worth noting if you fold: `AudienceParamScope` and `AudienceParamResource` lose their last consumer, which invalidates the reason this comment gives for keeping them declared. That would leave the const block holding the one value that means something, which seems like the honest end state.

solace-aross (changes requested):
The narrowing is the right call and it landed cleanly. Two things before merge, both from the same root: `scope` and `resource` left the schema, and two places still describe the world where they were in it.

**1. `[Unreleased]` does not record the schema narrowing.** I checked the delta against `main`. A config with `broker_oauth:` staged, `audience_parameter_name: scope`, and every broker on `auth.mode: basic` loads cleanly today, and at this head it fails with `broker_oauth.audience_parameter_name "scope" is not supported in this version`. That is a config that stops loading on upgrade, with nothing in the changelog to explain it. Suggested bullet in the inline comment.

**2. `internal/tokenexchange/from_config.go:152-155` now asserts something false.** `resolveAudienceParam` still returns `audience_parameter_name %q is schema-accepted but not yet implemented`. After this PR they are not schema-accepted. Details and two options in the inline comment on `config.go`, since `from_config.go` is not in the diff.

Everything else checks out. `internal/config`, `internal/tokenexchange`, and `internal/banner` pass locally at `a6e820d9`, and CI is green at that commit across build, lint, Analyze, and all five E2E suites. The flag removal itself is complete: no `ENABLE_UNRELEASED_BROKER_OAUTH` reference survives outside the `[0.5.0]` history block, `Hop2OAuthActive` reads as two clean structural preconditions, and the alignment invariant is now load-bearing where it belongs.

On Copilot's CHANGELOG thread: it re-fired on this commit and it is still wrong for the reason in my previous review. Those lines are 91 and 94, inside `[0.5.0]`, and they are accurate history. Resolve rather than act on it.

solace-aross (approved):
Both findings from my last review are fixed. I verified each against the code, not the commit messages.

**CHANGELOG (was Medium).** `CHANGELOG.md:63`, under `[Unreleased]` → `### Removed`, which is the right section for a narrowed allowlist. The bullet states the staged-block upgrade break and the migration step. Its load-time claim holds: `validateBrokerOAuthConfig` runs unconditionally from `validate()` at `internal/config/config.go:1094`, and the allowlist check at `:1365` fires from there whenever `broker_oauth:` is present, regardless of any broker's auth mode.

**Stale "schema-accepted but not yet implemented" (was Medium).** Gone. Folding `scope`/`resource` into the `default` case also retired the `AudienceParamScope` / `AudienceParamResource` constants, and nothing in the repo references either name now. The `AudienceFormat` enum is down to one value, so relabelling `request_test.go`'s case 2 to "out of range positive" is accurate rather than cosmetic.

Docs now quote the real error strings, including the `(must be one of [audience])` suffix that the validator appends. `docs/configuration.md:112`, `docs/authentication.md:318`, `docs/user-guide.md:223` and `broker-config.example.yaml:136` all agree on `audience` only.

`make check` green at `c3f980ac`: build, vet, golangci-lint, race-enabled tests.

**One non-blocking item, same drift class as the two strings `c3f980a` fixed:** `internal/config/config.go:112` still documents the field as `// required; one of {audience, scope, resource}`. That is the schema struct itself, so a maintainer reading `BrokerOAuthConfig` for the accepted values gets the wrong list. `validAudienceParams` sits 14 lines below and corrects the reader quickly, which is why this is not blocking. Worth a one-line fix while you are in here: `// required; must be AudienceParamAudience (see validAudienceParams)`.

Approving.

PR #228, opened 2026-07-28
SOL-151647: re-enable golangci-lint cache with a 1-day interval

solace-awaheed:
## Summary

  Re-enables the `golangci-lint-action` cache (disabled as a workaround in PR #157) now that the underlying flaky SA5011 is fixed at the source, recovering ~30s per CI lint run. CI-config change only — no product code touched.

  Fixes in SOL-151647

  ## Type of Change

  - [ ] Documentation update
  - [x] Performance improvement 
  - [ ] Dependency update

  ## Motivation and Context

  PR #157 set `skip-cache: true` on the lint step to stop a flaky SA5011 ("possible nil pointer dereference") that CI surfaced intermittently. The action's cache key hashes only `go.mod` — not the source tree or `.golangci.yml` — so a stale analysis cache could report findings against code that had since changed. `skip-cache` forced a full
  re-analysis every run (honest, but ~30s slower).

  That was a workaround. The real defect — test sites relying on staticcheck recognizing `t.Fatal` as a terminator — was fixed at the source in commit `669581b` (explicit `return`s added). With the finding gone from the tree, a stale cache can no longer resurface it, so caching is safe to re-enable.

  ## Changes Made

  - Removed `skip-cache: true` from the `golangci-lint` step in `.github/workflows/build-and-test.yml`, re-enabling the action's caching.
  - Added `cache-invalidation-interval: 1` to bound any staleness window to 1 day (vs. the 7-day default), since the action exposes no source-aware cache key (confirmed unchanged through v9).
  - Rewrote the accompanying comment to document the new caching rationale.

  ## Testing

  **Test Environment:**
  - Go version: 1.26.1
  - OS: macOS (darwin/arm64)
  - Broker version (if applicable): N/A (no broker interaction)

  **Tests Performed:**

  **Test Coverage:**
  - `make check` (build, vet, race tests, lint): passed, exit 0.
  - `golangci-lint v2.11.4 run --verbose` (mirrors the action's `args`): **0 issues**, with staticcheck active and a full analysis (368 raw findings → 0 after exclusion rules) — confirming the former SA5011 sites are genuinely clean, not masked by a cache.
  - `go build -v ./...`, `go vet ./...`, and the separate broker-driver `go vet ./...`: all exit 0.
  - Note: the action's cache round-trip (warm-cache vs. `skip-cache` agreement) can only be exercised on GitHub runners; it validates naturally in CI (first run = cache miss → full analysis; later runs = warm cache on a deterministic clean tree).

  ## Breaking Changes

  **Impact:** None.

  **Migration Guide:** N/A.

  ## Checklist

  ### Code Quality

  ### Security

  ### Testing
  - [x] Error paths tested 

  ### Documentation

  ### Git & CI
  - [ ] All commits are signed off (DCO: `git commit -s`) 
  - [x] Branch is up to date with main (branched off origin/main)
  - [x] No merge conflicts


solace-aross (dismissed):
Approving. The reasoning holds up and I verified the load-bearing claims rather than taking the description's word.

**The input is real.** `cache-invalidation-interval` is a declared input on `golangci-lint-action@v7` with default `'7'`, so both the field name and the "default is 7 days" claim in the comment are correct.

**The source fix is real.** `669581b` adds explicit `return`s after `t.Fatal` at three sites (`internal/semp/sempv1/envelope_test.go`, `truncate_test.go`, `internal/tools/composite_handler_test.go`). SA5011 cannot fire there now regardless of cache state.

**Best available evidence that the tree is genuinely clean, not masked.** The `lint` job on this PR was necessarily a cold, full analysis: `main` still carries `skip-cache: true`, so the action has never saved a cache under this key prefix and there was nothing to restore. That makes the passing 46s run a full uncached analysis on Linux, which is the exact platform where the flake used to fire and the one your local Darwin run could not cover. Stronger than the local verification in the description.

**One refinement to the comment's model, non-blocking.** From `src/cache.ts` at v7, `buildCacheKeys` pushes the key *before* appending the go.mod checksum, so the go.mod hash lands only in the primary key and the fallback restore key is `golangci-lint.cache-<OS>-[<workdir>-]<intervalBucket>-`. `restoreCache` prefix-matches on that, so when go.mod changes the action still restores a cache saved under a different go.mod hash from the same bucket. The go.mod component narrows the primary hit but does not guarantee invalidation on a dependency change. That strengthens rather than weakens the choice of `1`: the interval bucket, not the go.mod hash, is what actually bounds the fallback window. Worth folding into the comment so the next person reading it has the accurate model. Also note the bucket is `floor(epochSeconds / 86400)`, so it turns over at 00:00 UTC rather than being a rolling 24 hours.

**Second non-blocking note.** The `--verbose` comment immediately below now sits slightly at odds with the change. It says `--verbose` exists "so a '0 issues' result can be verified as full analysis rather than a silent skip." With caching on, a warm run legitimately analyzes less and that signal gets harder to read. `--verbose` is still worth keeping; the stated rationale is what needs a touch-up, and you are already editing the comment directly above it.

Neither note blocks. Only one `golangci-lint-action` usage exists in `.github/workflows/`, so there is no second call site to keep in sync, and full CI is green including all five E2E suites.

solace-aross (approved):
Re-approving. The only branch change since my last pass is the comment rewrite, and it lands both notes I raised.

Re-verified against `golangci-lint-action@v7` source rather than trusting my earlier pass:

- `buildCacheKeys` pushes the fallback key at `src/cache.ts:57`, before the go.mod sha1 is appended at `:66`. The fallback restore key really is `golangci-lint.cache-<OS>-[<wd>-]<bucket>-`, so "the go.mod hash only narrows the primary key" is correct as written.
- `getIntervalKey` at `src/cache.ts:36` is `Math.floor(secondsSinceEpoch / (days * 86400))`. With interval 1 that is a UTC-midnight bucket, not a rolling 24 hours. The comment matches.
- `cache-invalidation-interval` is still declared with default `'7'` at `action.yml:55`.

The `--verbose` rationale now claims only what it can prove with caching on, which is the right correction.

Functional change is unchanged from the version I already verified: `skip-cache` out, `cache-invalidation-interval: 1` in, one call site. CI green at bb753c0 with e2e-monitoring still running.

PR #230, opened 2026-07-28
fix(authz): [SOL-152508] warn when an RBAC write-tool grant is inert under enable_write_tools=false

bczoma:
## Summary

An RBAC grant of a write/action tool while `enable_write_tools: false` was accepted at startup in complete silence, even though the grant can never take effect. The startup validator now knows about the write-tool gate and emits one `WARN` per inert grant, naming the tool and the groups that reference it.

Jira: **SOL-152508**

## Type of Change

- [x] Documentation update

### The problem

With `mcp_client_auth.tool_authorization.enabled: true`, granting a write/action tool (`delete-queue-messages`, `disconnect-client`, `clear-queue-stats`, `clear-client-stats`, or any Config-API management tool) in `access_level_groups` while `enable_write_tools: false` produced no error and no WARN. The grant had zero effect — the tool is never exposed on the MCP server — so an operator had no way to tell "my policy is live" from "my policy is inert."

That was inconsistent with the two sibling cases the same validator already handled:

| Grant | Effect | Signal before | Signal now |
|---|---|---|---|
| Misspelled tool name | inert | fatal error | fatal error (unchanged) |
| `list-brokers` (structurally exempt) | inert | WARN | WARN (unchanged) |
| Write tool while `enable_write_tools: false` | inert | **nothing** | **WARN** |

### Root cause: the two-registry split

Two collections are populated at startup and they do **not** hold the same tools:

- **`mgr` (`*tools.ToolManager`)** holds every tool unconditionally. Populated by `NewToolManagerFromComposite` (`cmd/server/main.go:701`), `registerSEMPv1Tools(mgr)` (`:702`), `registerMixedTools(mgr)` (`:703`) — none of which take `enableWriteTools`.
- **The MCP `server`** holds only tools that pass the write gate.

`enable_write_tools` reached exactly one place: `tools.RegisterWithServer(...)` (`main.go:729`). Its gate at `internal/tools/register.go:157` does

```go
if isWriteTool(reg.meta.Annotations) && !enableWriteTools {
    slog.Info("write tool not registered (enable_write_tools is false)", ...)
    continue
}
```

so it skips `server.AddTool` but leaves the handler in `mgr`. `tools.ValidatePolicyToolNames` then built its known-tool set from `mgr.Handlers()` — the ungated superset — so a gated-out write tool was "known" and the grant passed silently.

## Changes Made

- `internal/tools/authorization.go`: `ValidatePolicyToolNames` takes a new `enableWriteTools bool`, computes the write-gated subset of `mgr.Handlers()` using the existing `isWriteTool` predicate, and emits one `slog.Warn` per inert grant with `gated_tool` and `referenced_by_groups`. Structure, naming, ordering and dedup mirror the existing `list-brokers` `exemptToGroups` WARN and the unknown-tool error rows.
- Extracted two small helpers used by both the new gated path and the pre-existing unknown path, rather than duplicating the pattern: `addGroupForTool` (populate a tool→groups index) and `sortedToolNames` (alphabetize such an index). No behavior change to the unknown-tool path.
- `cmd/server/main.go`: pass `cfg.EnableWriteTools` at the validation call site, the same value already handed to `RegisterWithServer` a few lines above.
- `docs/configuration.md`: new "Interaction with `enable_write_tools`" paragraph in the Tool Authorization section.
- `CHANGELOG.md`: entry under `[Unreleased]` → `Fixed`.
- Four new unit tests in `internal/tools/authorization_validate_test.go`, plus a `validateTestManagerWithWriteTools` helper that mirrors startup (mgr holds write tools regardless of the flag).

### Design decision: pass the `bool`, not a precomputed gated-tool set

I considered two ways for the gate to reach the validator:

1. **Add `enableWriteTools bool`** and let the validator derive the gated set from `mgr` — chosen.
2. Have the caller compute the set of gated-out tool names and pass it in.

Option 1 wins on three counts:

- **Consistency with existing signatures.** `RegisterWithServer(mgr, server, pool, enableWriteTools bool, policy, groupsClaimName)` already threads this exact flag as a bool from the same config value at the adjacent call site. The validator now reads identically, and `main.go` passes one value to both — making "these two must agree" obvious at a glance.
- **Single source of gate truth (DRY).** The predicate that decides what "gated" means stays in one place: `isWriteTool`, used by `RegisterWithServer` and now by the validator. Option 2 would require exporting `isWriteTool` (or reimplementing the annotation check) in `cmd/server`, giving the gate two implementations that can silently diverge — exactly the drift class this bug came from.
- **Locality / no leaked internals.** Computing the gated set in `main.go` would push a `tools`-package registration detail (annotations → gate) into the composition root. The validator already walks `mgr.Handlers()` and already reads metadata; adding the annotation check there costs one line and no new coupling.

The cost of option 1 is one more positional bool on an exported function. That is mitigated by the doc comment stating the precondition ("must be the same value handed to `RegisterWithServer`") and by the flag already existing in the sibling signature, so the two read as a pair.

**A third shape, acknowledged but not implemented.** `RegisterWithServer` could *return* the tool names it actually skipped, and that slice could be fed straight into the validator. That is the cleaner design on three counts: cross-site agreement becomes compiler-enforced rather than a convention documented in a doc comment (`authorization.go:173`); the duplicated derivation loop at `authorization.go:181-189` disappears; and the validator would report on what registration *did* rather than re-deriving what it *should have* done. I did not take it here — the shared-`isWriteTool` design already closes the practical drift (one predicate, one input collection, verified by the tests below), and changing `RegisterWithServer`'s signature and return contract is beyond this bug's scope. Recording it so the trade-off is on file if this code is revisited: if the write gate ever grows a second condition beyond the read-only annotation, prefer the returned-skip-list shape over threading another flag.

## Testing

**Test Environment:**
- Go version: go1.25.0 linux/amd64
- OS: Linux 6.6.87.2-microsoft-standard-WSL2 (WSL2)
- Broker version (if applicable): n/a — startup-validation logic, no broker interaction

**Tests Performed:**

**Test Coverage:**

New tests in `internal/tools/authorization_validate_test.go`:

- `..._WriteToolGrantWhileWriteToolsDisabled_WarnsNotError` — inert grant returns `nil` error and emits exactly one WARN carrying `gated_tool=delete-queue-messages`, `referenced_by_groups=Ops`, and the string `enable_write_tools` (the remedy).
- `..._WriteToolGrantWithWriteToolsEnabled_Silent` — same config with `enableWriteTools: true` → `nil` error and **zero** log output.
- `..._MultipleWriteGated_DedupedPerToolAndAlphabetized` — two gated tools across two groups → one WARN per tool, tools alphabetized (`clear-queue-stats` before `disconnect-client`), groups deduped and alphabetized in-row (`"Alpha, Zeta"`).
- `..._WriteGatedAndUnknown_BothSurface` — an inert grant next to a typo: exactly one error row for the typo (the gated tool is *not* reported as unknown) **and** one gated-tool WARN. The two signals do not mask each other.

Existing tests were updated to pass the new argument; all pre-existing assertions (unknown-tool fatality, `list-brokers` WARN, case-mismatch, empty-map silence) are unchanged and still pass.

Review follow-up: `..._ListBrokersGrant_WarnsNotError` is now table-driven over both `enable_write_tools` values and additionally asserts the emitted WARN is the `exempt_tool` one. Previously every write-gate test ran with the gate closed and every pre-existing test with it open, so the exempt path was never exercised under a closed gate. Branch order in the validator makes misclassification impossible, but AC 5's `list-brokers` half is now pinned down rather than inferred.

`make check` (build, vet, golangci-lint, `go test -race -v ./...`) exits 0 with zero failures:

```
$ make check ; echo "EXIT=$?"
go build -v ./...
go vet ./...
golangci-lint run
go test -race -v ./...
ok  	github.com/SolaceDev/solace-broker-mcp/cmd/server
ok  	github.com/SolaceDev/solace-broker-mcp/internal/auth
ok  	github.com/SolaceDev/solace-broker-mcp/internal/authz
ok  	github.com/SolaceDev/solace-broker-mcp/internal/banner
ok  	github.com/SolaceDev/solace-broker-mcp/internal/composite
ok  	github.com/SolaceDev/solace-broker-mcp/internal/composite/postprocess
ok  	github.com/SolaceDev/solace-broker-mcp/internal/composite/postprocess/handlers
ok  	github.com/SolaceDev/solace-broker-mcp/internal/config
ok  	github.com/SolaceDev/solace-broker-mcp/internal/idpclient
ok  	github.com/SolaceDev/solace-broker-mcp/internal/middleware/recovery
ok  	github.com/SolaceDev/solace-broker-mcp/internal/oauth/cache
ok  	github.com/SolaceDev/solace-broker-mcp/internal/observability/audit
ok  	github.com/SolaceDev/solace-broker-mcp/internal/observability/correlation
ok  	github.com/SolaceDev/solace-broker-mcp/internal/observability/health
ok  	github.com/SolaceDev/solace-broker-mcp/internal/observability/logging/sanitize
ok  	github.com/SolaceDev/solace-broker-mcp/internal/observability/metrics
ok  	github.com/SolaceDev/solace-broker-mcp/internal/observability/tracing
ok  	github.com/SolaceDev/solace-broker-mcp/internal/safego
ok  	github.com/SolaceDev/solace-broker-mcp/internal/semp
ok  	github.com/SolaceDev/solace-broker-mcp/internal/semp/auth
ok  	github.com/SolaceDev/solace-broker-mcp/internal/semp/correlationhdr
ok  	github.com/SolaceDev/solace-broker-mcp/internal/semp/resilience
ok  	github.com/SolaceDev/solace-broker-mcp/internal/semp/sempv1
ok  	github.com/SolaceDev/solace-broker-mcp/internal/semp/sempv2
ok  	github.com/SolaceDev/solace-broker-mcp/internal/tokenexchange
ok  	github.com/SolaceDev/solace-broker-mcp/internal/tools
ok  	github.com/SolaceDev/solace-broker-mcp/internal/tools/queuemetrics
ok  	github.com/SolaceDev/solace-broker-mcp/internal/tools/sempv1/brokerstatus
ok  	github.com/SolaceDev/solace-broker-mcp/internal/tools/sempv1/discardstats
ok  	github.com/SolaceDev/solace-broker-mcp/internal/tools/sempv1/redundancy
ok  	github.com/SolaceDev/solace-broker-mcp/test/integration
EXIT=0
```

Targeted run of the changed suite:

```
$ go test ./internal/tools/ -run TestValidatePolicyToolNames -count=1 -race
16 passed
```

## Acceptance criteria coverage

| AC | Status | Where |
|---|---|---|
| `ValidatePolicyToolNames` (or caller) is aware of `enable_write_tools` | Done | new `enableWriteTools` param; `main.go:743` passes `cfg.EnableWriteTools` |
| Write-tool grant with `enable_write_tools: false` → WARN naming tool(s) + group(s), no failure | Done | `..._WriteToolGrantWhileWriteToolsDisabled_WarnsNotError` |
| Multiple inert grants deduped per tool with groups listed, consistent with existing style | Done | `..._MultipleWriteGated_DedupedPerToolAndAlphabetized` |
| With `enable_write_tools: true`, same grant is silent | Done | `..._WriteToolGrantWithWriteToolsEnabled_Silent` |
| Unknown-tool grants remain FATAL; `list-brokers` keeps its WARN | Done | pre-existing tests unchanged and passing; `..._WriteGatedAndUnknown_BothSurface`; `..._ListBrokersGrant_WarnsNotError` now runs under both flag values |
| Unit tests: inert warns/no-fail, enabled is silent, inert + unknown surface independently | Done | four new tests above |
| `docs/configuration.md` describes the interaction | Done | "Interaction with `enable_write_tools`" in the Tool Authorization section |

## Breaking Changes

None. `ValidatePolicyToolNames` is exported but internal-only (`internal/tools`), with `cmd/server/main.go` as its sole non-test caller — updated in this PR. No config schema, tool name, or tool output-shape change.

**Impact:** none.

**Migration Guide:** none required. An operator who was relying on the silence now sees a startup WARN; setting `enable_write_tools: true` (or removing the write-tool grant) clears it.

### Code Quality
- [x] Self-review completed

### Security
- [x] Followed secure logging rules — ran the `/check-logs` scan over the diff: no CRITICAL/HIGH findings. The new WARN uses structured `slog.String` attributes only; the values are an MCP tool name and admin-authored group names from the server's own YAML (not caller-supplied token claims), matching how the adjacent `list-brokers` WARN already reports group names. No message-side `fmt.Sprintf`, no raw structs, no stdout.
- [x] Credential-carrying types implement `slog.LogValuer` — none added

### Testing
- [x] Error paths tested

### Documentation
- [ ] README.md updated — not user-facing at README level
- [x] docs/ updated (`docs/configuration.md`)
- [x] CHANGELOG.md updated
- [ ] Examples updated — deliberately **not**, see Additional Notes

### Git & CI
- [x] Branch is up to date with main (rebased onto `7768d7f`; 0 commits behind)
- [x] No merge conflicts
- [ ] CI checks passing — pending run on the rebased branch

## Additional Notes

**`broker-config.example.yaml` was deliberately NOT touched.** That file is owned by a separate in-flight branch, so documenting this interaction there would create an avoidable conflict. The operator-facing explanation lives in `docs/configuration.md` instead. If the example config should also carry a comment about the gate/RBAC interaction, that is best done as a follow-up on top of the branch that owns the file.

Scope was kept to the validator, its call site, its tests, and docs — no unrelated refactors. `internal/tools/register.go` was not modified: `isWriteTool` is package-private and the validator lives in the same package, so no export was needed.

---

**For Reviewers:**

**Focus Areas**: (1) the design choice of a positional `bool` vs. passing a gated-name set — see rationale above; (2) whether the WARN wording and attribute names (`gated_tool`, `referenced_by_groups`) sit right next to the existing `exempt_tool` WARN for anyone grepping startup logs; (3) the two extracted helpers, which now also serve the pre-existing unknown-tool path.

**Questions**: Should the gated-tool WARN also fire for a `list-brokers` grant's sibling case where a *read-only* tool is somehow absent from `mgr`? Today that stays a fatal unknown-tool error, which I believe is correct — flagging only in case there is a known deployment shape where it is not.


amitmorade (approved):
Traced this end-to-end against the actual registration/validation flow (`RegisterWithServer` in `register.go` and `ValidatePolicyToolNames` in `authorization.go`) to make sure the fix lines up with how the two-registry split actually behaves — it does.

Confirmed:
- `writeGated` is derived from the same `isWriteTool` predicate `RegisterWithServer` uses to skip `server.AddTool`, so the validator and the real registration gate can't drift apart.
- Branch order in the per-tool loop (`list-brokers` → `writeGated` → `known` → fatal-unknown) is correct: `writeGated` is always a subset of `known`, and checking it first is what lets a gated tool get its own WARN instead of silently falling into "known, skip" or wrongly into "unknown, fatal."
- The gated-tool WARN path is fully separate from `unknownToGroups`, so it can't mask a real typo — confirmed by the `WriteGatedAndUnknown_BothSurface` test.
- WARN shape (dedup per tool, alphabetized tool + groups) matches the existing `list-brokers` exempt convention, so this doesn't introduce a second log shape for the same class of "inert grant" signal.

Approving

solace-lootii (approved):
LGTM

helen-zhu-solace (approved):
LGTM as well! - Small nit: 
`sortedKeys` vs `sortedToolNames (internal/tools/authorization.go)` — they're the same operation over `map[string]struct{}` and `map[string]map[string]struct{}` respectively. A Go 1.18+ generic `sortedKeys[V any](m map[string]V) []string` would collapse them to one. Fine as-is; noting because the PR did extract helpers to avoid duplication, and this is the same pattern.

PR #213, opened 2026-07-23
SOL-152328:Add Kafka Receiver/Sender monitoring tools (list-kafka-receivers, get-kafka-receiver-status, list-kafka-senders, get-kafka-sender-status)

solace-lootii:
## Summary

Adds 4 read-only MCP tools for Kafka Receiver/Sender monitoring: `list-kafka-receivers`, `get-kafka-receiver-status`, `list-kafka-senders`, `get-kafka-sender-status`.

Tracked under SOL-152328.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

Continues the monitoring-tool coverage established for RDPs and Bridges (SOL-152124) to Kafka Receivers and Kafka Senders, following the same "list minimal + get-X-status detail" pattern used across the server. A Kafka Receiver pulls messages from an external Kafka cluster into a Message VPN; a Kafka Sender pushes messages from a VPN's queues out to an external Kafka cluster.

## Changes Made

- New composite postprocess handler `ListKafkaReceivers` (`internal/composite/postprocess/handlers/list_kafka_receivers.go`) — aggregates `downCount`, `disabledCount`, `byFailureReason` from the raw list, mirroring `list-rdps`'s single up/enabled/failureReason shape
- New composite postprocess handler `ListKafkaSenders` (`internal/composite/postprocess/handlers/list_kafka_senders.go`) — identical shape
- 4 new tool definitions in `internal/composite/definitions/tools.yaml`: `list-kafka-receivers`, `get-kafka-receiver-status`, `list-kafka-senders`, `get-kafka-sender-status`
- `list-kafka-receivers`/`list-kafka-senders`: paginated (`followPages`, default 100/max 500 via `maxResults`), curated `select:` field lists, summary block via postprocess
- `get-kafka-receiver-status`/`get-kafka-sender-status`: single-object detail tools including topic-binding health (`topicBindingUpCount`/`topicBindingCount`) or queue-binding health (`queueBindingUpCount`/`queueBindingCount`)
- Updated `internal/composite/loader_embedded_test.go`'s `wantToolCount` (30 → 34) to reflect the 4 new tools
[...]

PR #231, opened 2026-07-29
SOL-152370: Add E2E coverage for Kafka Receiver/Sender monitoring tools

solace-lootii:
## Summary

Adds full live-broker E2E coverage (`test/e2e-monitoring`) for the 4 Kafka Receiver/Sender monitoring tools added in SOL-152328, closing the verification gap that story left open — including a real Kafka broker in the test environment, not just placeholder empty-collection checks.

Tracked under SOL-152370.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

None of these quite fit — this is test-infrastructure-only work. No production code (`internal/`, `cmd/`) is touched; only `test/e2e-monitoring/{docker-compose.yml,helpers.sh,test-monitoring-tools.sh,README.md}` and `CHANGELOG.md`.

## Motivation and Context

SOL-152328 shipped `list-kafka-receivers`, `get-kafka-receiver-status`, `list-kafka-senders`, and `get-kafka-sender-status` spec-derived rather than lab-verified. This story set out to close that gap the way SOL-152231 did for bridges.

Initial scoping concluded that was blocked entirely by a Kafka-Bridging license/edition gate with no way around it. Further investigation during this same story found that conclusion was wrong: creating a Kafka Receiver/Sender is gated by two scaling settings that default to 0, not a license restriction — see Changes Made and Testing below for the full corrected picture.

## Changes Made

- **`docker-compose.yml`**: added `SYSTEM_SCALING_MAXKAFKABRIDGECOUNT=10` and `SYSTEM_SCALING_MAXKAFKABROKERCONNECTIONCOUNT=300` to both Solace brokers (the actual unlock — see Testing), and a new `kafka` service (`apache/kafka:3.7.0`, KRaft mode, no ZooKeeper) that both brokers bridge to
- **`helpers.sh`**: new F9/F10 fixture triads — `create/verify/cleanup_kafka_receivers_on` and the sender equivalents — each provisioning three real objects mirroring bridges' pattern: healthy (bound to the real Kafka broker), unreachable (`127.0.0.1:1`, feeds `downCount`), and admin-disabled (`enabled=false`, feeds `disabledCount`). Includes a `kafka_broker_available` guard so the sibling `e2e-llm` suite, which reuses this file's fixture code but has no Kafka service in its own compose, degrades gracefully (logs a warning and skips) instead of hanging on a container that will never appear
- **`test-monitoring-tools.sh`**: replaced the old empty-collection-only tests with real assertions — presence, pagination, and summary (`downCount`/`disabledCount`/`byFailureReason`) for `list-kafka-receivers`/`list-kafka-senders`; healthy (`up:true`) and failing (`up:false`, real `failureReason`) for `get-kafka-receiver-status`/`get-kafka-sender-status`. Kept the not-found error-translation tests from the prior version of this PR unchanged
- **`README.md`**: replaced the old "no-fixture note" with a full "F9/F10 — Kafka Receiver/Sender findings" section (mirroring the existing bridges findings section's format) documenting the corrected scaling-gate finding, the enum-restricted valid values, `failureReason` populating reliably, and binding cascade-delete behavior. Added F9/F10 rows to the fixture table and `KAFKA_PORT` to the port-allocation table
- **`CHANGELOG.md`**: rewrote the `[Unreleased]` entry to reflect the corrected findings rather than the earlier "blocked" framing

## Testing

**Test Environment:**
- Go version: 1.25.0
- OS: Linux (Fedora 6.14.5)
- Broker version: `solace/solace-pubsub-standard:latest` (this suite's image) + `apache/kafka:3.7.0` (KRaft mode)

**Tests Performed:**
- [x] Unit tests added/updated — N/A, this PR is E2E-only, no Go source changed

**Test Coverage:**
- **The earlier "license gate" conclusion was wrong.** Creating a Kafka Receiver/Sender is gated by two scaling settings that default to 0 on this broker image — not a license restriction. The confd env-backend keys are `/system/scaling/maxkafkabridgecount` and `/system/scaling/maxkafkabrokerconnectioncount`, set via `SYSTEM_SCALING_MAXKAFKABRIDGECOUNT`/`SYSTEM_SCALING_MAXKAFKABROKERCONNECTIONCOUNT` (note the `SYSTEM_` prefix — an initial wrong guess at `SOLACE_SCALING_*` silently no-ops rather than erroring). Both are **enums**, confirmed against the broker's own SEMP RPC schema: bridge count must be `0, 10, 50, 200`; broker-connection count must be `0, 300, 2000, 10000` — not free integers.
- With that unlocked: **`failureReason` was confirmed to populate reliably** — `"Shutdown"` (disabled), `"No remote-broker in UP state"` (unreachable), stable across 45s of polling — unlike bridges' `inboundFailureReason`, which never populates. A healthy Kafka Receiver/Sender was confirmed to converge to `up: true` against the real Kafka broker within 5-7 seconds. Topic/queue bindings work via `topicName`/`localTopic` (receiver) and `queueName`/`remoteTopic` (sender) fields; a queue binding's `enabled` defaults to `false` and must be set explicitly. Deleting a Kafka Receiver/Sender does not cascade its bindings (like RDPs, unlike bridges).
- **A suspected production bug turned out to be a false alarm on my end**: `get-kafka-receiver-status`/`-sender-status`'s `select` fields (`topicBindingCount` etc.) appeared to 400 when tested by hand against the broker's public `/SEMP/v2/monitor` API. Traced with temporary debug instrumentation (reverted, zero net diff on `client.go`) and confirmed this server's composite executor has always used the *private* `/SEMP/v2/__private_monitor__` API instead (the embedded spec's actual `basePath`, an existing pattern elsewhere in this codebase), where these fields work correctly. No code change was needed.
- Full suite run: **90/90 tests passed** (up from 66 before SOL-152370, 74 in this PR's earlier iteration), including all 24 new Tool 16-19 cases across both brokers.
- `go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 issues after clearing a known stale-cache SA5011 flake), `go test -race ./...` (31 packages, 0 failures) all clean.

## Breaking Changes

N/A — test-infrastructure-only change, no production code touched.

**Impact:**

**Migration Guide:**

### Code Quality

### Security
- [x] Followed [secure logging rules](../docs/secure-logging-rules.md) — N/A, no Go code/logging touched
- [x] Credential-carrying types implement `slog.LogValuer` — N/A

### Testing
- [x] Error paths tested

### Documentation
- [ ] docs/ updated (if architecture/design changes) — N/A, top-level `docs/` unaffected
- [ ] godoc comments added/updated for exported functions — N/A, no Go source changed
- [ ] Examples updated (if applicable) — N/A

### Git & CI
- [x] Branch is up to date with main (rebased if needed) — merged latest `main` (includes SOL-152508's RBAC warning and PR #224's DCO enforcement), resolved one real `CHANGELOG.md` conflict
- [x] No merge conflicts
- [ ] CI checks passing (lint, build, test, E2E) — pending CI run on this push; all checks pass locally

- [ ] Breaking changes documented in commit message — N/A
- [ ] Breaking changes documented in PR description — N/A
- [ ] Migration guide provided — N/A
- [ ] Deprecated functionality marked with comments and deprecation notices — N/A

N/A

## Additional Notes

This PR's scope grew significantly from its earlier iteration: what started as "empty-collection tests only, blocked by licensing" turned into full state-scenario coverage once the actual blocker (a scaling default, not a license) was found and fixed. The corrected findings are documented prominently in `README.md`'s F9/F10 section so nobody re-derives them.

## Related Issues/PRs

- Related to #213 (SOL-152328 — added the 4 tools this PR tests)
- Depends on: none (already rebased on latest `main`)
- Blocks: none

---

**For Reviewers:**

**Focus Areas**: The `kafka_broker_available` guard in `helpers.sh` — want a second opinion on whether skip-with-warning is the right behavior for the `e2e-llm` suite reuse case, versus some other approach.

**Questions**: None outstanding.


solace-aross (dismissed):
Approving. Test-only, no production path touched, and the eight cases are registered correctly against both brokers.

On the scope question you flagged: land it. A documented license gate, with the exact SEMP error and the spec check that ruled out a config knob, is worth more than holding the ticket open for a broker nobody has. The not-found cases are the real gain: they put live-broker coverage under the error translation that was previously spec-derived only.

One thing to close before it drops off the radar. The deferred scenarios (enabled/up, disabled, down-with-failureReason, and whether `failureReason` populates at all) are named in the PR body, the CHANGELOG, and the README, and tracked nowhere. Worth a follow-up ticket referenced from the no-fixture note, so the gap has an owner instead of a paragraph.

helen-zhu-solace (dismissed):
LGTM as well! e2e-monitoring passed

amitmorade (dismissed):
Verified the empty-collection assertions against `ListKafkaReceivers`/`ListKafkaSenders` in `internal/composite/postprocess/handlers/list_kafka_{receivers,senders}.go` — with a zero-item `data`, the handler produces `downCount: 0`, `disabledCount: 0`, `byFailureReason: {}`, `scanned: 0`, with `skipped`/`truncated` omitted when zero/false. That matches the new test assertions exactly, so this isn't guessed — it's mechanically derivable from the existing (unchanged) aggregation logic.

The not-found error assertions (`isError`, `retryable: false`, `sempCode: 6`, `sempStatus: NOT_FOUND`, `suggestions[0]`) are a faithful mirror of the existing `get-rdp-status` not-found test in `test/e2e-basic-mcp/test-standalone.sh`.

+1 to Andrea's point on filing a follow-up ticket for the deferred state-scenario coverage — agreed that's the one thing to close out before it drifts into just-a-paragraph territory.

Approving — scope reduction is well-justified given the license gate, and the coverage that does land is correct.

solace-lootii:
Thanks @amitmorade @solace-aross @helen-zhu-solace , I'm making some changes actually. I noticed something important as I looked into this further.

test/e2e-monitoring/helpers.sh
```diff
@@ -268,6 +268,203 @@ cleanup_bridges_on() {
...
+    while [ $attempt -lt "$max_attempts" ]; do
+        if docker exec kafka-e2e-mon /opt/kafka/bin/kafka-broker-api-versions.sh \
+            --bootstrap-server localhost:9092 >/dev/null 2>&1; then
+            log_ok "Kafka broker ready after ${attempt}s"
+            return 0
+        fi
+        sleep 1
+        attempt=$((attempt + 1))
+    done
+    log_warn "Kafka broker not ready after ${max_attempts}s — proceeding anyway"
```
helen-zhu-solace: `wait_for_kafka` log message contradicts control flow (real bug)
The message promises "proceeding anyway," but the function returns non-zero, and `create_fixtures` calls it directly (not in a conditional) with `set -e` in effect. On timeout the script will actually abort — the opposite of what the log says. Copilot already flagged this and I agree it deserves a fix before merge.

Pick one:
Fail loudly: keep `return 1`, change the message to something like "Kafka broker not ready after ${max_attempts}s — aborting F9/F10 setup" so the log matches the behavior; or
Truly proceed: `return 0`, and let the downstream `create_kafka_topics`/receiver creation surface the real error. Given that` create_kafka_topics` and the healthy fixture cannot succeed without Kafka, the first option matches actual behavior more honestly.

test/e2e-monitoring/helpers.sh
```diff
@@ -268,6 +268,203 @@ cleanup_bridges_on() {
...
+# this guard, create_fixtures would call wait_for_kafka there, which would
+# spend up to a minute retrying a container that will never appear, then
+# return non-zero and (this file runs under `set -e`) abort the entire
+# fixture set for that suite, not just the Kafka piece. A fast existence
+# check avoids that regardless of how any one caller's fixtures are
+# assembled. Cleanup functions need no equivalent guard: they only issue
+# semp_delete against the Solace broker (safe/idempotent on a 404 regardless
+# of whether kafka-e2e-mon ever existed), never docker exec against Kafka
+# itself.
+kafka_broker_available() {
```
solace-aross: This predicate answers "is the container running", but the decision you actually need is "does this suite expect Kafka at all". Those two come apart in exactly the case that matters.

The Tool 16-19 tests assert fixture presence unconditionally: `test-monitoring-tools.sh:1160-1163` requires all three receivers by name, `:1181` requires exactly 3, `:1193` requires `summary.scanned == 3`. So in this suite, if the `kafka` container fails to come up for any ordinary reason (image pull failure, the port clash on line 78, a healthcheck that never passes), the sequence is:

1. `kafka_broker_available` returns false.
2. `create_fixtures` logs one warning and skips F9/F10 (line 1033).
3. The 24 Tool 16-19 cases run anyway and fail on "test-kafka-receiver must be present".

An infrastructure failure surfaces as two dozen mystery assertion failures, with the real cause sitting in a single `log_warn` far up the log.

Gate on declared intent instead of runtime state. The compose file is already the source of truth for whether a suite has Kafka, and `SUITE_DIR` is already the established contract in this file (header comment, line 15):

```bash
# True when the current suite's compose stack declares a Kafka service at all.
# Intent, not runtime state: a declared-but-not-running Kafka is a failure worth
# surfacing, while a sibling suite with no kafka service is a legitimate skip.
kafka_expected() {
    docker compose -f "$SUITE_DIR/docker-compose.yml" config --services 2>/dev/null | grep -qx kafka
}
```

Then `if kafka_expected; then wait_for_kafka; ...` at line 1021. Under `e2e-llm`, `SUITE_DIR` is the LLM suite directory, whose compose declares no `kafka` service, so F9/F10 skip by declaration, which is the behavior you were after. Under this suite it is always true, so a Kafka that never becomes ready aborts at `wait_for_kafka` with a message that names Kafka, rather than 24 lines of downstream noise.

Your existing comment block is accurate about the `e2e-llm` hazard and worth keeping. It is the predicate, not the reasoning, that needs the change.

test/e2e-monitoring/docker-compose.yml
```diff
@@ -42,3 +52,33 @@ services:
...
+      - KAFKA_LISTENERS=PLAINTEXT://:9092,CONTROLLER://:9093
+      - KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://kafka-e2e-mon:9092
+      - KAFKA_CONTROLLER_LISTENER_NAMES=CONTROLLER
+      - KAFKA_LISTENER_SECURITY_PROTOCOL_MAP=CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT
+      - KAFKA_CONTROLLER_QUORUM_VOTERS=1@kafka-e2e-mon:9093
+      - KAFKA_INTER_BROKER_LISTENER_NAME=PLAINTEXT
+      - KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR=1
+      - CLUSTER_ID=ci4PjZ5aQzq5TZ4S3o1AeA
+    ports:
+      - "${KAFKA_PORT:-9094}:9092"
```
solace-aross: 9094 is already claimed by `e2e-llm`. `test/e2e-llm/targets/local-docker.env:8` sets `MCP_URL="${MCP_URL:-http://localhost:9094}"`, and `e2e-llm/README.md:143` documents "MCP :9094" for that suite.

That matters here specifically because `e2e-llm` is the suite that sources this file's sibling `helpers.sh` for its fixture code, so the two stacks are meant to coexist. The port table you extended states its own purpose as "Distinct from `e2e-basic-mcp` so both suites can run concurrently", and the same reasoning applies to the sibling that shares the fixture code. Every other port in this suite is deliberately partitioned already: SEMP 8090/8092 against the LLM suite's 8102/8104, SMF 55655/55656 against 55661/55662.

CI will not catch this, since the E2E suites run as separate jobs on separate runners. Local runs will: whichever stack binds second fails, and if Kafka wins the race then `e2e-llm`'s precheck sends an HTTP request to a Kafka listener and reports something unhelpful.

Suggest moving to a port outside both suites' claimed ranges (9096 or 9190), and adding an `e2e-llm` column to the README port table so the next person picking a port sees the whole picture.

solace-aross (changes requested):
Requesting changes on two points, both in the fixture plumbing rather than the coverage itself.

The coverage is the strong part of this PR and the scope growth is justified. Disproving the license-gate conclusion from the earlier iteration, and pinning the real cause to two enum-restricted scaling defaults, is the kind of finding that saves the next person a day. It also closes the follow-up-ticket ask from my earlier review by doing the work rather than deferring it, which is the better outcome. `README.md`'s F9/F10 section is the right home for it.

Two things to fix:

1. `kafka_broker_available` cannot tell "no Kafka by design" from "Kafka failed to start" (comment on `helpers.sh:318`). This is the focus area you flagged for reviewers. Short answer: skip-with-warning is right for the `e2e-llm` case and wrong for this suite, and one runtime predicate cannot serve both.
2. `KAFKA_PORT` defaults to 9094, which `e2e-llm` already claims for its MCP server (comment on `docker-compose.yml:78`).

I have not restated Helen's `wait_for_kafka` finding at `helpers.sh:341`. I verified it against the source and she is right: line 341 promises "proceeding anyway", line 342 returns 1, and line 1022 sits in the body of the `if`, so `set -euo pipefail` at line 8 aborts the run.

Worth connecting the two, though: point 1 decides which of her two options you want. Once the gate reflects intent, `return 1` plus the `set -e` abort is the correct behavior, so take her "fail loudly, fix the message" option rather than "truly proceed".

solace-aross (dismissed):
Approving. All three items are closed, and I checked the reasoning rather than just the diff.

`kafka_expected` is the right shape. Gating on the compose declaration instead of `docker ps` separates "no Kafka by design" from "Kafka failed to start", which is what the `e2e-llm` reuse needed. I confirmed the abort path behaves as the comment claims: `run-all.sh:39` calls `create_fixtures` unguarded under `set -euo pipefail`, and `wait_for_kafka` runs in the body of the `if`, not its condition, so `return 1` does end the run. The retry around `docker compose config` is a good catch; that failure mode would have rebuilt the same silent misclassification from a different direction.

`KAFKA_PORT` at 9096 is clear of 9094, and the README port table records why.

`wait_for_kafka`'s message now matches its return, which closes Helen's point.

On the coverage itself: the verification predicates go through `verify_monitor_object` with 60/30/15s windows rather than asserting straight after create, so the down and disabled states are not racing fixture setup. `e2e-monitoring` passes in 8m5s on this head.

solace-aross (approved):
Re-approving. My review at `fe31588` was dismissed by the merge, not by a change to the work.

The only commit since is the merge of `main` at `38706cc`, and `test/e2e-monitoring/` is byte-identical across it. I checked the `CHANGELOG.md` conflict resolution rather than assuming: the `[Unreleased]` block keeps all three entries that came from `main` (module-path rename, licence artifacts, RBAC inert-grant WARN) alongside your Kafka entry, and the link footer moved to `SolaceProducts` with them. Nothing was dropped in either direction. `e2e-monitoring` is green at 8m1s on this head.

One thing to ignore before it costs you time. Copilot's suppressed comment on `test-monitoring-tools.sh:1313` says `byFailureReason` is an object so `add` "can fail under jq". It does not. `add` reduces over `.[]`, which iterates an object's values, so `{"No remote-broker in UP state": 1} | add` is `1` and `{} | add` is `null`, which your `// 0` already handles. The assertion is correct as written.

`FOSSA Scan / SCA Scan` and `SCA gate` are red and it is not from this branch. #238 changes no dependencies either and fails the same way against the new `SolaceProducts_solace-broker-mcp` project. Neither context is required on `main-protection`, so it will not block your merge. I am raising it separately.

PR #232, opened 2026-07-29
SOL-152499: Automate DCO sign-off with a prepare-commit-msg hook

bczoma:
## Summary

Adds a `prepare-commit-msg` git hook, shipped in `.githooks/` and installed with `make hooks`, that appends the DCO `Signed-off-by` trailer automatically. SOL-152402 (#224) made sign-off mandatory; this makes it free.

SOL-152499

## Type of Change

- [x] Documentation update

## Motivation and Context

#224's `DCO sign-off` gate **detects** a missing sign-off after the fact: red check → amend or re-merge → force-push. Nothing helps a contributor get it right the first time, and for an external contributor that lands on their first interaction with the project.

The advice circulating internally — `git config --global format.signOff true` — does not work. `format.signOff` only affects `git format-patch`, not `git commit`; git's docs say so explicitly ("lets you enable the `-s`/`--signoff` option of **format-patch** by default") and there is no `commit.signOff` counterpart. Anyone relying on it still trips the gate.

**The hook is convenience, never the control.** It fails silently by design — a fresh clone does not have it, and a hook missing its exec bit is skipped with no warning. #224's CI gate remains the enforcement. The docs say so in as many words.

## Changes Made

- `.githooks/prepare-commit-msg` (new, 755) — appends the trailer via `git interpret-trailers --if-exists addIfDifferent`
- `.githooks/prepare-commit-msg.test.sh` (new, 755) — 40 assertions, following the `dco-check.test.sh` style. **Run by hand**, not by CI — see "No CI job" below
- `.github/CONTRIBUTING.md` — setup and verification under "How to Sign Off Commits"; `make hooks` in the Development Setup quick start
- `README.md` — new numbered Development Setup step, subsequent steps renumbered
- `CLAUDE.md` — two lines under "Before committing", so coding agents sign off unprompted
- `Makefile` — `make hooks` target
- `.github/CODEOWNERS` — explicit `.githooks/` rule

Seven files. No workflow is touched.

## Design decisions worth reviewing

**Install as an untracked copy under `.git/`, read from a trusted ref — never `core.hooksPath .githooks`, never a symlink.** This is a security requirement for a repo about to go public, not a style preference. With `core.hooksPath` pointed at tracked content, a fork PR can edit the hook and get code execution on a maintainer's machine during an ordinary git operation while that PR's tree is checked out. Reproduced during review: a plain `git commit` in a review worktree ran attacker-supplied code — no build, no tests, nothing a reviewer would classify as "running the PR's code". `git merge` fires `prepare-commit-msg` too, and linked worktrees share repo config, so this project's own worktree-based review flow is in scope. A symlink from the hooks dir into `.githooks/` reintroduces the whole thing, because it resolves to the checked-out file. Copying into the untracked hooks directory closes that. Review found a second step of the same hole: `make hooks` originally read the payload from the *checked-out tree*, so a hostile branch keeping the marker line satisfied the clobber guard and its payload landed under `.git/hooks/`, where it **outlives the branch**. `make hooks` now reads the blob from a trusted ref (`HOOKS_REF`, default `origin/main`), with fail-closed checks for an unresolvable ref, a missing file, and a blob without the marker. The full argument lives in the hook header; the other files point at it.

**`HOOKS_REF=HEAD make hooks` is required while testing this PR.** `origin/main` does not carry the hook until this merges, so the default refuses with an explicit message. It installs the *committed* HEAD, not a dirty worktree.

**`trailer.*` config is pinned on the `interpret-trailers` call.** The weld guard compares against the trailer the hook asked for, so one line of developer config (`trailer.separators='=:'`) silently defeated it and brought the weld back. Pinning `separators`, `where` and `ifmissing` keeps the emitted trailer byte-identical whatever the user has configured.

**The hooks directory is resolved with `git rev-parse --git-path hooks`, not hardcoded `.git/hooks`.** In a linked worktree `.git` is a file, so the naive form fails outright. It also honours a custom `core.hooksPath` outside the tree.

**`--if-exists addIfDifferent`, not `doNothing`.** It is a strict superset: no duplicate when `-s` was passed or on repeated amends, but it still adds your own line when a message carries a third party's sign-off.

**Installer refuses rather than clobbers.** A marker line identifies our own copy; a foreign `prepare-commit-msg` (ticket prefixers and commit-lint hooks are common) is left intact and the target exits non-zero. It also refuses to install when `core.hooksPath` resolves inside the working tree.

## No CI job — deliberate

An earlier revision of this PR added a `hook_selftest` job to `ci-pr.yaml`. **That has been reverted**; `.github/workflows/ci-pr.yaml` is now byte-identical to `main` and this PR touches no workflow at all.

The reasoning: a recurring job that is never expected to fail accrues cost on every PR for very little information, and the hook is a small shell script that changes rarely. The `DCO sign-off` gate from #224 already re-verifies every commit regardless of what ran locally, so nothing about enforcement depends on this suite.

The cost of that choice is stated honestly in the hook itself. See "The one regression CI cannot see" below.

## Testing

**Environment:** git 2.50.1 (host), git 2.39.5 (`debian:bookworm-slim` container, the version in the field report), and bash 3.2.57 (`bash:3.2`, what a stock Mac resolves). No Go touched.

**Coverage:** 40 assertions — plain editor mode, `-m`, `-s` (no duplicate), repeated `--amend --no-edit`, `git commit -v`, `commit.verbose=true`, `--cleanup=scissors`, `core.commentChar=';'` and `=auto`, `commit.template`/`-t`, CRLF buffers, clean and conflicted merges, `merge --signoff`, identity supplied via environment (bots/CI/agents), existing `Co-Authored-By` preserved, a third party's sign-off, hostile content in `user.name` round-tripping verbatim with nothing executed, hostile `trailer.*` repo config, an unresolvable committer identity, and the hook never aborting or corrupting a commit.

The suite is mutation-checked rather than assumed to have teeth. Against the first revision of the hook it reports 22 failures on git 2.50.1 and 24 on git 2.39.5; against what ships here, 40/40 on both — and 40/40 again on real bash 3.2.57 and with `GIT_COMMITTER_*` exported.

Run it after any edit to `.githooks/`:

```bash
.githooks/prepare-commit-msg.test.sh
```

## Review feedback addressed

**@solace-awaheed — trailer welded onto merge subjects.** Correct, and the root cause is a **git version** difference rather than macOS. On git ≤ 2.39, `git merge` hands the hook a message buffer with no trailing newline, and that version of `interpret-trailers` does not insert the blank separator, so the trailer welds under the subject:

```
subject: [Merge branch feat Signed-off-by: Dev Ex <dev@example.com>]
trailer: []
```

Reproduced in a container on git 2.39.5, where the suite reported 32 passed / 2 failed — exactly the two clean-merge cases. This affects anyone on Debian bookworm, Ubuntu 22.04, or RHEL 9, not only macOS. Their suggested fix is applied verbatim and measured 34/34 on both git 2.39.5 and 2.50.1 (the suite was 34 cases at that point; it is 40 now, and still green on both). I also tried generalising the existing structural guard to catch the weld wherever it lands and measured that at 16/34 — strictly worse, so it was dropped.

**@solace-aross — seven findings, all fixed in `7a01535`.** Each was reproduced before being changed, and each reply on the thread carries the evidence. The two that mattered most: `make hooks` reading the payload from the checked-out tree was a *persistent* code-execution path (the installed copy outlives the hostile branch), and `trailer.separators` in a developer's own config silently defeated the weld guard. Also fixed: the installer's `case` failing open for hooks directories outside the git dir (it would write this repo's hook into a shared global hooks dir), a bash 3.2 array idiom that stopped the suite running at all on macOS, the suite neutralising config but not `GIT_COMMITTER_*` in the environment (20/14 against a *correct* hook), a CR in the ident welding the subject, and an assertion with no reachable failure.

One of her suggestions was adjusted rather than applied — the replacement for that unreachable assertion, because git resolves the committer ident *before* running `prepare-commit-msg`, so no real commit reaches the hook in that state. Details in the thread.

**@Copilot** — all four addressed: the hooks-directory wording in `README.md` and `CONTRIBUTING.md` now says "the repository's hooks directory" rather than `.git/hooks/`, since the installer resolves it; the refusal hint offers both the local and `--global` unset plus `git config --show-origin --get core.hooksPath`; and `interpret-trailers` stderr is suppressed with failure treated as a no-op, with the trade-off written into a comment.

## The one regression CI cannot see

Worth knowing before approving, because it is the cost of having no CI job here:

| hook | git 2.50.1 | git 2.39.5 |
|---|---|---|
| without the newline normalisation | 40 / 40 | 38 / 40 |
| with it | 40 / 40 | 40 / 40 |

Deleting those three lines passes the suite on any current machine and breaks every contributor on an LTS distro. Neither CI nor a local run would catch it. The hook carries a comment saying exactly that, naming the git version and `git merge` as the source of the un-terminated buffer, so the next person to tidy that block has been warned in the only place they will be looking.

## Breaking Changes

None. `make hooks` is opt-in and per clone; nothing changes for anyone who does not run it.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code
- [x] No security vulnerabilities introduced — see "Design decisions"; the three attacks (`core.hooksPath`, symlink, worktree propagation) were each reproduced and the shipped install is immune

### Testing
- [x] Suite passes on git 2.50.1, git 2.39.5 and bash 3.2.57 (40/40 each); `dco-check.test.sh` still 42/42
- [x] Error paths tested

### Documentation
- [x] README.md updated
- [ ] CHANGELOG.md updated — deliberately not. `.github/scripts/production-surface.sh` scopes production surface to `internal/config/`, `internal/tools/`, and `internal/composite/definitions/tools.yaml`; no path here matches

### Git & CI
- [x] All commits are signed off (by the hook itself)
- [x] No merge conflicts

## Additional Notes

No Go files, so `/check-logs` has nothing to scan and `make check` is unaffected.

Follow-up worth considering separately: there is no shellcheck job in CI, so nothing lints shell in this repo.

## Related Issues/PRs

- Depends on #224 (merged) — the CONTRIBUTING section edited here is the one #224 rewrote, and the `DCO sign-off` check name is confirmed verbatim from a live run

---

**For Reviewers:**

**Focus Areas**: The install mechanism (copy vs `core.hooksPath`) is the load-bearing security decision — if you disagree with it, say so before the ergonomics. After that, the message-buffer handling in the hook: three review rounds each found a defect there, and it is the part that touches every commit a contributor makes.

**Questions**: Should `.githooks/` be limited to this one hook by policy? Without that, it becomes an attractive home for pre-commit lint hooks, and "the repo can run code on your machine at commit time" is a larger conversation than DCO sign-off.


.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,93 @@
...
+
+msg_file=$1
+
+ident=$(git var GIT_COMMITTER_IDENT 2>/dev/null) || exit 0
+# "Name <email> 1750000000 +0200" -> "Name <email>"
+signoff=$(printf '%s\n' "$ident" | sed -e 's/> [0-9]* [+-][0-9]*$/>/')
+
+trailer="Signed-off-by: $signoff"
+
+git interpret-trailers --if-exists addIfDifferent \
```
solace-awaheed: An issue related to MacOS default git version: Merge commits without a trailing newline can end up with a welded Signed-off-by on macOS, so Git treats it as part of the subject instead of a trailer. The current guard (81–88) doesn't catch this because the first line isn't blank. Normalizing the message to end with a newline before interpret-trailers avoids the issue and also fixes the failing merge self-tests.

Claude suggested the following fix: 
```
  trailer="Signed-off-by: $signoff"

  # interpret-trailers only inserts the blank line before an appended trailer when
  # the buffer already ends in a newline. A clean merge is handed to us WITHOUT one
  # (the buffer is a bare "Merge branch ..." line), so normalise first, else the
  # trailer welds directly under the subject.
  if [ -n "$(tail -c1 "$msg_file")" ]; then
        printf '\n' >>"$msg_file"
  fi

  git interpret-trailers --if-exists addIfDifferent \
        --trailer "$trailer" \
        --in-place "$msg_file"
```

solace-awaheed (changes requested):

amitmorade:
Did not find anything significant other than what the existing comments say

solace-wmaryam:
LGTM - only comment that seems worth fixing is @solace-awaheed's

.githooks/prepare-commit-msg.test.sh
```diff
@@ -0,0 +1,284 @@
...
+  shift || true
+  args=("$@")
+
+  dir=$(new_repo "$(basename "$dir")")
+  local i
+  for ((i = 0; i < ${#cfg[@]}; i += 2)); do
+    git -C "$dir" config "${cfg[i]}" "${cfg[i + 1]}"
+  done
+  stage "$dir" one
+  GIT_EDITOR="$WORK/ed-subject" git -C "$dir" commit -q "${args[@]}"
```
solace-aross: The suite cannot run on macOS. `${args[@]}` on an empty array under `set -u` is an unbound-variable error on bash 3.2.57, which is what `#!/usr/bin/env bash` resolves to on a stock Mac:

```
line 99: args[@]: unbound variable
```

It dies at `editor_case "plain"`, so no case runs. Linux CI is bash 5, which is why `Commit hook self-test` is green and this is invisible there.

You already use the correct idiom for the same problem on the `extra` array further down, so this is just an inconsistency:

```sh
GIT_EDITOR="$WORK/ed-subject" git -C "$dir" commit -q "${args[@]+"${args[@]}"}"
```

With that one token changed I get 34 passed, 0 failed on both git 2.54.0 and Apple Git 2.50.1, merge cases included. Worth fixing first, because I believe it is what produced the failing merge self-tests reported on this PR.

Makefile
```diff
@@ -28,6 +28,44 @@ COMPOSE_E2E_OAUTH := docker compose -f $(E2E_OAUTH_DIR)/docker-compose.yml
...
+	esac; \
+	dest="$$hooks_dir/prepare-commit-msg"; \
+	if [ -e "$$dest" ] && ! grep -q "$(HOOK_MARKER)" "$$dest"; then \
+	  echo "refusing to overwrite $$dest — it is not this repo's hook." >&2; \
+	  echo "Move it aside and re-run make hooks. To keep both, install ours under another" >&2; \
+	  echo "name in the hooks directory and call THAT from yours — never .githooks/, which" >&2; \
+	  echo "is the checked-out tree (see the header of .githooks/prepare-commit-msg)." >&2; \
+	  exit 1; \
+	fi; \
+	install -m 755 .githooks/prepare-commit-msg "$$dest"; \
```
solace-aross: This reads the source from the checked-out tree while the guard on line 59 only inspects `$$dest`, so `make hooks` is a path to exactly the code execution the hook header argues against.

A hostile fork PR keeps the marker on line 2 of the tracked file, `grep -q "$(HOOK_MARKER)" "$$dest"` succeeds against the reviewer's own previously installed copy, and the condition is false. The copy then happens unconditionally. Reproduced end to end: install legitimately, branch, insert `echo PWNED` into `.githooks/prepare-commit-msg` keeping line 2, run `make hooks` ("installed"), check out `main`, commit anything, payload runs. The copy under `.git/` outlives the branch, so this is more persistent than the `core.hooksPath` attack you ruled out.

Suggestion: install from a trusted ref by default rather than from the worktree, with an escape hatch for authors testing their own hook change.

```make
	ref="$${HOOKS_REF:-origin/main}"; \
	git cat-file blob "$$ref:.githooks/prepare-commit-msg" >"$$dest.tmp" && \
	  chmod 755 "$$dest.tmp" && mv "$$dest.tmp" "$$dest"; \
```

`HOOKS_REF=HEAD make hooks` then covers your own workflow on this PR. If you would rather keep the copy-from-tree behaviour, the header paragraph needs to name this residual path, and `CONTRIBUTING.md:269` should stop advising a re-run without qualifying which branch you are on.

Makefile
```diff
@@ -28,6 +28,44 @@ COMPOSE_E2E_OAUTH := docker compose -f $(E2E_OAUTH_DIR)/docker-compose.yml
...
+
+.PHONY: hooks
+hooks: ## Install the repo's git hooks (DCO sign-off); re-run after pulling hook changes
+	@set -e; \
+	hooks_dir="$$(git rev-parse --git-path hooks)"; \
+	mkdir -p "$$hooks_dir"; \
+	hooks_abs="$$(cd "$$hooks_dir" && pwd -P)"; \
+	git_dir="$$(cd "$$(git rev-parse --git-common-dir)" && pwd -P)"; \
+	top="$$(cd "$$(git rev-parse --show-toplevel)" && pwd -P)"; \
+	case "$$hooks_abs/" in \
```
solace-aross, line 58: This `case` has no default arm, so the refusal fails open in two configurations I could reproduce, both of which print `installed` and exit 0.

1. Linked worktree with an absolute `core.hooksPath` into the main checkout's tracked `.githooks`. `hooks_abs` is `main/.githooks`, so arm 1 (`$$git_dir/*`, i.e. `main/.git/*`) misses; `top` is the linked worktree, so arm 2 misses. It installs into a tracked directory of a checkout, which is precisely the configuration this block exists to prevent.
2. Any `core.hooksPath` outside the repo, which is the common global-dotfiles pattern. `make hooks` writes this repo's DCO hook into a shared hooks directory, so it signs off commits in every repo on the machine. The only signal is the printed path.

Fail closed instead:

```make
	case "$$hooks_abs/" in \
	  "$$git_dir"/*) ;; \
	  *) echo "refusing to install: hooks dir ($$hooks_abs) is outside this repo's git dir." >&2; \
	     echo "See the header of .githooks/prepare-commit-msg." >&2; \
	     exit 1;; \
	esac; \
```

The two-arm ordering itself is fine, for the record: the trailing slash in `"$$git_dir"/*` stops `$$top/.githooks/` matching `$$top/.git/*`, and `pwd -P` resolves a symlinked tree back to a path arm 2 catches. It is only the missing default.

`CONTRIBUTING.md:271` also overstates what this does today. "Refuses to install into a hooks directory inside the working tree" is one of the unsafe cases, not all of them.

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,93 @@
...
+# would already have.
+#
+# The test is deliberately structural — is line 1 blank with the trailer directly
+# under it? — and runs on the finished buffer, because everything about the buffer
+# ELSE is a moving target. "Is the message empty" cannot be answered by stripping
+# what we think git puts in there: comment lines (whose prefix is core.commentChar,
+# possibly `auto`), the un-prefixed diff below the `-v` scissors line, and whatever
+# a future git or --cleanup mode adds. This check is immune to all of them: it only
+# looks at the two lines whose contents would break.
+if [ "$(sed -n 1p "$msg_file" | tr -d '\r')" = "" ] &&
```
solace-aross, line 136: This compares line 2 against the trailer the hook *asked for*, not the one `interpret-trailers` actually *wrote*, so any git config that changes the emitted form silently defeats the check and the weld comes back.

`trailer.separators` in a developer's own config is enough. Editor mode, with `trailer.separators='=:'` set in the repo:

```
subject : [typed subject Signed-off-by= Dev Example <dev@example.com>]
trailers: []
```

Control with the same hook and no `trailer.separators`: subject `typed subject`, trailer parses. That is the exact regression this block exists to prevent, and per your own note in the test header it is invisible to the CI DCO gate, which greps line by line.

The suite structurally cannot catch this class, because `test.sh:26` nulls global config so no developer-config interference is ever in scope.

Cheapest fix is to stop letting config reach it, no logic change:

```sh
git -c trailer.separators=: -c trailer.where=end -c trailer.ifmissing=add \
	interpret-trailers --if-exists addIfDifferent \
	--trailer "$trailer" \
	--in-place "$msg_file"
```

Then the comment above about being immune to comment chars, scissors and cleanup modes is actually true, and a case setting a hostile `trailer.*` in repo config would pin it. (`trailer.ifexists=doNothing` in config is already correctly overridden by your CLI flag, and `trailer.where=start` is only cosmetic. It is the separator that breaks it.)

.githooks/prepare-commit-msg.test.sh
```diff
@@ -0,0 +1,284 @@
...
+HOOK="${SCRIPT_DIR}/prepare-commit-msg"
+
+WORK=$(mktemp -d)
+trap 'rm -rf "$WORK"' EXIT
+
+pass_count=0
+fail_count=0
+
+# Deterministic identities. Nothing here reads the developer's git config.
+export GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null
```
solace-aross, line 33: This neutralises config but not the environment, and env identity outranks config. `new_repo` only writes `user.name`/`user.email`, so anything that exports `GIT_AUTHOR_*`/`GIT_COMMITTER_*` makes the suite fail against a correct hook:

```
$ GIT_COMMITTER_NAME="CI Bot" GIT_COMMITTER_EMAIL="ci@example.org" \
  GIT_AUTHOR_NAME="CI Bot" GIT_AUTHOR_EMAIL="ci@example.org" ./prepare-commit-msg.test.sh
prepare-commit-msg self-test: 20 passed, 14 failed
```

That is a devcontainer, a corporate CI image, or agent tooling, which is the population your own comment at the top of the hook says the `git var` decision exists to serve. Fourteen red assertions there read as a hook regression.

```sh
unset GIT_AUTHOR_NAME GIT_AUTHOR_EMAIL GIT_COMMITTER_NAME GIT_COMMITTER_EMAIL
export GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null
```

Test 5 already sets those vars locally for the case that needs them, so unsetting here costs nothing. (`EMAIL=` alone is harmless, that one stays 34/34.)

.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,93 @@
...
+#    build, no tests — and because linked worktrees share repo config, the
+#    project's worktree-based review flow is in scope too. The untracked copy
+#    under `.git/` cannot be altered by checking out a hostile branch, and still
+#    fires in linked worktrees. This paragraph is the canonical statement of that
+#    reasoning; the Makefile, CONTRIBUTING.md, and README.md point here rather
+#    than restating it.
+
+msg_file=$1
+
+ident=$(git var GIT_COMMITTER_IDENT 2>/dev/null) || exit 0
```
solace-aross, line 70: Worth stripping CR here rather than at each comparison site. Git's ident sanitiser drops `\n`, `<` and `>` from both name and email, but not `\r`, so a CR survives into `$trailer`. The `tr -d '\r'` on lines 81 and 82 normalises the file's bytes but not `$trailer`, the compare never matches, the separator is never inserted, and the subject welds:

```
$ git config user.name "$(printf 'Ali\rce')"
subject : [real subject Signed-off-by: Alice <dev@example.com>]
trailers: []
```

Control with `user.name=Alice`: subject `real subject`, trailer parses.

Low severity, since it takes a deliberate `git config user.name $'a\rb'` or env var. A CRLF `.gitconfig` does not do it, git's config parser strips the CR. Raising it because the comment on line 79 claims immunity, and this is the one shape that gets through.

Keep it off the same line, so `|| exit 0` still keys on `git var` rather than on `tr`:

```sh
ident=$(git var GIT_COMMITTER_IDENT 2>/dev/null) || exit 0
ident=$(printf '%s' "$ident" | tr -d '\r')
```

.githooks/prepare-commit-msg.test.sh
```diff
@@ -0,0 +1,284 @@
...
+# --- 9. the hook never blocks a commit ---------------------------------------
+# A non-zero prepare-commit-msg aborts the commit. Even with no resolvable
+# identity at all, the hook must exit 0 — git's own identity error is what should
+# surface, not a hook failure.
+r=$(new_repo no_identity --no-identity)
+rc=0
+env -u GIT_AUTHOR_NAME -u GIT_AUTHOR_EMAIL -u GIT_COMMITTER_NAME -u GIT_COMMITTER_EMAIL \
+  EMAIL= GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null \
+  "$HOOK" /dev/null || rc=$?
+expect_equal "hook exits 0 with no resolvable identity" "0" "$rc"
```
solace-aross: This assertion has no reachable failure. Two mutations, both still 34/34:

- delete the final `exit 0` at `hook:93`
- delete the `|| exit 0` on `hook:58`

The hook cannot return non-zero by construction, so nothing here can go red. It also does not test the state its name claims: in this exact environment on macOS, `git var GIT_COMMITTER_IDENT` succeeds, because git happily derives `user@host.local` from the hostname. The failure branch on line 58 is never entered locally, and whether it is entered in CI depends on whether the runner's hostname has a dot.

To test the invariant that actually matters, drive a real commit in a repo where identity is genuinely unresolvable and assert git's own error surfaces and no commit was created:

```sh
git -C "$dir" config user.useConfigOnly true   # makes git var exit 128
```

That pins "the hook never blocks or corrupts a commit" instead of the script's exit code.

solace-aross (changes requested):
Requesting changes on five items, all reproduced locally on macOS. The design is right and the hook logic is sound. I ran your suite to 34/34 on both git 2.54.0 and Apple Git 2.50.1, and mutation-checked it rather than trusting the count: neutralising the anti-weld block takes it to 18 red, so that fix is genuinely pinned. What follows is the install path and the suite's blind spots, not the core idea.

The load-bearing one first, because it changes what the header can claim.

**`make hooks` is itself the execution path the header argues against.** `Makefile:66` copies the source from the checked-out tree, and the only guard, `Makefile:59`, inspects the destination. A fork PR keeps the line-2 marker, so for any reviewer who already installed legitimately the copy is unconditional. Reproduced: install, branch, add `echo PWNED` to the tracked hook keeping the marker, `make hooks`, check out `main`, and an ordinary commit prints the payload. It survives the branch being deleted, which makes it worse than the `core.hooksPath` variant you rejected, since that one dies with the checkout. What makes it live rather than academic is `CONTRIBUTING.md:269` telling people to re-run `make hooks` after pulling a change to `.githooks/`, which is exactly when reviewing a hook PR. Your paragraph at `prepare-commit-msg:50` is true and incomplete: checkout cannot alter the installed copy, but the documented install step can.

Four more inline: the suite cannot run on macOS at all (`test.sh:99`), the `core.hooksPath` refusal fails open (`Makefile:50`), the anti-weld check can be defeated by the developer's own config (`hook:81`), and the suite is not hermetic against `GIT_COMMITTER_*` (`test.sh:26`). `hook:81` is the one I would most want your read on, because it puts the weld back.

Two non-blocking notes.

`test.sh:280` has no teeth. Deleting the final `exit 0` (`hook:93`) leaves the suite 34/34, and so does deleting the `|| exit 0` guard (`hook:58`). The hook cannot return non-zero by construction, so there is no reachable failure for that assertion to catch.

`make hooks` has no automated coverage at all. Nothing in any shell script or workflow references it: 34 assertions on the 93-line hook, none on the 26-line installer that carries the security decision. I exercised every branch by hand today and they all behave, including the linked-worktree case resolving to the common `.git/hooks`. The risk is forward, not now. A later simplification of that `case`, or a change in `git rev-parse --git-path` behaviour, stops the refusal firing and nothing notices.

**On Asad's merge finding, I could not reproduce it, and I do not think the premise holds.** Two checks, on both git 2.54.0 and Apple Git 2.50.1:

- Real `MERGE_MSG` from a clean `--no-ff` merge ends with a newline. `od -c` gives `Merge branch 'side' \n` and nothing else.
- Handed a buffer with no trailing newline, `interpret-trailers` inserts the blank separator itself: `Merge branch 'side'` in, `Merge branch 'side'\n\nSigned-off-by: ...\n` out. No weld.

All three merge assertions pass here against the hook as shipped. My guess is you hit the `test.sh:99` item, which kills the suite at the first case on macOS system bash and would read as the merge tests failing. @solace-awaheed what git and bash version did you see it on? If there is a combination where `MERGE_MSG` genuinely arrives bare, I would rather pin it in the suite than add an unconditional write to the one path that touches every commit a contributor makes.

To answer your CODEOWNERS question: yes, hold `.githooks/` to this one hook by policy, and put that in the header. "The repo can run code on your machine at commit time" is a much larger conversation than DCO sign-off, and the `Makefile:66` item is a preview of why.

Not restating Copilot's four; all fair. Its `Makefile:55` point about `--global` gets more important once the refusal below actually fires on a global `core.hooksPath`.

.githooks/prepare-commit-msg
bczoma: Confirmed and applied verbatim in `fd33823`. Thanks — this was the one finding that would have shipped silently.

Root cause is narrower than macOS: it is **git version**. On git <= 2.39, `git merge` hands the hook a message buffer with no trailing newline, and that version of `interpret-trailers` does not insert the blank separator. Reproduced in a `debian:bookworm-slim` container on git 2.39.5:

```
subject: [Merge branch feat Signed-off-by: Dev Ex <dev@example.com>]
trailer: []
```

The suite there reported **32 passed, 2 failed** — exactly your two clean-merge cases. So this hits anyone on Debian bookworm, Ubuntu 22.04, or RHEL 9 as well, not only Mac.

Your fix measures **34/34 on both git 2.39.5 and 2.50.1** (now 40/40 after a later round). I tried generalising the existing structural guard to catch the weld wherever it lands instead, and measured that at **16/34** — strictly worse, so it is gone and your version stands.

One note for the record, since @solace-aross suggested her `test.sh:99` bash 3.2 finding may explain what you saw: I do not think it is the whole story. That error kills the script outright, so no case runs at all, whereas you reported merge cases *failing*. And with the newline block removed from the current hook the suite is 40/40 on git 2.50.1 but **38/40 on git 2.39.5**, independently of any array or config change. Both bugs are real and independent; both fixes are in.

.githooks/prepare-commit-msg
bczoma: Confirmed and fixed in `7a01535` with your patch verbatim. Reproduced first:

```
WITH trailer.separators='=:'  subject: [typed subject Signed-off-by= Dev Example <dev@example.com>]
                              trailer: []
control (no config)           subject: [typed subject]
```

After: subject `[typed subject]`, body line 3 `Signed-off-by: Dev Example <dev@example.com>`.

You are right that this is the exact regression the block exists to prevent, and that the suite structurally could not see it. Two cases now pin it — `trailer.separators='=:'` alone, and a hostile combination of `trailer.separators` / `where` / `ifmissing` / `ifexists`. Both fail if the `-c` pins are dropped.

The comment above the check no longer claims immunity it did not have: it now says it is immune to `trailer.*` **because** the call pins it, and the pin block explains that the check compares against the trailer the hook *asked for* rather than what was *written*.

One correction to the repro method, which does not change the finding: with a hostile `trailer.separators` in config, `git log --format='%(trailers)'` re-renders the separator from the **reader's** config, so a correct hook still reports `Signed-off-by=`. Part of what that command shows is the reader, not the hook. The suite now reads with `-c trailer.separators=:`. The underlying weld was real either way — the subject line proves it independently of `%(trailers)`.

Makefile
bczoma: Confirmed, and this is the most serious of the seven. Reproduced end to end exactly as you describe:

```
legit install done
# solace-broker-mcp:prepare-commit-msg   <- marker line intact
installed .git/hooks/prepare-commit-msg
payload fired on main? PWNED
```

You are also right that it is worse than the `core.hooksPath` path the header rules out, for the reason you give: the copy under `.git/` outlives the branch. An earlier review round waved this off as "already arbitrary code execution by consent, since you ran the fork's Makefile" — that misses that the header implies the install is safe and CONTRIBUTING actively tells reviewers to re-run it.

Took the trusted-ref option in `7a01535`. `HOOKS_REF` defaults to `origin/main`; `git cat-file blob` reads the payload from the ref, never the tree. Beyond your sketch: a `trap … EXIT HUP INT TERM` on the temp file so a partial install cannot survive, and three fail-closed pre-checks — ref unresolvable, file absent at the ref, and **blob missing the marker**. That last one is the answer to the question your comment implies: the marker check now follows the trusted ref rather than only `$dest`.

Re-tested: `payload fired? NO`, and the commit is still signed normally.

`HOOKS_REF=HEAD make hooks` is documented in the Makefile help, the hook header, and CONTRIBUTING, with the caveat that it installs the **committed** HEAD, not a dirty worktree.

One consequence worth flagging before merge: `origin/main` does not yet contain `.githooks/prepare-commit-msg`, so on this branch `make hooks` refuses with

```
refusing to install: .githooks/prepare-commit-msg does not exist at origin/main.
If you are on a branch that adds or changes it: HOOKS_REF=HEAD make hooks
```

That resolves itself on merge, but anyone testing this PR needs the escape hatch. It fails closed, which seemed the right side to err on.

Makefile
bczoma: Confirmed both configurations and fixed with your default arm in `7a01535`. The out-of-tree one is the more alarming of the two — reproduced:

```
installed /tmp/sharedhooks/prepare-commit-msg
```

Writing this repo's DCO hook into a shared hooks directory so it signs off commits in every repo on the machine is clearly wrong, and an earlier review round had verified that same behaviour as *correct*. Now refuses, rc≠0, shared dir left empty.

Noted for the record that the ordering was fine and only the default was missing — I kept the three-arm structure. Both refusal arms also print `git config --show-origin --get core.hooksPath`, since the local-vs-global distinction is what someone actually needs at that moment.

`CONTRIBUTING.md` no longer says the guard "refuses to install into a hooks directory inside the working tree" — you were right that this describes one unsafe case rather than all of them.

.githooks/prepare-commit-msg.test.sh
bczoma: Confirmed as an inconsistency and fixed with your one-token change in `7a01535`:

```sh
GIT_EDITOR="$WORK/ed-subject" git -C "$dir" commit -q "${args[@]+"${args[@]}"}"
```

Grepped the rest of the suite — line 99 was the only bare `"${arr[@]}"` on a possibly-empty array. Verified on a real `bash:3.2` image: **40/40**, where before the script died at `editor_case "plain"`.

On your hypothesis that this is what produced the failing merge self-tests reported here — I do not think it is the whole story, and it matters because it decides whether @solace-awaheed's newline fix stays. That error kills the script outright ("no case runs"), whereas the report was of merge cases *failing*. And with the newline block removed from the current hook, the suite is 40/40 on git 2.50.1 but **38/40 on git 2.39.5** — the two clean-merge cases — with no arrays or config involved. Your 34/34 was on Apple Git 2.50.1, where the weld does not occur; @solace-awaheed is on 2.39.3, where it does. Two independent bugs, both now fixed.

.githooks/prepare-commit-msg.test.sh
bczoma: Confirmed with your exact numbers and fixed as you wrote it in `7a01535`:

```
before: prepare-commit-msg self-test: 20 passed, 14 failed
after:  prepare-commit-msg self-test: 40 passed, 0 failed
```

`unset GIT_AUTHOR_NAME GIT_AUTHOR_EMAIL GIT_COMMITTER_NAME GIT_COMMITTER_EMAIL` plus `GIT_CONFIG_SYSTEM=/dev/null`. Test 5 still passes — it sets those vars locally for the one case that needs them, as you noted.

The point that this is the same population the `git var` decision exists to serve is well made: 14 red assertions in a devcontainer would read as a hook regression and send someone hunting in the wrong place.

.githooks/prepare-commit-msg
bczoma: Fixed in `7a01535` at the placement you specified, and for the reason you gave:

```sh
ident=$(git var GIT_COMMITTER_IDENT 2>/dev/null) || exit 0
ident=$(printf '%s' | tr -d '\r')
```

Kept on its own line so `|| exit 0` still keys on `git var` failing rather than on `tr`. Agreed on the severity — it takes a deliberate `$'a\rb'` — but you are right that the comment claimed immunity, and this was the shape that got through. The comment now names CR explicitly and says the sanitiser drops `\n`, `<` and `>` but not `\r`.

.githooks/prepare-commit-msg.test.sh
bczoma: You are right that the assertion had no reachable failure, and `user.useConfigOnly true` is in `7a01535`. I could not use the rest of the suggestion as written, and want to flag that rather than have it look half-applied.

Driving a real `git commit` in that state does not reach the hook. Verified on both git 2.39.5 and 2.50.1 with an instrumented hook printing to stderr: it never runs, because git resolves the committer ident **before** `prepare-commit-msg` and aborts with "Committer identity unknown". So "assert git's error surfaces and no commit was created" would have been unreachable in the same way as the assertion it replaced — it would pass whether or not the hook exists.

What is in there instead: your `user.useConfigOnly true` to make `git var` genuinely exit 128 (fixing the "does not test the state its name claims" half), then the hook invoked directly on a real message file, asserting exit status **and** that the message is byte-unchanged. That second assertion is what gives the case teeth — `Signed-off-by:` with an empty ident would be worse than no trailer. A third case fails loudly if `git var` ever succeeds there, so the setup cannot silently stop being the state under test.

Reachability proven both ways: `|| exit 1` on the hook reddens the status assertion, deleting the guard reddens the integrity assertion. Happy to change tack if you see a way to reach it through a real commit that I missed.

solace-aross (approved):
Approving. All seven findings from my last pass are fixed. I re-verified each rather than taking the summary on trust.

What I re-ran at `7a01535`:

- Suite: 40/40 on bash 3.2.57 (what a stock Mac resolves) with git 2.54.0, and 40/40 again with `GIT_COMMITTER_*` / `GIT_AUTHOR_*` exported. The unbound-variable crash is gone, so the suite actually runs on a Mac now.
- Mutation check on the `trailer.*` pins: swapping the pinned `git -c ...` call for a bare `git` drops the suite to 37/40. That fix has teeth.
- The persistent code-execution path I reported is closed, tested end to end. Fresh clone, legitimate install from a trusted ref, branch, insert `echo PWNED` into `.githooks/prepare-commit-msg` keeping the marker on line 2, commit, then plain `make hooks` as CONTRIBUTING tells a reviewer to do. The installed hook carries no payload; the installer read `origin/main` and ignored the checked-out tree.
- Fail-closed refusals hold in all three shapes: `core.hooksPath` into the working tree, `core.hooksPath` to a shared directory outside the repo (nothing written there), and an unresolvable `HOOKS_REF`. Each exits non-zero.

Reading `make hooks` at head, re-checking the marker on the fetched blob before the `mv` is a better guard than what I suggested. Good call.

CI is green at head including all five E2E suites. `mergeStateStatus` is BEHIND, so it needs an update before merge.

One correction to my own earlier finding, non-blocking. The CR strip at `.githooks/prepare-commit-msg:77` has no test coverage: deleting that line leaves the suite at 40/40, and I could not reproduce the weld on git 2.54.0 even against `d086b5e`. My original repro was weaker than I stated it. Keep the line, it is cheap and correct, but it is defensive rather than proven. Worth a case if you touch that block again.

On your question about scoping `.githooks/` to this one hook: yes, worth a stated policy. "The repo can run code on your machine at commit time" is a surface that grows one convenient pre-commit linter at a time, and the CODEOWNERS rule only helps if there is a bar for what may live there.

solace-awaheed (approved):
LGTM

bczoma:
Thanks both — and thanks for re-running everything rather than taking the summary on trust.

Three things from your approval, @solace-aross, and a plan for the split.

**1. The `.githooks/` policy — agreed, and split out as [SOL-152578](https://sol-jira.atlassian.net/browse/SOL-152578).**

Not folded in here for a specific reason: the `main-protection` ruleset sets `dismiss_stale_reviews_on_push: true`, so any further push would dismiss both of your approvals for a doc-only change. Your point that the CODEOWNERS rule only helps if there is a bar is the part I have carried into the ticket — the policy is written as criteria, not a count, and the load-bearing bullet is the one that excludes hooks which read or execute the working tree's code (linters, formatters, tests, codegen). That is the category "one convenient pre-commit linter at a time" falls into, so it needs excluding by name rather than by judgement. Canonical home is `.githooks/README.md`, pointed at from CONTRIBUTING and a CODEOWNERS comment, and it carries the reviewer instruction to treat a diff there like a workflow change.

**2. Your correction on the CR strip is right, and the comment is worse than you said.**

`.githooks/prepare-commit-msg` still asserts the weld as fact — "therefore survives into `$trailer` … and the subject welds". I accepted that finding without independently reproducing it, so the comment claims something neither of us has demonstrated. That is the same defect class this PR spent three rounds fixing, and leaving one in that file is not acceptable. Agreed the line stays — cheap and correct — but the comment has to say it is defensive and uncovered. In SOL-152578 as item 2, along with @Copilot's remaining `.git/hooks/` wording nit (now only wrong for linked worktrees and a custom in-`.git` hooks path, since the installer refuses anything outside the git dir).

**3. `BEHIND` collides with the ruleset, so the merge order matters.**

`strict_required_status_checks_policy: true` requires the branch to be current, and `dismiss_stale_reviews_on_push: true` dismisses approvals on push. So updating from `main` **also** drops both approvals. Sequence that works: update from `main` last, get one re-approval, merge before `main` moves again. Flagging it so it does not look like a broken check.

One thing outside this PR, while it is fresh: `main-protection` currently requires only `lint`, `build` and `FOSSA Scan`. Neither `DCO sign-off` nor `DCO check self-test` is registered, so #224's gate reports but does not block — and the required `FOSSA Scan` context is the one that reports *skipped* on PRs, exactly as your own ADMIN_SETUP note warns. Recorded as out of scope on SOL-152578 since it is an admin action against SOL-152402, not a code change.

[SOL-152578]: https://sol-jira.atlassian.net/browse/SOL-152578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #233, opened 2026-07-29
ci: report to Guardian via solace-cloud-manifest on push to main [DATAGO-145530]

mustafaal-sakkaf:
## Summary

- Adds `report_to_guardian` job to `build-and-test.yml` that runs after a successful FOSSA scan on the default branch
- Writes the dev entry (sha + version) to `solace-cloud-manifest` via `SolaceDev/maas-build-actions/.github/actions/update-dynamodb-release-manifest@master`
- Completes the CRA onboarding for `solace-broker-mcp` (FOSSA already existed; this adds the manifest write)

## Test plan

- [ ] Merge to `main` and confirm `report_to_guardian` job runs in the Actions tab
- [ ] Verify a dev entry for `solace-broker-mcp` appears in `solace-cloud-manifest`


.github/workflows/build-and-test.yml
```diff
@@ -3,7 +3,8 @@ name: Build and Test
 on:
   push:
     branches:
-      - '**'
+      - main
```
solace-aross: This narrowing blocks every PR merged after it.

`lint` and `build` are only ever produced by this file's `push` trigger. `ci-pr.yaml` does not call this workflow (it declares only `fossa_scan`, `dco_selftest`, `changelog`), and `release.yml:20`'s `workflow_call` surfaces the jobs as `test / lint`, not `lint`. The `main-protection` ruleset requires `lint`, `build`, `FOSSA Scan` with `bypass_actors: []` and `current_user_can_bypass: "never"`, and classic branch protection's context list is empty, so the ruleset is the whole gate. Required checks match on name only, which is why a name that is never emitted stays pending rather than being skipped.

After this merges, a PR from any branch other than `main` or `DATAGO-145530-fossa-guardian-onboarding` produces neither check and can never go green. Not even an admin can clear it without editing the ruleset.

`bc0e6d0` does not contain this change. Reverting `ced87b09` fixes it.

If narrowing the push trigger is a goal in its own right, the clean version is `push: branches: [main]` plus a `pull_request:` trigger. Job names stay the same so the ruleset keeps working, the duplicate runs go away, and it closes the fork-PR gap documented at `.github/ADMIN_SETUP.md:254-257`. Worth its own PR.

.github/workflows/build-and-test.yml
```diff
@@ -22,7 +23,6 @@ permissions:
 
 jobs:
   fossa_scan:
-    if: github.event_name == 'push' && github.ref_name == github.event.repository.default_branch
     name: FOSSA Scan
```
solace-aross, line 26: Removing this `if:` breaks the `FOSSA Scan` required context, independently of the trigger change. This one bites even if you keep `branches: '**'`.

`.github/ADMIN_SETUP.md:191-200` documents the mechanism: with the gate in place the job reports **skipped** on pull requests, GitHub counts a skipped check as passing, and a reusable-workflow job otherwise surfaces as `<caller> / <inner>`, never as the caller name alone.

Confirmed against real check runs:

```
PR #230 head:  FOSSA Scan | skipped   +  FOSSA Scan / SCA Scan | success
PR #227 head:  FOSSA Scan | skipped   +  FOSSA Scan / SCA Scan | success
PR #232 head:  FOSSA Scan | skipped   +  FOSSA Scan / SCA Scan | success
this PR head:  (no bare "FOSSA Scan")  +  FOSSA Scan / SCA Scan | success
```

Because the job now runs, the bare context is gone and the requirement can never be satisfied. Restored by the same revert of `ced87b09`.

Separately, the ruleset should really be holding `FOSSA Scan / SCA Scan`, per `ADMIN_SETUP.md:186`. That is an admin action, not yours, and not a reason to change this PR.

.github/workflows/build-and-test.yml
```diff
@@ -32,6 +32,23 @@ jobs:
       # Vault address is injected from an org/repo secret, not committed to the repo.
       VAULT_URL: ${{ secrets.VAULT_URL }}
 
+  report_to_guardian:
+    needs: fossa_scan
```
solace-aross, line 36: Is this intended as step 1 of 2? As written, nothing reaches Guardian.

The reference implementation has three pieces and this has one. `guardian_url` is not passed here and `.github/workflow-config.json` has no guardian section, so `sca-scan-and-guard.yaml:446` resolves it empty and skips the upload. `grep -rni guardian` across this repo's workflows, config and docs returns zero hits, and `gh api repos/.../actions/variables` returns `total_count: 0`. The `guardian_db_sync` job, which reconciles the FOSSA results into Guardian and manages the JIRAs, is not here either. So the manifest gets a SHA pointer for a repo Guardian has no scan data for.

Happy to be told this is deliberate sequencing. It just is not stated anywhere, and DATAGO-145530 reads as the whole onboarding.

Two smaller things while you are in this job:

`needs: fossa_scan` couples the manifest pointer to a `BLOCK`-mode scan. A newly published CVE against an unchanged dependency turns `fossa_scan` red on `main` with no code change, and the dev entry silently stops advancing. The reference impl has no `needs:` on its dev-manifest job for that reason; only `guardian_db_sync` needs the scan.

There is also no way to notice when this fails. It cannot be a required check (it only runs on `main`), and `workflow-config.json` sets no `slack_channel`, so a red Guardian write on `main` pages nobody.

.github/workflows/build-and-test.yml
```diff
@@ -32,6 +32,23 @@ jobs:
...
+      contents: read
+      id-token: write
+    steps:
+      - name: Update solace-cloud-manifest dev entry (sha + version)
+        uses: SolaceDev/maas-build-actions/.github/actions/update-dynamodb-release-manifest@master
+        with:
+          release_tag: dev
+          squad: ai
+          repository: ${{ github.event.repository.name }}
+          version: ${{ github.ref_name }}
```
solace-aross, line 49: Drop this line.

On the only path where this job runs, `github.ref_name` is `main`, and `update_release_manifest.sh` writes the input straight through as `"version":{"S":"${version}"}`. The dev entry would read `version: "main"`.

Deleting is the fix rather than deriving a version. `terraform-provider-solacecloud/.github/workflows/ci.yml` omits `version:` on the equivalent job deliberately, with the reason in a comment: the `dev` entry is sha-tracked, and the version-driven `production` entry is written from `release.yml`. No `git describe` and no full-history checkout needed.

solace-aross (changes requested):
Requesting changes. The `report_to_guardian` job itself is fine and the goal is right. The problem is the trigger surgery bundled in alongside it: it removes two of the three status checks the `main-protection` ruleset requires, which would block every PR merged after this one.

The useful part is that all of it sits in one commit. `bc0e6d0` (the feature) still has `branches: ['**']` and still has the `fossa_scan` `if:` gate. Both changes are in `ced87b09`, "ci: enable fossa+guardian jobs on feature branch for testing". Reverting `ced87b09` fixes findings 1 and 2 and leaves the 17-line feature intact.

Blocking:

1. Narrowing `push.branches` strands `lint` and `build` forever (comment on line 6).
2. Removing the `fossa_scan` `if:` deletes the `FOSSA Scan` required context (line 26).
3. `version: ${{ github.ref_name }}` writes the literal string `main` (line 49).

Needs an answer before approval: whether this is step 1 of 2, because the Guardian upload is not actually configured (line 36).

Also worth knowing: the scaffolding bought no test coverage. `branches: '**'` was already running this workflow on your branch (runs 30466870281, 30467015098), and `report_to_guardian` still reports `skipped` on run 30467442658 because its own `if:` demands the default branch. The new job has not executed anywhere yet. The "Build and Test" run at this head was **cancelled**, not failed, so the five red `e2e-*` checks are `needs: build` cascade artifacts rather than real failures. `build` has never completed on `ced87b09`.

DCO is red: `bc0e6d0` and `ced87b09` need `Signed-off-by`. `git rebase --signoff main && git push --force-with-lease`, or just amend if you are dropping `ced87b09` anyway. `make hooks` on `main` installs a hook that does this for you.

I have skipped the points Copilot already made on the `@master` ref. It is a fair note, but there are no tags or releases on `maas-build-actions` and `@master` is the org-wide norm for that action, so I would not block on it.

solace-aross (changes requested):
Thanks for the revert. That clears both blockers, and I verified it rather than assuming:

- `push.branches` is back to `'**'` (line 6), and this run produced `lint` and `build` again.
- The `fossa_scan` `if:` gate is back (line 25), and the bare `FOSSA Scan` context now reports `skipping` on this PR, which is what the ruleset needs.

Three things still open before I can approve.

**1. The Guardian upload still is not wired up** (line 36 thread). I can now name the missing piece. `terraform-provider-solacecloud/.github/workflows/ci.yml:117` carries a second job alongside the manifest write:

```yaml
guardian_db_sync:
  name: Guardian DB Sync
  needs: fossa_scan
  if: github.event_name == 'push' && github.ref == 'refs/heads/develop' && needs.fossa_scan.result == 'success'
  uses: SolaceDev/maas-build-actions/.github/workflows/guardian-db-sync.yaml@master
```

`report_to_guardian` writes the manifest pointer. `guardian_db_sync` is what reconciles FOSSA results into Guardian and creates/closes the JIRAs. At this head, `git grep -i guardian` over `.github/` returns exactly one hit, this job's own name, and `workflow-config.json` has no guardian section.

If the sync job is deliberately a follow-up PR, say so and I will approve this as step 1 of 2. My concern is only that DATAGO-145530 gets closed as "onboarded" while the reconcile half is absent.

**2. `version: ${{ github.ref_name }}` still writes the literal `main`** (line 49 thread). I confirmed the write-through path this time: `update_release_manifest.sh:65` emits `"version":{"S":"${version}"}` with no defaulting or validation.

One correction to my own earlier advice. I said "deleting is the fix" as if it were the only option. Two peer patterns are both valid:

- `terraform-provider-solacecloud/ci.yml:110-114` omits `version:` entirely (`release_tag`, `squad`, `repository`, `sha` only). The attribute then lands as an empty string, consistent with `image_tag` and `chart_version` already being empty on a sha-tracked dev entry.
- `solace-messaging-python-client/ci.yml:327`, which looks like where this step was copied from, passes a computed `steps.sca_params.outputs.fossa_branch` rather than the raw ref.

Either works. Passing the raw branch name is the one option that does not.

**3. `needs: fossa_scan` couples the manifest pointer to a blocking scan** (line 36 thread). `workflow-config.json` sets both `fossa.policy.mode` and `fossa.vulnerability.mode` to `BLOCK`. A newly disclosed CVE on an unchanged dependency therefore fails `fossa_scan`, `report_to_guardian` never runs, and the dev entry silently freezes at a stale sha. Worth a deliberate decision on whether the manifest write should depend on the scan verdict or just on the push succeeding. `update_release_manifest.sh:143` exits non-zero on a failed write, and since this job is not a required check and has no `slack_channel`, a red write on `main` would page nobody either way.

DCO is still red. None of `bc0e6d0`, `3901800`, `73231d6` carry `Signed-off-by`. `git rebase --signoff main && git push --force-with-lease` clears it, and `make hooks` on `main` installs a hook that adds the trailer from then on.

mustafaal-sakkaf:
> Thanks for the revert. That clears both blockers, and I verified it rather than assuming:
> 
> * `push.branches` is back to `'**'` (line 6), and this run produced `lint` and `build` again.
> * The `fossa_scan` `if:` gate is back (line 25), and the bare `FOSSA Scan` context now reports `skipping` on this PR, which is what the ruleset needs.
> 
> Three things still open before I can approve.
> 
> **1. The Guardian upload still is not wired up** (line 36 thread). I can now name the missing piece. `terraform-provider-solacecloud/.github/workflows/ci.yml:117` carries a second job alongside the manifest write:
> 
> ```yaml
> guardian_db_sync:
>   name: Guardian DB Sync
>   needs: fossa_scan
>   if: github.event_name == 'push' && github.ref == 'refs/heads/develop' && needs.fossa_scan.result == 'success'
>   uses: SolaceDev/maas-build-actions/.github/workflows/guardian-db-sync.yaml@master
> ```
> 
> `report_to_guardian` writes the manifest pointer. `guardian_db_sync` is what reconciles FOSSA results into Guardian and creates/closes the JIRAs. At this head, `git grep -i guardian` over `.github/` returns exactly one hit, this job's own name, and `workflow-config.json` has no guardian section.
> 
> If the sync job is deliberately a follow-up PR, say so and I will approve this as step 1 of 2. My concern is only that DATAGO-145530 gets closed as "onboarded" while the reconcile half is absent.
> 
> **2. `version: ${{ github.ref_name }}` still writes the literal `main`** (line 49 thread). I confirmed the write-through path this time: `update_release_manifest.sh:65` emits `"version":{"S":"${version}"}` with no defaulting or validation.
> 
> One correction to my own earlier advice. I said "deleting is the fix" as if it were the only option. Two peer patterns are both valid:
> 
> * `terraform-provider-solacecloud/ci.yml:110-114` omits `version:` entirely (`release_tag`, `squad`, `repository`, `sha` only). The attribute then lands as an empty string, consistent with `image_tag` and `chart_version` already being empty on a sha-tracked dev entry.
> * `solace-messaging-python-client/ci.yml:327`, which looks like where this step was copied from, passes a computed `steps.sca_params.outputs.fossa_branch` rather than the raw ref.
> 
> Either works. Passing the raw branch name is the one option that does not.
> 
> **3. `needs: fossa_scan` couples the manifest pointer to a blocking scan** (line 36 thread). `workflow-config.json` sets both `fossa.policy.mode` and `fossa.vulnerability.mode` to `BLOCK`. A newly disclosed CVE on an unchanged dependency therefore fails `fossa_scan`, `report_to_guardian` never runs, and the dev entry silently freezes at a stale sha. Worth a deliberate decision on whether the manifest write should depend on the scan verdict or just on the push succeeding. `update_release_manifest.sh:143` exits non-zero on a failed write, and since this job is not a required check and has no `slack_channel`, a red write on `main` would page nobody either way.
> 
> DCO is still red. None of `bc0e6d0`, `3901800`, `73231d6` carry `Signed-off-by`. `git rebase --signoff main && git push --force-with-lease` clears it, and `make hooks` on `main` installs a hook that adds the trailer from then on.

Onboarding to Guardian is a second step. Scheduled daily scans will pick up scan data once the manifest has a record of solace-broker-mcp

solace-aross (changes requested):
Sequencing answer works for me, that closes item 1. Onboarding as step 1 of 2, with the daily scans picking up data once the manifest holds a record, is a reasonable split.

Two items still open. The diff at this head is unchanged apart from the merge from `main`.

**`version: ${{ github.ref_name }}` (line 49)** still writes the literal `main` into the dev entry. This is the one that puts wrong data in a shared table, so it is what blocks. Drop the line, or pass a computed value the way `solace-messaging-python-client/ci.yml:327` does.

**`needs: fossa_scan` (line 36)** still couples the manifest pointer to a BLOCK-mode scan. Your answer sharpens this rather than settling it: if the daily scans only find the repo once the manifest holds a record, then a CVE on an unchanged dependency turning `fossa_scan` red is precisely what stops that record advancing. Either decision is fine, I just do not want it made by default.

DCO is red on all five commits. `git rebase --signoff main && git push --force-with-lease` clears it, and `make hooks` on `main` adds the trailer from then on.

PR #234, opened 2026-07-29
SOL-152405: refresh the flip checklist against live repository state

solace-aross:
## Why

PR #220 made `ADMIN_SETUP.md` accurate as of 2026-07-27. Three claims have gone stale in the two days since, and two of them instruct an admin to do the wrong thing. This closes out the last acceptance criterion on SOL-152405: *every remaining claim verified against repo and live repository state, including the pre-flip release item and the secret-scanning items.*

Docs only. No production surface, so no CHANGELOG entry (`.github/*.md` is outside `SURFACE_RE` in `.github/scripts/production-surface.sh`).

## What changed

**Secret scanning — the one that misleads.** The docs said secret scanning and push protection were both off and told the admin to enable them. The API now reports both on, plus validity checks and non-provider patterns. As written, the admin performs a no-op and a satisfied pre-flip item reads as an open blocker. Now says confirm rather than enable, with the read date and the `gh` command to re-read it.

**DCO confirmation is no longer deferred.** The doc said the two DCO contexts were "not yet confirmed against a live run" and told the admin to hold off registering them. `dco.yaml` landed in `7768d7f` and has run clean on four PRs. I ran the doc's own confirmation command against PR #232 (`29fe3ab`):

```
DCO check self-test        success
DCO sign-off              success
FOSSA Scan                skipped     ← the trap, now observed not inferred
FOSSA Scan / SCA Scan     success
lint build e2e-*          success
```

All 11 documented required contexts report as real check runs. Both DCO contexts are ready to register.

**Pre-flip release item.** Cited v0.5.0 and BREAKING entries. v0.6.0 was tagged 2026-07-28 and `[Unreleased]` now carries a Fixed entry. The item still stands (`main` is 8 commits past v0.6.0), but the versions and reasoning were wrong. Added the commands to re-derive both facts instead of trusting counts that drift.

**Fork-approval setting label.** The doc named the same radio button "all external contributors" in three places and "all outside collaborators" in two, including its own navigation path. An admin looks for that option by label. Now consistent. Prose uses of "external contributor" meaning a person are unchanged.

`OPEN_SOURCE_STATUS.md` carried the same three stale claims, so it gets the same corrections; leaving them would defeat the fix. Its dependency line also now names Renovate, since it previously mentioned only Dependabot.

## Verification

- `gh api repos/SolaceDev/solace-broker-mcp --jq '.security_and_analysis'` — both secret-scanning settings enabled
- `gh api .../commits/29fe3ab/check-runs` — all 11 contexts confirmed, bare `FOSSA Scan` skipped
- `git rev-list --count v0.6.0..origin/main` → 8
- `has_discussions: false` and `can_approve_pull_request_reviews: true` re-checked; existing text on both is correct and untouched

No Go or config files changed, so `make check` was not run.

## Out of scope

The `/discussions` links (SOL-152413, blocked on SOL-152428) and the Solace Community category remain open, as the doc already states.

## Separate finding

**SOL-150691** ("CI: run build/test/lint gates on pull_request so fork PRs are gated") is Closed/Done but its primary AC does not hold. At HEAD, `build-and-test.yml` still triggers on `push` + `workflow_call` only, and `ci-pr.yaml` never calls it, so a fork PR still gets no lint/build/e2e. It was closed 2026-07-02 from Needs Refinement with the comment "We already did this" and no PR reference. `ADMIN_SETUP.md:265-271` documents this gap as still open. Fork CI is in the SOL-152410 epic's definition of done — flagging rather than reopening.

### DIFF
--- a/.github/ADMIN_SETUP.md
+++ b/.github/ADMIN_SETUP.md
@@ -96,19 +96,27 @@ Confirm that landed, or expect broken links from day one.
 
 Navigate to: **Settings → Security → Code security and analysis**
 
-Current state, read from the repository API:
+Current state, read from the repository API on 2026-07-29. These settings change
+outside this file, so re-read before acting on the table rather than trusting it:
+
+```bash
+gh api repos/OWNER/REPO --jq '.security_and_analysis'
+```
 
 | Setting | Now | Action |
 |---------|-----|--------|
 | Dependabot alerts | On | None |
 | Dependabot security updates | On | None |
-| Secret scanning | **Off** | **Enable** |
-| Secret scanning push protection | **Off** | **Enable** |
+| Secret scanning | On | None |
+| Secret scanning push protection | On | None |
+| Secret scanning validity checks | On | None |
+| Secret scanning non-provider patterns | On | None |
 | CodeQL | Running - `Analyze (go)` passes on every PR | Confirm on the settings page |
 
-Enable both secret-scanning settings, not just the first. Alerts find credentials
-already committed; push protection rejects the next one before it lands. Both are
-free on public repositories.
+Both secret-scanning settings are already on, so there is nothing to enable here.
+Confirm rather than change: alerts find credentials already committed, push
+protection rejects the next one before it lands, and losing either is a
+regression. Both stay free on public repositories.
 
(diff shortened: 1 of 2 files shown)


helen-zhu-solace (approved):
Docs only change - LGTM!

PR #235, opened 2026-07-30
SOL-152553: Bump github.com/getkin/kin-openapi to v0.145.0

solace-lootii:
## Summary

Bumps `github.com/getkin/kin-openapi` from `v0.134.0` to `v0.145.0`, closing 2 Dependabot alerts.

Tracked under SOL-152553.

## Motivation and Context

Dependabot flagged 2 vulnerabilities in `openapi3filter`: a critical fail-open auth bypass (`ValidationHandler.Load()`, GHSA-r277-6w6q-xmqw) and a medium nil-pointer panic (schema-less `content` param, GHSA-jpcw-4wr7-c3vq). Both fixed in `v0.144.0+`.

Neither vulnerable path is actually reachable in this codebase — the only usage of this library is the `openapi2` subpackage (`internal/semp/sempv2/operation.go`), parsing our own embedded, trusted SEMP specs, never `openapi3filter`. Bumping anyway to close the alerts and stay current.

## Changes Made

- `go.mod`/`go.sum`: `kin-openapi` → `v0.145.0` (plus transitive bumps to `go-openapi/jsonpointer`, `go-openapi/swag/jsonname`, `oasdiff/yaml`, `oasdiff/yaml3`, and new compiled deps `santhosh-tekuri/jsonschema/v6` and `golang.org/x/text`)
- `THIRD_PARTY_LICENSES.md`/`NOTICE`: regenerated to match the new dependency closure — adds the new components above, and folds the `golang.org/x/text` CC-BY-SA finding in as a standing exception note (reachability verified independently by two reviewers; see the FOSSA/Legal discussion below)
- `CHANGELOG.md`: new `### Security` entry under `[Unreleased]`

## Testing

- `go build ./...`, `go vet ./...` — clean
- `go test -race ./...` — clean across all packages
- `golangci-lint run ./...` — 0 issues
- `.github/scripts/licenses-check.sh` — passes, inventory matches `go list -deps ./cmd/server`
- No code changes required; dependency-only bump

## Breaking Changes

N/A for the vulnerability itself (unreachable, per the reachability analysis above), but worth recording why the *fix* doesn't introduce risk either, since v0 semver carries no compatibility guarantee here. Across the two packages in our compiled graph (`openapi2`, which we import directly, and `openapi3`, which it pulls in), 9 of 12 non-test files changed in the first and 44 of 53 in the second — including a new JSON Schema validation backend gated behind `useJSONSchema2020`, which we never invoke since we only unmarshal and read struct fields, never call `Validate()`/`NewLoader()`/any `openapi3` symbol. Two changes in `openapi2` are behavioral rather than cosmetic: `Parameters` lost its `sort.Interface` implementation, and `Schema.UnmarshalJSON` dropped its date-format `Example` fix (upstream issue #697). Both are inert here — we never reference `openapi2.Parameters`, never read `.Example` in `internal/semp/`, and no embedded SEMP spec declares `"format": "date"`. Neither change would be caught by `build`/`vet`/`test`/`lint`, since nothing asserts on parameter ordering or `Example` values, which is why this reasoning is recorded here rather than left to CI. (Analysis by @amitmorade, verified independently by @solace-aross.)

## Additional Notes

No production code touched — dependency bump plus the licence-inventory files it obligates (`THIRD_PARTY_LICENSES.md`, `NOTICE`) and `CHANGELOG.md`.


solace-aross (commented):
Not approving yet, and it isn't the code. The bump is clean; the merge is waiting on the FOSSA licensing exemption.

I verified the x/text analysis independently and it holds:
- x/text is in the compiled graph, reachable as `openapi2` → `openapi3` → `santhosh-tekuri/jsonschema/v6` → `x/text/language`, 12 packages in total.
- The CC-BY-SA content sits in exactly three paths in v0.14.0: `unicode/norm/normalize_test.go`, `cases/map_test.go`, `internal/testtext/text.go`. None of the three is among those 12 compiled packages.

So the note at `THIRD_PARTY_LICENSES.md:75-82` is correct. One wording suggestion: "confirmed absent from this project's actual compiled dependency graph" reads on first pass as a claim about the module, and the module is present. Naming the compiled set instead makes the argument harder to dismiss, which matters if Legal leans on this note.

Also worth having before that conversation: bumping x/text does not clear the finding. v0.40.0 carries the same three files.

Nothing else from me. The reachability rationale for both advisories checks out, `openapi3filter` is not imported anywhere in the tree, and the CHANGELOG and licenses file are both updated. Glad to switch to approve once the exemption is recorded.

solace-aross (changes requested):
Switching from comment to request-changes. The bump itself is unchanged and I still hold to last pass's read of it. What has changed is that part of the blocker is now yours to move rather than Legal's.

**The branch conflicts with `main`.** It is 33 commits behind and GitHub reports `mergeStateStatus: DIRTY`. Two of the four files it touches were rewritten underneath it:

- `THIRD_PARTY_LICENSES.md` was regenerated wholesale by SOL-152414 (#242). Main's copy now carries `sony/gobreaker/v2`, `otter/v2`, `testify`, `go-spew` and `go-difflib`, and a completed `NOTICE`.
- `go.mod`'s module path became `github.com/SolaceProducts/solace-broker-mcp` (SOL-152669, #239).

**Regenerate the inventory rather than hand-resolving it.** `main` now carries `.github/scripts/licenses-check.sh` behind the `Third-party licenses current` check. It fails when the table stops matching `go list -deps ./cmd/server` in either direction, when a row does not parse, or when a dependency's own NOTICE goes unpropagated. Your table is already correct for your dependency set, x/text row and CC-BY-SA note included, so a regenerate should be clean. A hand-merged conflict is the likeliest way to trip the parse check, which is the failure mode that gate exists to catch.

Worth checking after the merge: whether `santhosh-tekuri/jsonschema/v6`, or anything else the bump pulls in, ships a NOTICE file. That propagation obligation was not gated when you opened this.

**The FOSSA licensing exemption is still the external blocker** and my x/text analysis stands. One thing around it has moved, and it will mislead you on the rerun: the scan now targets the `SolaceProducts_solace-broker-mcp` project, whose baseline is red on its own. #231 and #238 both change zero dependencies and both fail its licensing and vulnerability checks. So a red SCA result after your rebase is not evidence that x/text is still the cause. I am raising that baseline separately.

Every check on this PR is from 30 July, against the old FOSSA project and a `main` that has since moved. There is nothing current here to approve, which is the honest reason for the verdict rather than anything wrong in the diff.

amitmorade:
Verified the reachability argument independently and it holds — `openapi3filter`
has zero imports in the tree, and the only usage is the single `openapi2` import
at `internal/semp/sempv2/operation.go:11`.

One suggestion on **Breaking Changes: N/A**. That answers "can the vulnerability
hurt us?" but not "can the fix hurt us?" Two packages are in our compiled graph —
`openapi2`, which we import, and `openapi3`, which it pulls in — and across these
eleven `v0` versions 9 of 12 non-test files changed in the first, 44 of 53 in the
second, including a new JSON Schema validation backend.

Nearly all of it is unreachable for us, but it took checking to say so:

- The new validator sits behind `useJSONSchema2020`, a validation entry point. We
  never call `Validate()`, `NewLoader()`, or any `openapi3` symbol directly — we
  unmarshal and read struct fields.
- Two `openapi2` changes are behavioral, not cosmetic: `Parameters` lost its
  `sort.Interface` implementation, and `Schema.UnmarshalJSON` dropped its
  date-format `Example` fix (issue #697). The second is on our read path.
- Both are inert here: we never reference `openapi2.Parameters`, never read
  `.Example` in `internal/semp/`, and no embedded SEMP spec declares
  `"format": "date"`.

So `N/A` is the right answer — the ask is just that the reasoning live in the
description. At `v0` semver offers no compatibility guarantee, so the version
number isn't doing that work, and build/vet/test/lint wouldn't catch either
removal: nothing asserts on parameter ordering or `Example` values.

On @solace-aross's NOTICE question, now that the merge has landed:
`santhosh-tekuri/jsonschema/v6` is the one genuinely new compiled dependency. It
ships a `LICENSE` but no `NOTICE`, so there's no propagation obligation —
consistent with `Third-party licenses current` passing. That closes the item
rather than adding one.

Nothing blocking in the code. Worth noting the FOSSA results still showing here
are from the 30 July run against the old project and haven't re-run since the
merge, and GitHub still reports the branch `BEHIND` — so there may be another
sync needed before the checks mean anything.

solace-aross (dismissed):
Approved on the code. Everything I asked for last pass has landed. One merge condition below, and one correction to the FOSSA reading in this thread.

**My items are closed**

- The conflict is resolved and the branch is `MERGEABLE`. You merged `main` rather than hand-resolving, which was the right call.
- The licence artifacts were regenerated, not patched. I ran the repo's own gate at `c096964`: `✅ THIRD_PARTY_LICENSES.md and NOTICE match the 28 modules in the dependency closure of ./cmd/server (29 rows checked)`. `Third-party licenses current` is green in CI too.
- Folding the `x/text` analysis into `THIRD_PARTY_LICENSES.md:102-113` as a standing exception note is better than leaving it in PR comments. I re-verified it at this head rather than trusting the earlier pass: `go list -deps ./cmd/server` gives exactly 12 `x/text` packages, and none of the three CC-BY-SA-bearing files (`unicode/norm/normalize_test.go`, `cases/map_test.go`, `internal/testtext/text.go`) is present in any of them. The count and the reachability claim are both accurate.

@amitmorade's `santhosh-tekuri/jsonschema/v6` answer closes my NOTICE question: `LICENSE` and no `NOTICE` means no propagation obligation, which is why the gate is green. Agreed.

**Correction on the FOSSA state**

@amitmorade, the results have re-run since the merge. What is stale is the bot comment, not the scan. Run `30919347851` on `c096964` at 14:32 today reports `LICENSING_OUTCOME: failure` and `VULNERABILITY_OUTCOME: failure`, both in `BLOCK` mode. No SolaceProducts guard comment was posted, so the thread still shows only the 30 July SolaceDev one. The red is current, not left over.

That red is not evidence about `x/text`, though. #238 at the same baseline, with zero dependency changes in its diff, reports the identical pair of failures. The new `SolaceProducts_solace-broker-mcp` project is failing at baseline, which is tracked separately and needs an owner.

**Merge condition**

Do not merge on this approval while licensing is red. `main-protection` requires only `lint`, `build`, and the bare `FOSSA Scan` context, and that bare context reports `skipping` on pull requests, so it passes. The contexts actually failing, `FOSSA Scan / SCA Scan` and `SCA gate`, are not required. Nothing mechanical will stop this from going in, so the FOSSA/Legal exemption has to be the gate here. The branch is also `BEHIND` again and needs another sync.

Still worth doing, not blocking: put @amitmorade's "can the fix hurt us" reasoning in the description. At `v0` the version number carries no compatibility guarantee, and neither the `openapi2.Parameters` sort removal nor the dropped date-format `Example` fix would be caught by build, vet, test, or lint. That analysis is the only record that someone checked.

amitmorade (approved):
LGTM

THIRD_PARTY_LICENSES.md
```diff
@@ -92,6 +89,26 @@ under MPL-2.0.
...
+  license URL needs the `jsonname/` tag prefix, since that is where the tag
+  actually lives in the upstream repository.
+- `golang.org/x/text`'s declared module license is BSD-3-Clause (above). FOSSA
+  additionally reports CC-BY-SA as "discovered" for this dependency; that
+  content is Creative Commons license text quoted in several languages, used
+  only as sample Unicode test data in three files in `x/text`'s own test suite:
+  `unicode/norm/normalize_test.go`, `cases/map_test.go`, and
+  `internal/testtext/text.go`. `x/text` is reachable in the compiled dependency
+  graph via `openapi2` → `openapi3` → `santhosh-tekuri/jsonschema/v6` →
+  `x/text/language` — 12 packages in total — and none of those three files is
```
solace-aross, line 109: "12 packages in total" is correct, and I confirmed it. The problem is that the obvious way to check it disagrees.

`go list -deps ./cmd/server | grep golang.org/x/text` returns **16** lines here, not 12. Four of them are the Go toolchain's own vendored copies, reached through `net/http` → `x/net/idna`:

```
vendor/golang.org/x/text/transform
vendor/golang.org/x/text/unicode/bidi
vendor/golang.org/x/text/secure/bidirule
vendor/golang.org/x/text/unicode/norm
```

Those are not the module, and vendored stdlib copies carry no `_test.go` files, so they cannot bring the CC-BY-SA content in. The 12 in your note are the module packages, which is the right number for the claim being made.

Worth one clause saying so. This note exists to be re-checked by someone deciding a licensing question, and as written the first command they reach for contradicts it. The `vendor/golang.org/x/text/unicode/norm` line is the one most likely to cause a wrong conclusion, because `unicode/norm` is also the package holding `normalize_test.go`: a reader could reasonably decide the note is wrong and that the encumbered file *is* in the closure.

solace-aross (approved):
Approved. Nothing you wrote has changed since my last pass, and the two merges from `main` did not disturb the parts I cared about. I checked rather than assuming, since a merge is exactly where a regenerated artifact goes stale.

**Verified at `278d1d3`:**

| Claim | Evidence |
|---|---|
| Licence artifacts still match the graph after both merges | Ran the repo's own gate: `THIRD_PARTY_LICENSES.md and NOTICE match the 28 modules in the dependency closure of ./cmd/server (29 rows checked)` |
| `x/text v0.14.0` still arrives only through this PR | `go.mod` on `main` has no `x/text` at any version; it appears here via `kin-openapi 0.134.0 → 0.145.0` |
| The standing exception's package count is exact | Re-measured with `go list -deps ./cmd/server`: 12 `x/text` module packages, matching the note |
| None of the three CC-BY-SA files is compiled | `unicode/norm`, `cases`, `internal/testtext` are all absent from the module closure |
| Author-side commits unchanged | `git log --no-merges origin/main..278d1d3` is still just `6d73390` and `6b838a9`, both predating my last review |

One note on the exception text, inline. Not blocking, and worth fixing while the note is fresh rather than when someone is auditing under time pressure.

**On the red licensing check:** still not evidence about `x/text`, and I have same-day proof this time. #254 changes tests, a Makefile target and docs, touches no dependency, and reports the identical `FOSSA Scan / SCA Scan` and `SCA gate` failures. The `SolaceProducts` project is red at baseline. The remaining gate here is the FOSSA/Legal exemption, not anything you can do in the diff.

One thing that shifted today and is worth knowing: **#252 was closed unmerged.** That was the Dependabot root Go bump carrying the same `kin-openapi 0.145.0` and therefore the same `x/text v0.14.0`. With it closed, this PR is once again the only route for that dependency into the module, so the exemption decision lands entirely on this thread.

PR #237, opened 2026-07-30
SOL-152603: Add describe-semp-schema tool

solace-wmaryam:
## Summary

Add a `describe-schema` MCP tool that returns the SEMPv2 schema slice for a given operation's request-body definition, so an agent planning a create or update can enumerate every configurable attribute (with types, defaults, enum values, and writability flags) before invoking a write tool.

Closes [SOL-152603](https://sol-jira.atlassian.net/browse/SOL-152603)

## Type of Change

## Motivation and Context

Without `describe-schema`, a model planning an unfamiliar write can produce right-looking configs by reasoning through wrong theories and landing on plausible values by luck. With it, the model reaches for the schema when uncertain and produces verified configs instead of guessed ones — the value is _reliably_ correct outputs, not just correct outputs.

The tool is read-only and broker-independent (no `broker` parameter, no RBAC surface), registered standalone like `list-brokers`. Its response is derived entirely from the embedded OpenAPI specs and reveals nothing broker-specific.

## Changes Made

- New native Go handler `internal/tools/describe_schema.go` implementing `RegisterDescribeSchema`; parses all embedded SEMPv2 specs at startup and serves two views: `trimmed` (compact per-attribute list with name, type, description, enum, default, and writability flags from SEMP's `x-*` extensions) and `raw` (OpenAPI definition verbatim).
- Exported `DeriveSpecType` in `internal/semp/sempv2/operation.go` (was package-private) so the new handler can reuse spec-type derivation without duplicating it.
- Added a `describe-schema` pointer (with the SEMPv2 operation identifier) to the descriptions of the eight `create-*`/`update-*` write tools in `tools.yaml`.
- Lifted the `BEFORE INVOKING THIS TOOL` confirmation block to the top of every write/destructive tool description, so the confirmation requirement is the first thing an agent reads.
- `cmd/server/main.go`: registers `describe-schema` alongside `list-brokers` at startup, with a fatal exit if registration fails.
- `CHANGELOG.md` updated under `[Unreleased]` (Added + Changed entries).

## Testing

**Test Environment:**
- Go version: 1.26.3
- OS: Darwin 25.5.0 (arm64)
- Broker version (if applicable): N/A — tool is broker-independent

**Tests Performed:**

**Test Coverage:**
- `describe_schema_test.go` covers: trimmed and raw views, unknown operation error, operation with no body definition, attribute flag rendering (`writableOnCreate`/`writableOnUpdate`, identifying, requiredForCreate, sensitive, deprecated, requiresDisable), first-paragraph description truncation, and sorted attribute ordering.

## Breaking Changes

None. `describe-schema` is a new additive tool. The write tool description changes are text-only (no parameter or output-shape changes). `DeriveSpecType` export is backward-compatible.

### Code Quality

### Security

### Testing
- [x] Error paths tested

### Documentation

### Git & CI
- [ ] Branch is up to date with main (rebased if needed)
- [x] No merge conflicts

---

**For Reviewers:**

**Focus Areas**: Schema parsing logic in `buildSempSchemaMap` and `trimAttributes` (especially the `x-*` extension mapping); the `DeriveSpecType` export (previously package-private — verify no unintended callers are affected); write-tool description ordering changes in `tools.yaml`.

**Questions**: The branch is currently behind main by a few commits (SOL-152499, SOL-152508, SOL-152405). I'll rebase before marking ready for review.

[SOL-152603]: https://sol-jira.atlassian.net/browse/SOL-152603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solace-aross (changes requested):
Good tool to add, and registering it read-only and broker-independent is the right call. Two blocking issues before it goes in. Both verified on the branch at 89f0620.

## Blocking

**1. An RBAC grant naming `describe-schema` takes the server down.**

`internal/tools/authorization.go:190` builds the known-tool set from `mgr.Handlers()` plus `listBrokersToolName` only. `describe-schema` registers outside `mgr` (`cmd/server/main.go:741`), so `ValidatePolicyToolNames` classifies it as a typo and `main.go:757` exits 1. Reproduced locally against the branch:

```
describe-schema grant -> unknown tool "describe-schema" (referenced by groups: Ops)  [fatal]
list-brokers    grant -> nil + 1 WARN "grant has no effect; tool is exempt"          [boots]
```

The tool is in `tools/list`, and `docs/configuration.md:220` tells operators that a group granting every tool "must list every tool name explicitly". So adding it to a read-only group is the expected operator path, and it brings a running server down on a config change that is semantically a no-op. This is the failure class b669e9f (SOL-152508) closed one commit back.

Fix: add `describeSchemaToolName` to `known` and route it through the same exempt-tool WARN branch, with a test next to `TestValidatePolicyToolNames_ListBrokersGrant_WarnsNotError`.

**2. `$ref` attributes report fabricated writability, with no type and no way to resolve them.**

21 of the properties in the four definitions this PR points write tools at are bare `{"$ref": ...}` with no `type`, no `description`, and none of the `x-*` extensions. `describe_schema.go:203-206` reads those absent booleans and emits:

```json
{"name":"eventBindCountThreshold","writableOnCreate":true,"writableOnUpdate":true}
```

The comment just above says the flags are always emitted because omitting them "would let the agent read absence as writable". For these attributes, absence is exactly what is being read as writable. The real shape is `{clearPercent, clearValue, setPercent, setValue}` with mutually exclusive constraints, and it is unreachable through this tool: `raw` returns the `$ref` string verbatim, and `MsgVpnQueueEventBindCountThreshold` is not the body of any operation, so there is no follow-up call. Counts: `MsgVpn` 15 of 124, `MsgVpnQueue` 3 of 34, `MsgVpnTopicEndpoint` 3 of 31.

Either resolve one level of `$ref`, or emit `type: object` plus the ref and a note. Please don't synthesize a writability flag from a missing extension.

## Worth fixing before merge

**3. The trimmed view drops the fields that cause the rejections it exists to prevent.**

`describe_schema.go:190-197` truncates the description at the first `\n\n`. SEMP puts the `<pre>` block defining each enum value in a later paragraph, so it never survives: 3 of 3 enum attributes on `createMsgVpnQueue`, 14 of 14 on `createMsgVpn`. `permission` reaches the caller as `enum: ["no-access","read-only","consume","modify-topic","delete"]` with no definition of any value. The comment at `:187-189` says that block is "already surfaced structurally"; the values are, the meanings are not.

Also dropped: `pattern`, `maxLength`, `minimum`/`maximum` (14 of 34 `MsgVpnQueue` properties carry a numeric bound; `queueName` has `maxLength: 200` and `pattern: ^[^*?'<>&;]+$`), and `x-autoDisable`, non-empty on 9 of 34 `MsgVpnQueue` properties. `x-autoDisable` is how SEMP signals that changing an attribute on a live object forces `egressEnabled` to false. An agent that reads `permission: {writableOnUpdate: true}` calls `update-queue` and drops egress with no warning to the user.

**4. Two test gaps, two tests.**

Renaming `x-readOnlyPost` in the implementation leaves the suite green while every attribute of every operation reports `writableOnCreate: true`. Nothing asserts that field in either polarity, and the counterexample is already in the definition the tests load (`MsgVpnQueue.msgVpnName` has `x-readOnlyPost: true`).

Separately, nothing constructs an `mcp.Server`, so the handler's audit line is untested. `TestListBrokersEmitsAuditLog` (`register_test.go:267`) is the precedent, and that audit path is the whole reason the handler duplicates the defer contract.

**5. Docs.**

`describe-schema` appears in no markdown outside `CHANGELOG.md`. Now stale: `docs/tools-reference.md:24` and `README.md:79` ("Every tool except `list-brokers` takes a required `broker` parameter"), `docs/tools-reference.md:123`, `docs/configuration.md:222` and `docs/authentication.md:425` (`list-brokers` as the sole structural RBAC exemption), `docs/user-guide.md:94` (tool counts), and `CLAUDE.md:27`. `cmd/server/toollist_budget_test.go:60` claims it "can never drift from what main() registers", and it now does, since `registeredServer` doesn't call `RegisterDescribeSchema`.

## Two smaller notes

`describe_schema.go:31` justifies the RBAC exemption with "the embedded OpenAPI spec is public documentation". These are the private specs: `basePath` is `/SEMP/v2/__private_config__` and `info.description` labels it "Internal use only". I think the conclusion still holds, since the response carries no broker state, but the reason as written doesn't, and the CHANGELOG repeats it. "Spec content only, no broker state" would be accurate.

Lifting the confirmation block on `delete-*` and `disconnect-client` is an independent change to LLM-visible prompt text landing in a schema-discovery PR. `test/e2e-llm/scenarios/b4-*` and `b5-*` cover exactly those tools, and the workflow is `workflow_dispatch` only, so nothing ran on this branch. Either run those rows here, or split the reordering into its own PR where it can be judged on its own.

Also: branch name says SOL-152204, title and CHANGELOG say SOL-152603.

solace-aross (dismissed):
Approving. Everything from the last round is addressed, and the `$ref` fix took the better of the two options I offered.

What I checked:

- **RBAC crash.** `known[describeSchemaToolName]` is in the set, with its own WARN path alongside `list-brokers`, and `TestValidatePolicyToolNames_DescribeSchemaGrant_WarnsNotError` covers it. A grant naming the tool no longer takes the server down.
- **Fabricated writability.** The bare-`$ref` branch now sets `type: object`, resolves one level, inlines the sub-definition's properties, and `continue`s before the writability block, so nothing gets synthesised from absent extensions. Resolving beats annotating.
- **Trimmed view.** The `<pre>` block survives, and `pattern`, `maxLength` and `x-autoDisable` are carried through.
- **Tests.** `TestDescribeSchema_EmitsAuditLog` covers the audit path, and `toollist_budget_test.go` now calls `RegisterDescribeSchema`, so its "can never drift" claim is true again.
- **Docs.** README, CLAUDE.md and the four `docs/` pages are all in the diff.
- **Rationale.** `describe_schema.go` now reads "spec content only, no broker state", which is accurate where "public documentation" was not.

Withdrawing my request to split out the confirmation-block change. I read the hunk rather than assuming. On `delete-queue-messages` the confirmation text is word-for-word identical and has only moved to the front of the description, which is where all six sibling destructive tools already carry it. Nothing is weakened, and the odd one out is now consistent. Running the `b4` and `b5` rows is still worth doing at some point, but it is not a gate on text that did not change.

`build`, `lint`, `Analyze (go)` and all five `e2e-*` suites are green. FOSSA is red from the Vault outage hitting every PR since roughly 14:26 today, tracked as SOL-152670, not from anything in this change.

solace-aross (dismissed):
Re-approving. The only change since my last approval is the module-path rebase onto #239, and I diffed it rather than assuming.

- `describe_schema.go` and `authorization.go` differ solely in the import prefix.
- `internal/composite/definitions/tools.yaml` is byte-identical, so none of the LLM-visible tool text moved.
- No `SolaceDev` import remains in the files this PR owns.

`build`, `lint`, `Analyze (go)` and four of the five `e2e-*` suites are green, with `e2e-monitoring` still running. FOSSA is the Vault outage tracked as SOL-152670.

solace-aross (dismissed):
No changes wanted from me on the code. One mechanical blocker before this can merge: neither new commit carries a `Signed-off-by`, and `DCO sign-off` is a required check.

```
git rebase --signoff e4e35acd5dddf06725035ce3719646e9e1256dcf && git push --force-with-lease
```

That is the command the failing job prints. The force-push will dismiss this approval, so give me a shout when it is back and I will re-look at the same head.

**The rename is clean, and it is free**

`describe-schema` never shipped: its CHANGELOG entry is still under `[Unreleased]`, and the `[0.6.0]` block does not mention it. So no BREAKING marker is needed and no consumer can be holding the old name. Correct call not to add one.

Repo-wide grep for `describe-schema` at `613ea86` returns nothing, so the rename is complete across `tools.yaml`'s eight write-tool pointers, `authorization.go`, the docs set, `CLAUDE.md`, `README.md`, `main.go`, and both test files. `describe-semp-schema` is kebab-case and the `Name:` field matches the constant, per the house rule.

**Dropping the monitor spec is right, and I checked the premise**

The claim is that monitor operations would only ever return an empty attribute list. I parsed all three embedded specs and counted:

| spec | methods | operations with a request body |
|---|---|---|
| monitor | 258 GET | **0** |
| config | 197 GET, 92 POST, 65 PATCH, 64 PUT, 92 DELETE | 221 |
| action | 57 GET, 66 PUT | 66 |

Zero, exactly as stated. Nothing is lost by not indexing it.

`TestSempSchemaMap_ExcludesMonitorSpec` earns its green: I removed the `specType == "monitor"` skip and it failed on both assertions, reporting real indexed operations (`monitor/getMsgVpnKafkaSenderQueueBindings` and others). Restored, it passes.

I also checked for a dangling consumer, since `reg.specs` no longer holds a `monitor` entry. The only read is line 145, `r.specs[info.specType]["definitions"]`, and `info.specType` can never be `monitor` now that no monitor operation is indexed. No nil-map path.

`make check` exits 0 at `613ea86`.

**Two small notes, neither worth a round trip**

1. `specType` comes from `spec["basePath"]`, so the 3.4 MB monitor spec is still read and unmarshalled before being discarded at line 79. Boot-time only, and strictly better than before this PR (it used to be parsed *and* retained), so this is an observation rather than a request. Deriving the type from the filename to skip earlier would trade the authoritative `DeriveSpecType(basePath)` for a naming convention, which I would not take.

2. The `unknown operation` error at line 137 explains the expected form but not that monitor is out of scope. The `operation` parameter description does say it, so an agent reading the schema is covered; one that guesses `monitor/getMsgVpnQueue` gets a message that does not tell it why. Adding "monitor operations are not indexed" to that error would close the loop cheaply.

bczoma (changes requested):
I suggest improving the documentation for what this tool actually is: mainly a helper. Although one direct example when only this tool would be called is "What attributes can I set when creating a queue?", which is plausible but is the rare path. The common path is: a human says "create a queue with these settings" in ordinary language, and the agent calls describe-semp-schema unprompted because eight write-tool descriptions tell it to. An operator watching a trace will see a tool fire that they never asked for and that no document told them to expect. That's the one thing worth writing that isn't written.

Also would be good to add a basic test in one of the e2e suites.

solace-wmaryam:
Sounds good @bczoma, I'll make these changes now

bczoma (approved):
Looks good, thank you!

PR #238, opened 2026-07-30
SOL-152586: Use Renovate (not dependabot.yml) for Go module and GitHub Actions updates

solace-lootii:
## Summary

Originally added `.github/dependabot.yml` to enable Dependabot's scheduled version updates for Go modules and GitHub Actions, checking daily. Per review feedback (see thread below), that approach is dropped: **`.github/dependabot.yml` is removed**, and `.github/renovate.json` is extended instead to cover the same two ecosystems (`gomod`, `github-actions`), alongside the Claude Code CLI pin it already manages.

Tracked under SOL-152586.

## Motivation and Context

This is separate from the Dependabot security alerts already active on this repo — those fire when a CVE is published against a version we're using, regardless of this file. This adds routine, proactive version-drift checking so outdated dependencies (security-relevant or not) get flagged and proposed for update on a schedule, rather than only being noticed when a CVE lands.

**Why Renovate instead of Dependabot:** the required `DCO sign-off` check exempts nobody, including bots — an author-name exemption would be trivially forgeable. Renovate satisfies it today via its `commitBody` setting, which appends a `Signed-off-by` trailer to every commit it makes. Dependabot's `commit-message` config has no equivalent field (only `prefix`, `prefix-development`, `include: scope`), so every PR it opened would be permanently red on a required check, with no fix available on Dependabot's side.

## Changes Made

- Removed `.github/dependabot.yml`.
- Extended `.github/renovate.json`:
  - `enabledManagers` now includes `gomod` and `github-actions` (previously `npm` only).
  - New `packageRules`, carved out of the existing blanket `enabled: false` default: Go dependencies grouped into a minor/patch PR and a separate major PR, and GitHub Actions grouped into one PR — each with `"schedule": ["at any time"]` so they aren't limited to the existing Monday-only window used by the CLI-pin rule.
  - No per-directory config needed for Go: Renovate discovers all three `go.mod` files in the repo on its own (root, `test/e2e-common/broker-driver`, `test/e2e-basic-mcp/agent`), unlike Dependabot which would have needed an explicit `directories` list.
- Updated `.github/ADMIN_SETUP.md`'s description of Renovate's scope (it previously said Renovate covers only the CLI pin and that Go modules are "deliberately excluded"). `.github/OPEN_SOURCE_STATUS.md`'s existing "Renovate not Dependabot" guidance needed no change — it already matches this outcome.

## Testing

- Validated `renovate.json` with Renovate's own official `renovate-config-validator` (via `npx`) — passes schema validation.
- Ran the repo's `.github/scripts/dco-check.sh` locally against this branch's commit — confirms it carries a valid sign-off.
- Not fully testable pre-merge: whether a live Renovate run actually opens the expected grouped PRs (with sign-off intact) for `gomod`/`github-actions` can only be observed once this lands on `main` and Renovate's next run against the repo fires.
- Open item, unrelated to this diff but relevant to whether it does anything at all: this repo moved from the `SolaceDev` org to `SolaceProducts` recently (SOL-152415). GitHub App installations don't carry over an org transfer, and no Renovate PR has ever been opened on this repo under either org name — worth an org admin confirming Renovate/Mend is actually installed and has this repo selected under `SolaceProducts`.

## Breaking Changes

N/A — repo config only, no production code touched.

## Additional Notes

No `CHANGELOG.md` entry — this doesn't touch production/server behavior, just repo tooling config.


solace-aross (changes requested):
The motivation is right, and the description does a good job separating scheduled version updates from the security alerts already running. Two things block it as written, both about how this config lands in *this* repo rather than anything wrong with the YAML.

## Blocking: every Dependabot PR will fail the `DCO sign-off` gate

`DCO sign-off` is a required status check on `main-protection`, and `.github/scripts/dco-check.sh` exempts nobody by design. Its header comment says why:

> Bots are not exempt. An author-name exemption would be trivially forgeable (`git commit --author='renovate[bot] <...>'`), which is exactly the bypass this gate must not have. Renovate signs off its own commits instead, via the `commitBody` setting in `.github/renovate.json`.

Dependabot has no equivalent. Its `commit-message` block accepts only `prefix`, `prefix-development`, and `include: "scope"`. There is no field for a commit body or trailer, so its commits carry no `Signed-off-by` line and every PR this config opens goes red on a required check. Merging one then needs a human to amend and force-push, which also breaks Dependabot's own rebase handling on that branch.

Worth confirming against a real Dependabot PR after merge, but I'd expect this to be deterministic rather than intermittent, and it makes the feature non-functional rather than merely noisy.

## Blocking: this contradicts two checked-in policy docs

- `.github/ADMIN_SETUP.md:128` says "Dependency updates use Renovate, not Dependabot", then "Do **not** create `.github/dependabot.yml`. It would contradict that scoping."
- `.github/OPEN_SOURCE_STATUS.md:123` repeats it as an admin checklist item.

If the team has re-decided, that's fine, but the reversal needs to land in both files in this PR. Otherwise the next maintainer gets contradictory instructions and the SOL-152586 rationale lives only in a thread. Could you link the decision in the description?

## Suggestion: extend `renovate.json` rather than add a second updater

Renovate is already installed, already solves the DCO problem, and already carries the noise controls this repo decided it wanted. Adding `gomod` and `github-actions` to `enabledManagers` with their own `packageRules` reaches the same outcome with one tool, one config, and no red required check. It also keeps the "prevent runaway PRs" reasoning in one place instead of splitting dependency policy across two files with different philosophies.

If there's a specific reason Dependabot is the better fit here, say so in the description and I'll drop this.

## Worth fixing either way: the nested modules aren't covered

`directory: "/"` watches the root module only. Two others exist:

- `test/e2e-basic-mcp/agent/go.mod`
- `test/e2e-common/broker-driver/go.mod`

Both need their own `updates` entries, or a `directories:` list. Test harnesses, so lower stakes, but silently uncovered reads as covered. If they're meant to be excluded, a comment in the file saying so is worth the two lines.

## Minor: `open-pull-requests-limit` left at the default

The default is 5 across both ecosystems combined. With 10 direct Go deps and seven workflow files the volume is modest, so `daily` is reasonable and I'm not relitigating it. But a run that hits the cap just stops opening PRs, and the ones it skipped are invisible. Either raise the limit or add `groups` (one PR for Go minor/patch, one for actions) so a busy week can't quietly starve.

solace-aross (dismissed):
Approving. All four items from the last pass are addressed, and switching to Renovate is the right call rather than a workaround for the DCO gate.

What I checked rather than took on trust:

- Ran Renovate's own `renovate-config-validator` against this branch's `renovate.json`: `Config validated successfully`.
- `DCO sign-off` is green on the branch.
- `OPEN_SOURCE_STATUS.md:123-125` already says Renovate not Dependabot, so it genuinely needed no change.

One thing worth a half-sentence in the config, not blocking:

The "Renovate finds all three `go.mod` files on its own" claim holds, but it rests on a line that reads as redundant. `config:base` resolves to `config:recommended`, which pulls in `:ignoreModulesAndTests`, whose `ignorePaths` is `["**/node_modules/**", "**/bower_components/**", "**/vendor/**", "**/examples/**", "**/__tests__/**", "**/test/**", "**/tests/**", "**/__fixtures__/**"]`. The top-level `"ignorePaths": ["**/node_modules/**"]` in this file replaces that array, and that replacement is the only reason `test/e2e-common/broker-driver/go.mod` and `test/e2e-basic-mcp/agent/go.mod` stay in scope. It never mattered before because `gomod` was not an enabled manager. If a later cleanup drops that line as noise, both test modules fall out silently and the config still validates. A clause in the `description` naming it would stop that.

Also minor: `ADMIN_SETUP.md:130` says the three jobs are "each grouped into one PR", but Go splits into minor/patch and major. Copilot flagged the same wording. Small, but this is the file a future admin reads first.

On the red CI: `FOSSA Scan / SCA Scan` and `SCA gate` are failing and it is not from this branch. This PR changes no dependencies. #231 changes none either and fails the same way against the new `SolaceProducts_solace-broker-mcp` project, on licensing and vulnerability both. That reads as a fresh-project baseline rather than anything either branch introduced. Neither context is required on `main-protection`, so it will not block you. I will raise it separately.

solace-aross (approved):
Re-approved. Both notes from my last pass are addressed, and the only PR-authored change since `4cde5ea` is documentation.

- `renovate.json`'s `description` now names the `ignorePaths` dependency outright: that `config:base` resolves through `config:recommended` to `:ignoreModulesAndTests`, that the local list replaces rather than extends it, and that this is the only reason the two nested test `go.mod` files stay in scope. That was the load-bearing detail, and the "do not drop it as apparent redundancy" sentence is the part that will save the next reader.
- `ADMIN_SETUP.md:130` now matches what the config actually produces: one Go minor/patch PR, one Go major PR, one Actions PR.

Re-ran Renovate's own `renovate-config-validator` against this head: `Config validated successfully`. `DCO sign-off`, `lint`, `build`, `Analyze (go)`, and all five e2e suites green.

Two mechanics rather than review items. The branch is `BEHIND` main so it needs a merge before it can go in. `FOSSA Scan / SCA Scan` and `SCA gate` are red for the SolaceProducts-project baseline reason tracked separately, not for anything here: the diff changes no dependencies.

PR #239, opened 2026-07-31
Rename Go module path from SolaceDev to SolaceProducts

amitmorade:
## Summary

Renames the Go module path from `github.com/SolaceDev/solace-broker-mcp` to `github.com/SolaceProducts/solace-broker-mcp`, matching the repository's new home, and updates every internal import to match.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The repository moved from the SolaceDev org to SolaceProducts ahead of going public under Apache 2.0. GitHub's redirect covers git/web traffic but does not update the declared Go module path. Left unchanged, the first external `go get github.com/SolaceProducts/solace-broker-mcp` after the repo goes public would fail with Go's `unexpected module path` error, since the module still declares itself as `github.com/SolaceDev/solace-broker-mcp`.

## Changes Made

- Updated `go.mod`'s module declaration to `github.com/SolaceProducts/solace-broker-mcp`
- Updated all 283 internal import occurrences across 127 `.go` files to match
- Updated the module declarations in the two e2e test submodules (`test/e2e-basic-mcp/agent/go.mod`, `test/e2e-common/broker-driver/go.mod`), which declare their own module paths under the same prefix

## Testing

**Test Environment:**
- Go version: per `go.mod` (1.25.0)
- OS: macOS (local), Ubuntu (CI)

**Tests Performed:**
- [x] `go build ./...` verified clean at repo root and in both e2e submodules
- [x] Full grep confirms zero remaining references to the old module path in any `.go` or `go.mod` file

**Test Coverage:**
- Pure mechanical rename — no logic changed. One pre-existing timing-sensitive test (`TestBrokerClient_RateLimiterIsSharedAcrossProtocols`) failed once under full-suite load but passed 3/3 in isolation; confirmed unrelated to this change (real-clock rate-limiter flake, not caused by the path rename).

## Breaking Changes

**Impact:** Anyone already importing this module under the old `github.com/SolaceDev/solace-broker-mcp` path will need to update their import to `github.com/SolaceProducts/solace-broker-mcp`. No external consumers exist yet, so this is the cheapest point to make this change.

**Migration Guide:** Update import paths from `github.com/SolaceDev/solace-broker-mcp` to `github.com/SolaceProducts/solace-broker-mcp`.

## Related Issues/PRs

- Related to SOL-152415 (repo move to SolaceProducts)
- Blocks SOL-152415

---

**For Reviewers:**

**Focus Areas**: This should be a pure find-and-replace with no functional changes — worth spot-checking that no non-import occurrence (e.g. a string literal or comment referencing the old path) was missed or incorrectly altered.


amitmorade:
Fixing all the issues flagged by copilot

solace-aross (changes requested):
Requesting changes on one item. The rename itself is clean; the build-time version stamp goes silent after it.

## The version stamp breaks, and CI stays green

`internal/version/version.go:7` got its doc comment updated to the new path. The four places that actually pass `-X` did not:

- `Makefile:11`
- `Dockerfile:12`
- `.github/workflows/release.yml:83`
- `docs/internal/packaging-release.md:30`, `:48`, `:67`

Go silently ignores `-X` on a symbol that no longer exists. No error, so nothing downstream catches it. Verified on your branch:

```
main:         make build && ./solace-broker-mcp --version   ->  v0.6.0-10-gf4a81ff
this branch:  make build && ./solace-broker-mcp --version   ->  dev
make check on this branch:  exit 0
```

Where `dev` then surfaces:

| Consumer | Location |
|---|---|
| `--version` output | `cmd/server/main.go:528` |
| `server starting` startup log | `cmd/server/main.go:573` |
| MCP `Implementation.Version` advertised to clients | `cmd/server/main.go:694` |
| SEMP `User-Agent: solace/broker-mcp-server/<v>` | `internal/semp/sempv1/client.go:169`, `sempv2/client.go:327` |

Every binary and image the next release publishes would report `dev`, including in the broker-side logs support reads to identify which server version is talking to a customer's broker.

The fix is the same one-line swap in each file.

One correction to the Copilot comment on `internal/version/version.go`: those builds do not fail. Go accepts an unresolvable `-X` target without complaint, which is why this needs catching here rather than at release time.

## README's install line, worth a decision

`README.md:209` still reads:

```
go install github.com/SolaceDev/solace-broker-mcp/cmd/server@latest
```

After this merges, that produces the `unexpected module path` error your description quotes as the reason for the change. Lines 3, 4, 6, 173, and 309 carry the old org in badges and the clone URL.

If the docs sweep belongs to SOL-152415, reasonable. Line 209 is this PR's stated goal, so my preference is that it travels with the rename.

## Not this PR

The Copilot comment on `test/e2e-common/broker-driver/go.mod` flags a real mismatch (declared path says `e2e-monitoring`, the directory is `e2e-common`), but it predates you. `git show main:test/e2e-common/broker-driver/go.mod` has the same mismatch; this PR only swapped the org segment. Worth a follow-up ticket, not a change here.

The CHANGELOG footer compare links, `.github/{CODEOWNERS,CONTRIBUTING.md,SECURITY.md,ISSUE_TEMPLATE}`, `.fossa.yml`, and `workflow-config.json` all still name the old org. Fine under SOL-152415. `.ua/fingerprints.json` carries 234 old-path strings but it is generated, so regenerate whenever.

## What's good

`make check` exits 0 on the branch, both e2e submodules build clean, and no `.go` file has an old-path import left. The CHANGELOG entry is correctly marked BREAKING, sits under `### Changed` ahead of `### Fixed`, and the migration note matches what the diff does. Holding the diff to Go files plus `go.mod` kept 131 files reviewable.

amitmorade:
Thanks for the thorough review, Andrea! Fixed the version stamp issue — updated the `-X` ldflags target in `Makefile`, `Dockerfile`, and `release.yml`, plus the Docker image's OCI source label. Verified locally: `make build && ./solace-broker-mcp --version` now correctly shows the real version instead of silently falling back to `dev`.

Also updated `README.md` — the `go install` line plus the badges/release-link/clone-URL, since you flagged that one as worth traveling with this PR rather than waiting on the docs sweep.

Left the `e2e-common/broker-driver` go.mod mismatch and the other org references (CHANGELOG footer, CODEOWNERS, `.fossa.yml`, etc.) out of this PR per your note — those'll go under SOL-152415. Pushed, should be ready for another look.

solace-aross (approved):
Approving. The version stamp is fixed, and I verified it end to end rather than reading the diff:

```
make build && ./solace-broker-mcp --version
v0.6.0-17-ga1abbae
```

`dev` before, a real stamp now. `Makefile:11`, `Dockerfile:12` and `release.yml:83` all point at `github.com/SolaceProducts/...`, and the symbol genuinely resolves: `go.mod` declares that module path and `internal/version/version.go` still holds `var version` in package `version`.

README is fully swept, including the `go install` line at 209 that was the point of the change. No `SolaceDev` reference is left in it.

One item still open, not worth holding this up. `docs/internal/packaging-release.md` lines 30, 48 and 67 still document `-X github.com/SolaceDev/...`. That is the same one-line swap you just made three times, and anyone following that runbook for a manual build gets a silently `dev`-stamped binary, which is the failure this PR exists to remove. Worth fixing here if it is quick. If it slips to the docs sweep, keep it distinct from the org-URL cleanup: this one is a symbol path, so it fails silently rather than 404ing.

`build`, `lint` and `Analyze (go)` are green. The five `e2e-*` jobs were still running when I looked, so confirm those before merging. FOSSA is red from the vault outage hitting every PR since roughly 14:26 today, not from anything in this change.

amitmorade:
All actions ran successfully except the fossa check which is acceptable as known issue so merging. Thank you @solace-aross

PR #240, opened 2026-07-31
SOL-152411: let a fork pull request get a real CI verdict

solace-aross:
## What this does

Makes CI able to pass or fail an external contributor's pull request. Today it cannot: seven required checks never report on a fork PR, and the one supply-chain check that does run goes red for a missing secret the contributor could never have.

Closes [SOL-152411](https://sol-jira.atlassian.net/browse/SOL-152411).

## The three problems

**Seven checks never appeared.** `build-and-test.yml` triggered on `push` only, so a fork contributor's push fired in *their* fork and `lint`, `build`, and the five `e2e-*` checks never reported against this repository's commit. Required contexts that never report stay pending forever.

It now also triggers on `pull_request`, and `push` narrows to `main` so same-repo PRs don't run every check twice. I verified none of those jobs needs a secret: the e2e suites pull public images (`solace/solace-pubsub-standard`, `quay.io/keycloak/keycloak`), take broker credentials from committed `.env` files, and generate their own self-signed certs with `openssl`. A read-only `GITHUB_TOKEN` is enough.

**FOSSA went red for a reason no contributor could act on.** GitHub withholds repository secrets from a fork-triggered `pull_request` run, so `secrets.VAULT_URL` arrived empty and the shared workflow exited 1 on its own input validation. `fossa_scan` is now skipped for fork PRs.

**Skipping it is not enough on its own, and this is the part worth reviewing.** The ticket suggested just adding the `if`. That alone would have swapped a red check for a stuck one:

- A skipped **plain** job reports `skipped`, and GitHub counts that as passing.
- A skipped reusable-workflow **caller** is different: the inner job never runs, so the `FOSSA Scan / SCA Scan` context is never created at all. A required context that is never created stays pending forever.

Since `FOSSA Scan / SCA Scan` is the context `ADMIN_SETUP.md` tells admins to require, gating the caller would have made every external contribution unmergeable. Confirmed against live check runs on PR #232, where the already-gated caller in `build-and-test.yml` produced bare `FOSSA Scan` as `skipped` and no ` / SCA Scan` context.

So there's a new `sca_gate` job that always reports and is the context to require. It fails closed on a FOSSA failure **and** on a skip it cannot account for — it re-derives the fork condition itself rather than trusting `skipped`, so narrowing `fossa_scan`'s `if` later can't quietly switch the gate off on same-repo PRs.

## A concurrency block comes with the trigger change

Running on `pull_request` means a same-repo PR can have several runs in flight as
commits land, so this adds a `concurrency` block. Superseded PR runs cancel;
pushes to `main` never do, because every commit on `main` has to keep its own
FOSSA verdict and e2e result.

The non-PR fallback keys on `github.sha`, not `github.ref`, and that detail is
load-bearing. `cancel-in-progress: false` does not mean "let them all run":
GitHub holds at most one pending run per group and cancels the existing pending
run when a new one queues. Keying every push to `main` as `refs/heads/main` would
mean a burst of three merges runs the first, queues the second, and drops the
second when the third arrives, landing that commit on `main` with no verdict at
all. Per-commit groups avoid it without serializing anything. Copilot caught this
on review; the first version of this PR had the `github.ref` bug.

## What I deliberately did not do

No `pull_request_target`, and no `workflow_run` follow-up either. Both run untrusted code with an elevated token to buy a pre-merge scan on a path that already needs a maintainer to approve the workflow run and a code owner to approve the PR.

## Known gap, documented not papered over

A fork PR gets no pre-merge FOSSA verdict. The backstop is the full scan on push to `main` plus a human reading `go.mod`. That's detection after merge, not prevention before it.

The vulnerability half of that gap doesn't actually need a credential — `govulncheck` would run fine as an eighth job on a fork PR. I left it out to keep this change to the trigger and gate problem, and wrote it up in `ADMIN_SETUP.md` as the follow-up rather than leaving the gap looking unavoidable. Happy to file it.

## Also in here

Three things a review pass turned up that belong with this change:

- `transition_on_merge.yaml` would have gone red on every *closed* fork PR — it authenticates to Vault by OIDC, which a fork run cannot do. Same failure class this ticket exists to remove, so acceptance criterion 2 wasn't met without it. Gated to same-repo, which is also semantically right: an external contribution has no SOL ticket to transition.
- `build-and-test.yml`'s workflow-level `permissions` block granted `pull-requests: write`, `id-token: write`, `checks: write`, and `statuses: write` to all eight jobs. Only `fossa_scan` needs any of it. Not a regression from `push: '**'`, but it matters more now that the e2e jobs run shell from the PR's own ref, and `statuses: write` lets a step POST a green status under a required context name. Default is `contents: read`; the elevated scopes moved onto `fossa_scan`.
- Two docs had become false. `ADMIN_SETUP.md` still asserted reusable-workflow jobs "always surface as `<caller> / <inner>`, never as the caller name alone" — the exact behavior this change is built on disproving. `RELEASING.md` still advertised CI "on every branch push" and three e2e suites, marked `[Implemented]`.

## The trade-off to push back on if you disagree

Pushing a feature branch no longer runs CI on its own. Open the PR — a draft is enough — and every check runs.

The alternative was keeping `push: '**'` and gating each job on `github.event_name == 'pull_request' || github.ref_name == main`. That's strictly worse: it still can't run CI for a branch with no PR (the condition is false there), and it adds an all-skipped workflow run to every branch push. Deciding whether a push has an open PR needs an API call, so no declarative trigger preserves pre-PR branch CI.

## Verification

- `make check` — exit 0, 31 packages ok.
- `actionlint` v1.7.7 across all seven workflows — exit 0.
- `sca_gate`'s decision table unit-tested outside CI: `success`→pass, `failure`→fail, `cancelled`→fail, `skipped`+fork→pass, `skipped`+same-repo→**fail**.
- Claim (a) above confirmed against live check-run data on PR #232, not inferred.
- `workflow_call` retained, so `release.yml`'s tag path still works; `fossa_scan`'s `if` still evaluates false there, unchanged.

**Not verifiable from here:** no fork PR has ever been opened against this repo. `SCA gate` has no live run behind it. `ADMIN_SETUP.md` says so and lists what to confirm against a real external contribution before treating any of it as a gate.

No CHANGELOG entry — CI-only, outside the production surface `.github/scripts/production-surface.sh` defines.

### DIFF
--- a/.github/ADMIN_SETUP.md
+++ b/.github/ADMIN_SETUP.md
@@ -180,7 +180,9 @@ review.
 list is not: it currently holds `lint`, `build`, and `FOSSA Scan`, the last of
 which enforces nothing (see the warning below). Replace it with the following.
 These names are what GitHub actually reports, confirmed against the check runs on
-PRs #213, #216, and #217:
+PRs #213, #216, and #217 — except `SCA gate`, which SOL-152411 added and which has
+no live run behind it yet. Confirm that one against a real pull request before
+you rely on it:
 
 | Check | Source | Gates on |
 |-------|--------|----------|
@@ -191,21 +193,38 @@ PRs #213, #216, and #217:
 | `e2e-monitoring` | `build-and-test.yml` | E2E suite (known flaky fixture; rerun the job before investigating) |
 | `e2e-management` | `build-and-test.yml` | E2E suite |
 | `e2e-action` | `build-and-test.yml` | E2E suite |
-| `FOSSA Scan / SCA Scan` | `ci-pr.yaml` job `fossa_scan` | Licensing policy and critical/high vulnerabilities, diffed against the base branch |
+| `SCA gate` | `ci-pr.yaml` job `sca_gate` | The FOSSA verdict, or an accounted-for reason there is none. **Not** `FOSSA Scan / SCA Scan`; see the warning below |
 | `CHANGELOG updated` | `ci-pr.yaml` job `changelog` | Advisory today; see note below |
 | `DCO sign-off` | `dco.yaml` job `dco` | a sign-off on every commit the PR adds |
 | `DCO check self-test` | `ci-pr.yaml` job `dco_selftest` | the gate's own logic still working |
 
-⚠️ **Do not select `FOSSA Scan`.** It looks like the right entry and is not. A
-check by that exact name comes from `build-and-test.yml`, where the job carries
-`if: github.event_name == 'push' && github.ref_name == github.event.repository.default_branch`,
-so on every pull request it reports **skipped**, and GitHub counts a skipped
-check as passing. Requiring `FOSSA Scan` therefore enforces nothing. The context that
-gates is `FOSSA Scan / SCA Scan`: the `ci-pr.yaml` caller job (`name: FOSSA Scan`)
-plus the inner job of the reusable workflow it calls (`name: "SCA Scan"`), which
-exits non-zero on findings because `.github/workflow-config.json` sets both FOSSA
-modes to `BLOCK`. Reusable-workflow jobs always surface as
-`<caller job name> / <inner job name>`, never as the caller name alone.
+⚠️ **Two FOSSA-shaped entries are wrong, and the right one is neither.** Pick
+`SCA gate`.
+
+- **`FOSSA Scan` enforces nothing.** A check by that exact name comes from
+  `build-and-test.yml`, where the job carries
+  `if: github.event_name == 'push' && github.ref_name == github.event.repository.default_branch`,
+  so on every pull request it reports **skipped**, and GitHub counts a skipped
+  check as passing.
+- **`FOSSA Scan / SCA Scan` deadlocks fork pull requests.** It is the real
+  scanning context on a same-repo pull request: the `ci-pr.yaml` caller job
+  (`name: FOSSA Scan`) plus the inner job of the reusable workflow it calls
+  (`name: "SCA Scan"`), which exits non-zero on findings because
+  `.github/workflow-config.json` sets both FOSSA modes to `BLOCK`. But that caller
+  is now skipped for fork pull requests, because a fork gets no repository
+  secrets and the Vault-backed scan cannot run. A skipped *plain* job still
+  reports `skipped`; a skipped reusable-workflow *caller* never creates the
+  ` / SCA Scan` context at all. Required contexts that are never created stay
+  pending forever, so requiring this one makes every external contribution
+  unmergeable.
+- **`SCA gate` is the one to require.** `ci-pr.yaml` job `sca_gate` always
+  reports. It passes on a FOSSA success, fails on a FOSSA failure, and passes a
+  fork's skip with a logged reason. It re-derives the fork condition itself and
+  **fails** a skip it cannot account for, so narrowing `fossa_scan`'s `if` later
+  cannot quietly switch the gate off on same-repo pull requests.
+
+Reusable-workflow jobs always surface as `<caller job name> / <inner job name>`,
+never as the caller name alone.
 
 ⚠️ **Both DCO rows are required, and dropping either removes a control.** DCO
 stands in for a contributor licence agreement, so the repository is not covered
@@ -250,10 +269,11 @@ Four more notes on the list:
   `CHANGELOG_GATE_MODE: advisory`, so the script warns and exits 0. Requiring it
   makes the check's presence a merge precondition now, so flipping the mode to
   `blocking` later needs no ruleset change.
-- On a pull request, FOSSA runs in diff mode (`enable_diff_mode` is true for
-  `pull_request` events). It blocks on findings that are new relative to the base
-  branch, not on the full dependency inventory. That is what keeps `main` from
-  entering an irregular state; do not read a green PR as a clean full scan.
+- On a **same-repo** pull request, FOSSA runs in diff mode (`enable_diff_mode` is
+  true for `pull_request` events). It blocks on findings that are new relative to
+  the base branch, not on the full dependency inventory. That is what keeps `main`
+  from entering an irregular state; do not read a green PR as a clean full scan.
+  On a fork pull request it does not run at all — see the fork section below.
 - `Analyze (go)` (CodeQL) passes on every PR. Add it if you want code scanning to
   block merges, after settling the configuration question in the Security Settings
   section. It is not in the list above because that is a policy call, not a
@@ -262,25 +282,47 @@ Four more notes on the list:
   `transition_on_merge.yaml`, which runs on `pull_request: closed`, and the check
   name varies with the Jira key (e.g. `... Vault and JIRA Operations (SOL-152328)`).
 
-🚨 **CI cannot currently green-light a fork pull request.** Two separate problems,
-both of which need a CI change. Tracked separately.
+**How a fork pull request gets a verdict.** SOL-152411 fixed the two problems that
+made this impossible. What it changed, and what it deliberately did not:
 
-1. **Seven checks never appear.** `build-and-test.yml` triggers on `push`, not
-   `pull_request`. A fork contributor's push fires in their fork, so `lint`,
-   `build`, and the five `e2e-*` checks never report on this repository's commit.
-   Those required contexts stay pending forever.
-2. **`FOSSA Scan / SCA Scan` fails.** `ci-pr.yaml` passes `use_vault: true` and
-   `secrets.VAULT_URL`. GitHub withholds secrets from `pull_request` runs
-   triggered by a fork, so the reusable workflow finds an empty Vault URL and
-   exits 1. It runs, and it goes red.
+1. **The seven `build-and-test.yml` checks now report.** That workflow used to
+   trigger on `push` only, so a fork contributor's push fired in their fork and
+   `lint`, `build`, and the five `e2e-*` checks never reported on this
+   repository's commit — required contexts pending forever. It now also triggers
+   on `pull_request`. None of those jobs reads a secret: the e2e suites pull
+   public images (`solace/solace-pubsub-standard`, `quay.io/keycloak/keycloak`),
+   take their broker credentials from committed `.env` files, and generate their
+   own self-signed TLS certs with `openssl`. A read-only `GITHUB_TOKEN` is enough.
+2. **FOSSA no longer goes red for a missing secret.** `ci-pr.yaml` passes
+   `use_vault: true` and `secrets.VAULT_URL`, and GitHub withholds secrets from a
+   fork-triggered `pull_request` run, so the reusable workflow used to find an
+   empty Vault URL and exit 1 on every external contribution. The `fossa_scan`
+   job is now skipped for fork pull requests, and the always-reporting `SCA gate`
+   job is the required context instead.
+3. **`pull_request_target` was not introduced**, and neither was a `workflow_run`
+   follow-up. Both would run untrusted code with an elevated token to buy a
+   pre-merge scan on a path that already requires a maintainer to approve the
+   workflow run and a code owner to approve the pull request.
 
-Nobody has ever opened a fork pull request against this repository, so nothing
-below is observed. `CHANGELOG updated` should be fine: `ci-pr.yaml` gives that job
-only `contents: read` and it reads no secrets. `Analyze (go)` appears to come from
-CodeQL default setup, which the Security Settings section asks you to confirm.
-Copilot review is inconsistent even on same-repo pull requests, so do not count on
-it. Verify `CHANGELOG updated`, `Analyze (go)`, and `FOSSA Scan / SCA Scan`
-against a real fork pull request before you treat any of them as a gate.
+**The residual gap, stated so nobody assumes otherwise.** A fork pull request gets
+no pre-merge FOSSA verdict. A contributor who adds a dependency with a licensing
+conflict or a critical/high vulnerability is caught by the full scan on push to
+`main`, which is detection after merge, not prevention before it. Two things close
+that gap and both are human: read `go.mod` and `go.sum` in any fork pull request
+that touches them, and treat a red FOSSA on `main` as a revert, not a backlog item.
+
+Also worth knowing: on a same-repo pull request FOSSA runs in diff mode, so a green
+`SCA gate` there is not a clean full scan either — see the diff-mode note above.
+
+Nobody has ever opened a fork pull request against this repository, so none of the
+above is observed on a real external contribution. `CHANGELOG updated` should be
+fine: `ci-pr.yaml` gives that job only `contents: read` and it reads no secrets.
+`Analyze (go)` appears to come from CodeQL default setup, which the Security
+Settings section asks you to confirm. Copilot review is inconsistent even on
+same-repo pull requests, so do not count on it. Verify `SCA gate`,
+`CHANGELOG updated`, `Analyze (go)`, and the `lint` job's annotation behavior on a
+read-only token against a real fork pull request before treating any of them as a
+gate.
 
 **A third problem, and this one we create deliberately.** The GitHub Actions
 Permissions section below tells you to require approval for all outside
@@ -288,9 +330,11 @@ collaborators. That holds the entire workflow run, not just one job, until a
 maintainer approves it.
 
 The consequence, in order: an external contributor opens their first pull request,
-and **no checks run at all**. Not `FOSSA Scan / SCA Scan`, not `CHANGELOG updated`,
-not the seven that were already broken. Every required context sits pending and the
-pull request reads as stuck rather than as rejected or passing.
+and **no checks run at all**. Not `SCA gate`, not `CHANGELOG updated`, not the seven
+from `build-and-test.yml`. Every required context sits pending and the pull request
+reads as stuck rather than as rejected or passing. This is independent of the
+SOL-152411 fix: that made the checks *capable* of reporting on a fork pull request,
+and this setting decides *when* they are allowed to start.
 
 That is expected, not broken. Maintainers need to know it, or the first community
 contribution gets triaged as a CI outage. Two things follow:
@@ -303,8 +347,10 @@ This also constrains the required-check list. Every context you require is a
 context that reports only after a human approves the run, so requiring more of
 them lengthens the stall rather than tightening the gate.
 
-Until the CI fix lands, merge community contributions by pushing the branch into
-this repository yourself.
+Copying a community branch into this repository to test it is no longer necessary,
+and is now the worse option: a same-repo branch is a trusted context, so doing it
+runs the contributor's code with this repository's secrets. Approve the fork run
+instead.
 
 ---
 
@@ -373,9 +419,11 @@ writing; re-check, do not assume.
 - ✅ All secrets removed from git history (`git log --all -S "password"`)
 - ✅ `.gitignore` properly configured (`.env`, `broker-config.yaml`)
 - ✅ No sensitive data in issues or PRs
-- ⬜ **Required status checks corrected** per the Branch Protection section above,
-  including the `FOSSA Scan / SCA Scan` swap. The ruleset still holds the old
-  list.
+- ⬜ **Required status checks corrected** per the Branch Protection section above.
+  The ruleset still holds the old list. The supply-chain context to register is
+  `SCA gate` — not `FOSSA Scan`, which enforces nothing, and not
+  `FOSSA Scan / SCA Scan`, which never reports on a fork pull request and would
+  leave every external contribution pending.
 - ⬜ **"Allow GitHub Actions to create and approve pull requests" turned off.**
   Still on.
 - ⬜ **Fork pull request workflows set to require approval for all outside
--- a/.github/ADMIN_SETUP.md
+++ b/.github/ADMIN_SETUP.md
@@ -463,31 +463,60 @@ writing; re-check, do not assume.
   > today, which is the only reason a repo-wide FOSSA outage is not already
   > blocking everyone.
   >
-  > Check it, do not assume it. The scan runs almost entirely outside this
-  > repository: `ci-pr.yaml` calls
+  > Check it, do not assume it, and note that our SHA pin buys less than it looks
+  > like. `ci-pr.yaml` calls
   > `SolaceDev/solace-public-workflows/.github/workflows/sca-scan-and-guard.yaml`
-  > at a pinned SHA, and that workflow in turn pulls a nested action from `@main`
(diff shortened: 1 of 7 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -1,29 +1,89 @@
...
-# Minimal token permissions — read-only access to repo contents.
-# golangci-lint-action uses checks: write to post inline annotations on PRs.
-# FOSSA scan requires id-token, pull-requests, actions, packages, and statuses.
+# Cancel superseded pull-request runs, never superseded pushes to main: every
+# commit on main must keep its own FOSSA verdict and e2e result. `github.ref` is
+# the fallback for the push and `workflow_call` (tag) paths, where
+# `pull_request.number` is empty.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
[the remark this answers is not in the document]
solace-aross: Agreed, and the consequence is worse than queueing. GitHub holds at most one pending run per concurrency group and cancels the existing pending run when a new one queues, so three merges in quick succession run the first, queue the second, then drop the second when the third arrives. That commit lands on `main` with no FOSSA verdict and no e2e result, which is exactly what the comment above the block claims to protect.

Fixed in 2142b84 by keying the non-pull-request path on `github.sha`, as you suggested. `queue: max` would also close it, at the cost of serializing `main` behind the longest e2e run, so per-commit groups win on both counts.

One bounded cost, documented in the comment: tagging a commit that is already running as a `main` push puts both in the same group. That costs the release run a wait, not a verdict, since nothing is dropped until a third run queues.

bczoma (changes requested):
My particular concern is that before, we had run `build-and-test` by each push to any branch, so developers working on local branches received feedback before putting up a PR

See details/proposal in the next comment

.github/workflows/build-and-test.yml
```diff
@@ -1,29 +1,105 @@
...
+# fork, never against this repository's commit.
+#
+# `pull_request` fixes that, but keeping `push: '**'` alongside it would run
+# every check twice on same-repo pull requests — including five Dockerized e2e
+# suites. The alternative considered was keeping `push: '**'` and gating each job
+# on `github.event_name == 'pull_request' || github.ref_name == <default>`. That
+# is strictly worse: it still cannot run CI for a branch with no pull request
+# (the condition is false there), and it adds an all-skipped workflow run to
+# every branch push. Deciding whether a push has an open pull request needs an
+# API call, so no declarative trigger preserves pre-pull-request branch CI.
```
bczoma: Please rework this paragraph. It closes the question on the wrong alternative, and the conclusion is right for a reason that is not stated.

The alternative evaluated above is gating on `github.event_name == 'pull_request' || github.ref_name == <default>`, and you are correct that it is strictly worse. But that is not the option a reviewer reaches for. The obvious one keeps `push: '**'`, adds `pull_request`, and gates on **fork-ness** rather than event name:

```yaml
if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```

That preserves pre-PR branch CI, declaratively, with no API call:

- push to any branch runs, so pre-PR feedback survives
- a same-repo PR all-skips the `pull_request` run cheaply, and the required contexts are satisfied by the push run's check runs on the same head SHA — exactly how the seven contexts worked before this PR
- a fork PR executes the `pull_request` run, so the contexts report for the first time

So "no declarative trigger preserves pre-pull-request branch CI" is too strong as written, and anyone who spots this option will read the comment as having missed it.

The reason to reject it anyway is better than the one given: it puts **two check runs with the same name on the same SHA** — a real `lint` from the push run and a skipped `lint` from the PR run. If branch protection resolves a required context against the most recent check run of that name, the gate becomes order-dependent, and a skipped run landing last is a green gate over a red job. For a design whose whole thesis is GitHub's skip/report asymmetries, that is the argument worth making here. It is the one that actually rules the option out.

.github/workflows/build-and-test.yml
```diff
@@ -1,29 +1,105 @@
...
+#
+# The cost, stated plainly: pushing a feature branch no longer runs CI on its
+# own. Open the pull request — a draft is enough, `opened` and `synchronize` both
+# fire for drafts — and every check runs.
+#
+# `workflow_call` stays because release.yml calls this workflow on a `v*` tag
+# push. A `push` filter that names `branches` never matches a tag, so the
+# `tags-ignore: ['v*']` that used to sit here was already dead config and is
```
bczoma: This is inverted. GitHub's rule is:

> If you define only `tags`/`tags-ignore` or only `branches`/`branches-ignore`, the workflow won't run for events affecting the undefined Git ref.

The old config defined **both** `branches` and `tags-ignore`, so the tag ref was not undefined — a non-`v*` tag push *did* trigger this workflow, and `tags-ignore: ['v*']` was load-bearing in keeping `v*` tags out. It was not dead config.

The new behavior is still what you want: with `branches` as the only filter, no tag push triggers this workflow at all. But that follows from removing `tags-ignore` *and* narrowing `branches` together, not from a filter that never did anything. Given this PR's thesis is getting GitHub's filtering and reporting asymmetries exactly right, please correct the mechanism rather than leave it for someone to inherit.

.github/workflows/build-and-test.yml
```diff
@@ -1,29 +1,105 @@
...
+      - main
+  pull_request:
+    types: [opened, synchronize, reopened]
+    # Scoped to main to match dco.yaml, which is scoped for a correctness reason
+    # (see its header). Without this, a pull request against a throwaway base
+    # would spin up five Dockerized e2e suites and still not get a DCO verdict.
+    # Add a branch here if the project starts taking pull requests against
+    # release branches — and add it to dco.yaml at the same time.
+    branches: [main]
   workflow_call:
```
bczoma, line 65: Add `workflow_dispatch:` here. See the review summary — one line, restores running the full suite on a branch with no PR, matches the existing pattern in `fossa-scan.yaml` and `llm-eval.yml`, and directly answers the trade-off the description flags for pushback. Worth a line in the header comment too, so the "the cost, stated plainly" paragraph points at the escape hatch instead of leaving the reader with only the draft-PR route.

.github/ADMIN_SETUP.md
```diff
@@ -373,9 +448,11 @@ writing; re-check, do not assume.
 - ✅ All secrets removed from git history (`git log --all -S "password"`)
 - ✅ `.gitignore` properly configured (`.env`, `broker-config.yaml`)
 - ✅ No sensitive data in issues or PRs
-- ⬜ **Required status checks corrected** per the Branch Protection section above,
-  including the `FOSSA Scan / SCA Scan` swap. The ruleset still holds the old
-  list.
+- ⬜ **Required status checks corrected** per the Branch Protection section above.
```
bczoma, line 451: This checklist item needs a precondition. FOSSA is currently failing on every PR in this repo for a reason unrelated to your change: `.github/workflow-config.json` has no `secrets.vault.url`, the shared workflow reads the URL from that config file, and it logs `Vault URL: <not set>` and exits before scanning. Runs on 2026-07-30 were green, so this started today; the pinned shared-workflow SHA is unchanged here.

Branch protection currently has zero required contexts, so nothing is blocked today. But an admin who follows this line before the Vault URL is fixed makes `SCA gate` required while `fossa_scan` fails on every same-repo PR — and `sca_gate` correctly fails a `failure` result — so every PR in the repo becomes unmergeable. Please say so here, or fix the FOSSA config first.

To be clear, the gate behaving this way is correct. It is surfacing real breakage that leaving `FOSSA Scan / SCA Scan` unrequired had been hiding.

bczoma (changes requested):
Requesting changes. The fork-CI mechanics are sound and I checked the checkable parts (list at the end) — the objections are one missing trigger and two inaccurate claims that will mislead whoever touches this next.

## Required

**1. Add `workflow_dispatch:` to `build-and-test.yml`.** The PR is upfront that pushing a feature branch no longer runs CI, and frames it as the trade-off to push back on. Here is the pushback: it is largely avoidable, for one line.

```yaml
on:
  push:
    branches: [main]
  pull_request:
    types: [opened, synchronize, reopened]
    branches: [main]
  workflow_dispatch:
  workflow_call:
```

That gives `gh workflow run build-and-test.yml --ref my-branch` — the full seven-job suite on a branch with no PR. It is already the house pattern (`fossa-scan.yaml` and `llm-eval.yml` are both `workflow_dispatch` with a ref input), it costs nothing on security since dispatch requires write access, and `fossa_scan`'s `if` correctly evaluates false on a dispatch.

With that in, I am content with the trade. Local coverage is good (`make check`, a `make e2e-*-all` per suite) and the draft-PR path works — the claim that `opened` and `synchronize` fire for drafts is correct. Without it, the only way to get CI on a branch is to open a PR, which is a real regression in the daily loop for a change whose purpose is elsewhere.

**2. Fix the alternative the header comment rejects.** Inline. Right conclusion, wrong reason, and the reason given is refutable — which matters in a comment block whose job is to stop the next person reverting the trigger.

**3. Fix the `tags-ignore` claim.** Inline. The resulting behavior is correct; the stated mechanism is backwards.

## Not blocking, but do not merge the admin advice without it

`SCA gate` is red on this PR for a pre-existing reason. FOSSA is failing repo-wide:

```
- Vault URL: <not set>
LICENSING_OUTCOME: skipped   VULNERABILITY_OUTCOME: skipped   FOSSA_PROJECT_ID: (empty)
```

`.github/workflow-config.json` carries `secret_path` and `role` but no `secrets.vault.url`, and the shared workflow reads the URL from that config file. Every `ci-pr.yaml` run since roughly 14:26 today has failed identically; runs on 2026-07-30 were green and the pinned shared-workflow SHA is unchanged here. Not a regression from this PR — and `sca_gate` surfacing it is a point in the gate's favor. See the inline note on the checklist item for why it still needs a precondition.

## Checked and correct

- No `ghcr.io` images anywhere in `test/`, so dropping workflow-level `packages: read` is safe.
- `golangci-lint-action@v7` with no `only-new-issues`, so `contents: read` alone suffices. The claim that the old `checks: write` comment was wrong holds.
- `release.yml` has no `concurrency` block, so the new group in the called workflow cannot deadlock the tag path.
- `sca_gate`'s decision table is sound — `cancelled` and anything unexpected fall to `*)` and fail, and re-deriving `IS_FORK_PR` rather than trusting `skipped` does close the trap it claims to close.
- No `dependabot.yml` here, only `renovate.json` scoped to npm, so the Dependabot-gets-no-secrets variant does not apply. Worth knowing if one lands: `sca_gate` would fail those PRs as an unexplained same-repo skip.

.github/workflows/build-and-test.yml
solace-aross: Added in 9f0a689, and the header now points at it instead of leaving the draft-PR route as the only option.

I checked the two ways this could have broken a gate, since the whole point of the PR is that skips report inconsistently. `fossa_scan` in this file already requires `github.event_name == 'push'`, so it skips on a dispatch. `sca_gate` is in `ci-pr.yaml` behind a `pull_request` trigger, so a dispatch never creates it and there is no unexplained-skip path to trip. Fair call on the trade being avoidable for one line.

.github/workflows/build-and-test.yml
solace-aross: Reworked in 9f0a689. You are right that the paragraph closed the question on an option nobody would propose, and that "no declarative trigger preserves pre-pull-request branch CI" was too strong.

The header now names the fork-ness variant, states plainly that it does preserve pre-PR branch CI with no API call, and rejects it on the duplicate-context argument instead: a real `lint` and a skipped `lint` on the same head SHA make a required context order-dependent, and a skipped run landing last is a green gate over a red job. That is the argument that actually rules it out, and it is the same asymmetry the rest of the file is built on.

.github/workflows/build-and-test.yml
solace-aross: Corrected in 9f0a689. I checked the old trigger on `main` before rewriting: it defined `branches: ['**']` and `tags-ignore: ['v*']` together, so the tag ref was not undefined and a non-`v*` tag push did run this workflow. `tags-ignore` was load-bearing and calling it dead config was wrong.

The comment now says what actually stops tag pushes: narrowing `branches` to main and dropping `tags-ignore` together, because a filter list naming only `branches` leaves the tag ref undefined.

.github/ADMIN_SETUP.md
solace-aross: Precondition added in 9f0a689, and agreed on the risk: registering `SCA gate` while FOSSA fails repo-wide would make every pull request unmergeable.

One correction on the cause, because it changes who fixes it. The missing `secrets.vault.url` is real but is not what breaks the run. `ci-pr.yaml` passes `VAULT_URL: ${{ secrets.VAULT_URL }}` explicitly, the shared workflow prefers that over the config file, and validation passes:

```
15:27:37.52   SECRET_VAULT_URL: ***
15:27:37.52   CONFIG_VAULT_URL:
15:27:37.53   Vault configuration validated
15:27:37.53      URL: ***
15:27:37.53      Role: cicd-workflows-secret-read-role
```

The `- Vault URL: <not set>` line you quoted is the config-file parser echoing which keys the file carries, at 15:27:37.42, and that step's outcome is `success`. The scan dies half a second after validation, at the token exchange:

```
15:27:37.92 ##[error]failed to retrieve vault token. code: ERR_NON_2XX_3XX_RESPONSE,
message: Response code 400 (Bad Request), vaultResponse:
{"errors":["error validating claims: claim \"repository\" does not match any associated bound claim values"]}
```

The role `cicd-workflows-secret-read-role` binds on the `repository` claim, and the move to `SolaceProducts` changed it. So the fix is Vault-side, adding `SolaceProducts/solace-broker-mcp` to the role's bound claims, and no change to `workflow-config.json` will clear it. Byte-identical failure on #239 and #236; every run from today fails and yesterday's were green, which matches the org move rather than the pinned SHA.

The ADMIN_SETUP note records that as the cause and flags two things worth clearing at the same time without being the cause: `project_id` is still `SolaceDev_solace-broker-mcp`, which will misfile scans once auth works, and the absent config URL.

Agreed on the last point, and it is the part I care most about keeping: the gate is doing its job here. Leaving `FOSSA Scan / SCA Scan` unrequired is what had been hiding this.

solace-aross:
All three required items are in 9f0a689, rebased onto the module-path rename.

- `workflow_dispatch:` added, plus a header line pointing at it. Verified it cannot trip a gate: `fossa_scan` here requires `event_name == 'push'`, and `sca_gate` is in `ci-pr.yaml` behind a `pull_request` trigger, so a dispatch never creates it.
- The rejected-alternative paragraph now names the fork-ness variant, concedes it does preserve pre-PR branch CI declaratively, and rejects it on duplicate same-name check runs making a required context order-dependent.
- The `tags-ignore` mechanism is corrected. I checked `main`'s old trigger: `branches: ['**']` and `tags-ignore: ['v*']` were both defined, so it was load-bearing.
- ADMIN_SETUP checklist item carries the precondition.

One disagreement, in the thread on that last item. The missing `secrets.vault.url` is not the cause; `VAULT_URL` comes from a repository secret and validation passes. The run fails at the Vault token exchange on a `repository` bound-claim mismatch from the org move, so the fix is Vault-side. Log timestamps are in the thread.

Verified: `make check` exit 0, 31 packages ok. `actionlint` v1.7.7 across all workflows, exit 0. CI is green except the FOSSA chain.

.github/ADMIN_SETUP.md
solace-aross: Update, and it cuts against the cause I gave you an hour ago. The precondition is now written not to depend on any single cause (cd93d2c).

Vault was fixed org-side around 15:27Z. FOSSA authenticates fine now and fails later, pulling the scan container:

```
Unable to find image 'ghcr.io/solaceproducts/maas-build-actions@sha256:14d7b08fb346c895850c18304e790d5a93fb37655dba48d604bb8eedf4b055b6'
docker: Error response from daemon: manifest unknown
```

Identical on #240, #242 and #243, so it is still repo-wide and still not from any of those diffs. #243 repoints the FOSSA project id and is worth landing on its own, but it does not clear this.

The part worth keeping: we pin the shared workflow by SHA, but that workflow pulls a nested action from `@main` and runs its container by digest. Neither is frozen by our pin, so the scan's behaviour changes without a line moving here. Two unrelated FOSSA failures in one afternoon, with a green window between them, is the evidence. So the checklist item now says to confirm `SCA gate` green against a real pull request rather than naming a cause that goes stale, and it records both failures as history.

solace-lootii (dismissed):
Approving. Verified the trickier claims rather than taking them on faith — traced
the actual GitHub Actions skip semantics (a skipped reusable-workflow *caller*
never creates the inner job's context vs. a skipped plain job reporting
`skipped`/counted-as-passing), and confirmed `sca_gate`'s decision table is
implemented exactly as described: `success`→pass, `failure`/`cancelled`→fail via
the catch-all, `skipped`+fork→pass, `skipped`+same-repo→fail.

The concurrency-group reasoning (`github.sha` fallback instead of `github.ref`)
checks out too — walked through why per-commit groups avoid dropping the middle
commit in a merge burst, and the acknowledged tag-vs-push collision on
`workflow_call` is genuinely benign (a wait, not a dropped run).

One non-blocking nitpick: `ci-pr.yaml`'s workflow-level `permissions:` block
(pre-existing, untouched here) stays elevated by default rather than following
the same "read-only default, elevate per-job" pattern this PR just established
in `build-and-test.yml`. Not exploitable via a fork — GitHub caps fork tokens to
read-only regardless — so the exposure is limited to a compromised same-repo
branch, same residual class of risk `sca_gate`'s own comment already accepts.
Worth a follow-up, not a blocker here.

LGTM.

.github/ADMIN_SETUP.md
solace-aross: Correcting myself on two things, and withdrawing one concession I should not have made.

**First, I overcorrected in my last message.** I opened it with "this cuts against the cause I gave you an hour ago". It does not. The Vault bound-claim rejection was the cause of the failure you were looking at, and it was fixed org-side. A second, later failure with a different mechanism is not a correction to the first diagnosis. I framed a new event as an error on my part and that muddied the record.

**Withdrawing the `secrets.vault.url` concession.** I said setting it would be tidiness. That was wrong and it pointed the wrong way. The key used to hold `https://vault.maas-vault-prod.solace.cloud:8200` and was removed deliberately in `2d46757`, whose subject is "Vault URL hardening", while preparing this repository to go public. Re-adding it would put an internal hostname back into a repo that is about to be public. `ADMIN_SETUP.md` now says not to add it.

Worth adding: its absence cannot have caused anything. The key has been gone since #187, so if it were causal FOSSA would have been red since #187 rather than since 07-31. Your own observation that 07-30 was green rules it out.

**Second, and this is the part that matters for your checklist objection: `SCA gate` is not one infra fix away from green, and the next blocker is ours.** I had this wrong too, in the direction of being too quick to call everything infra.

PR #243 repointed `project_id` to `SolaceProducts_solace-broker-mcp`, which is a new FOSSA project carrying none of the old project's policy waivers. On `main` at `43a9a93`, `fossa test` already reports:

```
⚑ Denied by policy CC-BY-SA-4.0 on github.com/perimeterx/marshmallow@v1.1.5
⚑ MPL-2.0 license detected in github.com/hashicorp/go-cleanhttp@v0.5.2
⚑ MPL-2.0 license detected in github.com/hashicorp/go-retryablehttp@v0.7.8
Error: The scan has revealed issues. Number of issues found: 3
```

`block_on` includes `policy_conflict` and the guard maps "Denied by Policy" to it, so the `marshmallow` finding should block as soon as the guard can run. It is invisible today only because both guard steps die on the container pull before they can report, and the job's verdict is read from those steps. Which also means the `critical,high severity vulnerabilities detected` line in those runs is not a finding at all; nothing was scanned. Worth flagging upstream, since the shared workflow cannot tell "guard could not run" from "guard found violations".

An admin following my original checklist wording would have hit that after the infra fix landed, with the gate already required. That is exactly the outcome the item exists to prevent, so your objection was right even though the cause you named was not.

`1a59969` fixes the section: removes a false "green window" claim (FOSSA never went green on 07-31; the green `Build and Test` runs in that window had the caller skipped), corrects the timings, stops calling the two failures unrelated when they share one root cause, and adds the licensing row with the evidence above. It also states that our SHA pin freezes less than it appears to: the shared workflow references four nested actions at `@main` and derives the container registry namespace from `github.repository_owner` at runtime.

solace-lootii (dismissed):

bczoma (dismissed):
Looks good now!

solace-aross:
Re-approval needed, and I am sorry for the churn. One push, `48111bf`, containing two things:

1. **Balazs's ask.** CONTRIBUTING.md now has a "Getting CI to run on a branch" section: `make check` / `make e2e-all` locally, `gh workflow run build-and-test.yml --ref my-branch` for the full suite with no PR, or open a draft PR. It notes dispatch needs write access, so it is a maintainer route.
2. **Rebase onto `main`**, which was required anyway.

The rebase alone would have dismissed both approvals, so the note is bundled in rather than costing a second round. Nothing else changed.

Verified: `make check` exit 0, 31 packages, 0 failures. `actionlint` v1.7.7 exit 0. Both documented `make` targets confirmed present in the Makefile.

Worth flagging, because it is why this is awkward: the ruleset combines `strict_required_status_checks_policy` with `dismiss_stale_reviews_on_push`, and this repository has both `allow_update_branch` and `allow_auto_merge` disabled. So the only way to satisfy "up to date" is a push, and every push clears all approvals. With `main` taking merges every few minutes that can loop. Worth a follow-up to enable update-branch or auto-merge; not something to fix in this PR.

bczoma (approved):
Approving.

Not gating, consider following for clarity:

.github/ADMIN_SETUP.md, line 462 (at PR head 48111bf) — inside the Precondition: confirm SCA gate is green... .

Old sentence (wrong):

▎ Branch protection holds zero required contexts today, which is the only reason a repo-wide FOSSA outage is not already blocking everyone.

Corrected sentence:

▎ The main-protection ruleset requires lint, build, and FOSSA Scan today. The FOSSA outage is not blocking merges because the required context is FOSSA Scan — the caller job that reports skipped on every pull request and therefore enforces nothing, exactly as the warning above describes. The real scan, FOSSA Scan / SCA Scan, is failing and is not required. Registering SCA gate while FOSSA is broken replaces a vacuous green with a hard red on every pull request.

Why the original is wrong: it reads the legacy /branches/main/protection API, which returns an empty context list because the rules live in a ruleset. main-protection carries lint, build, and FOSSA Scan as required, with strict_required_status_checks_policy: true.

solace-lootii (approved):

PR #242, opened 2026-07-31
SOL-152414: make the licence artifacts match the binary, and gate the drift

solace-aross:
> **Stacked on #240.** Base is `aross/SOL-152411-fork-pr-ci`, not `main`, because both touch `ci-pr.yaml`, `ADMIN_SETUP.md`, and `RELEASING.md`. Review #240 first; GitHub retargets this to `main` when that merges. Read the diff against the base branch, not against `main`.

## What this does

`THIRD_PARTY_LICENSES.md` is the OSS compliance artifact that satisfies a Legal release condition. It had drifted from the binary it describes, `NOTICE` was incomplete, and neither file actually shipped with the release. Fixed all three, and replaced the sentence that was supposed to prevent recurrence with a check that does.

Closes [SOL-152414](https://sol-jira.atlassian.net/browse/SOL-152414).

## The drift

Three rows named components no longer in the binary: `maypok86/otter` v1.2.4 (replaced by `/v2`), plus `dolthub/maphash` and `gammazero/deque`, which were Otter v1 transitive deps.

Five components ship and were missing. Two the ticket predicted — `sony/gobreaker/v2` v2.4.0 (MIT) and `maypok86/otter/v2` v2.3.0 (Apache-2.0), both licences read off the generator rather than assumed. Three it did not: `stretchr/testify` v1.11.1 (MIT), `davecgh/go-spew` v1.1.1 (ISC), `pmezard/go-difflib` v1.0.0 (BSD-3-Clause).

**Why a test library is in a binary inventory**, since it looks like a mistake and someone will otherwise delete it: `otter/v2` v2.3.0 ships a file called `issue_test_1.25.go`. That name does not end in `_test.go`, so Go compiles it as ordinary package code, and its `testify/require` import pulls all three in. Confirmed from `go-licenses` and `go list -deps`; the reviewer went further and found 29 linked symbols from those modules in the built binary via `go tool nm`. All permissive, so this is accuracy, not licensing. The file says so, and says not to remove them.

## Two things the ticket did not ask about

**`NOTICE` was incomplete.** It omitted `github.com/oasdiff/yaml3`, a fork of `gopkg.in/yaml.v3` carrying the same Canonical attribution. It also paraphrased the CoreOS attribution rather than reproducing it, dropping the "This product includes software developed at" sentence, and gave Canonical's years as 2011-2019 (that is the LICENSE's range, not the NOTICE's). Apache-2.0 §4(d) asks for the notices *contained in* the dependency's NOTICE, so both now quote upstream verbatim.

**Neither artifact actually shipped.** This is the one I would most like a second opinion on, because it goes beyond the ticket. The release tarball carried `LICENSE` only. The container image carried none of the three. So every published artifact was missing the notices §4(d) requires travel with it, while the file in the repo described itself as "accompanying the release". The tarball now carries all three; the image carries them under `/licenses/`.

I fixed it rather than filing it because the alternative was knowingly shipping a false claim in a Legal-facing artifact days before the repo goes public. Say the word if you would rather split it out.

## The gate

`.github/scripts/licenses-check.sh` fails when the inventory stops matching `go list -deps ./cmd/server` in either direction, when a row's version is stale, when a row fails to parse, when a distinct-licence sub-package row disappears, or when a dependency's NOTICE goes unpropagated. It reads no secrets, so it gives a real verdict on a fork PR — where an unreviewed dependency is most likely to arrive.

It deliberately does not check licence *names* or *URLs*. That needs `go-licenses`, the network, and a warm module cache.

Runs on every PR as `Third-party licenses current`, and again at tag time as a `needs` of `build-binaries` and `build-docker` — mirroring how the CHANGELOG block is verified at the tag rather than trusted to PR CI. The tag-time run is what makes the guarantee real: the PR check is not yet a required status check, so until an admin registers it, a PR can merge with it red.

## What review caught, since it is most of this diff

The first version of the gate had three silent-pass bugs:

| Bug | Old behaviour |
|---|---|
| Wholly unparseable table | exit 1 with **zero output** — `grep -oE` returning 1 under `pipefail` killed the script before its own guard could print |
| **One** unparseable row | **passed green** — the row vanished from the parsed set, so the stale component it named became invisible |
| Distinct-licence sub-package row deleted | **passed green** — `go-jose/v4/json` is BSD-3-Clause where its parent is Apache-2.0, but the parent module was still covered |

And my self-test could not tell: it asserted on exit codes alone, and two cases proved nothing. The near-miss case claimed to test the `/` module boundary using a path that cannot match under any scheme. Mutation testing showed both it and the longest-prefix case survived a boundary-less matcher.

Rebuilt so every case is a shape where correct and broken implementations differ, and verified by mutation: removing the `/` boundary now turns the suite red. Ten cases.

Also: `sony/gobreaker/v2`'s licence link was a 404. `go-licenses` inferred a `v2/` subdirectory that does not exist. Checked all 33 links; it was the only one. The generator is authoritative on licence names, not URLs.

## Verification

- `licenses-check.sh` run against the pre-fix document reported all eight drift problems, independently of how they were found. Red before green.
- Self-test: 10 passed, 0 failed. Mutation-tested — boundary removal, always-true version check, and each drift direction disabled all turn it red.
- All four previously-silent failure modes now fail with actionable messages, including line numbers for an unparseable row.
- `make check` exit 0. `actionlint` exit 0 across all seven workflows.
- Licence names independently confirmed from `$(go env GOMODCACHE)`, not from the ticket.

**Verified, including the parts only a tag exercises.** Three surfaces in this PR are not touched by any pull-request check, because no workflow builds the image and `release.yml` only runs on a tag. All three were checked locally:

| Tag-only surface | How it was verified |
|---|---|
| `Dockerfile` `COPY ... /licenses/` | `podman build --format docker`, then `podman create` + `podman export`, since the distroless base has no shell. All three files land at `/licenses/`, mode 0644, byte-identical to the repo copies by `cmp`. |
| `release.yml` tarball members | Ran the same `tar czf` argument list; archive contains the binary, `broker-config.example.yaml`, `LICENSE`, `NOTICE`, `THIRD_PARTY_LICENSES.md`. |
| `release.yml` `licenses` job + its `needs` wiring | Verified by equivalence: its steps are byte-identical to the `ci-pr.yaml` job that is green on this PR, and `licenses-check.sh` reads no git history, so a detached tag checkout behaves the same. |

**What that does not cover**, stated so nobody reads more into it. The release builds with BuildKit via `docker/build-push-action` for `linux/amd64` and `linux/arm64` with a GHA cache; the local run was buildah on arm64, no cache. The instruction itself cannot diverge across those: a multi-source `COPY` requires a directory destination ending in `/` under BuildKit, buildah and the classic builder alike, this one does, `.dockerignore` cannot match any of the three files, and a new `COPY` vertex cannot cache-hit because its key covers the instruction text and source content digests. Three architecture-independent text files cannot differ per platform.

One new coupling worth naming: the release critical path now depends on the module proxy, because `licenses-check.sh` fails closed when a module directory cannot be resolved. A transient fetch failure can now block publishing where before it could not. That is the right direction to fail, and it is new.

Image-build coverage on pull requests is still worth its own ticket. It is the durable fix for this class, and without it every future `Dockerfile` change lands with the same gap.

## Note on CI

`FOSSA Scan / SCA Scan` is red here for a reason unrelated to this change: the org move left the Vault role's bound `repository` claim pointing at `SolaceDev`, so OIDC auth fails on every PR in the repo. Evidence on [SOL-152670](https://sol-jira.atlassian.net/browse/SOL-152670).

CHANGELOG entry added, though the production-surface rule does not require one — 0.6.0 shipped with an inaccurate inventory, so a downstream compliance consumer has reason to know their copy was wrong.


solace-lootii (changes requested):
Requesting changes — one blocker, everything else I'd approve as-is.

Traced `licenses-check.sh` by hand and it holds up: the module-level dedup in
checks 1a/1b is real (`go list -deps` reports per-package module info, `sort -u`
collapses same-module packages to one line), which is why `REQUIRED_COMPONENTS`
has to exist separately — deleting the `go-jose/v4/json` row alone would still
pass 1a/1b since the parent module stays covered. Confirmed the patch actually
closes that gap. Also hand-traced `resolve_module`'s `/`-boundary check and the
`change_version()` awk field-splitting against real table rows — both correct,
including why the near-miss test case is a genuine discriminator.

**Blocker: the Dockerfile change is unverified, and this is the wrong change to
ship unverified.** You said it yourself — no workflow builds the image on a PR,
so the `COPY --from=builder .../NOTICE .../THIRD_PARTY_LICENSES.md /licenses/`
step won't be exercised until the next tag. That means the first real test of a
compliance fix, for a public container image, going out days before the repo
opens up, is the release itself. I read through it and the reasoning holds
(builder does `COPY . .`, `.dockerignore` doesn't exclude either file), but
"reads correctly" isn't the bar I want for this one. Please run `docker build`
locally and confirm both files land under `/licenses/` before this merges — or
if that's not feasible right now, split the Dockerfile/tarball piece into its
own PR so the fully-tested inventory/gate fix isn't held hostage to it.

On your two second-opinion asks, unblocked by the above:
1. **testify/go-spew/go-difflib** — agree with your read; `go list -deps` keys
   off filename, not content, so the import is real and the "do not remove"
   note is correct.
2. **Scope** — given the Docker piece needs its own verification anyway, I'd
   now lean toward splitting rather than bundling: let the tested inventory fix
   merge on its own, and give the unverified shipping fix the scrutiny it
   needs without slowing this one down.

Also +1 on SOL-152670 blocking both PRs regardless — nothing here changes that.

solace-aross:
Verified rather than split. You were right to block on this.

Built the image and exported its filesystem, since distroless has no shell to exec into:

```
drwxr-xr-x      0  licenses/
-rw-r--r--  11358  licenses/LICENSE
-rw-r--r--   1630  licenses/NOTICE
-rw-r--r--   7402  licenses/THIRD_PARTY_LICENSES.md
```

All three are byte-identical to the repo copies (`cmp` clean against each). Build step 4/7 is the `COPY --from=builder`, so it is exercised, not inferred.

One tooling caveat worth stating: this machine has podman, not docker, so it was `podman build` against the same Dockerfile. The only difference that surfaced is a podman warning that `HEALTHCHECK` is ignored under the OCI image format, which is unrelated to the `COPY` and does not apply to the docker-format build the release does.

Checked the tarball half too, since it carries the same obligation. Ran the exact `tar czf` line from `release.yml`:

```
solace-broker-mcp
broker-config.example.yaml
LICENSE
NOTICE
THIRD_PARTY_LICENSES.md
```

On scope: the blocker was verify-or-split, and the verification is now done, so I have left it as one PR. Your lean toward splitting rested on the Docker piece being untested, and that reason has gone. Splitting now would also mean re-stacking this on top of #240 again, which has already been rebased twice today. Say the word if you still want it split and I will do it, but I would rather not pay that cost for a reason that no longer holds.

Agreed on your two second-opinion points, and agreed SOL-152670 gates both PRs regardless.

solace-aross:
Two corrections to my last reply, both against me.

**Retracting the rebase-cost argument.** I said splitting would mean re-stacking on #240 again. That was wrong, and it was the weakest reason I gave. Splitting does not change this PR's base; it means dropping about 14 lines here and opening a third PR off the same branch. Real work, but not a re-stack, and I should not have priced it that way to a reviewer who had just correctly caught me shipping something unverified.

**Your revert concern was right, and I have acted on it.** Both halves were in one commit, so `git revert` would have taken the tested gate fixes along with the untested shipping fix. That is the substance of "the tested fix isn't held hostage to the untested one", and replying "it's verified now" did not answer it. Split into two commits on the same branch:

- `85d0925` the three silent-pass gate fixes, the dead licence link, and NOTICE. Every line exercised by `Third-party licenses current` on this PR.
- `7ca3483` the Dockerfile `/licenses/` copy, the tarball members, and the release-time gate. Tag-only.

Independently revertable and independently reviewable, no third PR, no re-stack.

**Where I would still push back on splitting into separate PRs.** `THIRD_PARTY_LICENSES.md` line 5 has described itself as "the human-readable OSS compliance inventory that accompanies the release" since before this PR, and this PR does not change that sentence. If the inventory half merges alone and anyone cuts a tag, the release ships an accurate table that still falsely claims to travel with it. That reintroduces the defect the PR exists to close, so I would rather keep them in one PR as two commits than in two PRs.

**On the verification itself, one thing improved.** I rebuilt with `podman build --format docker` and the HEALTHCHECK warning disappears entirely; it was an OCI-format artifact, so that caveat was noise I should have removed with a flag instead of explaining. The result is unchanged: all three files at `/licenses/`, mode 0644, `cmp` clean.

I have also bounded the claim in the PR description rather than implying builder equivalence. Untested deltas are the BuildKit frontend, the amd64 leg, and the GHA cache. None can affect this instruction, and the description says why.

**A third tag-only surface I should have named, and did not.** The new `licenses` job in `release.yml` and its `needs` wiring are also exercised by no pull request. Not a defect, but its failure mode is worse than an incomplete artifact, because it blocks a release. Verified by equivalence: identical steps to the green `ci-pr.yaml` job, and the script reads no git history so a detached tag checkout behaves the same. It does add a module-proxy dependency to the release critical path, which is new and is now written down.

Your premise was right throughout: no PR workflow builds the image, and there is no image scanning anywhere. Image-build coverage on PRs is the durable fix and this will recur on every Dockerfile change until it exists.

solace-lootii (dismissed):
LGTM

solace-aross:
#240 merged, so this is now retargeted to `main` and needs one more approval. The retarget dismissed the previous one; nothing in the change moved.

`bbccc36` is the same three commits replayed straight onto `37f526c`. The diff is now only this ticket: 10 files, 524 insertions, no #240 content.

Worth noting, since it changes what you are looking at: this PR now gets the **full** check suite. While it was stacked on #240's branch it only ran the four `ci-pr.yaml` checks, because `build-and-test.yml` is scoped to pull requests against `main`. So `build`, `lint`, and the five e2e suites had never run against this change. They are running now.

Verified locally on `bbccc36`:

- `.github/scripts/licenses-check.test.sh` → 10 passed, 0 failed
- `.github/scripts/licenses-check.sh` → exit 0, matches the 31 modules in the closure of `./cmd/server`, 32 rows checked
- `actionlint` v1.7.7 across all workflows → exit 0
- `make check` → exit 0, 31 packages, 0 failures
- `ci-pr.yaml` carries both `release_notes_selftest` (from main) and `licenses` (from here), which was the conflict I resolved earlier

Still expect `FOSSA Scan / SCA Scan` and `SCA gate` red for the reasons in the #240 thread. Neither is a required context.

solace-lootii (approved):

PR #246, opened 2026-07-31
SOL-152543: attest release artifacts and republish the image under SolaceProducts

bczoma:
## Summary

Attach a GitHub build provenance attestation to every release artifact — the four binary archives and the multi-arch container image — and move the published image to `ghcr.io/solaceproducts/solace-broker-mcp` in the same change, so the first public release is verifiable from the start.

Jira: [SOL-152543](https://sol-jira.atlassian.net/browse/SOL-152543)

> [!IMPORTANT]
> **Ready for review, but do not merge yet.** Review feedback is welcome now; the merge is held until the preconditions below are met. The red FOSSA check is [DATAGO-146717](https://sol-jira.atlassian.net/browse/DATAGO-146717) and is unrelated to this change.

## Type of Change

- [x] New feature (non-breaking change which adds functionality) — the attestations
- [x] Breaking change — the container image path moves namespace
- [x] Documentation update

## Motivation and Context

`checksums-sha256.txt` proves the bytes are intact. It proves nothing about where they came from: a checksum file is only as trustworthy as whoever published it beside the artifacts. Internally that gap is invisible, because everyone who can push a tag is on the same team and the pipeline is the only way artifacts get built. Going public removes both assumptions — someone pulls the image into a production cluster with no way to distinguish it from an image an attacker pushed to a lookalike registry path, and no way to tell which commit produced it.

Timing matters too. If the first public releases ship unsigned and later ones are signed, the verification docs need a "signing starts at version X" caveat forever. Making the first public release the first signed one avoids that permanently.

### How attestations work, briefly

There is no key — nothing to store in Vault, rotate, or leak. The job asks GitHub for a short-lived token asserting which job, repo, and commit is running; Sigstore validates it and issues a ~10-minute signing certificate carrying those claims; the job signs a statement binding the artifact's digest to that build. Only the hash is signed — the artifact never leaves the runner. The signed record goes to a public transparency log and to GitHub, and the certificate then expires.

## Changes Made

**`.github/workflows/release.yml`**
- `attestations: write` added to the top-level permissions block (`id-token`/`packages` were already present)
- `build-binaries`: attest each archive after packaging, per matrix leg — provenance records the job that produced the subject, so attesting after download would name the release job rather than the builder
- `build-docker`: image path moved to the SolaceProducts namespace; `id: build` added so the push step's digest can be referenced; the image attested once on the multi-arch index digest (covering every tag, including `latest`), with `push-to-registry: true` so verification works without the GitHub API

No new job, no new secret, so nothing changes in the required-status-check list.

**Docs**
- `README.md` — `gh attestation verify` alongside the existing checksum step, for both the archive and the image; image path updated in the `docker run` example and the Compose snippet
- `RELEASING.md` — the conflated Stable-gate bullet split into three distinct claims (immutable tag `[Planned]`, signed *git tag* `[Planned]`, artifacts attested `[Implemented]`); pipeline description and post-tag verification checklist updated
- `CHANGELOG.md` — `Added` for the attestations, `Changed` with a **BREAKING** note for the image path
- Image path swept in `deploy/kubernetes/deployment.yaml`, `docs/user-guide.md`, `docs/internal/packaging-release.md`

## Testing

No Go code is touched, so the Go build and test surface is unaffected.

**Verified locally**
- `.github/workflows/release.yml` parses; all seven jobs intact, including the `licenses` gate added by SOL-152414
- No `.go`, `go.mod`, or `go.sum` changes
- No stale `ghcr.io/solacedev` reference remains, other than the deliberate migration note in the CHANGELOG entry
- Branch rebased onto current `main`

**Cannot be verified without cutting a release — reviewers should not read CI green as proof of these**
- that the attest steps succeed and `attestations: write` is sufficient
- that `steps.build.outputs.digest` is populated as expected
- that `gh attestation verify` passes against a published archive and image
- that the `ghcr.io/solaceproducts` namespace accepts the push (the package does not exist yet; the first release creates it)

## Breaking Changes

**Impact:** the published container image moves from `ghcr.io/solacedev/solace-broker-mcp` to `ghcr.io/solaceproducts/solace-broker-mcp`. Releases from this version forward publish only to the new path. The old path stops receiving new tags, so any `docker pull`, Compose file, or Kubernetes manifest still pointing there keeps resolving to the last image pushed and silently stays behind.

**Migration:** update every image reference to `ghcr.io/solaceproducts/solace-broker-mcp`. The tag scheme (`{version}`, `{major}.{minor}`, `latest`, `sha-<short-sha>`), image contents, config path, and health check are unchanged. Tags already published under the old path remain where they are, unsigned.

## Merge preconditions

Held deliberately until all of these are true:

1. **The repository is public.** Attestations on an internal repo require GitHub Enterprise Cloud and are signed by GitHub's internal Fulcio with no transparency log — a different trust root from what a public consumer verifies against. Merging early risks a tag being cut into that window.
2. **[DATAGO-146717](https://sol-jira.atlassian.net/browse/DATAGO-146717) is fixed.** `fossa-guard` derives its container image org from `github.repository_owner`, so the org move left it pulling from a namespace where the image does not exist. `release` declares `needs: fossa_scan`, so a release run cannot complete until this clears. Note this does *not* block merging — `SCA gate` is not a required check today — but it does block verifying the change.
3. **GHCR package visibility is handled.** A newly created GHCR package is private by default and does not become public when the repository does. Missed, the first public `docker pull` fails even though the release looks green.

The first release cut after merge is intended to be both the republish and the first signed release, so one release cycle verifies both.

## Additional Notes

Out of scope, deliberately: SBOM generation and SBOM attestation (owned by SOL-148440 — `actions/attest-sbom` is the natural follow-on); commit signing (the team chose DCO sign-off over required signed commits, SOL-152402 — this signs *artifacts*, a different mechanism against a different threat); signed git tags, which remain `[Planned]` and are only untangled from the artifact-signing claim here; cosign signatures; retroactive signing of past releases.

## Related Issues/PRs

- Implements SOL-152543
- Depends on DATAGO-146717 (verification), and on the public-repo flip (SOL-152410)
- Builds on SOL-152669 (module path rename) and SOL-152414 (licence artifacts now in the archive)


.github/workflows/release.yml
```diff
@@ -127,6 +158,22 @@ jobs:
             THIRD_PARTY_LICENSES.md
           echo "ARCHIVE=${ARCHIVE}" >> "${GITHUB_ENV}"
 
+      # Attested here, per matrix leg, rather than in the release job, for two
+      # reasons. The digest is bound to the archive before it round-trips
+      # through upload-artifact/download-artifact, so the attestation covers
+      # the bytes as built rather than the bytes as restored. And a failure to
+      # attest surfaces before anything is published, instead of after the
```
solace-aross, line 165: This principle is right, and it is the reason the second note below is worth raising: the same reasoning applied to `fossa_scan` shows the supply-chain gate is in the wrong place.

`build-docker` declares `needs: [test, release-notes, licenses]` and pushes with `push: true` unconditionally. `fossa_scan` gates only the `release` job. So on a tag push the image goes live on `ghcr.io` whether or not the SCA gate passes, and only the GitHub Release waits for it.

The `needs` graph is pre-existing and this PR does not change it, so this is not a defect you introduced. I am raising it here because this PR is what moves the image to the path that becomes public, and because the consequence is concrete right now: the FOSSA SCA contexts are currently failing at baseline on this repo, on PRs that change no dependencies at all. #254 shows the same pair of failures while touching only tests, a Makefile target and docs. So the first tag push under this workflow would most likely publish the image to the new public path and then fail before creating the Release, leaving a published image with no release behind it and no attestation visible on the Releases page.

Cheapest fix is adding `fossa_scan` to `build-docker`'s `needs`, which costs nothing on the happy path since the two already run in parallel. Gating the `push` step is the alternative if you would rather keep building the image for diagnostics on a failed scan.

.github/workflows/release.yml
```diff
@@ -177,9 +237,25 @@ jobs:
...
+      # bundle next to the image, which cosign can verify straight from
+      # ghcr.io. `gh attestation verify oci://…` still queries the GitHub API
+      # by default — it only reads the registry copy when passed
+      # `--bundle-from-oci`.
+      - name: Attest image provenance
+        uses: actions/attest-build-provenance@v4
+        with:
+          subject-name: ghcr.io/solaceproducts/solace-broker-mcp
+          subject-digest: ${{ steps.build.outputs.digest }}
+          push-to-registry: true
```
solace-aross, line 257: Worth adding a verification step that proves the attestation you just published actually verifies, before the release is announced.

The argument is the repo's own pattern rather than a general principle: this workflow already self-tests three pieces of its release machinery (`DCO check self-test`, `Release-notes extraction self-test`, and `licenses-check.sh` gating the inventory rather than trusting a "regenerate before each release" sentence). Attestation is now a fourth published guarantee, and it is the only one with no check behind it. A wrong `subject-name` casing or a digest that does not match what was pushed would produce a release whose README tells consumers to run a command that fails.

A `gh attestation verify` call in the `release` job against the downloaded archives, and against `oci://…@${digest}` for the image, would close that. It is the same command the README tells consumers to run, which is the point: if it passes in CI, the documented procedure is known to work for that release.

Not blocking, and not something this PR has to grow. The reason to do it before the next tag rather than after is that a broken attestation is discovered by whoever tries to verify it, which is the one audience this feature exists for.

solace-aross (dismissed):
Approved. Careful work, and the inline comments explain the non-obvious choices well enough that I could check them instead of guessing at intent. This workflow only runs on a tag push, so PR CI never parses it, which makes the unverifiable-until-release parts worth checking by hand. I checked the ones that would fail at tag time.

| Claim | Evidence |
|---|---|
| `artifact-metadata` is a real permission key | It is. GitHub's workflow-syntax scope list includes it alongside `attestations`. An invalid key here would fail the whole workflow at parse time, so this was the one worth confirming. |
| `actions/attest-build-provenance@v4` exists | v4 is the current major; latest release is v4.1.1. |
| The `test` job grant "mirrors the widest grant" `build-and-test.yml` makes | Exact, not approximate. `fossa_scan` is the only job in that workflow declaring a `permissions` block, so the union it asks for is precisely that block, and your grant matches it key for key with no shortfall and nothing surplus. |
| Dropping top-level to `contents: read` strips nothing silently | Enumerated all seven jobs. Every one carries an explicit block, including `release-notes` and `licenses` at `contents: read`. Nothing falls through to an unintended default. |
| `release` needs only `contents: write` | Correct. Checkout, two same-run `download-artifact` calls (no `actions: read` needed for same-run), `sha256sum`, and `softprops/action-gh-release`. |
| The repointing is complete | The only `ghcr.io/solacedev` left at this head is inside the CHANGELOG migration note, where it has to be. All six files carrying it on `main` are updated. |
| `--repo` alone is weaker than it looks | Confirmed against the `gh` manual: `--signer-workflow` enforces the signing workflow, `--repo` only scopes lookup. Worth having said three times. `--source-digest` and `--bundle-from-oci` behave as documented. |
| Image tags carry no leading `v` | `type=semver,pattern={{version}}` on `v1.2.0` yields `1.2.0`. The CHANGELOG's `1.2.0`/`1.2`/`latest`/`sha-<short-sha>` list is right. |

Two notes inline, neither blocking. The first is the one I would actually act on before the next tag.

**On Copilot's CHANGELOG comment:** ignore it or fold the headings, your call, but nothing depends on it. The duplicate `### Added` is already on `main` at lines 17 and 21, and this PR adds no heading line at all. I also ran the extractor against a renamed block to be sure: it copies the version block verbatim, so both `### Added` sections and every entry survive, including yours. Published notes will show two Added headings, which is cosmetic.

.github/workflows/release.yml
bczoma: Taken, in 17d21d8. `fossa_scan` is now in `build-docker`'s `needs`.

You're right that the graph is pre-existing, and right that it matters now rather than later. The asymmetry that decided it for me: a registry push cannot be withdrawn, whereas the binaries are only artifacts until `release` publishes them — and `release` already gated on the scan. So the image was the one published artifact with no SCA gate in front of it, and this PR is what makes that path public.

Your prediction of the failure mode is what tipped it from tidy-up to defect. With the SCA contexts red at baseline, the first tag push would have published to the new public path and then failed before creating the Release: a live image, no release behind it, and no attestation on the Releases page. That is a worse first impression than a failed run.

Gated the `needs` rather than the `push` step, on the grounds that a build whose scan failed has no diagnostic value worth a published image. Cost on a clean run is the longer of the two jobs, not the sum, since they already ran in parallel.

Two doc claims became false and went with it. `RELEASING.md` said the image is "pushed as soon as build-and-test passes, in parallel with the FOSSA scan". And the ASCII job diagram could not express the new graph unambiguously — it had also been omitting the `release-notes` and `licenses` gates entirely — so it is now an explicit list of what each job waits on.

The attest-after-push window is unchanged and still documented: attesting a registry digest requires the digest to exist, so that ordering is not fixable here.

Apologies for the dismissed approval — the push tripped `dismiss_stale_reviews_on_push`.

.github/workflows/release.yml
bczoma: Agreed on the substance, and deliberately not doing it in this PR.

The argument from the repo's own pattern is the convincing part. `DCO check self-test`, `Release-notes extraction self-test` and `licenses-check.sh` all exist because a published guarantee that nothing checks is a guarantee that quietly stops holding. Attestation is now the fourth, and the only one with nothing behind it — and the failure you describe is the expensive kind, because the person who discovers it is a consumer following instructions in the README.

Leaving it out here for two reasons. It is separable work rather than a fix to what this PR does, and this PR is already approved and rebased twice; growing it restarts a cycle for something that has no dependency on the rest of the change. I will file it as a follow-up under SOL-152410 rather than leave it in a comment thread.

One wrinkle worth recording for whoever picks it up: verifying in the `release` job proves the attestation resolves through the GitHub API, but not through the registry copy — `gh attestation verify` only reads the OCI referrer with `--bundle-from-oci`. Since `push-to-registry: true` exists specifically so consumers need not depend on the GitHub API, that path deserves its own assertion. And for the image it should verify `oci://…@${digest}` as you say, not a tag: a tag reference makes the check dependent on tag resolution, which is a different thing from what the attestation binds.

There is also a sequencing constraint — until the repo is public, attestations are signed by GitHub's internal Fulcio rather than the public-good instance, so a self-test written now would be exercising a trust root that changes at the flip.

CHANGELOG.md
```diff
@@ -14,10 +14,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+- **BREAKING**: The published container image moved from `ghcr.io/solacedev/solace-broker-mcp` to `ghcr.io/solaceproducts/solace-broker-mcp`, matching the repository's move to the `SolaceProducts` organization ahead of going public. Releases from this version forward publish only to the new path; the old path stops receiving new tags, so a `docker pull`, Compose file, or Kubernetes manifest still pointing at `ghcr.io/solacedev/solace-broker-mcp` will keep resolving to the last image pushed there and silently stay behind. Migration: update every image reference to `ghcr.io/solaceproducts/solace-broker-mcp` — the tag scheme (`{version}`, `{major}.{minor}`, `latest`, `sha-<short-sha>`), image contents, config path, and health check are unchanged. Tracked under SOL-152543.
```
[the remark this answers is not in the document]
bczoma: Not acting on this one — the premise is inverted. In this file `**BREAKING**:` (colon outside the bold) appears 11 times and `**BREAKING:**` once, so the new entry matches house style and the single outlier is the module-path entry directly above it. Applying the suggestion would move the file further from its own convention.

solace-aross (approved):
Approved. You went past what I asked for: I suggested adding `fossa_scan` to `build-docker`'s `needs`, and you also rewrote the RELEASING.md flow diagram to match the real graph and documented the one window that genuinely cannot be closed.

**Verified at `17d21d8`:**

- The `needs` graph matches the new diagram exactly: `build-binaries` on [test, release-notes, licenses], `build-docker` on those plus `fossa_scan`, `release` on [build-binaries, build-docker, fossa_scan]. No cycle.
- "Costs the longer of the two, not the sum" holds. `fossa_scan` declares no `needs`, so it starts with the workflow rather than after `test`. Same-head timings: SCA scan 48s against `build` at 3m50s, so it finishes roughly three minutes before `test` does and the added wall-clock is zero in practice.
- Keeping `build-binaries` ungated is the right asymmetry, and the RELEASING.md sentence names the reason correctly: archives are only artifacts until `release` publishes them, while a registry push is live immediately and cannot be withdrawn.

The push-before-attest paragraph is the right way to handle a gap you cannot close. Naming it, explaining why a consumer's failed verify means an incomplete release rather than tampering, and pointing at Rollback is more useful than a **[Planned]** marker.

**One consequence worth flagging, not a defect.** This makes the FOSSA baseline problem release-blocking. `FOSSA Scan / SCA Scan` and `SCA gate` are failing on this PR right now, and they fail on PRs that change no dependencies at all, so a tag push today would publish nothing: no image, and no Release. Before this change the same tag push would at least have published the image. That is the correct trade and exactly what you intended, but it moves an unowned problem onto the critical path for the next release, so it needs an owner before the next tag rather than after.

My other note, a `gh attestation verify` step proving the published attestation verifies, is still open. Carrying it forward rather than re-arguing it; it was a suggestion and the release machinery is better than it was.

.github/workflows/release.yml
solace-aross: Checked both wrinkles and they hold, so the follow-up should carry them rather than rediscover them.

The trust-root point is right, and it is worth more than a footnote. Private repositories get their certificate from GitHub's internal Fulcio on GitHub's own Sigstore instance, which has no transparency log and federates only with Actions. Public repositories get the Sigstore Public Good Instance, and the bundle is additionally written to a public immutable log. So the flip changes more than which key signs: a self-test written today could not exercise the log-inclusion path at all, because there is no log to be included in. That strengthens your sequencing argument rather than just supporting it.

Your `--bundle-from-oci` point is sharper than my note was. A verify step that only exercises the API path leaves the registry copy untested, and the registry copy is the entire reason `push-to-registry: true` is there. Agreed on `oci://…@${digest}` over a tag, for the reason you give.

SOL-152410 is the right home for it.

Nothing further from me on this PR.

PR #248, opened 2026-08-04
SOL-151811: Composite tool path parameters lack empty-string validation

solace-wmaryam:
## Summary

Required string parameters on composite tools now carry `minLength: 1` in their generated input schema, so an empty string is rejected at MCP schema validation instead of reaching the broker and failing there with a generic 4xx that names nothing.

Tracked under [SOL-151811](https://sol-jira.atlassian.net/browse/SOL-151811).

## Type of Change

## Motivation and Context

When a composite YAML tool declares a required string parameter (e.g. `msgVpnName`, `queueName`, `clientName`, `topicEndpointName`, `restDeliveryPointName`), an empty string passed MCP schema validation and reached the broker, which rejected it one round trip later with a generic 4xx that named nothing:

- When the empty value was a **path** segment, it templated into a double slash (`.../msgVpns//queues/foo/deleteMsgs`).
- When it was a **create-time object name carried in the request body** (`create-queue`'s `queueName`, `create-topic-endpoint`'s `topicEndpointName`, `create-rdp`'s `restDeliveryPointName`, `create-message-vpn`'s `msgVpnName`), it was sent as an empty attribute the broker refused.

The native SEMPv2 write-action handlers set `minLength: 1` on their path params and enforced it at schema-validation time — the request never left the client. When those tools were migrated to composite YAML in SOL-151646, the guardrail was dropped because the composite `ParameterDef` had no way to express it.

An earlier revision of this PR restored the guardrail only for path params, which left the four `create-*` object names (body fields, not path segments) still unguarded — [flagged in review](https://github.com/SolaceProducts/solace-broker-mcp/pull/248#discussion_r3714028355). The rule is now keyed off `required` + `string`, which subsumes the path-param case and also covers the create-time body names. Every required string in the catalog is an object identifier for which an empty value is never valid, so there are no false positives.

Not a correctness or safety regression — the request failed either way — but the error now surfaces at schema validation with a message naming the offending parameter, rather than one round trip later with a broker-side 4xx.

## Changes Made

- `internal/tools/composite_handler.go` — `buildCompositeInputSchema` emits `minLength: 1` for any `required` `string` parameter.
- `internal/tools/composite_handler_test.go` — unit test for the schema output (required string gets `minLength`, optional string and required non-string do not).
- `cmd/server/composite_schema_test.go` — production-surface test over the real embedded `tools.yaml` through the full registration pipeline, pinning `minLength` on every write tool's object-name params (create-time body names included) and asserting the optional `*Config` bodies stay unconstrained.
- `CHANGELOG.md` — `[Unreleased]` Fixed entry.

The rule reads the `required`/`string` flags already present on each YAML parameter definition — **no tool YAML changed** — so it applies uniformly to all 16 composite write tools (`create-queue`, `update-queue`, `delete-queue`, `create-message-vpn`, `update-message-vpn`, `delete-message-vpn`, `create-topic-endpoint`, `update-topic-endpoint`, `delete-topic-endpoint`, `create-rdp`, `update-rdp`, `delete-rdp`, `delete-queue-messages`, `clear-queue-stats`, `disconnect-client`, `clear-client-stats`) and to required string identifiers on read tools as well.

## Testing

**Test Environment:**
- Go version: go1.25.0
- OS: Linux
- Broker version (if applicable): n/a (schema-validation-layer change)

**Tests Performed:**

**Test Coverage:**
- `TestCompositeToolHandler_SchemaMinLength` — the schema output through `Metadata()`: a required string gets `minLength: 1`, an optional string does not, and a required integer does not.
- `TestWriteToolObjectNamesHaveMinLength` — over the real `tools.yaml` and the full registration pipeline (`registeredServer` + `tools/list` round-trip): every write tool's object-name params carry `minLength: 1`, including the four `create-*` body names, and the optional `*Config` bodies do not.

## Breaking Changes

N/A


internal/tools/composite_handler.go
```diff
@@ -39,16 +40,41 @@ func NewCompositeToolHandler(tool composite.CompositeTool, executor *composite.C
 	}
 }
 
+// pathParamNames returns the set of parameter names that are In:"path" on at
```
solace-awaheed: Not a blocker: This is correct for all write tools today since they all use identity arg templates. Just a small suggestion to document the assumption: if a future tool maps a path parameter from a differently named argument, the minLength guardrail would silently stop applying. A comment is probably enough for now, though a load-time test over the embedded YAML would be a stronger safeguard if we ever want one.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
...e) now carries `minLength: 1` in its generated input schema, so an empty string is rejected at MCP schema validation — before the request is built — with an error naming the offending parameter. Previously an empty string passed validation and was templated into the request URL, producing a double slash (`.../msgVpns//queues/foo/deleteMsgs`) that the broker rejected one round trip later with a generic 4xx that named nothing. The native write-action handlers set `minLength: 1` on these params until SOL-151646 migrated them to composite YAML, where the parameter definition had no way to express it; this restores the guardrail uniformly across every composite write tool (`create-queue`, `update-queue`, `delete-queue`, `create-message-vpn`, `update-message-vpn`, `delete-message-vpn`, `create-topic-endpoint`, `update-topic-endpoint`, `delete-topic-endpoint`, `create-rdp`, `update-rdp`, `delete-rdp`, `delete-queue-messages`, `clear-queue-stats`, `disconnect-client`, `clear-client-stats`), not just the four that were migrated. The fix lives in the schema-building framework by cross-referencing each parameter against its steps' operations, so no tool YAML changed. Tracked under SOL-151811.
```
solace-aross: This sentence is not true of the four `create-*` tools, and it is the sentence an operator will trust when auditing empty-string handling.

From the generated schema for all 39 registered tools:

| tool | params that get `minLength` |
|---|---|
| `create-message-vpn` | none |
| `create-queue` | `msgVpnName` only, not `queueName` |
| `create-topic-endpoint` | `msgVpnName` only, not `topicEndpointName` |
| `create-rdp` | `msgVpnName` only, not `restDeliveryPointName` |

The code is right, the prose is not. On a create operation the object name is a body field, so `In:"path"` is false for it, and `createMsgVpn` has no path parameters at all. `create-message-vpn` with `msgVpnName: ""` behaves exactly as it did before this PR.

Suggest either dropping the `create-*` tools from the enumeration, or keeping them and saying what they actually get: "restores the guardrail on every composite write tool whose object name is a path segment (`update-*`, `delete-*`, and the four action tools), and on `msgVpnName` for the `create-*` tools, whose object name is a body field and is therefore unaffected."

The PR description carries the same 16-tool list under "applies uniformly", so it needs the same correction.

internal/tools/composite_handler_test.go
```diff
@@ -268,6 +268,76 @@ func TestCompositeToolHandler_Metadata_FreshPointersAcrossCalls(t *testing.T) {
...
+	}
+
+	executor := composite.NewCompositeExecutor(ops)
+	handler := NewCompositeToolHandler(tool, executor)
+	props, ok := handler.Metadata().InputSchema["properties"].(map[string]any)
+	if !ok {
+		t.Fatal("expected properties map")
+	}
+
+	if got := props["msgVpnName"].(map[string]any)["minLength"]; got != 1 {
```
solace-aross: These tests do couple to the new logic, and they fail if the fix is reverted, so they earn their green. What is missing is a pin on the production surface: nothing asserts that `delete-queue-messages.msgVpnName` carries `minLength` once the real `tools.yaml` and the real SEMP catalog are loaded.

That matters more than usual here, because this exact guardrail was already lost in silence once. SOL-151646 (`edaeace`) deleted the `inputSchema()` functions in `queueactions/` and `clientactions/` that carried `minLength: 1`, and no test failed. The derivation added in this PR is name equality between a `ParameterDef.Name` and an operation's `In:"path"` parameter name, so the same silent loss recurs if a step's `args` key is ever renamed away from the tool parameter name.

The harness already exists. `registeredServer(t, true)` in `cmd/server/toollist_budget_test.go` builds the exact production pipeline, and a table over the real catalog is about a dozen lines:

```go
want := map[string][]string{
    "delete-queue-messages": {"msgVpnName", "queueName"},
    "disconnect-client":     {"msgVpnName", "clientName"},
    "delete-queue":          {"msgVpnName", "queueName"},
    // ...
}
```

Assert `minLength == 1` for each, and assert the `create-*` object names are absent. That puts the boundary from item 1 in a test instead of in prose, where it cannot drift.

internal/tools/composite_handler.go
```diff
@@ -39,16 +40,41 @@ func NewCompositeToolHandler(tool composite.CompositeTool, executor *composite.C
...
+// least one of the tool's steps' operations. Missing operations are skipped —
+// the executor surfaces those at run time.
+func pathParamNames(tool composite.CompositeTool, lookup func(string) (*sempv2.Operation, bool)) map[string]bool {
```
solace-aross: Note, not blocking. The derivation is name equality: a `ParameterDef` gets `minLength` when some step's operation has a path parameter of the same name. That is correct for every tool today. I scanned all of `tools.yaml` and every `args` key whose template references a tool parameter uses the identical name, so there is no rename to break it.

It is load-bearing and undocumented, though. `args: {msgVpnName: "{{.vpn}}"}` would silently drop the guardrail from `vpn`, and nothing would fail. The step's `args` mapping is the real source of truth for which tool parameter lands in which path segment. Worth either deriving from it, or stating the identity-mapping assumption in this comment.

internal/tools/composite_handler.go
```diff
@@ -39,16 +40,41 @@ func NewCompositeToolHandler(tool composite.CompositeTool, executor *composite.C
...
 // tool definition. The input schema is computed from the tool's Parameters; the
 // output schema is the generic step-keyed envelope shared by every composite
 // tool's collect strategy. Each call returns a freshly allocated value with
 // fresh maps inside, so callers cannot mutate shared state.
 func (h *CompositeToolHandler) Metadata() Metadata {
+	pathParams := pathParamNames(h.tool, h.executor.Operation)
+	params := make([]composite.ParameterDef, len(h.tool.Parameters))
+	copy(params, h.tool.Parameters)
+	for i := range params {
+		params[i].IsPathParam = pathParams[params[i].Name]
```
solace-aross: Note, not blocking. `IsPathParam` is derived state on a YAML DTO, which is why it needs `yaml:"-"` and why `Metadata()` needs the defensive copy above. Passing the set down instead keeps `ParameterDef` a faithful picture of the YAML and drops the copy:

```go
func buildCompositeInputSchema(params []composite.ParameterDef, pathParams map[string]bool) map[string]any
```

Same output, one less field that means nothing to a YAML author.

internal/tools/composite_handler_test.go
```diff
@@ -268,6 +268,76 @@ func TestCompositeToolHandler_Metadata_FreshPointersAcrossCalls(t *testing.T) {
 
 // TestCloneBoolPtr exercises cloneBoolPtr directly: nil-passes-through, and
```
solace-aross: The two new tests landed between this comment and the function it documents, so `TestCloneBoolPtr`'s doc comment now sits on `TestPathParamNames`. Move the new tests below `TestCloneBoolPtr`, or move the comment down with it.

solace-aross (changes requested):
The fix works, and I checked that rather than read it. Two things to change before merge, both in the spirit of the ticket.

## What I verified

Drove a real `tools/call` over the in-memory MCP transport against the production registration pipeline (`registeredServer` in `cmd/server/toollist_budget_test.go`, write tools enabled):

| call | result |
|---|---|
| `delete-queue-messages` `msgVpnName:""` | `parameter validation failed: msgVpnName: String length must be greater than or equal to 1`, 638µs, no dial |
| `delete-queue-messages` `queueName:""` | same, on `queueName` |
| `delete-queue` `msgVpnName:""` | same |
| `disconnect-client` `clientName:""` | same, on `clientName` |
| control, valid names | reaches the broker: `dial tcp [::1]:8081: connect: connection refused` |

So enforcement is genuinely server-side (`go-sdk/mcp/server.go:322` calls `applySchema`, which calls `resolved.Validate`), not advisory on the client. The description's central claim holds.

`go build ./...`, `go vet`, and `go test -race ./internal/tools/ ./internal/composite/` are all green at `3857cbb`.

I also dumped the generated input schema for all 39 registered tools. That pass is what produced item 1 below.

## Requested changes

1. `CHANGELOG.md:27` overclaims coverage for the four `create-*` tools. Details inline.
2. Nothing pins the guardrail on the real catalog, and this guardrail has already been lost silently once. Details inline.

Three non-blocking notes inline as well: the name-equality assumption, `IsPathParam` living on the YAML type, and a doc comment the new tests separated from its function.

CHANGELOG.md
solace-wmaryam: Good catch — and it's more than a prose issue, this is a real bug. You're right that on the `create-*` tools the object name is a body field, so the path-param guard never touches it. But that means `create-queue` with `queueName: ""` still passes MCP schema validation and fails at the broker one round trip later — the exact failure this PR set out to eliminate, just via a body field instead of a path segment. Same for `create-topic-endpoint`, `create-rdp`, and `create-message-vpn`.

So rather than only correcting the enumeration, I'm extending the guard: `minLength: 1` on every **required string** param, which includes the path-param rule and also catches the create-time object names in the body. I'll rework the code, tests, and CHANGELOG around that broader rule and re-request review.

internal/tools/composite_handler_test.go
```diff
@@ -268,6 +268,76 @@ func TestCompositeToolHandler_Metadata_FreshPointersAcrossCalls(t *testing.T) {
...
 // non-nil produces a fresh allocation with the same value.
```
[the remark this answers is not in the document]
solace-wmaryam: Fixed — moved the `TestCloneBoolPtr` doc comment back onto its function.

internal/tools/composite_handler_test.go
solace-wmaryam: Fixed — moved the `TestCloneBoolPtr` comment back onto its function; the two new tests sit above it now.

internal/composite/executor.go
```diff
@@ -86,6 +86,13 @@ func NewCompositeExecutor(operations map[string]*sempv2.Operation) *CompositeExe
 	}
 }
 
+// Operation looks up a catalog operation by prefixed ID (matching
+// Step.Operation). Exposed for schema-building without leaking the catalog map.
+func (ce *CompositeExecutor) Operation(prefixedID string) (*sempv2.Operation, bool) {
+	op, ok := ce.operations[prefixedID]
+	return op, ok
+}
```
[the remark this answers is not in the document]
solace-wmaryam: Resolved by removing the accessor. The guard no longer derives path params from the catalog — it keys off the `required`/`string` flags on the YAML parameter — so `CompositeExecutor` no longer exposes `Operation` and the catalog map stays fully encapsulated.

internal/tools/composite_handler.go
solace-wmaryam: This assumption no longer exists. The guard was reworked from "name-equality against a step's path params" to "any required string param," which reads straight off the YAML parameter definition and never consults the operation catalog — so a renamed `args` key can't silently drop it. It also closes the create-\* body-name gap. And the load-time test over the embedded YAML you suggested now exists: `TestWriteToolObjectNamesHaveMinLength` in `cmd/server`.

internal/tools/composite_handler.go
solace-wmaryam: Removed the derivation. The rule is now "required string → minLength:1," read from the YAML DTO, so there's no name-equality mapping left to drift. Confirmed against the catalog that every path param is already `required: true` (no regression) and every required string is an object identifier where empty is never valid (no false positives).

internal/tools/composite_handler.go
solace-wmaryam: Done — by removing `IsPathParam` and the whole path-param derivation rather than threading a set through. The rule keys off `ParameterDef.Required`, so `ParameterDef` is back to a faithful picture of the YAML with no computed field and no defensive copy.

internal/tools/composite_handler_test.go
solace-wmaryam: Added: `TestWriteToolObjectNamesHaveMinLength` in `cmd/server/composite_schema_test.go`, built on `registeredServer(t, true)` + a real `tools/list` round-trip. It pins `minLength` on every write tool's object-name params — the four create-\* body names included — and asserts the optional `\*Config` bodies stay unconstrained. On the silent-loss risk you flagged: the name-equality derivation is gone entirely, so the required/string flag this test checks is now the only source of truth.

solace-aross (dismissed):
Approved. Both blocking items are resolved, and you resolved the first one by fixing the code rather than the sentence, which is the better call.

Replacing the path-param derivation with `p.Type == "string" && p.Required` closes the `create-*` gap instead of documenting it, and it deletes the mechanism my other two notes were about: no `IsPathParam` on the YAML type, no `Operation()` accessor on the executor, no defensive copy in `Metadata()`. Thirty fewer lines for wider coverage. The misplaced doc comment is fixed too.

**What I verified**

The gap is genuinely closed, not reworded. Drove real `tools/call` requests over the in-memory transport:

| call | result |
|---|---|
| `create-message-vpn` `msgVpnName:""` | `parameter validation failed: msgVpnName: String length must be greater than or equal to 1` |
| `create-queue` `queueName:""` | same, on `queueName` |
| `create-topic-endpoint` `topicEndpointName:""` | same, on `topicEndpointName` |
| `create-rdp` `restDeliveryPointName:""` | same, on `restDeliveryPointName` |
| `list-queues` `msgVpnName:""` | same, so the read tools are covered as claimed |
| `get-bridge-status` `bridgeName:""` | same |

The new rule's load-bearing claim holds. I enumerated every string parameter in `tools.yaml`: 53 required strings across nine distinct names (`msgVpnName`, `queueName`, `clientName`, `topicEndpointName`, `restDeliveryPointName`, `bridgeName`, `bridgeVirtualRouter`, `kafkaReceiverName`, `kafkaSenderName`), and every one is an object identifier where empty is never valid. There are zero optional string parameters in the entire catalog, so the rule's scope is exactly those 53 and nothing else.

`cmd/server/composite_schema_test.go` is the pin I asked for, and it earns its green. I reverted the condition to `if false` and it failed on 30 assertions across all 16 write tools; restored, it passes. `make check` exits 0 on `04ea7a2`, including this test at 5.41s.

**Two notes, neither blocking**

1. The rule is now a convention about YAML content rather than something derived from the operation catalog: "every required string is an identifier." True for all 53 today, and the comment at `composite_handler.go:87-89` says so, but nothing enforces it. The day someone adds a required string where empty is meaningful (a filter expression, a topic pattern), they get a guard they did not ask for and the cause will not be obvious from the YAML. Fine as a trade, worth knowing you made it.

2. Your negative assertions use `create-message-vpn.msgVpnConfig`, `create-queue.queueConfig`, and `create-rdp.rdpConfig`, all of which are both `required: false` and `type: object`. They are excluded for two independent reasons, so the test cannot tell which gate did the work: dropping the `p.Type == "string"` half of the condition would still pass. `update-message-vpn.msgVpnConfig` is `required: true` and `type: object`, so adding that one row isolates the type gate and would catch a regression that started stamping `minLength` onto an object schema.

`FOSSA Scan / SCA Scan` and `SCA gate` are red for the SolaceProducts-project baseline reason, not this diff.

solace-aross (approved):
Re-approved. Both notes from my last pass are closed, and the test now discriminates in both directions.

**The negative row does the work I asked for**

Adding `update-message-vpn.msgVpnConfig` (`required: true`, `type: object`) isolates the type gate, and I confirmed that rather than assuming it. Changing the condition to `if p.Required` alone fails with `update-message-vpn.msgVpnConfig minLength = 1, want unset`. Against the previous table it would have passed, because every row there was excluded for two reasons at once.

The other half is pinned too, one layer down: changing the condition to `if p.Type == "string"` alone fails the unit test on `optional string filterText should not have minLength`. So `Required` is pinned by the synthetic parameter in `internal/tools`, and `Type` is pinned by a real one in `cmd/server`. Between the two layers neither half of the condition can be dropped silently. That is the right split.

**The `ok` checks close a real hole, not a cosmetic one**

The old negative loop read `schemas[name][param].MinLength`. A tool name that stopped being registered, or a parameter that got renamed, indexed into a nil map, yielded a zero-value `property` with a nil `MinLength`, and passed. The assertion would have gone quiet at exactly the moment it mattered. Naming both misses as failures fixes that. Same for the `prop()` helper in the unit test, which turns a panicking type assertion into a named failure.

`make check` exits 0 at `2940b5d`.

Nothing further from me. The remaining note from last pass stands unchanged and needs nothing: the rule is a convention about YAML content, and today all 53 required strings satisfy it.

PR #250, opened 2026-08-04
SOL-152808: revert Renovate Go/Actions coverage, re-add dependabot.yml, exempt Dependabot from DCO sign-off

solace-lootii:
## Summary

Reverts the SOL-152586 approach (Renovate covering Go modules and GitHub Actions) and replaces it with `.github/dependabot.yml`, since Renovate cannot be enrolled for a public repository under the org's current system. Dependabot's PRs are exempted from the `DCO sign-off` required check instead of trying to make them satisfy it, since Dependabot's commit-message config has no field for a trailer.

Tracked under SOL-152808 (supersedes SOL-152586, which is being closed/re-parented).

## Motivation and Context

Renovate was chosen over Dependabot in SOL-152586 specifically because Dependabot cannot attach a `Signed-off-by` trailer to its commits, and this repo's `DCO sign-off` required check exempts nobody by design. Since then, plat-eng confirmed Renovate cannot be enrolled for this repo once it's public (DATAGO-143266). That makes the Renovate-based approach non-viable long-term, so this PR reverses course: keep Dependabot for the ecosystems it can't avoid, and instead of trying to make its commits pass DCO, exempt it from that specific check.

## Changes Made

- **`.github/dependabot.yml`** (new): `gomod` across all three `go.mod` files (root, `test/e2e-common/broker-driver`, `test/e2e-basic-mcp/agent` — each needs its own entry, `directory` matches one path per ecosystem entry) and `github-actions`. Each entry groups minor/patch (or all actions) into one PR to avoid the default 5-PR cap silently starving updates, and sets `open-pull-requests-limit: 10` explicitly rather than leaving it at the default.
- **`.github/renovate.json`**: reverted to npm-only (just the pinned Claude Code CLI version), removing the `gomod`/`github-actions` `packageRules` SOL-152586 added.
- **`.github/workflows/dco.yaml`**: the `dco` job now carries `if: github.event.pull_request.user.login != 'dependabot[bot]'`. Keyed on GitHub's own record of who opened the PR (asserted via its API), not a commit author/committer field, which is exactly the forgeable shape this repo's DCO design already refuses to trust. Documented as a residual gap: this is PR-level, not commit-level, so a maintainer manually pushing a commit onto an open Dependabot branch would get that commit skipped too.
- **`.github/scripts/dco-check.sh`**: header comment updated — the script's own "bots are not exempt" logic is untouched (confirmed via its 42-case self-test suite, still 100% green), the exemption lives one layer up in the workflow trigger instead.
- **`.github/ADMIN_SETUP.md`** / **`.github/OPEN_SOURCE_STATUS.md`**: updated to describe the new split (Renovate: CLI pin only; Dependabot: Go/Actions scheduled updates plus its existing alerts/security role) and the DCO exemption. Also corrected the required-status-checks table's `DCO sign-off` row, which still described it as gating "every commit the PR adds" with no exception noted.

## Testing

- `renovate-config-validator` (Renovate's own tool, via `npx`) — passes against the reverted `renovate.json`.
- `actionlint` — passes against `dco.yaml`, confirms the new `if:` expression is syntactically valid.
- `.github/scripts/dco-check.test.sh` — 42/42 still passing, confirming the script's internal logic is unchanged.
- `go build`, `go vet`, `golangci-lint run` — clean.
- `.github/scripts/dco-check.sh` run locally against this branch's own commit — signed off, passes.
- Confirmed all three `go.mod` paths referenced in `dependabot.yml` exist.
- **Not verifiable pre-merge**: whether GitHub's Dependabot service parses this config without error server-side (no offline validator exists for `dependabot.yml` the way `renovate-config-validator` exists for Renovate — visible post-merge under Insights → Dependency graph → Dependabot), and whether a real Dependabot PR actually gets the DCO check skipped end-to-end. A follow-up spike is planned to confirm both once this is live.

## Breaking Changes

N/A — repo config only, no production code touched.

## Additional Notes

No `CHANGELOG.md` entry — repo tooling config, not production/server behavior.


.github/renovate.json
```diff
@@ -1,10 +1,10 @@
...
...o modules and GitHub Actions instead of using Dependabot, since Dependabot cannot satisfy the DCO sign-off gate on its own — but Renovate cannot be enrolled for a public repository under the org's current system, so that extension was reverted (SOL-152808) before this repo goes public. Go modules and GitHub Actions are covered by .github/dependabot.yml instead, with Dependabot's PRs exempted from the DCO gate at the workflow level (.github/workflows/dco.yaml) rather than via a sign-off Renovate can attach and Dependabot cannot. Bumps here follow npm's `stable` dist-tag (deliberately lagged from `latest`); see test/e2e-llm/README.md for the bump procedure. Top-level `commitBody` carries the DCO sign-off that the `DCO sign-off` CI gate requires, and is top-level rather than per-rule so it also covers lock-file maintenance and config-migration branches. The gate exempts nobody *by forging a commit's author field*: an author-name exemption would be forgeable with `git commit --author`. The address must match the identity Renovate commits under (the hosted GitHub App uses this one), otherwise its PRs go red on the gate — the fix is to correct the address here, never to exempt the bot.",
```
solace-aross: This description drops the `ignorePaths` warning that SOL-152586 added at review request, and `ignorePaths` is now *more* load-bearing than it was, not less.

Verified the chain at this head rather than from memory:

- The only npm manifest in the repo is `test/e2e-llm/package.json`. `git ls-files` finds no other `package.json`.
- `config:base` resolves to `config:recommended`, which folds in `:ignoreModulesAndTests`. I read the preset out of the installed Renovate package: its `ignorePaths` is `['**/node_modules/**', '**/bower_components/**', '**/vendor/**', '**/examples/**', '**/__tests__/**', '**/test/**', '**/tests/**', '**/__fixtures__/**']`.
- `**/test/**` matches `test/e2e-llm/package.json`. Line 6's `"ignorePaths": ["**/node_modules/**"]` replaces that array rather than extending it, and that replacement is the only reason Renovate sees the CLI pin at all.

Before this PR, dropping `ignorePaths` would have silently dropped two nested test `go.mod` files from a three-job config. After it, dropping `ignorePaths` silently switches off the only Renovate job in the repository. Higher stakes, and now with nothing in the file explaining why the line is there.

The warning needed rewording for the new scope, not deleting. Something like: "The `ignorePaths` below replaces `config:recommended`'s list (via `:ignoreModulesAndTests`), which covers `**/test/**`. That replacement is the only reason `test/e2e-llm/package.json` is in scope at all. Do not drop it as apparent redundancy: the single job this config exists for would silently stop running, with no signal."

.github/ADMIN_SETUP.md
```diff
@@ -125,20 +125,27 @@ returns 403. Something is producing this check and we cannot say from the API
...
-  sign-off` required check needs — they would go red on every PR.
-- Dependabot's role here is alerts and security-only updates. Those do not
-  conflict with Renovate and are already on.
+**Dependency updates split between Renovate and Dependabot.**
+
+`.github/renovate.json` covers exactly one job: the pinned Claude Code CLI
+version in the LLM e2e harness. `.github/dependabot.yml` covers scheduled
+version updates for Go modules across all three `go.mod` files (root,
+`test/e2e-common/broker-driver`, `test/e2e-basic-mcp/agent`) and GitHub
+Actions — grouped to keep volume down: one PR per Go module for minor/patch,
```
solace-aross, line 134: "one per module for major" is not what the config produces. `dependabot.yml` defines a single group per `gomod` entry, `go-minor-patch`, keyed on `update-types: ["minor", "patch"]` (lines 23-24, 36-37, 45-46). There is no major group, so Dependabot opens major bumps individually, one pull request per dependency.

That matters operationally rather than cosmetically. The sentence reads as "at most three Go PRs per module per cycle", which is the basis for calling `open-pull-requests-limit: 10` comfortable. The real worst case is one grouped minor/patch PR plus one PR per dependency with a major available, and the root module has roughly ten direct dependencies.

Either correct the sentence to say majors arrive individually, or add a second group so the config matches the description:

```yaml
    groups:
      go-minor-patch:
        update-types: ["minor", "patch"]
      go-major:
        update-types: ["major"]
```

I would take the second: it is what SOL-152586's Renovate config did (`Go dependencies (major)` was its own `groupName`), so grouping majors keeps the behaviour the revert is replacing rather than quietly changing it.

.github/workflows/dco.yaml
```diff
@@ -75,6 +75,38 @@ jobs:
...
+    # grant itself the exemption by claiming to be Dependabot.
+    #
+    # Residual gap, stated so it is not mistaken for an oversight: this keys on
+    # who opened the PR, not on who authored each commit in it. A maintainer
+    # who pushes an extra commit of their own onto an open Dependabot branch —
+    # rather than letting Dependabot manage it — gets that commit skipped too,
+    # since the whole job is skipped for the PR. Narrow in practice (Dependabot
+    # PRs are normally left for Dependabot to update), but real: do not use this
+    # PR as a place to hand-carry unrelated changes past the sign-off gate.
+    if: github.event.pull_request.user.login != 'dependabot[bot]'
```
solace-aross, line 109: Optional, not blocking, and only because you documented the gap honestly enough to make it worth answering.

The residual gap is real but needs write access to reach, so I would not hold the PR for it. If you ever want it closed, the shape is to keep the job running and skip per-commit rather than per-PR: gate on `github.event.pull_request.user.login == 'dependabot[bot]'` for eligibility, then still require that every commit in the range is one Dependabot produced. A maintainer's hand-pushed commit would then still need its sign-off, and the "do not use this PR to hand-carry unrelated changes" warning would be enforced rather than advisory.

The catch is that per-commit authorship is the forgeable field your design already refuses to trust, so this only works if you resolve it through the API rather than from commit metadata, which is more machinery than the gap warrants today. Recording the option, not asking for it.

solace-aross (changes requested):
The reversal is the right call, and the DCO exemption is designed correctly. Two documentation items to fix before merge, both one-sentence changes. No code defect.

## What I verified

**The exemption is not forgeable, and I checked the claim rather than took it.** `dco.yaml` triggers on `pull_request_target` scoped to `branches: [main]`, so the workflow definition and the checked-out script both come from the base ref. A pull request cannot add `if: false`, widen the condition, or claim to be Dependabot. The condition keys on `github.event.pull_request.user.login`, which is GitHub's own record of who opened the PR, not a commit field. `dco` is the only job in the workflow and its `name:` is `DCO sign-off`, matching the required context exactly, so a skip lands on the right check.

**Skipping does satisfy the required check here, and this repo already proves it.** `FOSSA Scan` is a required context that reports `skipping` on pull requests and counts as passing today. Same mechanic, so the exemption will behave as intended rather than leaving Dependabot PRs pending forever.

**Keeping the script honest was the right structural choice.** Leaving `dco-check.sh`'s "bots are not exempt" logic untouched and putting the exemption one layer up means the verifier never lies about what it checked. `dco-check.test.sh`: 42 passed, 0 failed at `6fd4611`.

`renovate-config-validator` passes on the reverted config, with no duplicate keys. `dependabot.yml` is valid YAML, and all three `go.mod` paths it names exist in the tree (`go.mod`, `test/e2e-common/broker-driver/go.mod`, `test/e2e-basic-mcp/agent/go.mod`).

## Requested changes

1. `renovate.json:4` drops the `ignorePaths` warning while `ignorePaths` is now more load-bearing than before, not less. Details inline.
2. `ADMIN_SETUP.md:134` describes major-version grouping the config does not do. Details inline.

One optional note on the residual gap, inline on the `if:` line.

solace-aross (approved):
Approved. Both items closed, and on the second one you fixed the config rather than the sentence, which is the better direction.

**Item 1.** The restored `ignorePaths` rationale is better than the wording I suggested. Naming `test/e2e-llm/package.json` as the only npm manifest, then the `config:base` → `config:recommended` → `:ignoreModulesAndTests` chain, then the before-and-after stakes, gives the next reader the whole argument without having to reconstruct it. The "would silently switch off the only job this config exists for, with no signal" line is the part that will stop the tidy-up.

**Item 2.** Adding `go-major` to all three `gomod` entries makes `ADMIN_SETUP.md:134-135` true as written. I confirmed the shape at this head: every entry now carries both `go-minor-patch` (`minor`, `patch`) and `go-major` (`major`), so "one PR per Go module for minor/patch, one per module for major, one for GitHub Actions" is exactly what the config produces. `renovate-config-validator` still passes on the reverted config.

Correcting the volume comment to say the default cap of 5 was never at risk is more honest than the original claim, and it is right: with both groups defined, every version update lands in one of them, so each directory tops out at two open PRs. `open-pull-requests-limit: 10` is decorative now, and the comment says so.

Nothing further from me. The optional note on the per-PR versus per-commit DCO skip stands as recorded, not as a request.

PR #254, opened 2026-08-04
SOL-150787: Unit-test quality audit + 85% coverage gate

solace-lootii:
## Summary

- **(a) Coverage gate**: `make test-cover` runs `go test -race -coverprofile` and fails under an 85% aggregate floor; wired into CI's `Test` step and into `make check`. Policy (no per-package exclusions, incl. `cmd/server`) documented in `docs/internal/unit-test-coverage.md` — aggregate measured 88.6% at the time this landed. The doc is explicit that this is a **backstop against collapse, not a fine-grained regression detector**: the aggregate can't isolate a small per-package drop (see the Kafka note below), so per-package `go test -cover` deltas in review are the actual regression-detection tool, and the doc says so plainly.
- **(b) Test-quality audit**: reviewed the existing suite for tautological asserts, mock-only tests, and duplicate coverage of the same generic mechanism across resource types (SEMPError propagation, `constructRequestBody`, pagination). Trimmed confirmed duplicates in `internal/composite` executor tests, `internal/tokenexchange`, `internal/config`, `internal/auth`, `internal/semp/resilience`, and `internal/tools`. Every deletion was verified individually against package-level coverage deltas before removal, not just trusted by name-similarity.
- **(c) Objective-gap backfill**: added `TestListBrokers_ResponseContainsOnlyAliases` — the Quality Plan (SOL-147161 §3.3) flagged that `list-brokers`' response had no test confirming it never leaks broker URLs/credentials to the calling client. Verified the new test actually catches a regression (sabotage-tested it against the handler, then reverted). TLS-flag and secret-handling (§3.7) were already well covered (TLS-pairing tests, `LogValue()` redaction tests on every credential-bearing config type) — no action needed there.

## Resolved during review

The executor-test consolidation in (b) originally landed unevenly: `executor_rdp_test.go` was cut from 6 tests to 1 (itself a duplicate of Queue's equivalent), and `executor_topic_endpoint_test.go` was deleted entirely — leaving all 6 create/update/delete write tools across RDP and topic-endpoint with zero executor-level unit coverage. Both are still functionally covered by `test/e2e-management`, but that's minutes against milliseconds and isn't part of `make check`.

Fixed: restored a minimal body-shape test per operation for both families (`TestExecute_{Create,Update,Delete}RDP_*`, `TestExecute_{Create,Update,Delete}TopicEndpoint_*` — 6 tests total, including a new update-topic-endpoint test that didn't exist even before this PR). Verified they aren't vacuous by sabotaging `constructRequestBody`'s path-param exclusion directly — 4 of 6 failed correctly (the 2 delete tests correctly stayed green, since delete has no body to break).

Also fixed in the same follow-up: the coverage floor was raised from an initial 80% to 85% and the doc rewritten for honesty (see (a) above); `tail -1` → `tail -n 1` for portability; the `list-brokers` alias check now tracks per-alias occurrence counts so a duplicate entry can't silently pass a count-only check.

## Test plan

- [x] `make check` (build, vet, lint, race tests, coverage gate) green at 88.6%, above the 85% floor
- [x] Verified the new `list-brokers` test fails when the handler is sabotaged to leak a URL, confirming it isn't vacuous
- [x] Verified every test deletion against per-package coverage deltas (caught and fixed a regression in `internal/composite/postprocess/handlers` — two branches in `ListKafkaReceivers`/`ListKafkaSenders` had gone untested; restored with `TestListKafka{Receivers,Senders}_SkipsMalformedRowAndSurfacesTruncation`)
- [x] Restored the 6 RDP/topic-endpoint executor tests described above; confirmed via targeted sabotage that they catch a real path-param/body-construction regression
- [x] Merged latest `main` in; no conflicts, `make check` still green after merge


docs/internal/unit-test-coverage.md
```diff
@@ -0,0 +1,44 @@
+# Unit Test Coverage Gate
+
+CI fails the `build` job if repo-wide unit test coverage drops below **80%** of
```
solace-aross: This is the line a future reader will take as the policy, and it does not say what the 80% is for.

Measured against this branch: actual is 88.6%, total statements 3715, covered 3293. So 321 statements can become uncovered before this trips, which is 9.7% of everything currently covered. The gate cannot see a regression that size.

Concretely, it would not have caught the one this PR found: the two untested branches in `ListKafkaReceivers`/`ListKafkaSenders` were caught by per-package delta comparison, not by the aggregate.

Please pick one and say it here: either the floor moves up to just under current with a note on when to ratchet it, or 80% stays and this section states that it is a backstop against collapse rather than a regression detector, and names per-package deltas as the thing that catches regressions. As written, a reader will assume the stronger guarantee.

Makefile
```diff
@@ -134,6 +134,13 @@ test: ## Run unit tests
 test-race: ## Run unit tests with the race detector (matches CI)
 	go test -race -v ./...
 
+.PHONY: test-cover
+test-cover: ## Run unit tests with coverage; fails if aggregate coverage drops below 80% (matches CI; see docs/internal/unit-test-coverage.md)
+	go test -race -v -coverprofile=coverage.out ./...
+	@total=$$(go tool cover -func=coverage.out | tail -1 | awk '{print $$3}' | tr -d '%'); \
+	echo "Total coverage: $$total%"; \
+	awk -v t="$$total" 'BEGIN { if (t+0 < 80.0) { print "FAIL: coverage " t "% is below the 80% floor"; exit 1 } else { print "OK: coverage " t "% meets the 80% floor" } }'
```
solace-aross: The threshold logic itself is correct. I checked the extraction (`go tool cover -func | tail -1 | awk '{print $3}'` picks up `88.6%` from the `total: (statements) 88.6%` line) and confirmed the comparison fires in both directions.

One robustness note: if `go tool cover` ever emits nothing (an empty or missing `coverage.out`), `$total` is empty and `awk -v t=""` evaluates `t+0` as `0`, which is below 80, so it fails closed. Good. Worth a one-line comment saying that is deliberate, since it reads like an accident and someone may later "fix" it by adding a guard that passes on empty input.

solace-aross (changes requested):
Strong work, and the self-flagging in the description is exactly right. One thing to add back, and one decision I'd like you to make rather than defer. Everything else checks out under verification.

## What I verified

**The deletions cost nothing.** I built per-package coverage profiles on `main` and on `fa5339d` and diffed them. **Zero packages moved by even 0.05 points** across all 1318 deleted lines. That corroborates your claim that you checked each removal against package deltas, including the Kafka handler restoration.

**The new `list-brokers` test is not vacuous.** I sabotaged `register.go` to append each broker's URL to its alias. The test failed on the shape assertion (`unexpected broker entry broker-a https://broker-a.example.com:8443`) and on three separate leakage assertions (`broker-a.example.com`, `broker-b.example.com`, `https://`). Restored, it passes. Your sabotage claim holds independently.

**The no-exclusion policy is sounder than I expected.** I went looking for the loophole in `docs/internal/unit-test-coverage.md`'s claim that the denominator is "every statement in the module": my hypothesis was that `go test -coverprofile` without `-coverpkg` only counts packages that have their own tests, so a future package with real logic and no test file would be invisible. I tested it by adding a 40-statement package with no test and never importing it. It appeared in the profile with 24 lines and pulled the aggregate from 88.6% to 88.0%. So the claim is accurate and the loophole I was worried about does not exist. Worth knowing the doc's central argument survives a deliberate attempt to break it.

`make test-cover` reports 88.6% and the threshold logic behaves correctly. `make check` exits 0 at `fa5339d`.

**`Analyze (go)` is red for an infrastructure reason, not yours.** The analysis completed and exported SARIF successfully; the failure is `##[error]Server Error` while *uploading* the code-quality results. A re-run should clear it.

## Requested

**Add back the minimal per-family executor tests.** You offered, and the answer is yes. The gap is a little wider than the description says: `executor_topic_endpoint_test.go` is gone entirely, and `executor_rdp_test.go`'s single survivor is `TestExecute_ListRDPs_TruncatesAtMaxResults`, which is a *list* test. So create/update/delete for both topic endpoints and RDPs, six shipping write tools, now have no executor-level test at all.

Coverage being unchanged is the point rather than the defence: `constructRequestBody`'s path-param exclusion is generic, so the lines stay covered by Queue's tests while the per-resource argument shape stops being asserted anywhere fast. `test/e2e-management` covers it, but that is minutes against milliseconds, and it is not what runs on `make check`. The same argument carried #248 last week: a generic mechanism plus a per-tool pin, because the pin is what catches a YAML edit that sends a path param in the body.

Your own "keep a minimum smoke test per tool family" bar is the right one. Applying it to these two families is all I am asking for.

## A decision I'd rather you made than left open

The floor is 80% against 88.6% actual, and the doc explains the no-exclusion choice thoroughly but never justifies that number. Concretely, **321 statements could become uncovered before CI goes red** — 9.7% of everything currently covered. For scale, the untested 40-statement package I added during this review cost 0.6 points, so roughly fourteen of those would fit under the floor unnoticed.

The sharper way to put it: **this gate would not have caught the regression this PR found.** Two branches in `ListKafkaReceivers`/`ListKafkaSenders` had gone untested, and you caught them by comparing per-package coverage deltas by hand. The aggregate never came close to 80%.

That is not an argument against the gate. A catastrophe backstop is worth having. It is an argument that the doc should say which of the two it is, because right now a reader will assume the gate protects against coverage regression and it does not. Either raise the floor to just under current with a note on when to ratchet it, or keep 80% and say plainly in the doc that it is a floor against collapse, that per-package deltas are what detect regression, and that reviewers should keep doing what you did here. I have a mild preference for the first, but I care more that the doc stops implying the stronger claim.

internal/composite/executor_queue_test.go
solace-awaheed, line 1: Same gap as RDP/topic-endpoint: TestExecute_CreateQueue_ConstructsBody and TestExecute_UpdateQueue_BodyExcludesPathParams were removed but not restored. loader_queue_test.go only validates the YAML, not the runtime path/body split, so a tools.yaml rename that moves a path param into the body would still pass make check and only show up in e2e. Can we restore these two tests to match the RDP/TE coverage?

internal/composite/postprocess/handlers/list_kafka_receivers_test.go
solace-awaheed, line 31: This covers the admin-disabled case, but not the recovered case (up=true, enabled=true, reason!=""). That's the only scenario proving bucketing respects !up, so moving byReason[reason]++ outside the if !up block would still pass the current suite. Can we add a recovered-with-stale-reason row here (and in the senders equivalent)?

internal/composite/executor_topic_endpoint_test.go
```diff
@@ -20,6 +20,17 @@ import (
...
 )
 
+// These three tests are deliberately kept even though constructRequestBody's
+// path-param-exclusion mechanism is already exhaustively tested via VPN's
+// canonical tests (executor_vpn_test.go) — that mechanism is generic, but it
+// stays correct regardless of resource name. What isn't generic is whether
+// THIS tool's own YAML-mirroring field names (topicEndpointName,
+// topicEndpointConfig) are wired to the right path/body split. A rename typo
+// in tools.yaml's create/update/delete-topic-endpoint definitions would pass
+// every other test in this package and only show up here, fast, without a
```
solace-aross: This sentence is not true of the code below it, and it is the sentence that makes these tests non-duplicative.

`createTopicEndpointTool()` and friends build `CompositeTool` literals in this file. Nothing here loads `tools.yaml`, so a rename in it cannot reach these tests.

I checked rather than assuming. Removed `topicEndpointName: "{{.Params.topicEndpointName}}"` from `update-topic-endpoint`'s step args in the real `tools.yaml`, which is a genuine wiring bug:

- `go test ./...` with the bug in place: all 31 packages ok, exit 0. `TestExecute_UpdateTopicEndpoint_BodyExcludesPathParams` passes.
- Same assertions with the tool loaded via `LoadTools(definitions.FS, "tools.yaml")` + the existing `findTool()`: fails with `args[topicEndpointName] = <nil>, want te-1`.

One import different, and only one of the two sees it.

Two corrections to my own framing, because they cut in opposite directions and both belong here.

The body-leak assertions at :181-186 cannot catch this bug class under any construction, catalog-backed or not. `constructRequestBody` derives `nonBodyParams` from `op.Parameters` (`executor.go:670`), not from step args, so a path param is excluded from the body regardless of what the step declares. Only the `args["topicEndpointName"]` assertion can see it. So the assertion this comment is built around is the wrong one for the failure it names.

Against that, the consequence is milder than "wrong write": the param vanishes and `buildURL` fails closed (`client.go:227`, `unfilled path parameters: topicEndpointName`). Broken tool, loud, not silent corruption. And `e2e-management` catches it on every PR. Minutes instead of milliseconds, which is exactly the trade this comment is honest about.

So: not blocking. Please make the comment say what the tests actually prove. The create-side test does carry real value the VPN set does not have, and it is worth naming: `config/createMsgVpn` is `POST /msgVpns` with no path params, so VPN never covers one scalar excluded as a path param plus a second scalar and an object spread into the body at the same time. Topic-endpoint create does.

Smaller: "deliberately kept" is not accurate for `TestExecute_UpdateTopicEndpoint_BodyExcludesPathParams`. It is new; `main` had only the create and delete tests here.

internal/composite/executor_rdp_test.go
```diff
@@ -48,6 +45,17 @@ func listRDPsTool() CompositeTool {
...
 }
 
+// The three tests below are deliberately kept even though
+// constructRequestBody's path-param-exclusion mechanism is already
+// exhaustively tested via VPN's canonical tests (executor_vpn_test.go) — that
+// mechanism is generic, but it stays correct regardless of resource name.
+// What isn't generic is whether THIS tool's own YAML-mirroring field names
+// (restDeliveryPointName, rdpConfig) are wired to the right path/body split.
+// A rename typo in tools.yaml's create/update/delete-rdp definitions would
+// pass every other test in this package and only show up here, fast, without
```
solace-aross: Same as the topic-endpoint file, and the two overlap: `create-rdp` has the identical parameter shape to `create-topic-endpoint`, so whichever of the two lands, the second covers no new combination.

Also worth a look before this settles: the restore is partial. `TestExecute_ListRDPs_SinglePage` and `TestExecute_CreateRDP_AlreadyExists` were on `main` and did not come back, and neither is mentioned in the description's list of deliberate removals. If dropping them was the intent, say so there; the audit's value rests on every deletion being an argued one.

solace-aross (dismissed):
Approved. Both items from my last pass are closed, and the coverage-floor one was answered rather than deflected.

**Verified, not taken on trust:**

| Claim | How I checked it |
|---|---|
| 85% floor, 88.6% actual | `make check` exits 0; `Total coverage: 88.6%` |
| "roughly 130 statements of headroom" | Computed from the profile: 3715 total, 3293 covered, 85% of 3715 = 3157.75, so 135. Accurate as written. |
| duplicate-alias check earns its green | Sabotaged `register.go` to emit the first alias twice. Fails on both rows: `"broker-a" appeared 2 times`, `"broker-b" appeared 0 times`. The dropped length check is fully subsumed. |
| Kafka regression actually fixed | `ListKafkaReceivers` and `ListKafkaSenders` both back at 100.0%. |
| `tail -n 1`, fail-closed awk comment | Both correct. |

The doc rewrite is the right outcome. Naming per-package deltas as the thing with the resolution the aggregate lacks, and the ratchet guidance about keeping a buffer below actual, are both better than what I asked for. The stress-test paragraph on untested packages appearing in the profile is worth having written down.

---

### One thing to fix, not blocking: the header comment on the two restored files claims a guarantee the tests cannot deliver

Detail inline. Summary: both files justify themselves on catching a `tools.yaml` rename, but they assert against hand-written `CompositeTool` literals, so `tools.yaml` cannot reach them. I injected a real wiring bug and confirmed the A/B.

I under-sold the severity in my inline note in one direction and over-sold it in another, so both corrections are there: the body-leak assertions cannot catch that bug class under *any* construction, but the failure mode is fail-closed at `client.go:227` rather than a misrouted write, and `e2e-management` runs unguarded on every PR and exercises `update-topic-endpoint`, `delete-topic-endpoint`, `update-queue`, `delete-queue`, `update-rdp`. So this is feedback latency, not a production escape. That is why it is a note and not a request for changes.

### @solace-awaheed on the queue tests

Your core point is right and it is the sharper half: this bug class passes `make check` and only surfaces in e2e. Two refinements.

First, restoring the queue tests in the same style as RDP/TE will not close the gap, because the RDP/TE pair does not close it either. Verified above.

Second, `loader_queue_test.go` does not validate the shipped YAML. It builds an inline `fstest.MapFS` copy, so it validates a duplicate. Same for `loader_topic_endpoint_test.go`. That is the cleanest way to see the actual hole: across the whole repo, exactly one test asserts real-catalog step args (`loader_embedded_test.go:112`, the `select` field of `get-replication-status`, a read tool). No write tool has its real path wiring pinned anywhere.

### Suggested remedy, which would make both the restore and the queue request unnecessary

One generic invariant over the real catalog: every path placeholder of every step's operation must be supplied by that step's `args` (or by `forEachKey`). I wrote it and ran it. On the clean catalog it passes in 0.44s, checking 54 placeholders across all 34 tools. With the bug injected it fails with:

```
update-topic-endpoint/updateTopicEndpoint (/SEMP/v2/__private_config__/msgVpns/{msgVpnName}/topicEndpoints/{topicEndpointName}): path param "topicEndpointName" not wired in args
```

About 30 lines, no per-tool maintenance, covers queue, RDP, topic endpoint, VPN, bridge and client at once, and grows automatically as tools are added. `loader_embedded_test.go` already parses the specs, so it belongs beside the existing `operations` subtest. `loader_idempotent_test.go:87-91` is the precedent for the reasoning and for the "update this guard rather than deleting it" note that keeps it from being brittle.

Treat that as a follow-up if you would rather not grow this PR. It is squarely the kind of gap the audit was for, so it is worth a ticket either way.

### Not yours

`FOSSA Scan / SCA Scan` and `SCA gate` are red at baseline on the `SolaceProducts` project, unrelated to this diff. Still unowned, and still invisible as a merge gate because `main-protection` requires only the bare `FOSSA Scan` context, which reports `skipping` and passes.

solace-aross (approved):
Approved. Both items from my last pass are closed, and the guard landed where I suggested, reusing the specs the `operations` subtest already parses.

**Verified by the same method the comment claims, rather than trusting it:**

| Check | Result |
|---|---|
| Guard passes on the clean catalog | `TestLoadTools_EmbeddedDefinitions/path-params-wired` PASS in 0.44s |
| Guard catches a real YAML wiring bug | Dropped `topicEndpointName` from `update-topic-endpoint`'s step args, the same injection I used last pass. Fails with `tool "update-topic-endpoint" step "updateTopicEndpoint" (…/{msgVpnName}/topicEndpoints/{topicEndpointName}): path param "topicEndpointName" not wired in args` |
| The hand-written tests still miss it | With that same bug in place, `TestExecute_UpdateTopicEndpoint_BodyExcludesPathParams` still passes. That contrast is the whole argument, and it now holds in code rather than in a comment |
| `make check` | exits 0, coverage 88.3% against the 85% floor (down from 88.6% because `main` merged in, still clear) |
| Catalog count | 34 tools, matching `wantToolCount`, so the guard covers every tool and step by construction |

The comment is now accurate on all four points I raised: it says these tests never load `tools.yaml`, it names the guard as the thing that does catch YAML bugs, it keeps the create-shape argument that genuinely earns them (one path param plus a scalar and an object spread into the body, which VPN's create never exercises because it has zero path params), and it corrects the "deliberately kept" framing for the update test that `main` never had.

Choosing not to add a `ForEachKey` exemption is the better call. Asserting that fan-out steps supply path params through `step.Args` today, and telling the next maintainer to update the check rather than delete it, is stricter than exempting the case up front and it prevents the confused deletion that would otherwise follow.

`PathParamNames` reads correctly: it de-duplicates, and on a malformed `{` with no `}` it stops, which matches the decision `unfilledPlaceholders` already documents for the same shape.

**@solace-awaheed's two asks are both closed, one of them differently than requested.** The Kafka recovered-case row is in and it discriminates, which I checked rather than assumed: moving `byReason[reason]++` outside the `if !up` gate, exactly the scenario you described, now fails with `got map[connection refused:3], want {connection refused:2}`. Before this row it would have passed.

On restoring the queue executor tests: the new guard supersedes that ask. It iterates every tool and every step in the real catalog, so queue, VPN, bridge and client are all covered for the failure mode you named, and covered against the shipped YAML rather than a hand-written copy. Restoring the queue tests in the synthetic style would not have closed it, which is the thing I got wrong in my first pass and worth saying plainly.

`FOSSA Scan / SCA Scan` and `SCA gate` are red at baseline on this repo and unrelated to this diff. #257 replaces that machinery and its gate fails closed, so this should stop being noise shortly.

solace-aross (approved):
Re-approved at `35b1ceb`. My approval was dismissed by a merge, not by new work: there are no PR-authored commits since `881a67f`, and diffing the two heads across this PR's own paths (`internal/`, `Makefile`, `docs/internal/unit-test-coverage.md`, `CLAUDE.md`, `build-and-test.yml`) returns nothing.

One thing I did re-run rather than wave through, because it is specific to what this PR adds. A merge is inert for your files but not necessarily for your gate: `main` has taken on the attestation work, the LLM-eval fixes and the transition-on-merge removal since I last looked, and any of that could have moved aggregate coverage under the new floor. It has not. `make test-cover` on the merged head reports **88.3%**, comfortably clear of the 85% floor and unchanged from my last measurement.

Everything else from my previous review stands, including the real-catalog path-param guard and the Kafka recovered-case row.

PR #255, opened 2026-08-05
SOL-152855: remove the transition-on-merge workflow, dead since the org move

solace-aross:
## Summary

`.github/workflows/transition_on_merge.yaml` has hard-failed on **every merged pull request since 31 July**. This deletes it.

[SOL-152855](https://sol-jira.atlassian.net/browse/SOL-152855)

> **Correction from the first version of this PR.** I originally wrote this as prevention: a workflow that *would* break on flip day. An Opus review pass caught that it broke six days ago, that the cause is the org move rather than the upcoming visibility change, and that the first people to see the red X were us. The change is the same; the reasoning and the urgency are not.

## Type of Change

## What is actually broken, and since when

| | |
|---|---|
| Last green run | `2026-07-30T21:28:40Z` |
| First failure | `2026-07-31T15:18:03Z` |
| Consecutive failures since | **16 of 16** |
| Jobs created per failing run | **0** |

The run name has degraded from `Update Ticket After Merge` to the raw path `.github/workflows/transition_on_merge.yaml` — GitHub's startup-failure signature. `gh run view` prints it plainly: *"This run likely failed because of a workflow file issue."*

Affected branches include `wmaryam/SOL-152204_describe_schema`, `awaheed/SOL-152740`, a Dependabot PR, and `lkiri/sol-152553-kin-openapi-bump` — the kin-openapi security bump merged today.

## Why

Line 29:

```
uses: SolaceDev/re-workflows/.github/workflows/transition-pr-on-merge.yaml@release/v3
```

`SolaceDev/re-workflows` is internal, and after the move it sits in a different organisation from the caller. An internal reusable workflow is reachable only from repositories its Actions access policy admits, and this caller left that scope on 31 July. Same date and mechanism as the Vault OIDC breakage already recorded in `ADMIN_SETUP.md`: the move changed the `repository` claim.

The upstream file still exists, so this is not a deletion at the other end.

**Going public is a second, independent blocker that has not fired yet.** A public repository cannot call an internal reusable workflow at all, regardless of how Actions access is configured. Both are in the commit message because fixing only the first would not be enough, and because someone reading "public repos can't load it" as the whole story might restore the file while we are still internal.

## Why no `if:` can save it

Reusable workflows are resolved when a run is queued, before any job-level `if:` is evaluated. The existing `head.repo.full_name == github.repository` guard correctly skips fork pull requests and cannot help here.

The 16 failing runs each created **zero jobs**, which is what that ordering looks like from outside: there was no job in which to evaluate a condition.

## What we give up

The automatic Jira status transition on merge — which has not worked since 31 July, so nothing currently working is being removed. Jira's GitHub integration still links the PR to its ticket, so traceability stays.

**On restoring it later, corrected.** The first version of this PR argued against reimplementing locally because it would "put a Jira credential on a public repository's merge path." That argument does not hold: this repo already does exactly that, on the strictly more exposed *pre*-merge path, in `ci-pr.yaml` (`use_vault: true`, `secrets.VAULT_URL`, on `pull_request`).

The honest argument is cost/benefit — a Vault-plus-Jira integration is not worth carrying for a status transition.

And there is a third option I missed: move the logic into `SolaceDev/solace-public-workflows`, which is **public** and which this repo already calls three times. A public reusable workflow resolves fine from a public caller, and the Vault OIDC pattern it would need is already proven here. That is the cheapest route if the squad wants the feature back.

## Changes Made

- Deleted `.github/workflows/transition_on_merge.yaml`
- Removed the `ADMIN_SETUP.md` bullet telling admins not to require `run-pull-request-checks / *`, and corrected the surrounding count from four notes to three

## Testing

Verified against the live GitHub API, not documentation.

```
$ gh api repos/SolaceDev/re-workflows --jq .visibility
internal

$ gh api repos/SolaceProducts/solace-broker-mcp/rulesets/13942241 \
    --jq '.rules[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context'
lint
build
FOSSA Scan
```

The legacy branch-protection rule was checked separately and carries an empty contexts list, so no required check is removed by either mechanism and no merge can block on a check that will never report.

`make check` passes, exit code 0. On the first local run it failed with what looked like a `staticcheck` finding; it was a stale golangci-lint cache pointing at a deleted worktree. Cleared and re-ran clean.

**Test Environment:** Go 1.26.2, macOS (darwin/arm64), broker n/a.

## Known remaining references

Two, both in committed but generated `/understand` artifacts: `.ua/knowledge-graph.json` and `.ua/fingerprints.json`. They are rewritten wholesale by the next `/understand` run. Regenerating a knowledge graph inside a workflow-deletion change would bury the diff, so they are left.

The first version of this PR claimed "no dangling references left". That grep was scoped to `*.yaml`, `*.yml`, and `*.md` and structurally could not see `.json`. Everything else is genuinely clean: nothing in scripts, CODEOWNERS, README, `docs/`, or other workflows.

## Breaking Changes

None for users. Squad-internal: merging a PR no longer transitions its Jira ticket, which has been true since 31 July.

## Checklist

- [x] Self-review completed
- [x] No credentials in code
- [x] All tests pass locally
- [x] docs/ updated (`.github/ADMIN_SETUP.md`)
- [x] CHANGELOG.md not updated, deliberately: `.github/workflows/` is not production surface per `.github/scripts/production-surface.sh`

## Additional Notes

**Sequencing with #256** (corrected — the earlier note named the wrong sibling). That PR adds `THIRD_PARTY_BUILD_TEST.md`, which lists this reusable workflow, and a gate that fails on a row describing something no longer used. I verified the interaction rather than assuming: merging the two branches is conflict-free, and the gate names the stale row precisely. Whichever lands second deletes one row.

## Related Issues/PRs

- Part of SOL-152410 (open-source release epic)
- Adjacent to #256 (SOL-152861)
- SOL-152412 owns ruleset changes

---

**For Reviewers:**

**Focus Areas**: Whether anyone wants the Jira transition back badly enough to route it through `solace-public-workflows`. If yes, that is a follow-up ticket rather than a reason to hold this — the current file cannot be made to work in either org configuration.


amitmorade (commented):
Approving. I verified the load-bearing claims against the live API rather than reading the body, and they hold.

**The workflow is dead, not merely suspect.** `gh run list --workflow transition_on_merge.yaml` returns 16 consecutive `failure` runs — last green `2026-07-30T21:28:40Z`, first failure `2026-07-31T15:18:03Z`. Your table matches exactly. Nothing currently working is being removed.

**No required check depends on it.** The `main-protection` ruleset (13942241) requires only `lint`, `build`, `FOSSA Scan`. The legacy branch protection on `main` carries `"contexts": []` and `"checks": []`. `run-pull-request-checks / *` is required by neither mechanism, so this cannot strand a PR on a check that will never report — which is the failure mode that would have made this risky.

**The removal is isolated.** Two files, no production surface. A case-insensitive grep for `transition[-_]on[-_]merge|transition-pr-on-merge|re-workflows|run-pull-request-checks` across the branch, excluding `.ua/`, returns nothing. The remaining five workflows all call `SolaceDev/solace-public-workflows` (public, SHA-pinned), so the deleted file was the only caller of the internal `re-workflows` and no shared dependency is severed. Skipping CHANGELOG is right — `.github/workflows/` is not production surface.

The `.ua/` exception is the correct call. Both files are generated wholesale by the next `/understand` run and nothing in CI reads them; regenerating a knowledge graph here would bury a 38-line diff.

The doc edit is self-consistent: "Four more notes" → "Three more notes" matches the single bullet removed, and no surviving bullet references the deleted workflow.

On the focus area: routing the Jira transition through `solace-public-workflows` is the right shape if the squad wants it back, but it belongs in its own ticket. The current file cannot be made to work in either org configuration, so there is no version of holding this that preserves the feature.

amitmorade (approved):
LGTM

PR #256, opened 2026-08-05
SOL-152861: inventory the build-time and test-time third-party components

solace-aross:
## Summary

Adds `THIRD_PARTY_BUILD_TEST.md`, the second of the two third-party lists the Solace public-repository Legal checklist requires, plus a CI gate that keeps it honest.

[SOL-152861](https://sol-jira.atlassian.net/browse/SOL-152861)

## Type of Change

- [x] Documentation update
- [x] Bug fix (non-breaking change which fixes an issue) — closes a compliance-artifact gap

## Motivation and Context

The [CPD Legal checklist](https://sol-jira.atlassian.net/wiki/spaces/CPD/pages/2841706565) asks for two lists under "No copyright infringement":

> List of all used 3rd party products that are part of the release artifacts
> List of all used 3rd party products used at build and test time.

`THIRD_PARTY_LICENSES.md` is the first. It computes the dependency closure of `./cmd/server`, by design, because that file carries the Apache-2.0 section 4(d) obligation and ships inside the tarball and the container image. Nothing covered the second list.

## Changes Made

- `THIRD_PARTY_BUILD_TEST.md` — 27 components across three sources: the Go modules of the two e2e submodules, the twelve GitHub Actions CI uses, and the five container images pulled by the build and the e2e fixtures
- `.github/scripts/build-test-licenses-check.sh` — fails on drift in either direction across all three sources
- `.github/scripts/build-test-licenses-check.test.sh` — eleven cases, each asserting the gate actually fails
- Wired both into the existing `Third-party licenses current` job in `ci-pr.yaml` and `release.yml`
- `CHANGELOG.md` entry under `[Unreleased]`

**Verdict: nothing here constrains distribution.** All 27 components are permissive (MIT, BSD-3-Clause, Apache-2.0). No strong-copyleft licence (GPL, LGPL, AGPL, EPL, CDDL) appears anywhere in the build or test path.

Every licence was read from the component's own licence file or from `gh api repos/OWNER/REPO --jq .license.spdx_id`. None was inferred from a package name or copied from a neighbouring row — that precision is the point of the artifact, and SOL-152414 is the evidence for why.

## Three findings worth a reviewer's attention

**1. The root module contributes nothing, and that is the interesting part.** `go list -deps -test ./...` on the root module resolves to exactly the same external module set as `go list -deps ./cmd/server`. Every test dependency already ships and is already listed.

That is the reverse of the error in SOL-152414, where `testify`, `go-spew`, and `go-difflib` were excluded from the release inventory as "test-only" and in fact reach the binary via a file named `issue_test_1.25.go` that Go compiles as ordinary package code. Both directions of that assumption have now been wrong once, so the file states the result rather than leaving an empty section a reader would read as an oversight.

**2. `golang.org/x/oauth2` is pinned at two versions in this repository.** v0.35.0 in `test/e2e-basic-mcp/agent`, v0.36.0 in the root module. Separate `go.mod` files, so nothing forces them to agree. Both BSD-3-Clause, so there is no licensing consequence, but it is real and is now recorded instead of being smoothed to one row.

**3. The distroless base is the one build input that ships, and neither inventory enumerates it.** `gcr.io/distroless/static-debian12:nonroot` is the runtime base of the container image we publish, so its layers are distributed to users. `THIRD_PARTY_LICENSES.md` covers Go modules linked into the binary and does not describe base-image OS packages.

This is a gap in the *release* inventory, not this one, and I have deliberately not expanded scope to close it here. Distroless `static` is minimal — no shell, no package manager, essentially CA certificates, timezone data, and a passwd file — which is why it has not bitten. The file says so plainly rather than implying coverage it does not have. Worth its own ticket if a reviewer agrees.

## Testing

**The gate is proven to fail, not merely observed to pass.** Eleven cases, each mutating a copy of the inventory and asserting an exit code:

```
$ .github/scripts/build-test-licenses-check.test.sh
build-test-licenses-check.sh self-test
  ok       committed inventory passes unmodified (exit 0)
  ok       missing Go module row fails (exit 1)
  ok       wrong Go module version fails (exit 1)
  ok       Go module row for an unused module fails (exit 1)
  ok       missing row for a module that also ships fails (exit 1)
  ok       missing action row fails (exit 1)
  ok       action row for an unused action fails (exit 1)
  ok       missing container image row fails (exit 1)
  ok       image row for an unreferenced image fails (exit 1)
  ok       a row the strict parser skips is reported, not ignored (exit 1)
  ok       a fully unparseable table fails loudly (exit 1)

11 passed, 0 failed
```

The last two are the shapes `licenses-check.test.sh` found as real silent passes in its own subject after review. They are asserted here from the start rather than after relearning the lesson.

**Everything else:**

```
$ .github/scripts/build-test-licenses-check.sh
✅ THIRD_PARTY_BUILD_TEST.md matches the build and test inputs (27 component(s) checked, 27 rows parsed).

$ .github/scripts/licenses-check.sh        # unchanged, still passes
$ .github/scripts/licenses-check.test.sh   # unchanged, still passes
$ make check                                # exit 0
```

**Test Environment:**
- Go version: 1.26.2
- OS: macOS (darwin/arm64)
- Broker version: n/a

**Tests Performed:**
- [x] Unit tests added/updated (the self-test)

Not run: `shellcheck`, not installed locally. CI does not currently lint shell scripts either — that gap is SOL-152856.

## Breaking Changes

None. New file, new script, two added CI steps in an existing job.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code

### Testing
- [x] All tests pass locally
- [x] Error paths tested

### Documentation
- [x] docs/ updated
- [x] CHANGELOG.md updated

### Git & CI

## Additional Notes

**Design decision, since it is the one a reviewer would most reasonably challenge.** Several submodule dependencies also ship and therefore appear in both files. That is deliberate. The checklist asks for a list of all products used at build and test time, not the delta against the release list, and a reader should not have to diff two files to satisfy it. The root module is the exception, and only because its closure is *identical* — repeating all 31 rows would create two lists to keep in sync for no gain.

I built the gate the other way first, skipping anything already in the release inventory. The gate immediately failed against my own file, which is how the contradiction surfaced. Recording that because it is the kind of thing a green-from-the-start gate would have hidden.

**Sequencing with #255.** That PR removes `transition_on_merge.yaml`. This file lists its reusable workflow. Whichever lands second drops one row; the gate will say exactly which.

## Related Issues/PRs

- Part of SOL-152410 (open-source release epic)
- SOL-152860 needs this to complete the Legal checklist
- DATAGO-140399 (Balazs) covers FOSSA auto-pushing attribution output; see the Drift section for how the two relate
- Adjacent to #255 (SOL-152855)

---

**For Reviewers:**

**Focus Areas**: The three findings above, particularly whether the distroless gap deserves its own ticket now. Also the `sed -E 's/:[^:]*$//'` tag-stripping in the image check — it assumes no registry port in an image reference, which is true for every image we use but would break on `localhost:5000/foo`.

**Questions**: Is a separate file the right call versus a second section in `THIRD_PARTY_LICENSES.md`? I chose separate because that file ships and this one must not be mistaken for something carrying a distribution obligation. Happy to be argued out of it.


solace-aross:
Verified the interaction with #255 rather than leaving it asserted in the description.

Merging the two branches is conflict-free, and the new gate catches the row that goes stale:

```
::error file=THIRD_PARTY_BUILD_TEST.md::`SolaceDev/re-workflows/.github/workflows/transition-pr-on-merge.yaml` is listed but nothing in the repository uses it any more. Drop the row.
```

So whichever of the two lands second deletes one row and CI says exactly which. No coordination needed beyond that.

solace-aross:
An Opus review pass ran the gate against mutated **inputs** rather than a mutated document — the dimension the self-test never exercised. It found five drift shapes that passed silently and one factual error that falsified the file's own conclusion. All fixed in `dce6861`.

## The inventory was wrong, not just incomplete

`test/e2e-llm/package.json` pins `@anthropic-ai/claude-code`, and `llm-eval.yml` installs it with `npm ci`. Nine packages, none listed, and their declared licence is `SEE LICENSE IN README.md` — Anthropic's commercial terms, not an OSI licence.

So the file asserted "all components are permissive" while omitting the one source that is not. That is the SOL-152414 defect class exactly: a component missing from a compliance artifact because nobody looked at that source. Uncomfortable to find it in the PR whose whole point is closing that class of gap.

Fixed: npm is now a fourth gate source with its own section, and the Verdict says what is actually true — nothing here constrains how we license or distribute the product, with the three qualifications that claim needs (not everything is open source; one entry ships; image layers are out of scope).

## Gate fixes

| Defect | Why it mattered |
|---|---|
| `- uses:` steps invisible | `^[[:space:]]*uses:` cannot match `- uses: foo@v1`. This repo writes `- name:` / `uses:` throughout, so a new third-party action entering CI would have reported green |
| Versions unchecked for actions and images | Only Go modules were compared. Dependabot bumps Actions **daily** — the highest-frequency drift source was the unchecked one |
| Discovery hardcoded | Two submodule paths, one Dockerfile. A third submodule or second Dockerfile would add components the gate cannot see. Now derived with `find -L` |
| `go list` failure swallowed | `2>/dev/null` inside process substitution is invisible to `set -e` and `pipefail`. The closure emptied and the reverse check then told the maintainer to **delete every legitimate row** |
| Three greps lacked `\|\| true` | Empty extraction killed the script at an assignment: exit 1, no output, indistinguishable in CI from real drift. The header claimed "every grep" and was wrong |
| Image filter failed open | `^[a-z]+$` dropped any registry-less name with a hyphen — `eclipse-temurin`, `open-liberty`. Replaced with an explicit exclusion list |

## Self-test fixes

The "fully unparseable table fails loudly" case was **vacuous**. It asserted only an exit code, and the bug it exists to catch — a dropped `|| true` — also exits 1, silently. It now captures output and asserts on the message.

Worse: the harness symlinked its fixtures while the gate used `find` without `-L`, so several cases were passing for the wrong reason. Fixed in the gate rather than the harness, because a checkout can legitimately symlink subtrees and a gate that sees nothing through a symlink fails **open**.

Six new cases: npm both directions, action and image versions, the `- uses:` form, a second Dockerfile. **Eleven cases became seventeen**, and the baseline now genuinely passes rather than appearing to.

## CI

Both new steps get `if: always()`. GitHub skips later steps once one fails, so a red `licenses-check.sh` would have stopped them running at all — one gate masking the other, the exact coupling the separate script exists to avoid.

## Verification

```
$ .github/scripts/build-test-licenses-check.test.sh
17 passed, 0 failed

$ .github/scripts/build-test-licenses-check.sh
✅ THIRD_PARTY_BUILD_TEST.md matches the build and test inputs (36 component(s) checked, 36 rows parsed).

$ .github/scripts/licenses-check.sh        # unchanged, passes
$ .github/scripts/licenses-check.test.sh   # unchanged, passes
$ make check                                # exit 0
```

36 components, up from 27.

## Not changed, with reasons

- **`normalize_version`.** Correctly identified as not needed here: this document records versions literally on both sides, so a pseudo-version is satisfiable by recording it verbatim, and it fails closed the first time.
- **Two submodules converging on one module at different versions.** Would deadlock — no set of rows can satisfy both. Fails closed, and does not occur today. Left as-is rather than adding machinery for a hypothetical.
- **~35 lines shared with `licenses-check.sh`.** Real duplication. Extracting a shared library touches the gate protecting the *shipping* artifact, which I would rather not do in the same change. Worth its own ticket.

solace-aross:
All three addressed. Thanks — the second one was a real defect, not a wording nit.

**Verdict contradicted Scope** — fixed in `dce6861`. The Verdict now leads with the question the file actually answers (nothing here constrains how we license or distribute the product) and carries three explicit qualifications: not everything here is open source, one entry does ship, and image layers are out of scope.

**"none of them is redistributed by us"** — fixed in `a413496`. You were right, and it was worse than inconsistent prose: the document said the correct thing while the script's own header said the opposite. `gcr.io/distroless/static-debian12` is the runtime base layer of the image we publish, so it is redistributed. The header now states the exception and points at Scope rather than offering a tidy sentence someone could take as a compliance conclusion.

**Registry port mistaken for a tag** (the suppressed comment) — fixed in `a413496`. Splitting now happens on the last path segment, so a colon before the final `/` is treated as a port. Digest pins split on `@`.

```
localhost:5000/foo         -> image=localhost:5000/foo                 tag=(untagged)
localhost:5000/foo:v1      -> image=localhost:5000/foo                 tag=v1
golang:1.25-alpine         -> image=golang                             tag=1.25-alpine
foo@sha256:abc123          -> image=foo                                tag=sha256:abc123
```

That one needed a test that asserts on the *message*, not the exit code: the bug renames a component rather than dropping it, so the gate exits 1 either way and only the reported name tells a correct gate from a broken one. 18 cases now, all green, and the gate still passes on 36 components.

solace-aross:
Caught the three suppressed comments from the later review passes as well. All fixed in `3f64e99`.

None of them changed what the gate *detects*. All three changed what it tells a maintainer to *do*, and each pointed somewhere wrong — which is the kind of defect that survives every test.

**The licence hint named the workflow file, not the repository.** `basename` on `SolaceDev/solace-public-workflows/.github/workflows/sca-scan-and-guard.yaml` produced `gh api repos/SolaceDev/sca-scan-and-guard.yaml`, a repo that does not exist. Owner/repo is the first two path segments:

```
before:  gh api repos/SolaceDev/sca-scan-and-guard.yaml
after:   gh api repos/SolaceDev/solace-public-workflows
```

Now covered by a case that asserts on the hint text itself. A wrong hint costs the reader a detour and teaches them to distrust the message, so it is worth a test.

**The parse-failure message demanded four columns.** The parser reads only the first two, and two tables here carry three by design (internal workflows, and one other). A reader following that message would have reformatted rows that were already correct. It now states what is actually required and says the rest is free-form.

**"Any prefix" was not what the code did.** The SHA comparison requires at least seven characters. That floor is deliberate — shorter prefixes collide often enough that a stale row could match a new pin by accident — so the comment now says that rather than describing behaviour the code does not have.

19 cases, all green, gate still passing on 36 components.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
...inferred. Nothing found constrains how we license or distribute the product — every Go module and GitHub Action is permissive (MIT, BSD-3-Clause, Apache-2.0) — but the file deliberately does not claim universal permissiveness: the Claude Code CLI used by the LLM eval suite ships under Anthropic's commercial terms and the Solace broker test fixture is proprietary, both being tools we run rather than code we link or redistribute. Kept honest by `.github/scripts/build-test-licenses-check.sh`, which fails CI on drift in either direction across all four sources, including component *versions*, with a seventeen-case self-test. Discovery is derived with `find` rather than hardcoded, so a new submodule, lockfile, or Dockerfile cannot introduce components the gate is blind to. Two further facts surfaced and recorded rather than smoothed over: `golang.org/x/oauth2` is pinned at two different versions across the repository's modules (no licensing consequence, both BSD-3-Clause), and the distroless runtime base is the one build input that reaches a published artifact, so its OS-package layers are a gap in the *release* inventory that neither file currently enumerates. Tracked under SOL-152861.
```
amitmorade: **Pre-existing on `main`, noting it here since this PR touches the file:** `[Unreleased]` has two `### Added` blocks (86575f8) — they arrived from separate PRs that each opened the section. Not caused by this change.

Worth collapsing into one, ideally before the next release: `cut-release` promotes `[Unreleased]` wholesale, and a repeated heading tends to quietly malform a release note rather than error.

.github/scripts/build-test-licenses-check.test.sh
```diff
@@ -0,0 +1,278 @@
+#!/usr/bin/env bash
```
amitmorade, line 1: One gap worth noting, non-blocking: the pass path is only ever exercised against the committed file. The fifteen mutation cases all move *away* from valid, so nothing covers the gate staying quiet on a change that's legitimately correct — bump a version in the repo and update its row, expect exit 0. That's the direction maintainers hit often, and it's the one the baseline is carrying alone.

THIRD_PARTY_BUILD_TEST.md
```diff
@@ -0,0 +1,234 @@
...
+
+### `test/e2e-basic-mcp/agent`
+
+A standalone module. It builds the MCP client that drives the end-to-end suite.
+
+| Component | Version | License | License text |
+|---|---|---|---|
+| `github.com/google/jsonschema-go` | v0.4.2 | MIT | [license](https://github.com/google/jsonschema-go/blob/v0.4.2/LICENSE) |
+| `github.com/modelcontextprotocol/go-sdk` | v1.5.0 | Apache-2.0 | [license](https://github.com/modelcontextprotocol/go-sdk/blob/v1.5.0/LICENSE) |
+| `github.com/segmentio/asm` | v1.1.3 | MIT | [license](https://github.com/segmentio/asm/blob/v1.1.3/LICENSE) |
```
amitmorade, line 74: One licence correction: `github.com/segmentio/asm` is listed as MIT, but its LICENSE opens with "MIT No Attribution" — that's MIT-0, a distinct SPDX identifier (`gh api repos/segmentio/asm --jq .license.spdx_id` agrees).

No distribution consequence — MIT-0 drops the attribution requirement, so it's more permissive than what's recorded. But the file's stated rule is that every licence was read from source rather than inferred, and MIT-0 → MIT is the normalisation that rule exists to catch.

Worth noting it sits in the one dimension the gate structurally can't check — name and version only, never licence — so the row is in sync and still wrong. Not an argument against the gate, just its documented limitation showing up on the first artifact.

Spot-checked seventeen other rows against their LICENSE files and the API: no other discrepancies, including `modelcontextprotocol/go-sdk`, where NOASSERTION comes from the relicensing preamble and your row is more precise than the API can be.

.github/scripts/build-test-licenses-check.sh
```diff
@@ -0,0 +1,297 @@
...
+    [[ "$image" =~ $IMAGE_EXCLUDE_RE ]] && continue
+    [[ "$image" =~ $IMAGE_OURS_RE ]] && continue
+    expect "$image" "$tag" "Container image" "Name the licence of the project that publishes it."
+done < <(
+    {
+        while read -r df; do
+            [ -n "$df" ] || continue
+            # `FROM x AS y` — keep the image, drop the stage alias.
+            { grep -hoE '^FROM[[:space:]]+[^[:space:]]+' "$df" || true; } | sed -E 's/^FROM[[:space:]]+//'
+        done < <(find -L . -name 'Dockerfile*' -type f -not -path './.git/*' 2>/dev/null | sort || true)
```
bczoma: **Minor, optional.** `-not -path` filters results but doesn't stop `find` descending, so this walks gitignored trees that hold *other branches'* checkouts. In my working copy:

```
./.claude/worktrees/SOL-152543+sign-release-artifacts/Dockerfile
./.claude/worktrees/sol-148441-config-example/Dockerfile
./Dockerfile
```

Both extra paths are gitignored (`.gitignore:31-33`) precisely because they're separate worktrees. A prune clause fixes it and also stops the descent rather than cleaning up afterwards, which matters more given the deliberate `-L`:

```bash
find -L . \( -path ./.git -o -path ./.claude -o -path ./.worktrees -o -name node_modules \) -prune \
     -o -name 'Dockerfile*' -type f -print 2>/dev/null | sort
```

That drops the output to `./Dockerfile`. The same prune arguably belongs on the two `test/` discoveries — a stray `package-lock.json` under `node_modules` would inject phantom rows once anyone runs the LLM eval suite locally. Verified it still finds all three real inputs (`test/e2e-basic-mcp/agent/go.mod`, `test/e2e-common/broker-driver/go.mod`, `test/e2e-llm/package-lock.json`).

Severity is genuinely low: it fails closed, and CI checkouts have neither directory, so this is a local-run-only false positive — and only when a worktree's Dockerfile differs from main's. Flagging it because the argument for derived discovery is that it behaves predictably as the tree grows, and worth a short comment saying *why* the prune is there so it doesn't get simplified back.

.github/scripts/build-test-licenses-check.sh
```diff
@@ -0,0 +1,297 @@
...
+
+It is one of the two third-party lists the Solace public-repository Legal
+checklist requires, so a mismatch is a compliance defect rather than a
+formatting nit.
+
+If an error says a row is unused, check that the corresponding check above did
+not fail first: a broken 'go list' empties a whole closure and every row it
+covers then looks unused. Fix the upstream error before deleting any row.
+
+To rebuild the expected sets by hand:
```
bczoma: **Requesting a change here.** These four commands are the only rebuild procedure for `THIRD_PARTY_BUILD_TEST.md`, and they exist solely inside this heredoc — reachable only by failing the gate.

`THIRD_PARTY_BUILD_TEST.md` has no code block at all. Its Drift section describes the four sources in prose but never gives a runnable command, and nothing outside the two workflows and `CHANGELOG.md` references either the file or this script.

That's an asymmetry with the file this one is modelled on. `THIRD_PARTY_LICENSES.md` puts its generator in its own header — `go run github.com/google/go-licenses@v1.6.0 csv ./cmd/server` — visible to anyone who opens it, no failure required. Here, someone asked to rebuild the inventory for an audit opens the doc, finds prose and no command, and either reverse-engineers a 297-line script or asks you.

Suggest lifting these into a short "Rebuilding this file" section in `THIRD_PARTY_BUILD_TEST.md`, keeping the note that licences are read by hand and never inferred, and having this heredoc point at that section instead of duplicating it. One copy to keep correct.

Two reasons to do it in this PR rather than later. First, the file's own standard is to state things plainly rather than imply coverage it doesn't have, and "how to regenerate" is the one thing a compliance artifact's reader most needs. Second, the `Dockerfile` one-liner three lines below carries the same non-pruning behaviour as the discovery above — so today there are already two copies and both are wrong in the same way.

bczoma (changes requested):
Two comments, both narrow. The mechanism itself reads well — discovery derived with `find` rather than hardcoded, drift caught in both directions, the row-parse check running *before* the inventory checks, and a self-test that asserts the gate fails rather than observing that it passes. The `if: always()` reasoning on the CI steps is right and the comment explaining it saves the next person from "simplifying" it away.

Requesting changes for one thing only, and it's a docs move rather than a code change: the rebuild procedure currently exists only on the failure path. The `find` prune below is optional — take it or leave it.

bczoma:
The gate went red after main merged in, and it's correct — first live catch.

`actions/attest-build-provenance@v4` came from #246 (mine), which added it to `release.yml` at two call sites. Build-time only: it runs on the runner to produce attestations, isn't linked and isn't redistributed, so it needs a row here and nothing in `THIRD_PARTY_LICENSES.md`. MIT, confirmed from the LICENSE file itself as well as the API:

```
| `actions/attest-build-provenance` | v4 | MIT | [license](https://github.com/actions/attest-build-provenance/blob/main/LICENSE) |
```

Alphabetically first in the Actions table. `CHANGELOG.md` says "fourteen GitHub Actions and reusable workflows" — that becomes fifteen. Since the script accumulates failures rather than exiting at the first, that one row is all of the drift #246 introduced.

**One thing worth changing beyond the row.** The same stale row also failed the self-test (`18 passed, 1 failed`) — the baseline "committed inventory passes unmodified" case, the only one that isn't hermetic. So one missing row produces two red steps, and the first thing a maintainer reads is *the self-test is failing*, which normally means the checker is broken. The real diagnosis is buried in that case's captured output.

That's this gate's own documented limitation showing up live, and the fix is cheap: either pin the baseline case to a frozen fixture so self-test health decouples from inventory freshness, or have it say "the committed inventory is stale — this is the gate doing its job, not a self-test defect." Same territory as @amitmorade's note that the pass path is carried entirely by the live repo.

Sequencing note, since it cuts both ways: had this merged before #246, mine would have been the PR going red and I'd have added the row there — which is the intended steady state. But `Third-party licenses current` isn't a required check on main today (only `lint`, `build`, `FOSSA Scan` are), so #246 could have merged red and put main into exactly the state this file exists to prevent. Worth its own ticket; not yours to fix here.

.github/scripts/build-test-licenses-check.sh
solace-aross: Done in 1e440b8. The four commands now live in a "Rebuilding this file" section in `THIRD_PARTY_BUILD_TEST.md`, linked from the header so it is visible on open, and this heredoc points there instead of carrying its own copy.

Two things your comment surfaced that I had not seen.

**The single copy was not faithful.** The Go and Actions commands were missing filters the script applies: `grep -v '^\./'` for this repository's own workflow calls, and the exclusion of our own module paths. Run as documented they printed a `./.github/workflows/build-and-test.yml` line and two self-module lines, none of which may become rows. Someone rebuilding for an audit would have added them and then hit "listed but nothing in the repository uses it any more". Both filters are in, and I ran all four commands against the tree: they now print exactly what the tables contain.

**The `Dockerfile` one-liner three lines below carried the same non-pruning behaviour**, as you said. It is the pruned form now, and there is one copy of it rather than two.

.github/scripts/build-test-licenses-check.sh
solace-aross: Taken rather than left. Reproduced it first: with `.claude/worktrees/fake/Dockerfile` present, the old form returned that path alongside `./Dockerfile`, the pruned form returns only `./Dockerfile`. The same applies to `test/e2e-llm/node_modules`, where the `package-lock.json` discovery had no exclusion at all, so a local `npm ci` would have injected phantom component rows.

All three discoveries now go through one `discover()` helper, so the prune list has a single definition instead of three.

One correction to the reasoning, because it changes what a maintainer might do with it. The descent is not caused by `-L`. `find` walks a real excluded subdirectory whether or not `-L` is given, and I checked both ways. That matters: anyone who believes `-L` is the cause has an obvious cheaper fix available, dropping `-L`, which does nothing here and reintroduces the symlink fail-open that `-L` exists to prevent. The comment now states the two properties separately and says not to fuse them.

There is also a self-test case now asserting that a sibling worktree's `Dockerfile` is ignored, so the prune cannot be simplified back without turning the suite red.

THIRD_PARTY_BUILD_TEST.md
solace-aross: Checked this closely and I am keeping MIT. Your reading of the LICENSE text and the API is accurate for what they show; both describe the default branch rather than v1.1.3.

v1.1.3 is MIT. Upstream relicensed to MIT-0 in v1.2.1 on 2023-11-07, about two years after the version we pin.

| Route | Result |
|---|---|
| `raw.githubusercontent.com/segmentio/asm/v1.1.3/LICENSE` | `MIT License`, `Copyright (c) 2021 Segment`, attribution clause present |
| `$(go env GOMODCACHE)/github.com/segmentio/asm@v1.1.3/LICENSE` | byte-identical to the above |
| `gh api repos/segmentio/asm/license?ref=v1.1.3 --jq .license.spdx_id` | `MIT` |
| `THIRD_PARTY_LICENSES.md` | `MIT` for the same module and version, generated by `go-licenses` |

Your own command returns MIT once it carries a ref. Without one it answers for the default branch, which is the trap.

Worth noting the direction too: recording MIT-0 here would have understated an obligation rather than overstated one, and it would have contradicted the licence link in the same row.

The row is unchanged, but the comment earned more than a no-change. I added the relicence as a note, because the check compares names and versions and never licences, so a bump past v1.2.0 changes what this column should say without turning CI red. 1e440b8.

.github/scripts/build-test-licenses-check.test.sh
solace-aross: Added in 1e440b8, and you were pointing at more than I first credited.

The case I wrote first, bump a version and update its row expecting exit 0, buys no coverage of the version comparison itself: the baseline already exercises the exact-match branch on every action row. What it does buy is a control on the workflow-copy mutations. It is the only exit-0 case that replaces the symlinked workflows, so a truncated or corrupted `unlink_workflows` surfaces there and nowhere else, while the three cases that mutate workflows all expect exit 1 and stay green against a broken copy. The comment says that now instead of overclaiming.

The pass path that was genuinely unguarded is the short-SHA floor. `version_matches` accepts a documented prefix of at least 7 characters against a 40-character pin, and I could weaken that floor to 1 with the whole suite still green, which would let a one-character stale row satisfy any future re-pin. Three cases cover it now: a longer correct prefix passes, a correct prefix below the floor fails, a wrong prefix of acceptable length fails. Each one turns red against the matching mutation.

CHANGELOG.md
solace-aross: Collapsed in 1e440b8.

One correction to the mechanism, since it affects how much this needs to worry anyone. Nothing is lost today. `extract-release-notes.sh` matches on `## [` boundaries and is agnostic to `###`, so both blocks and all four entries reach the published body intact, with the heading merely repeated. I ran it against a two-block fixture to confirm.

Where the concern does land is the curated summary. `cut-release` has an LLM draft `.github/release-notes/vX.Y.Z.md` from the promoted block "grouped by the same categories", and its verification checklist has no check for a duplicate category heading. That is the artifact that could quietly drop an entry, and it is the one with no gate on it.

Fixed here rather than deferred because it is two deleted lines in a file this PR already touches, and we are nine days past v0.6.0 on a two-to-three-week cadence.

THIRD_PARTY_BUILD_TEST.md
```diff
@@ -0,0 +1,174 @@
...
+Kept honest by `.github/scripts/build-test-licenses-check.sh`, which fails CI
+when this file stops matching what the repository actually uses. See
+[Drift](#drift) for what it does and does not cover.
+
+## Verdict
+
+Nothing here constrains how we license or distribute the product. All components
+are permissive (MIT, BSD-3-Clause, Apache-2.0), and none of them ships. No
+strong-copyleft licence (GPL, LGPL, AGPL, EPL, CDDL) appears anywhere in the
+build or test path.
```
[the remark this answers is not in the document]
solace-aross: Already addressed. The Verdict section now carries an explicit "One entry does ship" paragraph naming `gcr.io/distroless/static-debian12` and linking to Scope, rather than claiming nothing here ships.

.github/scripts/build-test-licenses-check.sh
```diff
@@ -0,0 +1,206 @@
...
+# one for what ships, and one for what we use at build and test time. The first
+# is THIRD_PARTY_LICENSES.md, gated by licenses-check.sh. This gates the second.
+#
+# It exists as a separate script rather than another section of licenses-check.sh
+# because the two answer different questions against different inputs, and the
+# failure of one should not mask the other. licenses-check.sh reasons about the
+# dependency closure of a single main package and the Apache-2.0 4(d) NOTICE
+# obligation. This one reasons about three unrelated sources — submodule closures,
+# workflow `uses:` references, and container image tags — none of which has a
+# NOTICE obligation because none of them is redistributed.
```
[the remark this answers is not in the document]
solace-aross: Already addressed. The script header states the distroless exception explicitly rather than claiming nothing is redistributed.

amitmorade (commented):
LGTM, @bczoma to approve

.github/scripts/build-test-licenses-check.sh
```diff
@@ -0,0 +1,329 @@
...
+    { grep -nE '^\| `' "$DOC" || true; } | { grep -vE '^[0-9]+:\| `[^`]+` \| [^ |]+ \|' || true; } >&2
+fi
+
+documented=$(
+    { grep -oE '^\| `[^`]+` \| [^ |]+ \|' "$DOC" || true; } |
+        sed -E 's/^\| `([^`]+)` \| `?([^ |`]+)`? \|$/\1 \2/' |
+        sort -u
+)
+documented_names=$(awk '{print $1}' <<<"$documented" | sort -u)
+
```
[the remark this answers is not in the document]
solace-aross: Confirmed and fixed in 49fe91b. This one was real, and it reproduces.

```
| `apache/kafka` | 3.7.0 | ...   (correct, in use)
| `apache/kafka` | 4.9.9 | ...   (stale, appended)   -> gate exits 0
```

The reverse-direction check cannot see it either, because it compares names and the name is legitimately in use. So a stale row sits beside a correct one with CI green.

`doc_version_of()` is unchanged. A duplicate name is now rejected outright, checked alongside the row-parse check and before anything reads a version, since a name resolving to the wrong row makes every verdict downstream meaningless. The error names the versions in conflict, and says that two modules genuinely needing different versions means extending the check rather than duplicating the row.

Worth recording that the bug is asymmetric, because it changes what a test has to assert. A duplicate whose version sorts *before* the real one already failed, incidentally, on the version comparison. Only the sorts-after case failed open. A single case would therefore have passed against the unfixed script, so there are two, one per direction, and I verified that removing the new check turns exactly one of them red.

bczoma (approved):
Approved, thanks for the updates!

PR #257, opened 2026-08-05
ci(DATAGO-147232): enroll solace-broker-mcp in Guardian without Vault

artyom-morozov:
## What

Enrolls this repo in **Guardian** and moves FOSSA + Prisma scanning off the Vault-backed reusable workflow to a native pipeline that uses **GitHub Environment (`guardian`) secrets** instead of Vault — required because this repo is public.

- **`.github/actions/guardian-scan`** (composite): builds the image, runs FOSSA SCA + license gate, Prisma scan, and (on trunk) uploads to Guardian, syncs the DB/Jira, and runs the vulnerability gate.
- **`guardian-scan.yaml`**: PR → report/diff only; **push to `main` → upload to Guardian + block on its verdict**.
- **`release-readiness-check.yaml`**: pre-release gate (also called by `release.yml`), with admin-only `skip_security_checks` bypass.
- Dropped the Vault FOSSA jobs from `build-and-test.yml` / `ci-pr.yaml`, deleted `fossa-scan.yaml` and `workflow-config.json`, and pointed `release.yml` at the readiness gate.
- SHA-pinned all third-party actions (kept `SolaceDev/*` at their existing pin).
- Updated `ADMIN_SETUP.md` / `CONTRIBUTING.md` / `OPEN_SOURCE_STATUS.md`.

## Already done (out of band)

- Product `solace-broker-mcp` registered in Guardian under squad **`broker`**.
- `guardian` Environment created with the six secrets (`FOSSA_API_KEY`, `GUARDIAN_URL`, `GUARDIAN_API_TOKEN`, `PRISMACLOUD_ACCESS_KEY_ID`, `PRISMACLOUD_SECRET_KEY`, `PRISMACLOUD_CONSOLE_URL`), no required reviewers, no branch restriction.

## Admin follow-ups

- **Branch protection:** require **`Guardian scan gate`** (drop the old `SCA gate` / `FOSSA Scan`). See `ADMIN_SETUP.md → Guardian Scanning Setup`.
- **Waive the pre-existing FOSSA license findings** (`marshmallow` CC-BY-SA-4.0, two MPL-2.0) on the `SolaceProducts_solace-broker-mcp` FOSSA project, or the license BLOCK will legitimately block `main`.
- Out of scope: `transition_on_merge.yaml` still uses Vault for Jira transitions.


.github/workflows/release.yml
```diff
@@ -128,64 +130,32 @@ jobs:
...
 
       - name: Upload artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: binary-${{ matrix.goos }}-${{ matrix.goarch }}
           path: ${{ env.ARCHIVE }}
 
   build-docker:
     needs: [test, release-notes, licenses]
```
solace-aross: Add `release-readiness` here:

```yaml
needs: [test, release-notes, licenses, release-readiness]
```

As written the image reaches `ghcr.io` whether or not Guardian blocks, and `latest` moves with it. Only the GitHub Release waits. A registry push cannot be withdrawn; the binary archives are only artifacts until `release` publishes them, so the one irreversibly-published artifact is the one left ungated, in the PR whose purpose is gating.

Two reasons this is worth a round trip rather than a note. `release.yml` triggers only on `push: tags: v*`, so no PR run and no CI anywhere surfaces it before a real tag; the first person to find it is whoever tags and watches `latest` move while `release` fails. And you are already rewriting this exact wire, deleting `fossa_scan` and writing its replacement, so this is finishing the substitution rather than fixing an unrelated pre-existing defect. I would not ask for attestations or an SBOM here.

Fail-closed also cannot wedge an emergency release, because you already built the escape hatch: an admin can set `skip_security_checks` and publish over a Guardian block. That combination, fail closed by default with an audited admin override, is the right shape.

For the record, `origin/main` has the same ungated shape today, so this is not a regression you introduced. It is a defect the substitution should carry across.

solace-aross (changes requested):
Good work, and the part I have been waiting months for is here: the gate fails closed. Requesting changes on one thing, plus a sibling-PR interaction that will bite whoever merges second.

**Credit where it is due, verified at `493f144`:**

- The aggregate `gate` job requires `success` from `build`, `fossa-license`, `fossa-vuln` and `prisma`, plus `guardian-gate` on trunk, under `if: always()`. That is the exact trap the outgoing setup fell into, where the required `FOSSA Scan` context reported `skipping` and therefore counted as passing while `FOSSA Scan / SCA Scan` failed unrequired. Every PR in this repo could merge on a red supply-chain verdict. This fixes it.
- `ADMIN_SETUP.md`'s "Require `Guardian scan gate`, and nothing FOSSA-shaped" is the right instruction and the thing that actually closes the hole.
- The `skip_security_checks` bypass is genuinely enforced: `getCollaboratorPermissionLevel` on `context.actor`, honoured only when `permission === 'admin'`, and it warns either way.
- `pull_request` rather than `pull_request_target`, fork PRs skipped rather than handed secrets, `persist-credentials: false`, third-party actions SHA-pinned.

I also dropped a finding before writing this: I thought the six `guardian` secrets were over-exposed to PR runs given upload only happens on trunk, but `fossa-license`, `fossa-vuln` and `prisma` all need their own credentials on the PR path, so the environment has to be reachable there.

---

### Blocking 1: the image publishes without waiting for the security verdict

Inline. `build-docker` needs `[test, release-notes, licenses]` and passes `push: true`; `release-readiness` is absent from that list and has no `needs:` of its own, so it runs in parallel and gates only `release`.

I looked for a gate elsewhere before raising it. `build-image.yaml`'s `if: ${{ inputs.push }}` conditions switch modes rather than gate on a verdict; there is no `environment:` on `build-docker` or on `build-image.yaml`'s `build` job, so no protection rule covers the push path.

### Blocking 2: this will conflict with #246, and the easy resolution drops an approved security fix

#246 is approved and pending. Both branches share merge-base `e58feb01` and neither contains the other, so they are siblings and the conflict is guaranteed: #246 rewrites `build-docker`'s `needs:` line to add `fossa_scan` and puts a rationale comment directly above it, while this PR keeps that line and replaces the ~45 inline docker steps beneath it with the reusable-workflow call. Same hunk.

The trap is that the mechanical resolution compiles. `fossa_scan` does not exist here, so "keep the new version" is the only resolution that parses, and it silently discards #246's gating plus its `attestations`/`id-token` provenance block, for which `build-image.yaml` has no equivalent. The result is valid YAML that passes every check.

Fixing Blocking 1 inside this PR is what makes the resolution safe in either merge order, which is the main reason I am asking for it here rather than filing it.

---

### Notes, not blocking

**Doc drift from the parallel-jobs restructure in `f36a6a3`.** `ADMIN_SETUP.md`'s required-check table still says job `guardian-scan-gate`, now `gate`, and the secrets section still refers to "the `Guardian scan` job" singular where there are now five. The left column, the check name an admin actually types into branch protection, is correct, which is the part that matters.

**Fork PRs merge with no SCA verdict.** You document that the scan skips and the gate accounts for it, so this is the consequence rather than the mechanism: today almost every PR is same-repo and the carve-out rarely fires, but once the repo is public, fork PRs become the normal contribution path and each one merges unscanned. `OPEN_SOURCE_STATUS.md`'s "require approval for outside collaborators" protects the secrets, not the verdict. Worth an owner before the repo flips rather than after.

**`const skip = ${{ inputs.skip_security_checks }};`** in the `authorize` job is safe only because the input is typed `boolean` in both trigger blocks. Passing through `env:` is the robust idiom, and this is the one control standing between a Guardian block and a publish.

solace-aross:
Correction to my review above. I credited this PR's gate with fixing the "a skipped scan counts as a pass" hole. That was overstated, and the distinction matters for what this PR is actually buying.

`main` already has a fail-closed gate. `ci-pr.yaml`'s `sca_gate` job, named `SCA gate`, treats an unexplained skip on a same-repo PR as a failure and carries the comment "THIS is the required status check for supply-chain scanning, not `FOSSA Scan / SCA Scan`."

The real gap is branch protection, not workflow logic. `main-protection` requires exactly three contexts: `lint`, `build`, `FOSSA Scan`. `FOSSA Scan` reports `skipping` on pull requests and therefore passes, while `SCA gate` and `FOSSA Scan / SCA Scan` are not required and are both red today.

So this PR replaces one non-required fail-closed gate with another non-required fail-closed gate. The thing that closes the hole is the admin action you already list under "Admin follow-ups" — requiring `Guardian scan gate` and dropping the FOSSA-shaped contexts. Worth being precise about because the ordering matters: merging this without that ruleset change leaves the repo exactly as exposed as it is now, and the new gate's correctness is not what fixes it.

Your gate design is still better than what it replaces, for a reason that survives the correction: it aggregates four scan jobs plus the trunk gate under `if: always()`, where the outgoing arrangement gated on one job. My two blocking findings are unaffected.

artyom-morozov:
Thanks for the review — addressed:

**@solace-aross — release image push was ungated (blocking).** Fixed in `2c8725c`: `build-docker` now `needs: [test, release-notes, licenses, release-readiness]`, so the irreversible multi-arch push to `ghcr.io` (and `latest` moving) waits for the Guardian gate. The admin `skip_security_checks` escape hatch in `release-readiness-check.yaml` still allows an authorised publish over a block. `build-binaries` is intentionally left ungated — those are only artifacts until the `release` job (which depends on `release-readiness`) publishes them.

**Branch protection (your correction).** Agreed — the new gate is fail-closed but *not required* yet, so merging this alone leaves the repo as exposed as today. Closing the hole is the ruleset change: require `Guardian scan gate` and drop the FOSSA-shaped contexts (`FOSSA Scan`, `FOSSA Scan / SCA Scan`). That's documented in `ADMIN_SETUP.md → Guardian Scanning Setup` and the "Admin follow-ups" in the PR description as a merge prerequisite, not a code change here.

**Copilot's 3 comments are obsolete.** They target `.github/actions/guardian-scan/action.yml` and its README (the `github-token` default + unpinned README example). That composite action was removed when the scan was restructured into parallel jobs — the `github_token` is now passed as `${{ secrets.GITHUB_TOKEN }}` directly in each job, so the literal-default gotcha no longer applies.

solace-aross (changes requested):
Blocking 1 is closed, and the comment you added is better than the change alone: naming the irreversibility and pointing at the admin escape hatch means the next person reading `needs:` knows why the extra dependency is there and that it cannot wedge an emergency release.

Blocking 2 is still open, and it stopped being hypothetical. **#246 merged while this PR sat**, so what I described as a risky conflict resolution is now a live regression:

| | attestation steps | top-level permissions |
|---|---|---|
| `origin/main` (with #246) | **3** | `contents: read` |
| this PR at `2c8725c` | **0** | `contents: write`, `packages: write`, `id-token: write`, `pull-requests: write`, `actions: read`, `issues: write`, `checks: write`, `statuses: write` |

`2646dd6` is not an ancestor of `2c8725c`, so this branch still predates #246 entirely. Merging it as-is would delete all three `attest-build-provenance` steps from `release.yml` and restore the broad permission grant that #246 replaced with per-job scoping.

I checked the one thing that would have refuted this, in case attestation had simply moved when you extracted `build-image.yaml`: it has not. `git grep attest-build-provenance` across `.github/` at this head returns nothing at all.

What is needed is a merge of `main` and a deliberate reconciliation, not a mechanical one. Specifically worth preserving:

- the three `attest-build-provenance` steps, four binary archives and the image, with the image one carrying `subject-digest` and `push-to-registry: true`;
- `attestations: write` and `id-token: write` on whichever job now builds the image, which after your refactor is `build-image.yaml`'s `build` job rather than `release.yml`'s `build-docker`;
- `artifact-metadata: write`, which #246 added because `attest-build-provenance` creates a storage record under an organization account;
- the top-level drop to `contents: read` with per-job scoping.

The last two are the easiest to lose, because they are permission lines rather than steps and nothing fails loudly when they go missing. The attestation itself would at least go red on the next tag.

Nothing else from my previous review changed. The gate design, the admin bypass enforcement, `pull_request` over `pull_request_target`, and the SHA pinning all still stand as credited, and the three notes (ADMIN_SETUP drift from the parallel-jobs restructure, the fork-PR carve-out becoming the normal path once public, and the `${{ inputs.skip_security_checks }}` interpolation) are unchanged.

.github/workflows/guardian-scan.yaml
```diff
@@ -0,0 +1,330 @@
...
+        default: true
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+
+env:
```
reubenjds, line 50: if this is here do we need .fossa.yml

reubenjds:
> **Guardian vuln diff vs solace-broker-mcp live baseline** — added 0, resolved 0, changed 0 (no severity changes)
> 
> ### Per-image CVE breakdown
> Severity	Baseline (live)	Candidate (PR)	Δ
> **Total**	**0**	**0**	**+0**

Remove this for public repos

artyom-morozov:
Second review round addressed.

**Blocking 2 — conflict with #246 (reconciled).** #246 is already merged, so I merged `main` into this branch and reconciled rather than leaving it for the merge button. The trap you described — "keep new version" silently drops #246's image attestation — is now structurally avoided:
- `build-docker` is the reusable `build-image.yaml` call **and** `build-image.yaml` (push mode) now emits the image build-provenance attestation itself (`actions/attest-build-provenance`, `push-to-registry`), with `id-token`/`attestations`/`artifact-metadata` on the job. So the reusable refactor preserves #246's provenance instead of discarding it, in either merge order.
- `build-docker.needs` is `[test, release-notes, licenses, release-readiness]` — #246's gating intent (`fossa_scan`) carried across to the Guardian gate, which is also Blocking 1.
- #246's per-leg binary attestations + least-privilege permissions are kept as-is.
- `transition_on_merge.yaml`: `main` already deleted it (SOL-152855); the merge took that. `MERGEABLE` now.

**`skip_security_checks` interpolation (important).** Fixed — the `authorize` github-script now reads it from `env:` (`SKIP_SECURITY_CHECKS`) instead of interpolating `${{ inputs… }}` into the script body. Audited the other jobs I added: `setup` and `gate` already pass every context value through `env:`; no remaining `${{ }}` interpolation in any `run:`/`script:` body.

**Doc drift (ADMIN_SETUP).** Fixed: the required-check table and prose now say job `gate` (the aggregator's id after the parallel-jobs restructure); the secrets/fork wording names the five parallel jobs (`build`, `fossa-license`, `fossa-vuln`, `prisma`, `guardian-gate`) instead of a singular "Guardian scan job". The left column — the check name an admin types, `Guardian scan gate` — was already correct and is unchanged.

**Fork PRs merge unscanned.** Strengthened the residual-gap section: called out that once public, fork PRs are the normal path (not the edge case), that "require approval for outside collaborators" protects the secrets and not the verdict, and flagged it explicitly as an owner decision to make **before** flipping public.

For reference, the pre-existing `cache-poisoning`/`artipacked` zizmor findings on `release.yml`'s `setup-go`/checkout steps are inherited from #246/`main` and left untouched to avoid unrelated churn; the new files I added are zizmor-clean.

solace-aross (changes requested):
Both earlier blocking items are properly closed, and the second one was closed better than I asked. But there is a third that I should have caught two passes ago: **the Guardian Scan workflow has never run.** Not once, across four commits.

### Closed, verified

| Item | Evidence |
|---|---|
| Image push gated on the verdict | `build-docker` needs `release-readiness`, and it survived the `main` merge. |
| #246's work preserved | Attestation count is still 2 across the pair: binaries in `release.yml`, image now in `build-image.yaml` where the image is actually built. That is the right home for it after the refactor, not merely a rescue. |
| The permission lines I flagged as easiest to lose | All four present on `build-image.yaml`'s `build` job, including `artifact-metadata: write`. `release.yml` top-level is still `contents: read`. |
| Bypass input hardening | `env: SKIP_SECURITY_CHECKS` plus `process.env.SKIP_SECURITY_CHECKS === 'true'`. Exactly the robust form, and the string comparison is right. |
| ADMIN_SETUP drift | The required-check table now names job `gate`. |

### Blocking: `Guardian Scan` fails at startup, and has since `f36a6a3`

```
1c5c6d6e  success           uses: ./.github/actions/guardian-scan     (composite)
a88a39b7  success           uses: ./.github/actions/guardian-scan     (composite)
f36a6a31  startup_failure   uses: ./.github/workflows/build-image.yaml  <-- restructure
493f1444  startup_failure
5cf8b0d6  startup_failure
```

Green until exactly the commit that introduced the reusable-workflow call, and red at startup ever since. Run `31110957779` created **zero jobs** (`jobs.total_count: 0`), so nothing in this pipeline has ever executed. Every other workflow succeeds on the same commits, which is why this hides so well: the PR looks all-green because the one workflow that matters never produced a check at all.

**The cause is a permission escalation, and the comment in the code has it backwards.** `build-image.yaml`'s `build` job declares five scopes (`contents: read`, `packages: write`, `id-token: write`, `attestations: write`, `artifact-metadata: write`). `release.yml`'s `build-docker` grants all five. `guardian-scan.yaml`'s `build` grants only `contents: read`. A called workflow's job requesting more than its caller granted is rejected at run creation, not silently capped, so:

```
# Callers cap these: the scan caller grants
# only contents:read, so the attest step (push only) never runs there.
```

is inverted. It is also the same rule @reubenjds documents in #259, from the other direction: a caller granting less fails at job creation, before any step runs.

Fix is either granting the same five scopes on the scan caller, or splitting `build-image.yaml` so the scan path's job declares only `contents: read`. Actions has no conditional permissions, so one shared reusable workflow forces every caller to grant the superset.

One caveat on evidence: startup-failure runs expose no logs, jobs, or check runs through the API, so I could not capture the literal error text. It is in the UI banner on run `31110957779`, and worth reading directly rather than taking my inference.

### Why this is worth holding the PR for, in either merge order

`FOSSA Scan` no longer appears in any workflow at this head, and `main-protection` requires exactly `lint`, `build`, `FOSSA Scan`. A required context whose workflow no longer exists stays pending rather than being skipped, so merging as-is blocks every subsequent PR.

The ruleset swap you already list as an admin follow-up does not rescue that. Requiring `Guardian scan gate` while the workflow cannot start converts "context missing" into "context pending forever". Both orderings brick the branch until the permissions are fixed, and that is the argument rather than either ordering on its own.

### Note

Fixing the permissions will get the run started, but it will not tell you the pipeline works. Nothing downstream of `setup` has ever executed, so the Prisma artifact handoff and both `fossa-guard` steps are entirely unexercised. Expect the first genuinely-started run to surface more, and budget for that rather than treating a green permissions fix as done.

solace-aross (changes requested):
Still blocked, for the same reason as my last review, plus one new item that came in with this commit.

**The startup failure is unchanged at `b7b6145`.** `Guardian Scan` is `startup_failure` on this head too, making five consecutive commits. The permission mismatch is byte-identical: `guardian-scan.yaml`'s `build` grants `contents: read` while `build-image.yaml`'s `build` declares five scopes. This commit did not touch either side, so nothing in my previous review needs restating — it all still applies.

**Credit for the typed-boolean work,** which is a real class of bug and not one I raised. Routing derived flags through `env` and reserving `fromJSON` for string values is right, and the comment explaining that `${{ inputs.push }}` is already type-safe because GitHub rejects a non-boolean from a caller — so `fromJSON` on it would actively error — is the kind of note that stops someone "fixing" it later. Good change.

**New: the Guardian setup runbook was deleted, not moved.** This commit removes 47 lines from `ADMIN_SETUP.md`, and that section was the only place documenting how to configure the thing this PR depends on: creating the `guardian` Environment, the reasoning for no required reviewers and no branch restriction, the table of six secrets, and the Guardian product registration.

I checked whether it relocated, and it did not. At this head no markdown file in the repository names `FOSSA_API_KEY`, `GUARDIAN_API_TOKEN`, or `PRISMACLOUD_ACCESS_KEY_ID` — I grepped all of them across `*.md` and got nothing. What survives documents only which status check to require, and it still leans on the environment existing: the fork carve-out explains that fork PRs get "no Environment secrets" without saying anywhere what those secrets are or how the environment came to be.

That matters more than a usual docs nit because your own description says the environment and its six secrets were created out of band. Once this merges, the only record of that configuration is in someone's memory and the GitHub settings UI. Nobody can audit it, rotate against it, or rebuild it after an org move — which is exactly the failure this repository just spent a week recovering from, when the `SolaceDev` to `SolaceProducts` rename broke a Vault OIDC role binding that `ADMIN_SETUP.md` did document.

If the section moved somewhere I did not look, point me at it and I will withdraw this. If it was dropped to keep the file short, the secrets table and the environment-settings rationale are the two parts worth keeping at any length.

solace-aross (dismissed):
Approved. All three blocking items are closed, and the startup-failure fix is better than the one I suggested.

I proposed raising the caller's grant. You split the reusable workflow instead: `build-image.yaml` is now a scan-only builder that never pushes and needs `contents: read`, while the release's multi-arch build, push and attestation moved back inline into `release.yml` where the elevated scopes legitimately belong. That keeps every caller minimal rather than making the scan path carry registry and attestation scopes it never uses. The comment now states the rule correctly too — a caller must grant at least what the callee's jobs declare, and under-granting fails at run creation rather than being capped.

**Verified at `92ecd89`:**

| | |
|---|---|
| `Guardian Scan` starts | Yes. Five commits of `startup_failure` ended at `356681d`. |
| The pipeline actually runs | `Resolve context` success, `Build image / build` success, `Prisma scan` success. First time anything downstream of `setup` has executed. |
| #246's work intact | 2 `attest-build-provenance` steps in `release.yml`, `build-docker` holding all five scopes including `artifact-metadata: write`, top-level still `contents: read`. |
| Image push still gated | `build-docker` needs `[test, release-notes, licenses, release-readiness]`. |
| The gate fails closed on a real verdict | `Guardian scan gate` = failure because `FOSSA licensing` and `FOSSA vulnerability` both failed at their guard steps. `Guardian gate` correctly skipped, being trunk-only. |

**The red is real, and it is the most useful thing this PR has produced.** Those two failures are at the FOSSA *guard* steps, meaning the scan authenticated, ran, and returned a BLOCK verdict on both licensing and vulnerability. That independently confirms through an entirely different credential path and pipeline what #254 showed on the old Vault path: this repository has a genuine FOSSA policy and vulnerability verdict against it, not a misconfiguration. Every prior red I have looked at was ambiguous between those two explanations. This one is not.

The consequence is worth stating plainly since it lands on whoever requires the new context: once `Guardian scan gate` is required, that verdict blocks merges. That is the intended behaviour and the whole point of the change, but it means someone has to resolve the actual FOSSA findings rather than the plumbing. Not a code change, and not yours.

**One thing I would still like before merge, and it is cheap.** The Guardian setup runbook is still gone — no markdown in the repository names `FOSSA_API_KEY`, `GUARDIAN_API_TOKEN`, or `PRISMACLOUD_ACCESS_KEY_ID` at this head. I was wrong to imply the content would be lost: it is recoverable from `b7b6145^`, so restoring the secrets table and the environment-settings rationale is a cherry-pick rather than a rewrite. Worth doing because the environment and its six secrets were created out of band, so that section is the only description of a security-critical configuration nobody can otherwise audit or rebuild — which is the failure this repository just spent a week recovering from when the org rename broke a Vault OIDC role binding.

Not holding the PR for it. The rest is done and working, and the FOSSA verdict needs an owner regardless of when this merges.

solace-aross (approved):
Re-approved at `06d78f2`. Additive, and the rule is applied correctly: `packages: read` on both FOSSA jobs plus the ceiling propagated through `release-readiness-check.yaml` and `release.yml`, which is necessary because a caller has to grant at least what the callee declares.

I checked the propagation is complete rather than assuming. Four files mention `guardian-scan.yaml`; only two of them call it, and both were updated. `build-and-test.yml` and `ci-pr.yaml` only reference it in comments explaining that scanning moved out, so they correctly stay at `contents: read`.

That reduction is worth crediting on its own, and the comment gives the right reason: those workflows now grant `contents: read` to every job where they used to hand `pull-requests`, `id-token`, `checks` and `statuses` write to all of them, and `statuses: write` really is the sharp one once `pull_request` is a trigger and the e2e jobs execute shell from the PR's own ref.

**A caveat on my previous review, because this commit undercuts something I stated too confidently.** I said the two FOSSA guard failures showed the scan had authenticated and returned a genuine BLOCK verdict on licensing and vulnerability, and that this independently confirmed a real supply-chain problem rather than a misconfiguration. This commit says those jobs could not pull the private `maas-build-actions` scan container — which is a competing explanation for exactly the same red, since a pull failure surfaces at the same guard step a verdict would.

I cannot settle it yet. The run for this head is still in progress and job logs are not retrievable until it completes, so treat my "independently confirms a real FOSSA verdict" claim as unproven for now. The run that finishes on this commit decides it: if the guards go green, the earlier red was the container pull and I was wrong. If they fail the same way with the pull fixed, the verdict is real and the point stands.

Everything else from my approval holds, including the ask to restore the Guardian setup runbook from `b7b6145^` before merge.

solace-aross:
Heads-up so you don't debug a phantom: the red `Guardian Scan` on this head is a GitHub transient, not your change. Run `31119217271`'s `Resolve context` job failed at `Set up job` with two annotations, `Failed to resolve action download info.` and `Service Unavailable`. That is a 503 resolving an action, so every downstream job skipped and `Guardian scan gate` correctly failed on the aggregate. Nothing to fix; a re-run should clear it.

That also means my caveat from the previous review is still open. The FOSSA licensing and vulnerability jobs were skipped this time, so this run produced no evidence about whether their earlier failures were a real BLOCK verdict or just the private-container pull you fixed in `06d78f2`. The next run that actually reaches those guards settles it, and I would like to see it before treating either explanation as established — I stated the verdict reading too confidently two reviews ago and it is still unproven.

My approval stands; this is informational.

One other thing worth knowing, since it changes this PR's position rather than its content: **#259 is now closed unmerged**, so the competing approach is off the table and this is the only path for moving FOSSA off Vault. The merge-order question I raised against it no longer applies. What does still apply is the ruleset swap — `FOSSA Scan` is gone from every workflow here while `main-protection` still requires it, so requiring `Guardian scan gate` and dropping the FOSSA-shaped contexts has to land with this, not after it.

PR #259, opened 2026-08-05
DATAGO-147338: FOSSA: authenticate with the FOSSA_API_KEY Actions secret

reubenjds:
https://sol-jira.atlassian.net/browse/DATAGO-147338

### What is the purpose of this change?

The org now distributes the FOSSA API key as a GitHub Actions secret instead of through Vault, so the Vault path in our scan workflows is the wrong one. `FOSSA_API_KEY` already exists as an org-level secret visible to this repo.

### How was this change implemented?

The four callers of the shared `sca-scan-and-guard` workflow (`ci-pr.yaml`, `build-and-test.yml`, `release.yml`, `fossa-scan.yaml`) now pass `use_vault: false` and `secrets.FOSSA_API_KEY` in place of `secrets.VAULT_URL`. The `sca_scanning.secrets.vault` block comes out of `workflow-config.json`, since its secret path and role existed only to locate the FOSSA token. Comments and admin docs that named Vault as the reason for the fork skip are reworded, and the internal Vault hostname still committed in `ADMIN_SETUP.md` is removed (it was the last copy in the tree).

### Key Design Decisions

`id-token: write` stays in all four callers even though nothing uses it now. The shared workflow declares its permissions at workflow level including that scope, and its `sca_scan` job declares none of its own, so that is what the nested job requests. A called workflow cannot request more than its caller grants, so a caller granting less fails at job creation, before any step runs. I confirmed this against a repo that omits it: `pubsubplus-kubernetes-helm-quickstart` calls the same workflow without `id-token: write` and its run failed having created zero jobs. The reason is written into `build-and-test.yml` so the scope is not later removed as dead weight.

### Is there anything the reviewers should focus on/be aware of?

Three things:

1. **This does not fix the currently red FOSSA check, and was never going to.** Vault was working fine before this change (run 31039440782 shows both Vault steps green). The failure is `Block on FOSSA Policy/Vulnerability Failures`, because both guards fail to pull `ghcr.io/solaceproducts/maas-build-actions` with `manifest unknown` (DATAGO-146668). The guards are `continue-on-error`, so the summary misreports that as findings.
2. **Overlaps with #257**, which removes these callers outright and moves FOSSA to the `guardian` Environment secret. Every file here is one it rewrites or deletes. Landing this smaller step first means #257 needs a rebase where taking its version is correct throughout. Happy to close this instead if that is preferred.
3. **`VAULT_URL` is now orphaned.** No workflow reads it: `transition_on_merge.yaml` still uses Vault for Jira, but it passes no `secrets:` block and no `secrets: inherit`, so its address comes from `re-workflows`. Left in place rather than deleted, since removing a secret is not reversible.

Verified by dispatching the manual scan on this branch (run 31040933545): the Vault steps are skipped, the log reports `FOSSA API key retrieved from GitHub Secrets`, and `fossa analyze` authenticated and uploaded (it emits the project dashboard URL), so the key is valid and not just present. Steps 1 to 18 pass; only the two pre-existing guard failures above remain. No CHANGELOG entry, since this touches no published contract.

### DIFF
--- a/.github/ADMIN_SETUP.md
+++ b/.github/ADMIN_SETUP.md
@@ -225,7 +225,7 @@ against a real pull request before you rely on them:
   (`name: "SCA Scan"`), which exits non-zero on findings because
   `.github/workflow-config.json` sets both FOSSA modes to `BLOCK`. But that caller
   is now skipped for fork pull requests, because a fork gets no repository
-  secrets and the Vault-backed scan cannot run. A skipped *plain* job still
+  secrets and the credentialed scan cannot run. A skipped *plain* job still
   reports `skipped`; a skipped reusable-workflow *caller* never creates the
   ` / SCA Scan` context at all. Required contexts that are never created stay
   pending forever, so requiring this one makes every external contribution
@@ -315,11 +315,13 @@ made this impossible. What it changed, and what it deliberately did not:
    take their broker credentials from committed `.env` files, and generate their
    own self-signed TLS certs with `openssl`. A read-only `GITHUB_TOKEN` is enough.
 2. **FOSSA no longer goes red for a missing secret.** `ci-pr.yaml` passes
-   `use_vault: true` and `secrets.VAULT_URL`, and GitHub withholds secrets from a
-   fork-triggered `pull_request` run, so the reusable workflow used to find an
-   empty Vault URL and exit 1 on every external contribution. The `fossa_scan`
+   `use_vault: false` and `secrets.FOSSA_API_KEY`, and GitHub withholds secrets
+   from a fork-triggered `pull_request` run, so the reusable workflow used to find
+   an empty credential and exit 1 on every external contribution. The `fossa_scan`
    job is now skipped for fork pull requests, and the always-reporting `SCA gate`
-   job is the required context instead.
+   job is the required context instead. Moving the key off Vault onto an Actions
+   secret did not close this gap: GitHub withholds every repository secret from a
+   fork run, whatever its source.
 3. **`pull_request_target` was not introduced**, and neither was a `workflow_run`
    follow-up. Both would run untrusted code with an elevated token to buy a
    pre-merge scan on a path that already requires a maintainer to approve the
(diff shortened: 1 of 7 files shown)


solace-aross (approved):
Approved on the wiring. The migration is demonstrably correct, and this PR also happens to remove the last copy of an internal hostname from the tree, which matters with the repo going public.

**Verified at `4289239f`:**

| Claim | Evidence |
|---|---|
| The credential path actually works | Job steps: `Validate Vault Configuration: skipped`, `Retrieve Secrets from Vault: skipped`, `Validate FOSSA API Key: success`. Both Vault steps correctly out, and the key validates rather than merely being passed. |
| All four callers converted consistently | `ci-pr.yaml`, `build-and-test.yml`, `release.yml`, `fossa-scan.yaml` all on `use_vault: false` + `FOSSA_API_KEY`, with no straggler still passing `VAULT_URL` or `use_vault: true`. |
| The hostname really was the last copy | A repo-wide grep at this head finds no Vault hostname anywhere. `https://vault.maas-vault-prod.solace.cloud:8200` is gone from `ADMIN_SETUP.md`, and nothing else in the tree carries one. |
| `id-token: write` retention is justified | Your reasoning holds: a called workflow's job requests what the callee declares, a caller granting less fails at job creation, and you cited a repo that omits it as the counter-check. Keeping it is the safe side of that trade. |

**One thing to sort before merge: `SCA Scan - FOSSA` was cancelled, not failed.** It ran 30m22s and the job concluded `cancelled`, with every FOSSA policy and vulnerability step skipped after it. For comparison, the same job on other pull requests completes in 12 to 55 seconds. So nothing here is a FOSSA verdict — the scan hung or was cancelled after authenticating successfully. That needs one clean completion before merge, because a cancelled scan is not evidence the migration works end to end, only that the key is accepted. If it hangs again, the 30-minute wall is itself worth reporting to whoever owns the shared workflow.

**A gap this PR inherits rather than introduces, worth knowing about.** Moving off Vault does not fix the scan for Dependabot pull requests. I confirmed the repo has `VAULT_URL` and `LLM_SERVICE_API_KEY` as Actions secrets while the **Dependabot secret store is empty**, and a Dependabot-triggered `pull_request` run resolves `secrets.*` against that empty store. `FOSSA_API_KEY` will arrive empty there exactly as `VAULT_URL` does today, so `fossa_scan` still runs (the `if:` only excludes forks, and a Dependabot branch is same-repo) and still fails before reaching FOSSA. Every Dependabot PR keeps a permanently red `SCA gate`. The fix is the same shape as the fork carve-out you already reworded: teach the gate that `github.actor == 'dependabot[bot]'` is an accounted-for reason there is no verdict.

**The thing I cannot settle by reviewing.** This and #257 are two different answers to the same question. This PR keeps the shared `sca-scan-and-guard` pipeline and swaps the credential source; #257 replaces the pipeline with a native Guardian one on Environment secrets and deletes `fossa-scan.yaml`, which this PR edits. Merging both is incoherent, and whichever lands second conflicts. That is a call for you and plat-eng, and it should not be settled by whichever review lands first. If #257 is the direction, most of this PR becomes moot except the hostname removal, which is worth keeping either way.

reubenjds:
Original Fossa scan failed due to fossa outaged.
image

PR #18, opened 2026-04-27
DATAGO-132977: Add transition_on_merge workflow

clarkbains:
## Summary

Adds reusable workflow caller that transitions JIRA tickets when PRs are merged. Uses SolaceDev/re-workflows transition-pr-on-merge.yaml@release/v3.

Fixes DATAGO-132977

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Dependency update

## Changes Made

-
-
-

## Testing

**Test Environment:**
- Go version:
- OS:
- Broker version (if applicable):

**Tests Performed:**
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated (if applicable)
- [ ] E2E tests added/updated (if applicable)
- [ ] Tested locally with `go test -race ./...`
- [ ] Tested with real Solace broker

**Test Coverage:**

-

## Breaking Changes

**Impact:**

**Migration Guide:**

### Code Quality
- [ ] Code follows the [coding standards](../.github/CONTRIBUTING.md#coding-standards)
- [ ] Self-review completed (checked for typos, logic errors, edge cases)
- [ ] Code is well-commented (explains "why", not "what")
- [ ] Complex logic has explanatory comments
- [ ] No commented-out code or debug statements left in

### Security
- [ ] **No credentials in code** (checked all files, commits, and logs)
- [ ] Followed [secure logging rules](../docs/secure-logging-rules.md)
- [ ] Credential-carrying types implement `slog.LogValuer`
- [ ] No security vulnerabilities introduced (checked with golangci-lint/gosec)

### Testing
- [ ] All tests pass locally: `go test -race ./...`
- [ ] No race conditions detected
- [ ] Edge cases covered by tests
- [ ] Error paths tested
- [ ] Test names clearly describe what is being tested

### Documentation
- [ ] README.md updated (if user-facing changes)
- [ ] docs/ updated (if architecture/design changes)
- [ ] godoc comments added/updated for exported functions
- [ ] CHANGELOG.md updated (if applicable)
- [ ] Examples updated (if applicable)

### Git & CI
- [x] All commits are signed off (DCO: `git commit -s`)
- [x] Commits have clear, descriptive messages
- [x] Branch is up to date with main (rebased if needed)
- [x] No merge conflicts
[...]

PR #263, opened 2026-08-06
SOL-152902: Stop machine hostnames reaching commit metadata, and settle the 7 already in history

helen-zhu-solace:
## Summary

Adds a CI gate that fails any PR whose commits carry a non-routable author or committer email, and records the seven pre-existing commits already on `main` as a scoped exception.

Fixes SOL-152902

## Type of Change

- [x] Documentation update

## Motivation and Context

Publishing this repository publishes its whole history, permanently, and a fork keeps a copy we cannot reach. Seven commits on `main` carry author addresses that are internal machine hostnames rather than routable ones. Author email is not stored in any file, so no secret scan finds it — the Open Source Solace Software Checklist lists it as one of the items that was invisible in documentation claiming the control was in place.

Two independent pieces of work:

1. **Stop the next one.** Add a forward-only CI gate so no new PR can land another leak.
2. **Settle the seven.** Record an exception per checklist §5 rather than rewriting history — a rewrite would invalidate the SHA of nearly every commit and break every existing reference in Jira, PR discussions, and the CHANGELOG. Exposure is one internal dev hostname and one laptop name, both already attached to those contributors' own commits.

The gate matters because DCO does not catch this class of leak. DCO validates the *sign-off trailer*; the identity check validates the *author field*, which is what git clients, GitHub, `git log`, and `git blame` all render. Both fields default to `git config user.email`, so a single bad config satisfies DCO ("sign-off matches author") while still leaking the hostname forever.

## Changes Made

- `.github/scripts/dco-check.sh` — new `identity_domain_ok()` helper and a second per-commit failure path. Rejects author or committer domains ending in `.local`, `.sol-local`, `.internal`, `.lan`, or a bare hostname with no dot at all. Denylist rather than allowlist so ordinary outside contributors are not blocked. Failure output is a separate block from DCO's, so an identity-only failure does not print misleading "missing DCO sign-off" text. Header comment explains the second control and the exception it complements.
- `.github/scripts/dco-check.test.sh` — 12 new cases: each rejected shape, `example.com` still passing, `users.noreply.github.com` still passing, committer-side check, fix-advice strings, and identity-only failure not emitting the DCO header. 41 existing cases untouched.
- `.github/CONTRIBUTING.md` — new *Author identity* subsection under *Enforcement*, stating the requirement and the fix commands.
- `.github/OPEN_SOURCE_STATUS.md` — new *Exceptions* section with E1 recording the seven legacy SHAs, the "accept, do not rewrite" decision with reasoning.
## Testing

**Test Environment:**
- Go version: n/a — the change is bash + docs, no Go code touched
- OS: Linux (Fedora)
- Broker version (if applicable): n/a

**Tests Performed:**
- [x] Unit tests added/updated — 12 new cases in `dco-check.test.sh`, the script's self-test suite
- [ ] Tested locally with `go test -race ./...` — n/a, no Go changes
- [ ] Tested with real Solace broker — n/a, CI-only change

**Test Coverage:**
- `./.github/scripts/dco-check.test.sh` → **53 passed, 0 failed** (41 pre-existing + 12 new)
- Cases cover: `.sol-local`, `.local`, `.internal`, `.lan`, bare-hostname → reject; `solace.com` and `users.noreply.github.com` → accept; `example.com` → accept (denylist trade-off); committer-side check; fix-advice strings appear; identity-only failure does not print the DCO "missing sign-off" header.

## Breaking Changes

**Impact:** none. The gate is forward-only — it walks `HEAD..BASE` on new PRs, not history already on `main`. Every commit currently on `main` other than the seven listed uses a routable domain, so no in-flight branch is retroactively affected. The seven pre-existing commits are covered by the recorded exception, not by rewriting.

**Migration Guide:** n/a.

### Code Quality

### Security
- [x] Followed [secure logging rules](../docs/secure-logging-rules.md) — n/a, no Go logging changed
- [x] Credential-carrying types implement `slog.LogValuer` — n/a
- [x] No security vulnerabilities introduced — the change *adds* a control

### Testing
- [x] All tests pass locally: `./.github/scripts/dco-check.test.sh` → 53/53
- [x] No race conditions detected — bash, single-threaded
- [x] Edge cases covered by tests (committer-side, identity-only, fix-advice presence)
- [x] Error paths tested

### Documentation
- [x] README.md updated (if user-facing changes) — n/a, contributor-facing
- [x] docs/ updated (if architecture/design changes) — n/a
- [x] godoc comments added/updated for exported functions — n/a, no Go
- [ ] CHANGELOG.md updated — **intentionally skipped**: the change touches CI plumbing (`.github/`), not production surface (`internal/config/`, `internal/composite/definitions/tools.yaml`, or `internal/tools/`) per `CLAUDE.md`.
- [x] Examples updated (if applicable) — n/a

### Git & CI
- [x] No merge conflicts
- [x] CI checks passing (lint, build, test, E2E) — will confirm after push

- [x] Not applicable

n/a — CI-only change. Sample failure output from the self-test:

```
::error::1 of 1 commit(s) in this pull request use an author or committer email whose domain is not routable (.local, .sol-local, .internal, .lan, or a bare hostname). These publish permanently with the git history.

  <sha> add a thing
    author:    Alice Example <alice@machine>
    committer: Alice Example <alice@machine>

Set your commit identity to a routable address and rewrite the offending
commits:

  git config user.email "you@solace.com"
  git commit --amend --reset-author
```

## Additional Notes

**Why denylist, not allowlist:** the ticket text pointed both directions. Denylist blocks the specific leak shapes we have seen and does not restrict who can PR to a public repo — a `.com`, `.io`, or personal-domain address contributes fine. Allowlist would have restricted every outside contributor to GitHub-noreply or a project-recognised domain, which is a policy call bigger than this ticket. Denylist can be tightened later if needed.

**Why one CI job, not two:** the identity check runs on the same commit walk as DCO and reads the same identity fields. One job means one clear failure surface for a contributor rather than two red checks about the same commit. Trade-off noted in the script header.

## Related Issues/PRs

- Related to open-source-readiness tracker in `.github/OPEN_SOURCE_STATUS.md`
- Blocks: repository going public

---

**For Reviewers:**

**Focus Areas**:
- `identity_domain_ok()` in `dco-check.sh` — is the denylist the right set of suffixes? Any shape you have seen in the wild that should be added?
- The `OPEN_SOURCE_STATUS.md` E1 wording — is that the record you want, and who are the four approvers?
- Whether one CI job or two is the right call (see script header for the argument).

**Questions**:
- Confirm you are OK with denylist over allowlist. Denylist accepts `example.com` and any other routable domain; allowlist would restrict to `solace.com` + GitHub noreply. Chosen because this repo will accept outside contributions.


solace-lootii (changes requested):
Requesting changes — two correctness bugs, both reproduced locally.

**1. The identity check inherits DCO's merge-exemption, which defeats it exactly
where CONTRIBUTING.md says `git merge main` is fine (`dco-check.sh:257`).**
Merge commits that `merge_content_reason` deems content-free never enter
`$commits`, so their author/committer identity is never checked either. Repro'd:
a clean `git merge main` with committer `alice@laptop.local` → script prints
"all 1 commit(s) ... signed off — OK", exit 0. The leak this PR exists to catch
ships straight through the one merge shape you tell contributors to use.

**2. The documented bulk-fix clobbers other people's authorship
(`CONTRIBUTING.md:410`, same command in `dco-check.sh:402`).**
`git rebase --exec 'git commit --amend --reset-author --no-edit' main` resets
every replayed commit's author to whoever runs it, not just the offending one.
Repro'd on a 2-commit branch (Bob's legit commit + Alice's leaky one): after
Alice runs the exact documented command, Bob's commit now shows her as author.
This is copy-pasteable straight from CI's failure output.

Minor, non-blocking: the DCO-failure print block is now an unindented ~90-line
if/else (`dco-check.sh:287-376`) — worth a pass before merge, easy to break
silently later.

Please fix 1 and 2 before merge — 2 in particular is destructive advice a
contributor could follow verbatim.

solace-lootii:
Re-reviewed — both bugs are fixed, and I re-verified each independently rather
than just reading the diff.

- **Merge-exemption identity leak**: fixed. `identity_commits` now walks merges
  unconditionally, separate from DCO's content-exempt range. Re-ran my original
  repro plus the trickier "PR is solely one content-free merge" edge case —
  both correctly fail now.
- **Destructive `rebase --exec` advice**: fixed. The guard means only the
  offending commit gets amended — re-ran my Bob/Alice repro, Bob's commit is
  now untouched. Bonus: it also fixes the `Signed-off-by:` trailer, not just
  the author. `CONTRIBUTING.md` matches.
- Unindented failure block: fixed, extracted into `report_dco_failures()`.

One gap I'd still like closed before merge: no test case covers a content-free
merge with a bad identity — the exact regression just fixed. All `identity_*`
cases use plain commits; all merge cases check DCO, not identity. Given how
thorough the `evil_merge_*` matrix already is, one more fixture closes the loop
so a future refactor can't silently reintroduce this.

Nice fix all around — LGTM once that test's in, and I'm comfortable approving
without it if you'd rather rely on my manual verification above instead.

.github/scripts/dco-check.sh
```diff
@@ -162,6 +187,14 @@ for merge in $(git rev-list --merges "$HEAD_REV" "^${BASE_REV}" $EXCLUDE_BASE_RE
   fi
 done
 
+# The identity check's own range: every commit the PR contributes, merges
+# included unconditionally. See the merge-exemption note in the header — the
+# DCO exemption is about certifying content and does not extend to the author
+# and committer fields, which a content-free merge still carries and still
+# publishes.
+# shellcheck disable=SC2086 # as above
+identity_commits=$(git rev-list "$HEAD_REV" "^${BASE_REV}" $EXCLUDE_BASE_REF)
```
solace-aross, line 196: This is the exact line that fixes solace-lootii's merge-exemption bug, and it's the only thing standing between the shipped behavior and the regression. Verified by reverting it to `identity_commits="$commits"` (the pre-fix coupling) and running the self-test suite: 61/61 still pass. No test builds a merge commit with a bad-domain identity, so this line can silently revert to the reported bug and CI stays green. See the review body for a validated test case to add.

solace-aross (changes requested):
Both bugs from solace-lootii's review are genuinely fixed, verified by injection rather than by reading:

- Merge-exemption bug (`dco-check.sh:196`): recoupled `identity_commits` back to the DCO-exempted `$commits` range to reproduce the exact regression, then ran a real `git merge --no-ff main` with an `alice@laptop.local` committer (main advanced independently, so it's a genuine merge, not a no-op). Sabotaged: exit 0, `Identity: all N commit(s) ... OK`, leak ships. Restored: exit 1, names the merge commit. Fix holds.
- Destructive bulk-rebase advice: ran the documented guarded command (`git rebase --exec '... grep -qiF "$BAD" || exit 0; ...'`) on a real two-commit branch (Bob's legit commit + Alice's leaky one). Bob's authorship came out untouched; only Alice's commit was rewritten. The old unguarded `--reset-author` would have clobbered Bob's — this one doesn't.

One gap, and it's the same shape as the bug you just fixed: **the merge-exemption fix has no test pinning it.** I reverted `dco-check.sh:196` to `identity_commits="$commits"` — the exact coupling solace-lootii flagged — and ran the self-test suite: 61 passed, 0 failed. Every test stays green with the SOL-152902 scenario reintroduced. I grepped the full identity section (`dco-check.test.sh` #18/#19) and the existing merge fixtures (`merge_exempt`, `evil_merge_repo`, etc.) — all of them commit under a routable address, so none exercise a bad-domain identity through a merge.

This matters more here than in most CI scripts: the script's own header says the suite exists so "the gate's logic is verified on every PR rather than trusted," and `ci-pr.yaml` makes the self-test a required check for exactly that reason. Right now that guarantee has a hole shaped exactly like the bug this PR exists to close, and the only thing that caught it was a human reviewer's eye — which is the failure mode SOL-152902 is about in the first place.

Validated addition (passes 62/62 on the shipped script, fails 61/1 on the sabotage — confirmed against a second independent pass before posting this):

```bash
# --- 18b. the identity check is NOT exempted by the DCO merge exemption ------
r=$(new_repo identity_merge_not_exempt)
git -C "$r" checkout -q -b feature
content_commit "$r" feat.txt "feature work

Signed-off-by: $ALICE_NAME <$ALICE_EMAIL>"
git -C "$r" checkout -q main
content_commit "$r" main.txt "main work

Signed-off-by: $ALICE_NAME <$ALICE_EMAIL>"
git -C "$r" update-ref refs/remotes/origin/main "$(git -C "$r" rev-parse HEAD)"
git -C "$r" checkout -q feature
GIT_AUTHOR_NAME="$ALICE_NAME"    GIT_AUTHOR_EMAIL="alice@laptop.local" \
GIT_COMMITTER_NAME="$ALICE_NAME" GIT_COMMITTER_EMAIL="alice@laptop.local" \
  git -C "$r" merge -q --no-ff -m "Merge branch 'main' into feature" main
if assert_parents "$r" 2; then
  expect 1 "a content-free merge's identity is still checked" "$r" "not routable"
fi
```

Drop it before section 19, ahead of the drift test.

Everything else — the wording/domain-consistency fixes, the OPEN_SOURCE_STATUS.md redaction, the single-line `--exec` command, the CHANGELOG skip per CLAUDE.md (CI plumbing, not production surface) — checks out against the code.

solace-lootii (approved):
Verified — re-ran the merge-exemption repro, the guarded rebase fix, and the
new test 18b; also mutation-tested 18b myself (reverting the fix flips exactly
that one test red, nothing else). All fixed, no gaps left.

LGTM.

.github/scripts/dco-check.test.sh
```diff
@@ -594,6 +594,168 @@ else
...
+if grep -qF "missing a Developer Certificate of Origin sign-off" <<<"$out"; then
+  printf '::error::SELF-TEST FAILED: identity-only failure printed the DCO-missing header.\n'
+  printf '%s\n' "$out" | sed 's/^/    | /'
+  fail_count=$((fail_count + 1))
+else
+  printf 'ok  identity-only failure does not print the DCO header\n'
+  pass_count=$((pass_count + 1))
+fi
+
+# --- 18b. the DCO merge exemption does not extend to the identity check ------
```
solace-aross, line 695: Confirmed this catches the regression: reverting `identity_commits` to `$commits` in `dco-check.sh:196` turns this from a pass into `SELF-TEST FAILED: a content-free merge's identity is still checked — expected exit 1, got 0`, and the rest of the suite stays green (61/61) around it. That's exactly the blind spot flagged last pass — good fix.

solace-aross (approved):
Second pass. The gap from my last review is closed: `efa993983930` adds the exact regression test I gave (18b), pins `identity_commits` to a range independent of `$commits`, and extracts `report_identity_failures()` so the DCO and identity blocks stay separately guarded.

Verified by injection, not by reading:

- Baseline at this head: `./.github/scripts/dco-check.test.sh` → 62/62.
- Sabotaged the exact line the fix depends on — reverted `dco-check.sh:196` to `identity_commits="$commits"` (the pre-fix coupling solace-lootii's bug reintroduces) — and reran: 61 passed, 1 failed. The new test 18b is what catches it: `SELF-TEST FAILED: a content-free merge's identity is still checked — expected exit 1, got 0`.
- Restored the file, reran: back to 62/62.

That also resolves the footnote from last time — 18b's fixture is itself a content-free merge (`git merge --no-ff` with no conflicting content) with a bad-domain identity, so the previously-unexercised "content-free merge falls through to the identity check" branch is now covered by the same test.

CI is green across the board, including `DCO check self-test` and `e2e-management`. Nothing else outstanding. Approving.

PR #266, opened 2026-08-07
SOL-152888: filter tools/list by the caller's authorization

amitmorade:
## Summary

Narrows the MCP `tools/list` response to the tools a caller's groups actually grant, reusing the policy that already gates `tools/call`. Off by default behind a new `filter_tools_list` flag.

Tracked under [SOL-152888](https://sol-jira.atlassian.net/browse/SOL-152888).

## Type of Change

- [x] Documentation update

## Motivation and Context

SOL-151440 shipped claims-based tool RBAC, and its Decision 5 deliberately deferred the discovery half: `tools/list` still returns every registered tool to every caller regardless of what they may invoke.

So an agent is handed tools it will be denied. It spends context on their descriptions and schemas, and when a call is refused it tends to misdiagnose the cause — during the spike, Claude Code blamed the transport and advised re-authenticating.

**This is discovery hygiene, not an access control.** `tools/call` remains the only authorization boundary and is unchanged here. An unlisted tool is still callable by name, because the SDK resolves `callTool` against the global tool set rather than against whatever a previous list returned. Please review it on those terms — read as a security fix, the review asks the wrong questions.

Measured on the local stack: **24 tools → 3** for a caller in one group. At the ~400 tokens/tool figure `TestToolListBudget` (SOL-152125) pins, roughly 8,400 tokens of dead context per turn.

## Changes Made

- **`internal/config/config.go`** — new `FilterToolsList *bool` on `ToolAuthorizationConfig` plus a nil-guarded `FilterToolsListEnabled` reader. `*bool` matches the sibling flags and keeps absent distinguishable from explicit `false`, which the planned default flip will need. `applyDefaults` deliberately leaves it nil.
- **`internal/tools/listfilter.go`** — the middleware. Installed via `Server.AddReceivingMiddleware`, since `Server.listTools` discards its context parameter and offers no per-caller hook.
- **`internal/tools/identity.go`** — `requestGroups` widened from `*mcp.CallToolRequest` to `*sdkauth.TokenInfo`, so the call path and the middleware read the groups claim through one function instead of two that can drift. 12 lines became 1; one caller updated, no test churn.
- **`internal/tools/authorization.go`** — new `isExemptFromToolAuthorization` predicate, now the single point of truth for which tools bypass the policy.
- **`cmd/server/main.go`** — installs the middleware after `buildToolPolicy`, and states the filtering posture on every boot.
- **Docs** — `configuration.md` carries the reference; `README.md`, `authentication.md`, and `user-guide.md` point at it.

### Design notes

**Gated on the compiled policy, not on `ToolAuthorizationEnabled`.** The filter needs a policy to call `Authorize` against, and `buildToolPolicy` already folds the enabled check into whether one exists. Checking the thing the code actually uses leaves no way for the two to disagree. That is also where the precedence between the two config keys lives — nothing in the config schema encodes it.

**Runs after `next`.** Unlike `withAuthorization`, which decides before dispatching, the list must exist before it can be narrowed.

**Declines other methods explicitly.** Receiving middleware is process-wide — `serverMethodInfos` is one flat table, so `tools/call`, `initialize`, `ping`, and every notification traverse this chain. "Out of scope" required a method guard, not just intent.

**Copies the result.** `ListToolsResult` belongs to the SDK, so the filter returns a shallow copy rather than narrowing in place. `NextCursor` survives, and a test pins it.

**Flag on with authorization off logs WARN and continues.** The state is inert rather than incorrect, and it is what an operator produces by flipping the master switch during an incident without touching anything else — failing startup there would turn a rollback into an outage.

### A bug found while wiring

The filter initially exempted `list-brokers` only. But `describe-semp-schema` is also registered without a policy, so `Authorize` returns a zero-value deny for it — filtering dropped it from every caller's list while it stayed perfectly callable. That breaks listed ⟺ callable, the invariant this feature exists to preserve.

The invariant test missed it too, because it hardcoded one exempt name. It now derives them from the shared predicate. That still leaves the predicate itself unpinned — filter and test would agree while both were wrong, which is exactly how this survived — so a separate test checks it against the registration sites in `main.go`. Fixed in `d7e24ff`.

## Testing

**Test Environment:**
- Go version: go1.26.0 darwin/arm64
- OS: macOS 26.5.2
- Broker version: soltr 10.25.0.259 (`solace/solace-pubsub-standard`), Keycloak via `solace-local-infra`

**Tests Performed:**

**Test Coverage:**

`make check` green, total coverage 88.5% against the 85% floor.

Each middleware test asserts **both** halves of the contract — the tools returned *and* the audit record. Split apart, a case could pass its output assertion while logging the wrong reason, which is precisely what an operator would hit.

| Case | Returns | Logs |
|---|---|---|
| Granting group, partial match | granted tools + exempt | `filtered`, INFO |
| Groups match no grants | exempt only | `not_permitted`, INFO |
| No groups claim | exempt only (fail closed) | `missing_claim`, **WARN** |
| Nil `Extra` / nil `TokenInfo` | exempt only, no panic | `missing_claim`, **WARN** |
| Grants cover everything | full list | `unfiltered`, INFO |
| Two callers, one middleware | different lists | two records, different reasons |
| `tools/call`, `initialize`, `ping`, notifications | pointer-identical passthrough | **no record** |
| `next` returns an error | error propagated | **no record** |
| Nil tool in the SDK result | survivors returned | ERROR + count, plus the normal record |

Silence is asserted as part of the contract: a broken method guard would otherwise flood the log on every ping.

Also covered: the input `ListToolsResult` is unmutated and `NextCursor` survives the copy; caller group names never appear in any record; nil policy panics at wiring time; the four install postures; the reason/level classification directly; and the listed ⟺ callable property across tools × group sets, verified by running the real `withAuthorization` rather than re-asking the policy.

**Manual E2E** against Keycloak + two brokers, `filter_tools_list: true`:

```json
{"level":"INFO","msg":"tool list filter","event":"tool_list_filter",
 "decision_reason":"filtered","groups_present":true,
 "tools_before":24,"tools_after":3,
 "sub":"ac043549-...","iss":"https://localhost:18443/realms/mcp-test-realm",
 "jti":"onrtrt:e81082d2-...","correlation_id":"019fdc6f-..."}
```

3 = the group's one granted tool plus both exemptions — which also confirms the `describe-semp-schema` fix end to end. Startup logged the two posture lines adjacently, as intended:

```
INFO tool authorization is enabled       policy={tool_grant_count:5, group_count:2}
INFO tools/list filtering is enabled
```

## Breaking Changes

None. The flag defaults off; with it off `AddReceivingMiddleware` is never called and the dispatch path is unchanged. `tools/call` behaviour is untouched in every configuration.

### Code Quality
- [x] Self-review completed

### Security
- [x] Credential-carrying types implement `slog.LogValuer` — identity goes through `Identity.LogValue()`

### Testing
- [x] Error paths tested

### Documentation
- [x] README.md updated
- [x] docs/ updated
- [x] CHANGELOG.md updated
- [x] Examples updated

### Git & CI
- [x] No merge conflicts
- [ ] CI checks passing — pending on this push

## Additional Notes

**Follow-up worth filing: the two audit paths should share one schema.** `tool_authorization` and `tool_list_filter` answer the same question — may this caller use this tool? — at two points in a request's life, and they share the `decision_reason` vocabulary. But they grew independently, so the overlap is convention rather than enforcement. This PR adds an `event` field so its own records are self-identifying; the call path has no equivalent, since touching it is out of scope here. The real fix is a shared audit emitter owning the field names and reason vocabulary, with `mcp_method` as the dimension. That touches SOL-151440's audit contract, which `authorization_audit_schema_test.go` pins deliberately.

**Also considered and rejected:** telling the caller anything about what was filtered. `ListToolsResult` carries only `Tools`, `NextCursor`, and `_meta`; the model's context is built from names, descriptions, and schemas, so `_meta` never reaches it, and `ServerSession.Log` returns nil unless the client called `logging/setLevel`. The only mechanism that would work is advertising a fake tool whose description explains the situation — which breaks listed ⟺ callable and tells an unauthorized caller that hidden tools exist. GitHub's MCP server hit the same wall and accepted the same limitation. One deliberate difference from theirs: when they cannot fetch a token's scopes they skip filtering and return everything. That is fail-open. Ours fails closed.

**Not in scope:** `tools/call` changes; notification / cache-invalidation machinery (the SDK has no per-session `tools/list_changed`, and most clients ignore it); per-broker policy, wildcards, scope hierarchies; protocol `2026-07-28`'s `cacheScope`/`ttlMs`, which needs an SDK upgrade — `cacheScope: "private"` will matter for filtered lists once available.

**Deliberately promoted to WARN, considered but left out:** the plan suggested promoting `internal/auth/middleware.go`'s DEBUG "groups claim not found in token" line to WARN. This PR's own `missing_claim` WARN covers the diagnostic need for the list path; the auth-middleware line is a different file and a different concern, cleaner as its own small ticket.

---

**For Reviewers:**

**Focus Areas**:
1. **The framing.** Everything downstream depends on this being discovery hygiene rather than access control — including the decision to WARN rather than fail startup on an inert flag. If you disagree with the framing, say so first; several decisions flip with it.
2. **`internal/tools/listfilter.go`** — the exemption must be checked before the policy, or exempt tools vanish while staying callable.
3. **The gate in `installToolListFiltering`** — is gating on `policy != nil` rather than on `ToolAuthorizationEnabled` the right call?
4. **Log levels.** `missing_claim` is WARN, everything else INFO, on the reasoning that only a missing claim means something is actually wrong. Every filtered list emits a record regardless of outcome — volume filtering is the operator's job, not ours.

**Questions**:


solace-aross (changes requested):
Nice work on this, especially catching the describe-semp-schema gap mid-PR — that's exactly the kind of self-review that should happen before a reviewer sees it.

One blocking issue before merge.

`TestIsExemptFromToolAuthorization_MatchesUnpolicedRegistrations` (listfilter_test.go:706) doesn't do what its comment and d7e24ff's commit message say. It restates `isExemptFromToolAuthorization`'s own two constants and checks the predicate against itself — it never reads `main.go`, so it can't "pin the predicate against the registration sites."

Verified by injection: added a third unpoliced `AddTool` in `main.go` with no policy and no predicate entry. Full suite green. That tool silently drops from `tools/list` while staying callable by name — the same bug class d7e24ff fixed, back open and untested.

Today's two-name enumeration is correct. It's just not held there by anything. Two ways to close it:

- Make exemption a property of registration: have `RegisterListBrokers` and `RegisterDescribeSempSchema` contribute to a set the predicate reads, so a new unpoliced site forces the decision.
- Or add a `cmd/server` test that wires a real server and asserts every registered tool is either in `mgr.Handlers()` or in the exempt predicate.

Non-blocking: the `missing_claim` WARN in `listfilter.go` drops `expected_claim`, which `withAuthorization`'s equivalent WARN carries. Small parity gap in the one field an operator needs to triage a claim-mapper misconfig.

Everything else held up under a from-scratch verification pass: the discovery-hygiene framing, exemption-before-policy ordering, gating on `policy != nil`, fail-closed parity between the two paths, concurrency, and the docs all check out against the code.

amitmorade:
> Nice work on this, especially catching the describe-semp-schema gap mid-PR — that's exactly the kind of self-review that should happen before a reviewer sees it.
> 
> One blocking issue before merge.
> 
> `TestIsExemptFromToolAuthorization_MatchesUnpolicedRegistrations` (listfilter_test.go:706) doesn't do what its comment and [d7e24ff](https://github.com/SolaceProducts/solace-broker-mcp/commit/d7e24ff23786eceb47ce020fbdbb8b177bd9400e)'s commit message say. It restates `isExemptFromToolAuthorization`'s own two constants and checks the predicate against itself — it never reads `main.go`, so it can't "pin the predicate against the registration sites."
> 
> Verified by injection: added a third unpoliced `AddTool` in `main.go` with no policy and no predicate entry. Full suite green. That tool silently drops from `tools/list` while staying callable by name — the same bug class [d7e24ff](https://github.com/SolaceProducts/solace-broker-mcp/commit/d7e24ff23786eceb47ce020fbdbb8b177bd9400e) fixed, back open and untested.
> 
> Today's two-name enumeration is correct. It's just not held there by anything. Two ways to close it:
> 
> * Make exemption a property of registration: have `RegisterListBrokers` and `RegisterDescribeSempSchema` contribute to a set the predicate reads, so a new unpoliced site forces the decision.
> * Or add a `cmd/server` test that wires a real server and asserts every registered tool is either in `mgr.Handlers()` or in the exempt predicate.
> 
> Non-blocking: the `missing_claim` WARN in `listfilter.go` drops `expected_claim`, which `withAuthorization`'s equivalent WARN carries. Small parity gap in the one field an operator needs to triage a claim-mapper misconfig.
> 
> Everything else held up under a from-scratch verification pass: the discovery-hygiene framing, exemption-before-policy ordering, gating on `policy != nil`, fail-closed parity between the two paths, concurrency, and the docs all check out against the code.

Thanks Andrea — the injection was the right way to test it, and you're right that
the test didn't hold anything.

Worse than that, actually: I named the exact failure mode in `d7e24ff`'s message
("filter and test would agree while both were wrong") and then wrote a test with
that property. Naming a risk isn't closing it.

## Blocking finding — fixed in 760a0f6

Went with your second option. `TestEveryRegisteredToolIsGatedOrExempt` now lives
in `cmd/server/tool_authorization_coverage_test.go`: it builds the real
registration pipeline (including this package's own `registerSEMPv1Tools` /
`registerMixedTools`, so a native tool added later is picked up automatically),
drives a real `tools/list` over an in-memory transport, and requires every
exposed tool to be either in `mgr.Handlers()` or exempt.

The package move is the substantive part. A test in `internal/tools` can't see
`main()`, so it can only ever compare the predicate to a copy of itself — that's
the structural reason the old one was vacuous, not an oversight in how I wrote it.

I reproduced your injection before trusting the new test:

```
zz_injection_probe_test.go:37: PROBE OK: detected [injected-unpoliced-tool]
```

Also added `TestExemptToolsAreActuallyRegistered` for the reverse direction — an
exemption naming a tool that no longer exists. Not a live bug, but it would
absorb a rename silently: old name keeps its exemption, new name gets filtered.

**Stated plainly in the test's own doc comment: this detects, it does not
prevent.** A new unpoliced registration still compiles; CI is what catches it.

## On your first option

Your first suggestion is the better design and I'd like to do it, just not here.

The version worth building isn't a registry the exempt functions append to — a
new `AddTool` site still compiles without touching it, so that's the same bug with
the polarity flipped. It's making registration *require* the answer: one internal
registration function taking a gated-or-exempt argument, with the exempt set
derived from what actually registered rather than enumerated. Then forgetting is
a compile error and the predicate has nothing to maintain.

Two reasons I left it out of this PR:

It reaches further than the bug. `ValidatePolicyToolNames` hardcodes both exempt
names *again* (`authorization.go:190-191`, `212-216`) — a third independent copy.
Doing this properly means all three consumers read one derived set, which changes
`ValidatePolicyToolNames`' per-exempt-tool WARN output, which
`authorization_validate_test.go` pins. That's SOL-151440's surface, not this
ticket's.

And it would rewrite `register.go` and `describe_semp_schema.go`, neither of which
this feature otherwise touches — a diff shape I'd push back on if I were reviewing.

Filing it as its own ticket, and I'll link it here. If you'd rather it block this
PR, say so and I'll fold it in.

## Non-blocking — `expected_claim`

Fixed in the same commit; Copilot flagged it independently, which I took as a
signal. `WithListFiltering` now takes the configured claim name and emits
`expected_claim` on the missing-claim record only — omitted elsewhere, since the
claim read fine there and naming it would imply a fault. Sanitized via
`sanitize.Claim`, same as the call path. Test covers both directions.

## One thing you prompted that wasn't in your review

Checking for comments made stale by `d7e24ff` turned up three that still
described a single exemption — including `listfilter.go:69`, the one Copilot
suppressed with exactly the warning that it would "mislead future maintainers back
into the original bug." Fixed all three.

`make check` green, 88.5% coverage.

solace-aross (approved):
Blocking finding from my last pass is fixed, and the fix is the stronger option I called out rather than the smaller patch.

Verified independently in a throwaway worktree at f0bc9bcf, not by reading:

- Injected the exact defect the finding described (a third unpoliced `AddTool`, no policy wrapper, no exemption entry) into `gatedAndExposedTools`'s registration pipeline. `TestEveryRegisteredToolIsGatedOrExempt` failed: `tool(s) with neither a policy wrapper nor an exemption: [injected-unpoliced-tool]`. Reverted, back to green. This is the real reproduction I asked for last time — the old test in `internal/tools` couldn't see `main()`, so it was checking the predicate against itself.
- Sabotaged the `expected_claim` fix (dropped the field from the missing-claim log line). `TestWithListFiltering_ExpectedClaimOnMissingClaimOnly/present_on_missing_claim` went red: `expected_claim = <nil>, want "custom_group_claim"`. Reverted, back to green. Confirms that test earns its green rather than passing vacuously.
- `make check` on the worktree: 88.6% coverage, 85% floor, clean vet/lint.

On `TestExemptToolsAreActuallyRegistered` (the reverse-direction test) and the design note on why this belongs in `cmd/server` rather than `internal/tools`: agreed, and the doc comment on `gatedAndExposedTools` states the reasoning correctly.

The follow-up ticket for making exemption a property of registration rather than an enumerated predicate is the right call to defer — it touches `ValidatePolicyToolNames`'s pinned WARN output and two files this PR doesn't otherwise touch. Link it here once filed so it isn't lost.

Approving.

PR #267, opened 2026-08-07
SOL-152641: prove claims-based tool RBAC end to end, and fix the hop-1 token that carried no groups claim

solace-awaheed:
… carried no groups claim

## Summary

Adds end-to-end coverage for claims-based tool RBAC (SOL-151440) — a real Keycloak token, through the MCP server, to an allow or a deny on `tools/call` — and fixes the realm gap that would have made every one of those assertions fail.

Jira: [SOL-152641](https://sol-jira.atlassian.net/browse/SOL-152641)

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

> The primary category — **test coverage** — isn't in the list above. No production
> Go code changed; this is E2E tests, a Keycloak fixture, CI, and docs.

## Motivation and Context

Tool RBAC had unit-test coverage only. Nothing caught a regression in the full chain:
Keycloak issues a token with a `groups` claim → the server verifies it → resolves the
claim → `policy.Authorize` decides.

Writing those tests surfaced why nobody had noticed the path was untested: **the hop-1
token carried no `groups` claim at all.** A Keycloak client's dedicated protocol mappers
fire only when that client is the *requesting* client, and every realm-role→`groups`
mapper lived on `mcp-server-client` and the two `solace-broker*` clients — all on the
hop-2 side. `agentic-app-client`, which mints the subject token the MCP server actually
reads, had an audience mapper and nothing else.

That is why the brokers worked fine while tool RBAC saw nothing: the exchanged token is
requested by `mcp-server-client`, so its mapper fires and `solace-admins` reaches the
broker. One claim name, two consumers, filled in by different clients.

## Changes Made

- **`realm-export.json`** — added a realm-role→`groups` mapper to `agentic-app-client`
  (the actual fix); added realm role `Ops` assigned to `test-admin-user`; added client
  `agentic-app-client-nogroups`, identical but without the mapper, so the missing-claim
  scenario has a token that authenticates cleanly yet carries no `groups`. Committed to
  the fixture, not clicked in the Keycloak UI — UI changes live only in that container's
  database and vanish on the next import.
- **`test-rbac-scenarios.sh`** (new) — 8 RBAC scenarios plus the disabled-mode no-op,
  driven over the real MCP JSON-RPC wire.
- **`helpers.sh`** — parameterised `tool_authorization` config blocks; `mint_token`
  client override; audit-log assertion helpers (`log_mark`, `authz_records_since`,
  `assert_authz_decision`); `assert_server_rbac_mode` positive control; server restart
  with log rotation; process-liveness check in the readiness poll.
- **`run-all.sh`** — three phases across two server configs; runs every phase even if an
  earlier one fails, then prints a per-phase verdict table.
- **`test-oauth-scenarios.sh`** — the cache-invalidation scenario's broker-audience
  restore moved into its `EXIT` trap (see Additional Notes).
- **`build-and-test.yml`** — `timeout-minutes: 30` on the `e2e-oauth` job; collects
  rotated server logs and generated configs on failure.
- **`README.md`** — documents the two-token property, why the policy grants on `Ops`, and
  why assertions read the log rather than the wire.

## Testing

**Test Environment:**
- Go version: `go1.26.5 darwin/arm64`
- OS: macOS (Darwin 25.6.0)
- Broker version: `solace/solace-pubsub-standard:latest` ×2, Keycloak `26.2.5`

**Tests Performed:**

**Test Coverage:**
- `make e2e-oauth-all` from clean volumes: **15/15 pass, exit 0**, all three phases green
- `make check`: exit 0 — lint 0 issues, **88.3%** against the new 85% floor
- Scenarios cover: allow (asserting *which* group granted it), tool-scoped grant, broker
  authorising independently, both deny reasons distinguished on the audit event, the two
  denies being indistinguishable to the caller, deny short-circuiting before the broker,
  `list-brokers` structural exemption, and RBAC disabled as a genuine no-op
- **Mutation-tested twice**, because a passing test proves nothing until it has been seen
  to fail for the right reason:
  - grant moved to `solace-admins` → allow scenario fails with
    `matched_groups=["solace-admins"], want ["Ops"]`
  - second tool added to the policy → tool-scoping scenario fails
  Both exit non-zero; both restored.

## Breaking Changes

**None to product surface.** One developer-facing change: `run-all.sh` no longer aborts at
the first failing scenario script — it runs all three phases and aggregates the exit code.
Anyone scripting against it sees a different exit path. Rationale in Additional Notes.

### Code Quality
- [x] Self-review completed (plus a five-lens principal review — see Additional Notes)

### Security
- [x] **No credentials in code** — no *new* ones; the test-realm secret and passwords in
      `.env`/`realm-export.json` are pre-existing and unchanged. The only `.env` addition
      is a client ID.
- [x] Followed secure logging rules — n/a, no Go logging changed
- [x] Credential-carrying types implement `slog.LogValuer` — n/a, no Go changed
- [x] No security vulnerabilities introduced (`golangci-lint` → 0 issues)

### Testing
- [x] Error paths tested (both deny reasons, exemption, feature-off)

### Documentation
- [x] README.md updated — `test/e2e-oauth/README.md`; root README unchanged (no
      user-facing behaviour change)
- [ ] docs/ updated — n/a, no architecture change
- [ ] godoc comments added/updated — n/a, no Go changed
- [ ] CHANGELOG.md updated — **n/a**: nothing under `internal/`, `tools.yaml`, or native
      tool packages changed, so no `[Unreleased]` entry is required
- [ ] Examples updated — n/a

### Git & CI
- [x] Branch is up to date with main (rebased onto `cc7a109`)
- [x] No merge conflicts
- [ ] CI checks passing — verified locally; CI pending on this PR (see Additional Notes)

## Additional Notes

**Why `Ops`.** The policy grants `list-vpns` to a group no broker maps. Granting on
`solace-admins` instead would make a passing test meaningless — it could not distinguish
"the MCP server authorized this" from "the broker would have allowed it anyway", and would
keep passing with the RBAC layer deleted. `test-admin-user` holds both groups, so the allow
scenario asserts `matched_groups` is exactly `["Ops"]`. To be precise: `Ops` *does* reach
the broker in the exchanged token; it is simply unmapped there.

**Why the fail-fast removal forced a second change.** The cache-invalidation sc
poisons broker A's required audience and restored it only on its success path. With
fail-fast gone, a lost restore would leak into the later phases — where every RBAC scenario
calls broker A — turning one hop-2 defect into a wall of unrelated failures. The restore is
now unconditional in that script's `EXIT` trap.

**Local caveat for reviewers.** Changing `realm-export.json` requires
`make e2e-oauth-down` first: Keycloak imports the realm only when its *container is
created*, so `up -d` against an older container silently keeps the stale realm.
`mint_token` now prints a hint when it detects the symptom. CI is unaffected.

**Runtime caveat.** Verified on rootful Podman via the Docker-compatible socket
Docker Desktop was unavailable on the authoring machine. Nothing here is runtime-specific
and CI runs Docker, but that exact path is unexercised locally — this PR is what proves it.

**Review depth.** Ran a five-lens principal review (architect, QA, security, support, SRE).
Four vacuous-pass risks surfaced where a scenario would have passed while provi
opposite of its name; all are fixed and re-verified. One reported finding was dropped as
factually unsupported.


test/e2e-oauth/test-rbac-scenarios.sh
```diff
@@ -0,0 +1,363 @@
...
+    fi
+    return 0
+}
+
+# ── Scenario: list-brokers is structurally exempt ───────────────────────────
+# The policy grants test-readonly-user nothing, and list-brokers is not in the
+# policy at all — yet it must still work. Exemption is structural, not a rule:
+# RegisterListBrokers takes no policy argument, so the gate is never wrapped
+# around it and never runs. Hence the second assertion: no audit record.
+test_list_brokers_exempt() {
```
amitmorade: **`test-rbac-scenarios.sh`** — scenario 8 needs a second assertion, not just a comment

`test_list_brokers_exempt` is the right shape: real data back, plus no audit record. The
absence of the record is what actually distinguishes "structurally exempt" from "happened
to be granted."

The problem is that the exemption set has two members and the test pins one.
`isExemptFromToolAuthorization` (`internal/tools/authorization.go:54`) returns true for
`list-brokers` **and** `describe-semp-schema`, and both get there the same way:
`RegisterListBrokers` and `RegisterDescribeSempSchema` are both called from
`cmd/server/main.go` (770, 775) outside the policy-wrapping path, neither taking a policy
argument.

So `describe-semp-schema` is exempt by exactly the mechanism this scenario exists to prove,
and nothing in the suite touches it. Move it inside the wrapped registration path and
scenario 8 stays green while a tool that should never reach the gate starts getting denied
for every caller the policy grants nothing.

Could you extend the scenario to cover both? The assertion is the same shape —
`describe-semp-schema` takes a required `operation` string and returns
`StructuredContent`, so something like `{"operation":"config/createMsgVpnQueue"}` gives you
a real success to assert on alongside the empty `authz_records_since`.

Ideally driven off the predicate rather than a second hand-written block, so adding a third
exempt tool later fails loudly here instead of silently going uncovered — that's the same
single-source-of-truth reasoning `RBAC_GRANT_GROUP` already applies.

solace-aross (dismissed):
Approving. This is genuinely excellent E2E test work — the two-token/two-vocabulary distinction, the positive controls before every absence assertion, and the mutation-testing evidence in the PR description are the right instincts, and it holds up under reading.

One open point from @amitmorade's comment on scenario 8, checked against the actual code rather than taken on faith: `describe-semp-schema` is exempt from the RBAC gate by the identical mechanism as `list-brokers` (`cmd/server/main.go` registers both outside `RegisterWithServer`, no policy arg — `internal/tools/authorization.go`'s `ValidatePolicyToolNames` doc names both as structurally exempt). Scenario 8 only exercises `list-brokers`. I confirmed the gap is real by injection: wrapping `describe-semp-schema` in the policy in a throwaway worktree left `go build`, `go vet`, and the full `go test ./...` suite green — nothing catches it today, including `TestRegisterListBrokers_NeverComposesWithAuthorization` in `register_test.go`, which only covers the one name.

Not blocking, for two reasons: this PR ships no production Go, so it's a strict improvement over today regardless: and the two `RegisterX` call sites in `main.go` are small, adjacent, and each already carries a comment naming the exemption, so ordinary review is a real net even without a test. But since you asked for the second assertion: the cheaper home is making `TestRegisterListBrokers_NeverComposesWithAuthorization` table-driven over both exempt names (same in-memory-transport harness, same policy-grants-nothing setup) rather than a 9th E2E scenario — that runs on every `make check` instead of only in the Dockerized OAuth job, which is the property actually worth having. Your call whether that's this PR or a fast-follow.

Separately, two things worth knowing before merge, neither about the test logic:
- CI is red on `Third-party licenses current` and the three `autobots/pr-checker/jira-*` checks. The licenses failure looks like branch staleness — main picked up two FOSSA/Guardian-alignment commits (a554999, 158af9f) after your last rebase onto cc7a109, and that's exactly the class of check it trips. A rebase should clear it. The Jira checks want the PR title to carry the ticket prefix (e.g. `SOL-152641: ...`) — right now it's linked in the body but not the title.
- Copilot's two comments: the `helpers.sh` one about `set -e` swallowing `mint_token`'s curl-failure diagnostics doesn't hold up — I tested it directly (bash, `inherit_errexit` is off by default, so a failing pipeline inside a function invoked via `$(...)` doesn't abort mid-function; execution reaches the diagnostic and returns normally). The `test-oauth-scenarios.sh` one about the swallowed restore error is a fair, minor diagnosability suggestion, already visible on the thread — nothing to add.

test/e2e-oauth/test-rbac-scenarios.sh
```diff
@@ -0,0 +1,388 @@
...
+# scenario stayed green — and it would start denying every caller the policy
+# grants nothing.
+test_exempt_tools_bypass_the_gate() {
+    local token entry name args predicate mark resp content records
+    local covered source_set
+
+    # Coverage guard first. If the source declares an exempt tool this table
+    # does not cover, fail here rather than silently testing a subset.
+    covered=$(printf '%s\n' "${RBAC_EXEMPT_TOOLS[@]}" | cut -d'|' -f1 | sort -u)
+    source_set=$(exempt_tool_names_from_source)
```
solace-aross, line 304: Verified this against the source directly: `grep -rhoE 'ToolName[[:space:]]*=[[:space:]]*"[^"]+"' internal/tools/` returns exactly `describe-semp-schema` and `list-brokers`, matching `RBAC_EXEMPT_TOOLS`. Good anchor choice — it'll fail loud if a third exempt tool shows up uncovered.

test/e2e-oauth/helpers.sh
```diff
@@ -190,25 +301,235 @@ start_oauth_server() {
...
-    response=$(printf '%s' "$password" | curl -sf --cacert "$KEYCLOAK_CERT" -X POST "$KEYCLOAK_TOKEN_ENDPOINT" \
+    response=$(printf '%s' "$password" | curl -s --max-time 15 --cacert "$KEYCLOAK_CERT" -X POST "$KEYCLOAK_TOKEN_ENDPOINT" \
         -H "Content-Type: application/x-www-form-urlencoded" \
         -d "grant_type=password" \
-        -d "client_id=${HOP1_CLIENT_ID}" \
+        -d "client_id=${client_id}" \
         -d "username=${username}" \
-        --data-urlencode "password@-")
-    token=$(jq -r '.access_token // empty' <<<"$response")
+        --data-urlencode "password@-") || curl_status=$?
```
solace-aross, line 334: Confirmed this was a real gap, not just defensive polish: `set -euo pipefail; response=$(curl <unreachable>)` aborts the script before any diagnostic runs, empirically. The `|| curl_status=$?` here is the correct fix. (Correcting my own prior notes, which had called Copilot's version of this finding a false positive — it wasn't.)

solace-aross (approved):
Approve. This commit closes out both open threads from the last round; checked each rather than taking the description's word for it.

- amitmorade's coverage-gap finding: confirmed by grepping `ToolName = "..."` under `internal/tools/` myself. It returns exactly `describe-semp-schema` and `list-brokers`, matching `RBAC_EXEMPT_TOOLS` and matching `main.go:736,741` where both are registered with no policy argument. The coverage guard (`covered` vs `source_set`) will trip if a third exempt tool is ever added without a table entry — good, that's the failure direction you want.
- Copilot's `set -e` finding on the old `mint_token`: I owe a correction here. My own notes from the last pass called this a false positive. It isn't — I tested it directly (`set -euo pipefail; response=$(curl <unreachable>)`) and the script aborts before reaching any diagnostics, exit code passed straight through. The fix (`|| curl_status=$?` at helpers.sh:334, structured status handling after) is the correct pattern and actually needed, not just defensive polish.
- The trap fix in test-oauth-scenarios.sh (log_warn instead of swallowing the restore failure) is a straightforward improvement, nothing to add.

No Docker in this session, so I couldn't re-run e2e-oauth myself to sabotage-test the new scenario the way the last pass did. Leaning on the static source cross-check above plus CI's own green e2e-oauth run on this head as evidence instead — flagging that gap rather than asserting a live verification I didn't do.

PR #268, opened 2026-08-07
ci(SOL-152951): refresh the build/test third-party inventory for Guardian SHA pins

solace-aross:
## What

`main` has been red on **Third-party licenses current** since #257 merged. This makes it green and fixes two silent-pass defects in the self-test that the red baseline was hiding.

`THIRD_PARTY_BUILD_TEST.md` is one of the two third-party lists the Solace public-repository Legal checklist requires. #257 SHA-pinned every action, added five Solace composite actions plus `actions/github-script`, and dropped the SCA reusable workflow. The inventory was not updated to match, so the compliance artifact described thirteen actions at the wrong ref, omitted five, and listed one component nothing uses.

The gate was right. The data it guards was stale.

## Changes

**`THIRD_PARTY_BUILD_TEST.md`**

- Actions table carries the short SHA the check compares, plus the release tag that SHA resolves to. A SHA alone tells an auditor nothing about which version they are reading.
- Every licence re-read **at the pinned ref**, not at the default branch, and licence links now resolve at the tag. That is the exact mistake the file already documents for `segmentio/asm`, which relicensed after the version we consume. `actions/github-script` uses `LICENSE.md`, so the obvious URL 404s; there is a note.
- The Solace table moves from one reusable workflow to the five composite actions that replaced it.

**`.github/scripts/build-test-licenses-check.test.sh`**

All four self-test failures reported `expected exit 0, got 1`. They were one defect, not four: every case asserting a clean run inherits the stale baseline. Two more only became visible once the baseline was green, and both are the silent-pass shape this suite exists to catch.

| Defect | Effect | Fix |
|---|---|---|
| `bump_action_version` rewrote the literal `actions/checkout@v4` | Matched nothing after SHA-pinning. One case asserted exit 1 against an untouched tree; its pair asserted exit 0 against a row change with no workflow change behind it. | Matches the pin by shape, and proves it substituted |
| `change_version` rebuilt a fixed four columns | Silently truncated the fifth column from the widened actions table, leaving a row that still parsed | Rebuilds from `NF` |

`drop_row` and `change_version` now fail loudly when they match nothing, so a case naming a since-renamed row cannot go vacuous. That is not hypothetical: it is how the licence-hint case broke.

**Each new guard was verified by injecting the defect it is meant to catch**, not by reading it. All three reported `the test's own mutation failed` rather than passing.

## Verification

| Command | Result |
|---|---|
| `make check` | pass, coverage 88.3% (85% floor) |
| `build-test-licenses-check.sh` | pass, 41 components, 41 rows |
| `build-test-licenses-check.test.sh` | 26 passed, 0 failed |
| `licenses-check.sh` / `licenses-check.test.sh` | pass (other two steps of the same job) |

No CHANGELOG entry: nothing under `internal/config/`, `internal/tools/`, or `tools.yaml` is touched, and this file does not ship.

## Two things I did not fix here

**Dependabot will fail this gate on every action bump.** It re-pins SHAs and does not edit Markdown, so each bump PR needs a manual row update. That is the gate working as designed, but the ergonomics deserve a decision rather than a surprise. Not in this PR.

**Composite actions under `.github/actions/` are invisible to the check.** Check 3 greps `.github/workflows/` only. The directory does not exist today so there is no live exposure, but #257's own description claims to add `.github/actions/guardian-scan`, and if that lands its third-party `uses:` would go undocumented with the gate green. That is a fail-open in the property this script calls its most important one. Left out deliberately: it is a discovery-semantics change that deserves its own test and review, not a rider on a data fix.

SOL-152951

### DIFF
--- a/.github/scripts/build-test-licenses-check.test.sh
+++ b/.github/scripts/build-test-licenses-check.test.sh
@@ -39,6 +39,11 @@ REPO_ROOT=$(git rev-parse --show-toplevel)
 CHECK="$REPO_ROOT/.github/scripts/build-test-licenses-check.sh"
 DOC="THIRD_PARTY_BUILD_TEST.md"
 
+# A syntactically valid 40-character SHA that no action is pinned to, used by the
+# version-bump cases. A constant rather than a literal at each use site, so the
+# mutation and the row it expects cannot drift apart.
+BUMPED_SHA="0123456789abcdef0123456789abcdef01234567"
+
 pass=0
 fail=0
 
@@ -47,7 +52,16 @@ fail=0
 # sed delimiters a source of bugs in the test itself.
 
 drop_row() { # <tmp> <name>
+    # Confirm a row actually went. `grep -v` that matches nothing exits 0 and
+    # copies the file unchanged, so a case naming a row that has been renamed or
+    # deleted would assert against a pristine tree. That is not hypothetical: the
+    # licence-hint case below named the SCA reusable workflow, and when
+    # DATAGO-147232 removed it the mutation quietly became a no-op.
+    local before after
+    before=$(wc -l <"$1/$DOC")
     grep -vF "\`$2\`" "$1/$DOC" >"$1/t" && mv "$1/t" "$1/$DOC"
+    after=$(wc -l <"$1/$DOC")
+    [ "$after" -lt "$before" ]
 }
 
 add_row() { # <tmp> <name> <version>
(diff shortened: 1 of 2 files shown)


amitmorade (approved):
Approving. CI is also green

I verified the claims independently rather than reading them:

- All 14 action SHAs match what's actually in `.github/workflows/` — no omissions, no extras.
- All 14 SHA→release tag mappings resolve correctly via the GitHub API.
- Every licence link returns 200 **at the pinned tag**, and every SPDX identifier matches GitHub's own detection.
- `build-test-licenses-check.sh` → 41 components / 41 rows. Self-test → 26 passed, 0 failed.

The part that earns the approval is the test fix, not the data refresh. I reconstructed the pre-PR state and confirmed the vacuous case directly: the old `drop_row` naming the since-removed SCA workflow row leaves the check with nothing to complain about, and it exits 0. I also added probe cases calling the new `change_version` and `drop_row` with a nonexistent component — both now report `the test's own mutation failed` rather than asserting against an unmodified tree.

PR #274, opened 2026-08-07
SOL-152962: add persist-credentials: false to remaining checkouts

solace-aross:
## What this does

Adds `persist-credentials: false` to every `actions/checkout` step that was missing it — `release.yml`, `llm-eval.yml`, `build-and-test.yml`, `ci-pr.yaml`, `dco.yaml`. `build-image.yaml` and `guardian-scan.yaml` already had it. Without it, the job's `GITHUB_TOKEN` stays in the local git config for the rest of the job, readable by any later step, script, or dependency.

## The one non-mechanical part

`dco.yaml`'s real gate does a live `git fetch` against `origin` right after checkout (this repo is internal, not public, so that needs auth). Turning off persisted credentials there would break it, so that one command now authenticates itself instead — with a Basic auth header, matching how `actions/checkout`'s own git-auth-helper does it (GitHub's git-over-HTTPS frontend only accepts Basic, not Bearer). Verified directly against this repo with a real token before relying on it. Added a same-repo self-test in `ci-pr.yaml` (`dco_selftest` job) that exercises the identical mechanism on every PR, so a future break in the auth scheme fails fast on the PR itself rather than deadlocking `main` once merged — the ruleset's bypass list is empty, so a broken required check there has no escape hatch.

## Also in this PR

`.github/ADMIN_SETUP.md` and `CHANGELOG.md` updated to reflect three related items closed directly via the GitHub API in the same pass (no code diff, so no separate PR): Actions can no longer approve pull requests (SOL-152958), the default Actions workflow token permission is now read-only (SOL-152959), and the `main-protection` ruleset now requires Code Owner review and resolved threads (SOL-152961). Fork PR workflow approval (SOL-152960) stays open — no REST endpoint exposes that setting, so it needs an org admin in the UI.

## Testing

`make check` passes (88.3% coverage). All 5 touched workflow files parse as valid YAML. The `dco.yaml` auth fix was verified live against this repo with a real token before merging it into the workflow.

Raised from a Claude-assisted security scan by Reuben D'Souza: https://solacedotcom.slack.com/archives/C0BLSM4NFRS/p1786118844182969

Tracks SOL-152962, parent epic SOL-152410.


bczoma (approved):
No correctness or security issues found — the non-mechanical piece (dco.yaml's re-authenticated fetch) is the right fix, done the right way, with a self-test to catch regressions. Good to merge as-is.

PR #275, opened 2026-08-07
SOL-150785: SEMP output-schema drift guard (v2 monitor spec-consistency + SEMPv1 fixture coverage) 

solace-lootii:
## Summary

- **(B) SEMPv1 fixture coverage** — complete. Fresh audit of all SEMPv1 read calls found exactly one with no round-trip test at all: `internal/tools/queuemetrics/queue_response.go` had no `queue_response_test.go` (every sibling package has one), and the incidental coverage it got via `handler_test.go` missed 2 of 6 fields (`highWaterMarkBytes`, `deliveredUnackedMsgCount`). Added `queue_response_test.go` covering all 6 fields against the real fixture, plus an absent-fields case. Verified it catches a regression (sabotaged an XML tag, confirmed both new tests fail, reverted).

  Four other partial-coverage findings from the same audit (`redundancy`, `spool_response`, `hardware_response`, `discardstats`) are explicitly **out of scope** for this ticket — they already have a round-trip test, just not exhaustive field coverage, which is closer to SOL-150787's test-quality territory than this ticket's "does each call have a test" ask. Not fixed here; happy to note them somewhere if useful.

- **(A) SEMPv2 monitor spec-consistency — Phase 1 of 5, foundation only.** Added `extractResponseFields` to `internal/semp/sempv2/operation.go`, mirroring the existing `extractBodyFields` pattern: resolves an operation's 200 response schema through the `{"data": ...}` envelope (unwrapping array-of-item shapes too) to the real embedded spec's per-field types — something the embedded `__private_monitor__` spec always had but nothing in the codebase read until now. `Operation.ResponseFields map[string]string` is the new surface; **nothing consumes it yet**.

  Verified against real data, not just unit tests: ran extraction across all 891 parsed operations with no panics and no malformed output (726 resolved, 165 correctly nil for meta-only responses). Confirmed empirically (not assumed) that every operation across all three embedded specs uses `"200"` as its sole success code. Caught and fixed two real issues in review before this landed: three defensive branches with zero test coverage (added a synthetic internal test file, now 100%), and a doc comment that overclaimed present-tense "callers do X" behavior when no caller exists yet.

  **Remaining for (A), not in this PR:** Phase 2 (schema generator + wiring 2 pilot tools + an e2e checkpoint before wider rollout — this changes runtime validation behavior, not just adds tests), Phase 3 (roll out to the rest of the monitor tools + `queuemetrics`), Phase 4 (drift-negative tests: unexpected/missing/type-changed field rejection), Phase 5 (the CI breakage guard itself). additionalProperties policy already decided in planning (type-check-if-present, not required, except each tool's own resource-identifier field) but not yet applied anywhere.

- **(C) config/action (write tools) — deferred to a separate ticket(https://sol-jira.atlassian.net/browse/SOL-152947), not in this PR's scope.** The ticket's own deferral condition ("revisit when write tools land") already fired — write tools + an embedded config/action spec both landed (SOL-148460 on 2026-06-29, SOL-151646 on 2026-07-13) after this ticket was written but before it was picked up. Building the config/action equivalent of part A is comparable in scope to all of part A itself, so it's being filed as its own follow-up ticket rather than folded in here.

## Test plan

- [x] `make check` (build, vet, lint, race tests, coverage gate) green at 88.4%, above the 85% floor
- [x] Sabotage-tested the new `queue_response_test.go` (broke an XML tag, confirmed both tests fail, reverted)
- [x] Verified `extractResponseFields`/`resolveSchemaRef`/`schemaTypeName` at 100% branch coverage via synthetic malformed-input tests (dangling `$ref`, array with no `items`, etc.) — none of these paths are reachable with real spec data today, so real-spec tests alone couldn't exercise them
- [x] Ran extraction across all 891 real parsed operations — no panics, no empty/malformed types
- [x] Merged latest `main` in (which now includes the merged SOL-150787 coverage-gate work); no conflicts, `make check` still green after merge


solace-aross (approved):
Verified the two new test suites earn their green rather than just passing by accident. Worktree at 819e43e:

- Sabotaged the `highWaterMarkBytes` XML tag in `queue_response.go` — both `TestQueueDetailResponse_RoundTrip` and `_OmitsAbsentFields` failed as expected, then passed clean on revert.
- Sabotaged the array-unwrap branch in `extractResponseFields` (operation.go) — `TestParseSpecs_ResponseFields` and `TestExtractResponseFields_SingleObjectAndArrayAgree` both caught it.
- Full `make check` at head: green, 88.4% coverage.

`ResponseFields` is additive with zero callers today, so there's no behavior change beyond the extraction logic itself. That logic is well-documented and matches the empirical claims in the description (891 ops parsed, 726/165 resolved/nil split). Approving.

PR #276, opened 2026-08-07
docs: add STRIDE threat model, link from architecture.md

solace-aross:
## What

Adds `docs/internal/threat-model.md` — STRIDE across the seven trust
boundaries scoped in SOL-152900: MCP client → server, server → broker,
OAuth token exchange, config and secrets, embedded OpenAPI specs
(`describe-semp-schema`), multi-broker isolation, and supply chain.
Linked from `architecture.md`.

Every threat carries a `file:line`-cited mitigation, verified by reading
the code, or an explicit accepted-risk statement. All accepted risks are
consolidated in §8 for exception review.

## Why

Threat-modelling a design and recording it is a PRODUCTS-tier requirement
on the [Open Source Solace Software Checklist](https://sol-jira.atlassian.net/wiki/spaces/DC/pages/6616154181/Open+Source+Solace+Software+Checklist),
and this repo (now SolaceProducts) has none. The 23 July four-approver
review passed without this artifact.

## Findings worth a look before Compliance signs off

- **§6, accepted risk #17** — authorization is tool-scoped only
  (`internal/authz/authz.go:27-123`); there is no per-broker authorization
  at the MCP layer, so any caller authorized for a tool can target any
  configured broker. Every other accepted risk in this doc is a scoped
  tradeoff — this one reads as a structural gap and probably deserves its
  own decision rather than a blanket accept.
- **§7, accepted risk #18** — `Third-party licenses current` runs in CI
  but isn't in the live required-status-checks ruleset on `main`, despite
  `ADMIN_SETUP.md` saying it should be. Confirmed via `gh api .../rulesets/13942241`.
- **§3 correction** — `architecture.md` never mentioned the token-exchange
  circuit breaker (`internal/tokenexchange/circuitbreaker.go`), which is
  real and on by default. Called out and fixed in passing.

## Verification

Docs-only change (one new file, a 3-line link addition to
`architecture.md`); no `.go` files touched. `go build ./...` and
`go vet ./...` pass. No CHANGELOG entry — doesn't touch production
surface (config, tool schemas, or native tool packages).

## Not in this PR

Sharing with Compliance and recording their response on SOL-152900, and
any exception sign-off for §6/#17 — both are next steps on the ticket,
not something this PR can do on its own.

SOL-152900


bczoma (approved):
Accurate, well-sourced, no fabricated citations found. Good to merge; the two flagged items (#17 especially) are the right things to route to the four-approver/Compliance review the PR description says is next, not something this PR needs to resolve itself.

Notes:
Two findings actually matter and are correctly flagged as needing a decision, not a rubber stamp:
- #17 (§6): Authorization is tool-scoped only — no per-broker authorization at the MCP layer, so any caller authorized for a tool can hit any configured broker. Verified against internal/authz/authz.go — the Policy struct genuinely carries no broker dimension. This is real and structural, not a scoped tradeoff like the others; it only matters for multi-tenant deployments (single-tenant-per-instance sidesteps it), but the doc is right to call it out separately for an explicit accept/fix decision.
- #18 (§7): Third-party licenses current is documented as required in ADMIN_SETUP.md but isn't in the live ruleset — confirmed via the API. Correctly framed as a one-line fix rather than a design question.

PR #279, opened 2026-08-09
SOL-152979: sanitize broker URLs in log output as defense-in-depth

solace-aross:
## What

Several runtime log sites logged a broker URL raw, bypassing the `sanitizeURLString` sanitizer this codebase already built for exactly this purpose (`BrokerConfig.LogValue`):

- `internal/semp/pool.go`: the once-per-alias "broker connection created" log line.
- `internal/semp/resilience/sender.go`: `Sender.brokerURL`, sanitized once at construction (covers all 11 log sites in `sender.go`/`retry.go` that read it — the field is logging-only).
- `internal/semp/sempv1/client.go`, `sempv2/client.go`: `baseURL` in both `HTTPClient.LogValue` implementations.

Exported `config.sanitizeURLString` as `config.SanitizeURLString` so `internal/semp` and `internal/semp/resilience` can reuse it.

## Why

`validateBrokerURL` already rejects any broker URL with embedded userinfo at config load, in every mode — so this is **defense-in-depth hardening, not an active credential leak today**. Each of these log sites bypassed both the sanitizer and the `ReplaceAttr` redaction safety net (whose redacted-key list doesn't include `url`/`base_url`/`broker`), leaving no independent defense if that upstream validation guarantee is ever weakened or bypassed by a future change.

This ticket's severity/framing was corrected mid-flight: it was originally filed as an active leak, then reframed after a plan-review pass found the upstream validation guarantee.

## Testing

- New regression test `internal/semp/pool_internal_test.go` (`TestGetOrCreate_SanitizesURLInLog`): builds a `BrokerPool` around a fake `BrokerSource` carrying a credentialed URL (bypassing `validateBrokerURL` deliberately, since no production caller can) and asserts the log line never carries the credential.
- Revert-verified: reverting the fix makes the new test fail with the credential visible in the log output; reapplying makes it pass.
- `make check` green, 88.5% coverage held (same as baseline).

## Scope note

`internal/tokenexchange/types.go:167` (and `Params.String()`/`GoString()` at lines 153-160) has the same pattern from a different source (IdP token URL, not broker URL) — noted as a follow-up, not in this fix's scope. `internal/banner/banner.go:84` (OIDC issuer) is the same pattern too. Will file a follow-up ticket for both.

Tracked under SOL-152979.


internal/semp/resilience/sender.go
```diff
@@ -136,7 +136,13 @@ func New(httpClient *http.Client, sempCfg *config.SEMPConfig, authn auth.Authent
...
 		sem:           sem,
 		rateLimiter:   limiter.C(),
-		brokerURL:     brokerURL,
+		// Sanitized once here rather than at each of the many log sites below —
+		// brokerURL is logging-only (see the field doc), so nothing downstream
+		// needs the raw form. Defense in depth: validateBrokerURL already
+		// rejects credentialed URLs at config load, but this keeps that
+		// guarantee from being the only thing standing between a broker URL
+		// and the log stream (SOL-152979).
+		brokerURL: config.SanitizeURLString(brokerURL),
```
bczoma, line 145: `brokerURL` changed meaning from raw to sanitized while keeping the name and type `string`, and the invariant is only established here in `New` — `sender_test.go` assigns the field directly in several places, so the constructor isn't the only writer. A future change that reads this to build a request URL (the natural reading of the name) silently gets a re-serialized value or `<unparseable url>`.

Consider `type logURL string` with the sanitizer as its only constructor; the direct test assignments then have to opt in, and the guarantee survives refactoring instead of depending on the field comment being read.

internal/config/config.go
```diff
@@ -1564,12 +1564,15 @@ func sanitizeURL(u *url.URL) string {
 	return cp.String()
 }
 
-// sanitizeURLString parses s and returns the URL with any userinfo stripped.
+// SanitizeURLString parses s and returns the URL with any userinfo stripped.
```
bczoma, line 1567: Small contract gap, no reachable impact: `url.Parse("admin:pw@host:943")` yields `Scheme="admin"`, `Opaque="pw@host:943"`, `User==nil`, so this returns the string verbatim — "returns the URL with any userinfo stripped" is unqualified but not universally true. Not reachable in production (`validateBrokerURL` rejects non-http/https schemes before the userinfo check, and `LoadConfig` aborts on that error), so this is a comment defect rather than a leak.

Worth the two lines anyway, because the comments at :438 and :455 justify the sanitizer as cover for logging a config *before* validation — precisely the unvalidated case it doesn't handle:

```go
if u.Opaque != "" || (u.Scheme != "http" && u.Scheme != "https") || u.Host == "" {
    return "<unparseable url>"
}
```

A table test over empty / no-userinfo / `user@host` / `user:pass@host` / schemeless `user:pass@host:port` / `//user:pass@host` / unparseable would pin the contract.

bczoma (commented):
Nice change, and the framing is what makes it easy to review — arguing the severity *down* and naming `validateBrokerURL` as the upstream guarantee saved me from over-escalating. I checked that claim and it holds: one production `BrokerConfig` source, `validate()` unconditional from `LoadConfig`, userinfo rejected before the dev/prod branch, no CLI or dynamic path around it. The "logging-only" claim on `Sender.brokerURL` holds too — all reads across `sender.go`/`retry.go` are `slog.String("broker", ...)`, so sanitizing once in `New` is the right call. And the revert-verification on the new test is the detail that makes it trustworthy.

Three things I'd like before merge, then some follow-up material.

**1. `docs/internal/secure-logging-rules.md:62` now contradicts the code.** Rule 4's "Prefer" example is `slog.String("url", config.URL)` — logging a broker URL raw, which is exactly what this PR removed from `pool.go`. That's the doc `/check-logs` enforces and that these code comments cite as normative, so as it stands the standard instructs the next contributor to reintroduce SOL-152979. Suggest `slog.String("url", config.SanitizeURLString(cfg.URL))`, keeping the `slog.Any` contrast intact. This is the one I'd hold on.

**2. Test coverage — and a correction to the sempv2 comment above.** Copilot's suggestion to mirror SEMPv1's `TestHTTPClient_LogValue_ExcludesCredentials` won't give you what you want: that test builds its client from `srv.URL`, which carries no userinfo, so it passes with the sanitizer removed. Worth fixing the model before copying it. Also uncovered: `resilience.New` — the highest-leverage site, since one sanitize there feeds ~12 log calls. Suggested set:

- add a credentialed-`baseURL` case to the sempv1 table test (fixing the existing blind spot), then mirror for sempv2;
- assert `New(..., "https://u:p@h:943", ...).brokerURL` has no credential but retains the host.

**3. `CHANGELOG.md:55` says "three runtime sites" but enumerates four** (`pool.getOrCreate`, `Sender`, sempv1 `LogValue`, sempv2 `LogValue`). Separate from the identifier-name point above — "four runtime sites", or "three subsystems" if you meant to group them.

**Architectural note, not a blocker.** This hardens four sites by hand and the Scope note names three more out of scope, while the one central choke point — `redactSecretAttr`/`ReplaceAttr` in `cmd/server/main.go:58-89` — went untouched. That shape doesn't converge; the fifth site gets found by the next audit. What would close the class rather than the instances is a *value-transform* (not `[REDACTED]`, which would destroy triage value) in `redactSecretAttr` for url-ish keys, with the per-site calls kept as the second layer. Paired with a line in Rule 2 stating that URL-valued attributes must route through `config.SanitizeURLString` — that rule currently exists only in Go comments, and since Rule 3's key list deliberately excludes `url`/`base_url`, Rule 2 is the only control for these fields.

**One doc nit that can't be an inline comment** (file isn't in the diff): `docs/internal/threat-model.md:55` cites `sempv1/client.go:49-53` / `sempv2/client.go:73-77` for `HTTPClient.LogValue`, but those are now `:67` and `:91`, and the row doesn't record that `base_url` is sanitized. Since that's the audit artifact for this control, worth appending "...and routes it through `config.SanitizeURLString`", and possibly a row for the `Sender`/pool sites this PR brings under the same mitigation.

Happy to approve once 1-3 land.

amitmorade (commented):
Balazs's architectural note lines up with something I was thinking about too, from a slightly different angle.

My comment above was about the call-site side: `SanitizeURLString` is a free function, so calling it is opt-in, and a future log site could miss it the same way these did before this PR. What Balazs raised looks at the catch-all side: `redactSecretAttr`/`ReplaceAttr` would normally be the safety net for this, but its key list currently excludes `url`/`base_url`, so it wouldn't catch a miss there either.

Not suggesting a change to this PR, just flagging that these might be two pieces of the same gap, in case it's worth a closer look down the line.

CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+- Broker URLs logged at three runtime sites — the once-per-alias "broker connection created" line, the `Sender`'s retry/failure logging (`internal/semp/resilience`), and both SEMPv1/SEMPv2 `HTTPClient.LogValue` implementations — now go through the same `sanitizeURLString` sanitizer `BrokerConfig.LogValue` already applied, instead of logging the URL raw. `validateBrokerURL` already rejects any broker URL with embedded userinfo at config load in every mode, so none of these could carry a credential today; this closes the gap where that validation guarantee was the *only* thing standing between a broker URL and the log stream, with no independent defense at the log sites themselves and no coverage from the `ReplaceAttr` redaction safety net (its redacted-key list doesn't include `url`/`base_url`/`broker`). No behavior change for a valid, non-credentialed URL. Found in a Claude-assisted security sweep. Tracked under SOL-152979.
```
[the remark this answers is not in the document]
solace-aross: Fixed in 5814813 — `sanitizeURLString` → `SanitizeURLString`. Also reworded "three runtime sites" to "four" per bczoma's review; the entry undercounted them.

internal/semp/sempv2/client.go
```diff
@@ -82,11 +82,15 @@ type HTTPClient struct {
...
+// SanitizeURLString for the same reason BrokerConfig.LogValue routes URL
+// through it: validation already rejects credentialed URLs, but this keeps
+// that guarantee from being the only thing standing between a broker URL and
+// the log stream (SOL-152979).
 // See docs/secure-logging-rules.md Rule 2.
 func (c *HTTPClient) LogValue() slog.Value {
 	return slog.GroupValue(
-		slog.String("base_url", c.baseURL),
+		slog.String("base_url", config.SanitizeURLString(c.baseURL)),
 	)
```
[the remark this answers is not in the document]
solace-aross: Added in 5814813. Also added a credentialed-baseURL case — building from srv.URL alone carries no userinfo, so a straight mirror of sempv1's original test wouldn't have exercised the sanitizer at all (bczoma caught the same gap in his review).

internal/semp/resilience/sender.go
solace-aross: Declined, after running a dedicated skeptical pass first (checked all 12 reads of d.brokerURL — sender.go:226/375/392, retry.go:184-304 — every one is slog.String("broker", ...); nothing in this package builds a request URL from it, so there's no live read site the type would protect). It's also unenforceable where it'd matter: logURL("raw") is a legal same-package conversion, so the 7 direct test assignments in sender_test.go stay legal with a one-token cast — the guarantee would be nominal, not enforced. You'd already verified the logging-only contract holds in your review body, and the field comment plus the constructor rationale at sender.go:139-144 state it at both the declaration and its only production writer. Happy to revisit if a future change adds a non-logging read of the field.

internal/config/config.go
solace-aross: Fixed in 5814813 — tightened SanitizeURLString to also reject non-http(s)/no-host input, including the opaque schemeless form, matching validateBrokerURL's own contract, and added the table test (empty/no-userinfo/user@host/user:pass@host/schemeless/scheme-relative/non-http(s)/no-host/unparseable).

One correction from a self-review pass on my first cut at this: the Opaque != "" check I added is actually dead — Opaque != "" implies Host == "" in every case I could produce, so it's the scheme/host check that rejects the schemeless form, not the Opaque check. Kept the Opaque check anyway as its own explicit guard rather than relying on that being coincidental, but fixed the doc comment, which had misattributed the rejection to it.

solace-aross:
@bczoma @amitmorade — addressed in 5814813 (also covers the two inline threads separately).

Your three before-merge items:
1. Rule 4's example — fixed, now routes through `SanitizeURLString`.
2. Test coverage — added the credentialed-baseURL case to sempv1's test (the original only used `srv.URL`, so it never exercised the sanitizer's actual stripping), mirrored it in sempv2 from scratch, and added a test on `resilience.New` — the highest-leverage site.
3. CHANGELOG — "four sites", not three; fixed the stale function name too.

On the architectural note: agreed it's a real gap, but the `redactSecretAttr` fix doesn't close it the way it looks like it would. The `broker`-keyed log sites — 12 of the 15 this PR hardens, all in `sender.go`/`retry.go` — log an alias under that key, not a URL: `SanitizeURLString("prod-broker")` returns `<unparseable url>`. A url-ish key transform in `redactSecretAttr` either misses those 12 sites or blanks the alias if `broker` gets added to the key list. Ran that past a dedicated skeptical pass before deciding not to touch it here — it needs to resolve what `broker` means across the log taxonomy first, which is its own design question.

Took the one piece of it that was cheap and in scope: Rule 2 now states URL-valued `LogValuer` attributes must route through `SanitizeURLString`. That immediately put `internal/tokenexchange/types.go`'s `idp_endpoint` field in violation with no exception recorded, so filed [SOL-153000](https://sol-jira.atlassian.net/browse/SOL-153000) for it and noted the exception in the doc in the meantime.

Skipped the threat-model.md line-reference nit per standing policy on nits.

solace-aross:
Rebased onto main (picked up SOL-150813/#245 and SOL-152902/#263, no conflicts). CI rerunning; pushed 6193b76.

bczoma (dismissed):
Looks good now!

solace-aross:
Rebased onto main (`9d75d63`) to pick up #278. One conflict, in `docs/internal/secure-logging-rules.md`: main's docs pass had already turned the "when Story 1B adds OAuth config" stub into concrete completed bullets, while this branch kept the stub and appended the new URL-sanitization rule after it. Resolved by keeping both real additions and dropping the stale stub — no content lost either side.

`make check` green post-rebase (88.6% coverage). This force-push dismissed your approval, @bczoma — re-requested.

solace-aross:
Rebased onto main (`c0f5c6d`) to pick up v0.7.0. Same changelog issue as #280: the rebase misplaced this PR's entry into the already-released `[0.7.0]` section instead of `[Unreleased]`. Fixed and force-pushed. `make check` green (88.6%). @bczoma re-requesting isn't needed — your review request is still standing from the earlier rebase.

bczoma (approved):
Approved

PR #280, opened 2026-08-10
SOL-152980: return structured results, not bare errors, from CallTool's local failure paths

solace-aross:
## What

`CallTool` had one correct error-return idiom and several incorrect ones. Per `errors.go`'s own documented policy on `buildErrorResult`: "Per the MCP spec, tool execution errors should be returned as results (not protocol-level errors) so the LLM can see them and self-correct." Only the `execution_error` branch followed that; every other failure branch returned a bare Go error, bypassing `buildErrorResult` entirely.

The three branches that matter most fire **after** `handler.Handle` has already run: `nil_result`, `output_validation_error`, and `marshal_error`. For a destructive tool, the broker-side mutation may have already succeeded by the time one of these fires, but the caller got a bare protocol error with no `retryable` signal and no way to tell the action already took effect.

## Fix

Two helpers, since the failure branches aren't all the same shape:

- `buildLocalErrorResult(err)` — for errors this package constructs directly and can vouch for verbatim (missing broker, unknown broker alias, parameter validation, nil/invalid handler result, output validation, marshal failure). Deliberately distinct from `buildErrorResult`, whose type-switch exists to suppress arbitrary broker/handler-originated text behind a generic message (`TestCallTool_PlainError_IsErrorResult` pins that suppression) — routing a pre-crafted local message through that switch would silently replace it with the misleading generic one.
- `buildBrokerResolutionErrorResult` — dispatches `classifyBrokerError`'s two outcomes correctly: `unknown_broker` is pre-crafted (→ `buildLocalErrorResult`), but `broker_init_error` wraps an arbitrary underlying construction error (today cookie-jar/authenticator setup, potentially a future OAuth token-exchange `ExchangeError`) that is not vouched-for text, so it goes through `buildErrorResult` instead.

`unknown_tool` is intentionally left as a protocol-level error: the MCP spec puts an unknown tool name in the protocol-error bucket, and go-sdk's server already rejects it before `CallTool` is reached in production.

Updated two comments that stated an invariant — "the agent only ever sees the sanitized message" — that becomes false once `buildLocalErrorResult` exists.

## Testing

- Strengthened the 6 existing tests that asserted directly on the bare error return, to assert on the returned `*mcp.CallToolResult` instead, via a new `callToolResultText` test helper. `TestCallTool_OutputValidationFails` is the ticket's namesake regression test.
- Revert-verified: all 6 fail against the pre-fix code with the exact bare-error behavior the ticket describes, and pass after.
- `make check` green, 88.5% coverage held.

## Review

Went through two independent skeptical plan/diff reviews (Fable). The plan review caught a real defect in my first draft: routing `broker_init_error` through the same verbatim helper as `unknown_broker` would have echoed unvouched-for construction-error text to the agent. Fixed before implementation.

Tracked under SOL-152980.


internal/tools/manager_test.go
```diff
@@ -152,34 +152,55 @@ func TestRegister_DuplicatePanics(t *testing.T) {
...
+	}
+	sc, ok := result.StructuredContent.(map[string]any)
+	if !ok {
+		t.Fatalf("StructuredContent = %T, want map[string]any", result.StructuredContent)
+	}
+	if sc["retryable"] != false {
+		t.Errorf("retryable = %v, want false", sc["retryable"])
+	}
+	return result.Content[0].(*mcp.TextContent).Text
+}
```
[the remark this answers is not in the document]
solace-aross: Fixed in b6bf9fc — guarded both the length check and the type assertion (Fatalf on mismatch instead of an unguarded index/cast), and made the retryable check fatal too so a mismatch stops the test immediately instead of continuing into what was the panic-prone line.

internal/tools/manager.go
```diff
@@ -299,10 +311,13 @@ func logToolResult(ctx context.Context, tool string, broker *string, start time.
...
-	// it is the one place the raw text is kept (the agent only ever sees the
-	// sanitized message from buildErrorResult). Folding it onto this line, rather
-	// than a separate emit, keeps one tool error in one record at a single (ERROR)
-	// level.
+	// it is the one place the raw text is kept. For a broker/handler-originated
+	// error the agent only ever sees the sanitized message from buildErrorResult;
+	// for a local error the agent sees this same err.Error() text verbatim via
+	// buildLocalErrorResult, which is safe precisely because that text is one
+	// this package wrote itself, never broker- or handler-originated
+	// (SOL-152980). Folding "detail" onto this line, rather than a separate
```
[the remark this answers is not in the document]
solace-aross: Fixed in b6bf9fc — "that text is one this package wrote itself" -> "that text is something this package wrote itself".

solace-aross:
Rebased onto main (`1f31664`) to pick up v0.7.0. The rebase's context-matching silently placed this PR's changelog entry into the already-released `[0.7.0]` section instead of a fresh `[Unreleased]` block — same '### Fixed' heading text existed in both places pre/post-release, so patch application matched the wrong one. Caught it, moved the entry back, and force-pushed with the fix included. `make check` green (88.5%).

internal/tools/errors.go
```diff
@@ -157,6 +162,62 @@ func (m *ToolManager) buildErrorResult(err error, brokerAlias string) *mcp.CallT
...
+// retryable; an unrecognized construction error is not).
+func (m *ToolManager) buildBrokerResolutionErrorResult(errorType string, err error, brokerAlias string) *mcp.CallToolResult {
+	// errorType is classifyBrokerError's own return value — the two string
+	// literals it can produce ("unknown_broker", "broker_init_error") are
+	// asserted against directly here rather than via a shared const, matching
+	// every other errorType value in manager.go.
+	if errorType == "unknown_broker" {
+		return buildLocalErrorResult(err)
+	}
+	return m.buildErrorResult(err, brokerAlias)
```
helen-zhu-solace, line 218: The `broker_init_error` path has no test. If that routing ever flips, the suite stays green and unvouched error text reaches the agent — the exact thing the plan review caught. The `unknown_broker` side is already covered by TestCallTool_UnknownBroker_PreservesCallerCasing; it's just this half that's bare.

solace-aross: Fixed in 5527011 — added TestClassifyBrokerError_BrokerInitError_SuppressesUnvouchedText, driving classifyBrokerError/buildBrokerResolutionErrorResult directly with a synthetic non-ErrUnknownBroker error (the branch is not reachable through CallTool with today's BrokerPool, per classifyBrokerError's own doc comment). Confirmed it catches the regression: forcing the unknown_broker branch to also handle broker_init_error made the test fail with the underlying dial error leaking verbatim; reverted, and it now asserts the generic message.

solace-wmaryam (dismissed):
I agree with Helen's analysis — the `broker_init_error` coverage was the right thing to pin, and the fix in 5527011 handles it well. Otherwise looks good to me.

internal/tools/manager.go
```diff
@@ -197,26 +204,31 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
...
 
-	// Guard against nil results from handler.
+	// Guard against nil results from handler. This and the two checks below
+	// fire after handler.Handle has already run — for a destructive tool, any
+	// broker-side mutation may have already taken effect — so each returns a
+	// structured result (not a bare error) precisely so the agent gets a
+	// retryable=false signal instead of a protocol error it might mistake for
+	// "nothing happened, safe to retry" (SOL-152980).
 	if toolResult == nil || toolResult.StructuredContent == nil {
 		errorType = "nil_result"
```
helen-zhu-solace, line 214: `nil_result` has no test, and it's the one post-handler branch that's both
reachable and unpinned. `output_validation_error` got its regression test, and
`marshal_error` can't fire (`ValidateOutput` already marshals the same
`StructuredContent` a few lines up, so anything that would break
`MarshalIndent` here surfaced as `output_validation_error` instead).

That leaves this branch carrying the PR's own argument with nothing holding it
in place: `handler.Handle` has run, a destructive tool may have already
mutated the broker, and `retryable: false` is what keeps the agent from
retrying an action that already took effect. If a later refactor flips this
back to `return nil, toolErr`, the suite stays green.

Cheap to pin with the helper you already added — both halves of the condition,
since a non-nil result with nil `StructuredContent` is the likelier shape in
practice:

```go
func TestCallTool_NilResult(t *testing.T) {
	for _, tc := range []struct {
		name   string
		result *ToolResult
	}{
		{"nil result", nil},
		{"nil structured content", &ToolResult{}},
	} {
		t.Run(tc.name, func(t *testing.T) {
			mgr := NewToolManager(newTestPool(t))
			handler := newStubHandler("test-tool")
			handler.handleFn = func(context.Context, *ToolContext, map[string]any) (*ToolResult, error) {
				return tc.result, nil
			}
			mgr.Register(handler)

			result, err := mgr.CallTool(context.Background(), "test-tool", map[string]any{
				"broker":     "dev",
				"msgVpnName": "default",
			}, Identity{})
			text := callToolResultText(t, result, err)
			if !strings.Contains(text, "returned nil result") {
				t.Errorf("text = %q, want 'returned nil result'", text)
			}
		})
	}
}
```

solace-aross: Fixed in fd4e6d3 — added TestCallTool_NilResult covering both shapes (nil result, nil StructuredContent).

solace-aross:
@helen-zhu-solace all feedback addressed (nil_result test in fd4e6d3/rebased to 7f96f6e), CI green, rebased onto main. Ready to merge once you get a chance to re-review.

helen-zhu-solace (dismissed):
LGTM!

solace-aross:
Rebased onto main to clear a conflict from #296 landing (CHANGELOG.md context only). No change to this PR's own diff — `internal/tools/` is byte-for-byte identical to the commit you approved. Sorry for costing you a re-click; re-requested review since the push dismissed it.

helen-zhu-solace (approved):

solace-aross:
Rebased onto main again (#301 landed, CHANGELOG.md context only). `internal/tools/` is unchanged from the commit you approved. Sorry, another mechanical re-click — this repo's strict up-to-date + per-push approval rules mean any approved PR that sits for a bit needs this. If it's useful, merging right after approval avoids the churn.

PR #285, opened 2026-08-10
SOL-153003: SIGPIPE + pipefail silently inverts results across the e2e shell suites

helen-zhu-solace:
## Summary

`… | grep -q` pipelines under `pipefail` silently invert their result when grep short-circuits: grep exits on match, the writer takes SIGPIPE and exits 141, and `pipefail` reports 141 as the pipeline's status — so a successful match reads as a failure. This replaces those pipelines with herestrings in the e2e shell suites.

Fixes SOL-153003

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The `e2e-monitoring` job fails intermittently with ~24 assertion failures across Tools 16–19 — `test-kafka-receiver must be present` (expected `true`, actual `false`), `summary.scanned` expected `3` actual `0`, and `jq: parse error: Invalid numeric literal` on empty payloads.

The real cause sits ~200 lines earlier in the log:

```
helpers.sh: line 345: printf: write error: Broken pipe
[WARN] Kafka service not declared in this compose stack — skipping F9/F10 Kafka Receiver/Sender fixtures
```

The `kafka` service *is* declared and `kafka-e2e-mon` starts healthy. `kafka_expected()` tested the compose service list as `printf '%s\n' "$services" | grep -qx kafka`. Bash's `printf` builtin emits **one `write()` per line** (strace-verified), and `kafka` is the *first* of the three services compose prints — so grep matches and exits with two writes still outstanding. Losing that race gives `printf` SIGPIPE, and `pipefail` (helpers.sh:8) turns 141 into the pipeline's status. The predicate reads "Kafka not declared", F9/F10 are skipped, and Tools 16–19 assert against bridges that were never created.

It is scheduler-dependent, which is why it passes on idle machines and surfaces under CI load. **A rerun going green does not indicate an environmental flake.**

The same pattern was present in the shared assertion helpers, so the fix spans both suites.

## Changes Made

- `test/e2e-monitoring/helpers.sh` — `kafka_expected()` matches via herestring instead of `printf … | grep -qx`. This is the confirmed defect from the CI logs.
- `test/e2e-common/lib.sh` — same fix in `assert_contains` and `assert_not_contains`. Reachable but never observed firing; see Testing for the reachability analysis.
- `test/e2e-common/lib.sh` — `echo "$haystack" | head -5` in `assert_contains`'s failure path also becomes a herestring. Cosmetic only: it cannot invert a result there, but it prints broken-pipe noise into a failure report, which is exactly the noise that obscured this bug's diagnosis.
- Updated the comment above `kafka_expected()`. It already argued for keeping compose's exit status out of the grep pipeline; the leftover pipeline is precisely what failed, so it is now documented as a third cause of the same misclassification.

No production code changed — `test/` only, so no CHANGELOG entry is required per `CLAUDE.md`.

## Testing

**Test Environment:**
- Go version: go1.25.12 linux/amd64
- OS: Linux 7.0.10-101.fc43.x86_64 (Fedora)
- Broker version (if applicable): `solace/solace-pubsub-standard:latest`, 2 brokers + `kafka-e2e-mon`

**Tests Performed:**

No Go code changed, so the Go test boxes are not applicable. **No regression test was added** — see Additional Notes for why.

**Test Coverage:**

- Full `e2e-monitoring` suite against real brokers: **90/90 MCP tool tests passed, 20/20 fixture verifications passed**, matched against a clean-tree baseline run that also passed 90/90. No `Broken pipe` in the output, no skip warning, and Tools 16–19 ran **24/24 PASS** against created F9/F10 fixtures on both brokers.
- `kafka_expected()` verified directly on both branches: returns 0 against the suite's compose file (kafka declared), returns 1 against a compose file with no kafka service — so the fix does not simply force the predicate true.
- Mechanism confirmed by strace: `write(1,"kafka\n",6)`, `write(1,"solace-a\n",9)`, `write(1,"solace-b\n",9)` — three writes, match on the first.
- Failure mode reproduced deterministically at payload sizes above the 64KB pipe buffer: old pattern 200/200 inverted, herestring 0/200. The small-payload race did **not** reproduce locally even under full CPU load (0/2000), consistent with the ticket's "passes on an idle one" — the CI logs are the evidence that it fires.
- `lib.sh` reachability, measured rather than assumed: 2/1000 inversions on a realistic 407-line / 15KB `MarshalIndent`-shaped haystack with the needle on an early line; 0/1000 with the herestring.

One intermediate run failed 90/90 at the MCP handshake. That was traced to a stale server left on port 9090 by an earlier aborted run, not to this change — re-running from a clean state reproduced the baseline exactly.

## Breaking Changes

N/A — test scaffolding only, no production surface touched.

**Impact:**

**Migration Guide:**

### Code Quality

### Security

No Go code changed, so the logging and lint items are unaffected by this diff.

### Testing
- [x] Error paths tested

Both predicate branches and both assertion directions were exercised; no test names changed.

### Documentation

None apply — no user-facing, architectural, or exported-API change.

### Git & CI
- [x] No merge conflicts

N/A

## Additional Notes

**No regression test.** The house precedent for this bug class is `.github/scripts/extract-release-notes.test.sh` (Case 9), which works because that script is standalone and CI runs its `.test.sh`. There is no equivalent harness for the e2e helpers and nothing in CI would execute a new one, so a test file added here would rot rather than guard. Building that harness is a larger change than this fix warrants; happy to file it as follow-up if reviewers disagree.

**A green CI run is weak evidence for this fix.** The race is rare per execution — the failure surfaced on 1 of 2 attempts of the same unchanged commit. The meaningful signal is the sustained absence of `printf: write error: Broken pipe` over time, not this PR's build going green.

**Two remaining `| grep -q` pipelines are intentionally untouched:** `licenses-check.sh:187` and `license-header-check.sh:154`. Both feed grep from a *process* rather than a shell builtin, and neither result is load-bearing in the same way; they are out of scope here.

## Related Issues/PRs

- Related to SOL-153003
- Related to #245, #280 — the runs where this first surfaced

---

**For Reviewers:**

**Focus Areas**: The `test/e2e-common/lib.sh` changes are the ones worth scrutiny — they are a *latent* fix, not a confirmed one. I found no evidence those sites ever fired in CI. The case for changing them is that the precondition is guaranteed rather than hypothetical: tool text content is produced by `json.MarshalIndent`, so the 11 call sites passing `$output`/`$msg`/`$content`/`$text` always have multi-line haystacks. The 25 sites passing raw single-line `$response` were never at risk below 64KB. If you would rather keep this PR to the confirmed defect, the lib.sh hunks split out cleanly.

### DIFF
--- a/test/e2e-common/lib.sh
+++ b/test/e2e-common/lib.sh
@@ -671,12 +671,17 @@ assert_contains() {
     local msg="${3:-Response should contain '$needle'}"
     # -F: treat needle as a fixed string, not a regex. Without this, metacharacters
     # (. * [ ] etc.) in tool responses or expected values silently change matching.
-    if echo "$haystack" | grep -qF "$needle"; then
+    # Herestring, not `echo … | grep -qF`: echo emits one write() per line, so on a
+    # multi-line or >64KB haystack grep -q can match and exit with writes still
+    # pending, and the writer's SIGPIPE (141) becomes the pipeline's status under
+    # pipefail — inverting the result. Here that would be a spurious failure; in
+    # assert_not_contains below, a silent pass.
+    if grep -qF "$needle" <<<"$haystack"; then
         return 0
     else
         log_fail "$msg"
         log_fail "  Expected to find: $needle"
-        log_fail "  In: $(echo "$haystack" | head -5)"
+        log_fail "  In: $(head -5 <<<"$haystack")"
         return 1
     fi
 }
@@ -685,7 +690,9 @@ assert_not_contains() {
     local haystack="$1"
     local needle="$2"
     local msg="${3:-Response should NOT contain '$needle'}"
-    if echo "$haystack" | grep -qF "$needle"; then
+    # Herestring for the reason given in assert_contains above — load-bearing here,
+    # since the inverted result is a silent pass on a guard (e.g. credential leak).
+    if grep -qF "$needle" <<<"$haystack"; then
         log_fail "$msg"
         return 1
     fi
--- a/test/e2e-monitoring/helpers.sh
+++ b/test/e2e-monitoring/helpers.sh
@@ -338,11 +338,17 @@ F10_KAFKA_REMOTE_TOPIC="e2e-monitoring-kafka-sender-topic"
 # fix, just from a different cause. So the compose command's own exit status
 # is checked explicitly, with a couple of retries for the transient case,
 # rather than folding it into the same pipeline as the grep.
+#
+# The match is a herestring for the same reason. As `printf … | grep -qx` it
+# hit the misclassification above from a third cause: printf emits one write()
+# per line and "kafka" is compose's first, so grep matched and exited with two
+# writes left, and the resulting SIGPIPE (141) became the pipeline's status
+# under pipefail. A herestring returns grep's own status.
 kafka_expected() {
     local services attempt
     for attempt in 1 2 3; do
         if services=$(docker compose -f "$SUITE_DIR/docker-compose.yml" config --services 2>&1); then
-            printf '%s\n' "$services" | grep -qx kafka
+            grep -qx kafka <<<"$services"
             return $?
         fi
         sleep 1


solace-aross (approved):
Approve. Verified the SIGPIPE/pipefail mechanism independently instead of trusting the write-up: built a >64KB multi-line haystack with the needle on an early line and ran the old `echo | grep -qF` pattern under `pipefail` 200 times. 200/200 false failures. The herestring replacement: 0/200. The mechanism is real, not just plausible.

No production code touched, CI fully green. The `lib.sh` hunks are a latent fix (never observed firing in CI), but the reasoning holds: assertion haystacks come from `json.MarshalIndent`, which is always multi-line.

Copilot's note on the write()-per-line wording is a fair precision point on the comment text, not a correctness issue. The fix doesn't depend on that specific mechanism claim. Not blocking.

PR #287, opened 2026-08-10
docs: name the release jobs RELEASING.md actually has

bczoma:
## Summary

`RELEASING.md` named `fossa_scan` in four places. That job no longer exists — it became `release-readiness` when the shared `SolaceDev` SCA workflow was replaced by the local Guardian gate. Docs only; no behaviour change.

Doing this ahead of cutting `v0.7.0`, because the block being corrected is the one a releaser reads to know what waits on what.

## Changes Made

- The pipeline dependency block now names `release-readiness` and describes it as the Guardian gate, cross-checked against `release.yml`: `build-binaries` waits on `test`, `release-notes`, `licenses`; `build-docker` and `release` additionally wait on `release-readiness`.
- "Runs the FOSSA scan against the tag" → "Runs the release readiness check (the Guardian gate) against the tag".
- Realigned the monospace block, since `release-readiness` is longer than the name it replaced.

## Deliberately not changed

The Publish gate bullet still says a fork pull request gets no scan "because GitHub withholds the credential". The Guardian workflow reads its secrets from an Environment rather than from the caller, so that may no longer be the mechanism — but I have not verified what fork behaviour is now, and guessing in a runbook is worse than leaving a claim that was true when written. Worth a separate look.

## Testing

No code. Verified by parsing `release.yml` and diffing the `needs` graph against every job named in the doc; `grep fossa_scan RELEASING.md` is now empty.

## Checklist

- [x] Self-review completed
- [x] No production surface touched, so no CHANGELOG entry (`production-surface.sh` scopes that to `internal/config/`, `internal/tools/`, and `tools.yaml`)

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -70,7 +70,7 @@ git push origin v0.4.0
 Pushing the tag runs `.github/workflows/release.yml`, which:
 
 1. Re-runs the full build-and-test suite.
-2. Runs the FOSSA scan against the tag.
+2. Runs the release readiness check (the Guardian gate) against the tag.
 3. Builds binaries for `linux` and `darwin` × `amd64` and `arm64`, attesting each archive's build provenance in the job that built it.
 4. Builds and pushes a multi-arch image to `ghcr.io/solaceproducts/solace-broker-mcp` (`{version}`, `{major}.{minor}`, `latest`, `sha-<short-sha>` tags), and attests the image digest, pushing the attestation to the registry alongside it.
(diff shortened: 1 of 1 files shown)


solace-aross (approved):
Approve. Verified against the actual workflow rather than taking the doc's word for it: `.github/workflows/release.yml` on main confirms the job is `release-readiness` (not `fossa_scan`) and the `needs` graph matches exactly what's now documented — `build-binaries: [test, release-notes, licenses]`, `build-docker` and `release` both add `release-readiness`.

Checked for other leftover `fossa_scan` references: none. The one remaining "FOSSA scan" mention (line 114, license-inventory reconciliation) is a different, still-accurate use, not a missed rename.

Docs-only, no functional risk.

PR #286, opened 2026-08-10
SOL-152412: describe the repository as it is configured, not as a transition

bczoma:
## Summary

Brings `.github/ADMIN_SETUP.md` in line with the repository settings applied on 2026-08-07, and rewrites it — plus two sibling documents — to describe a public repository in the present tense instead of narrating a transition that has happened.

Closes acceptance criterion 5 of [SOL-152412](https://sol-jira.atlassian.net/browse/SOL-152412): "the live settings match `.github/ADMIN_SETUP.md`, and the checklist items in that file are ticked off truthfully."

## Type of Change

- [x] Documentation update

## Motivation and Context

The document actively misdescribed the live state. It told a reader that branch protection held zero required contexts and that the corrected list was still to be registered — the opposite of the truth, and the sort of claim that ends up in a go/no-go answer. Now that the repository is public, these files are world-readable, so a guide that describes the repo as internal and understates its own controls is a factual misstatement about security posture.

### The correction that matters most

SOL-152412's acceptance criterion 2 says a pull request with a licensing conflict or a critical/high vulnerability cannot merge to `main`. **It can.** Read from `guardian-scan.yaml`:

- `setup` sets `enforce=true` only when `is_trunk` is true, so the FOSSA licensing guard runs `BLOCK` on push to `main` and `REPORT` on a pull request.
- `guardian-gate`, which carries vulnerability enforcement, requires `is_trunk` and is therefore always skipped on a pull request.
- On a fork pull request the scan jobs are skipped for want of Environment secrets and `gate` exits 0 with a notice.

The real posture is **detection on `main`, not prevention at the pull request**. This PR states that plainly in one place rather than leaving it to be assembled from three sections, and names the compensating human controls. The AC itself needs a matching correction; that is a separate Jira edit.

## Changes Made

- **`ADMIN_SETUP.md`** retitled **Repository Admin Reference**, reorganised around an **Outstanding actions** table at the top — seven open items, each with its ticket and a link to the section explaining it. Everything else describes configuration already in place.
- **Required status checks** section records the applied state, with per-row `integration_id` pinning. It is the single authoritative list; the Rulesets section deliberately does not repeat it.
[...]

PR #294, opened 2026-08-11
SOL-153050: retire the DCO workflow, keep the identity check it carried

bczoma:
## Summary

DCO enforcement has moved to the CNCF `dco2` GitHub App. This deletes the workflow and scripts it replaced, and updates the ten files that still described them as the control.

It also keeps a control that would otherwise have been deleted by accident — see below.

Implements [SOL-153050](https://sol-jira.atlassian.net/browse/SOL-153050).

## Type of Change

- [x] Documentation update

## Motivation and Context

`DCO` is the required status check, pinned to the `dco2` App. The workflow-based `DCO sign-off` and `DCO check self-test` were removed from the `main-protection` ruleset on 10 August, so the files that produced them were doing nothing — while `CONTRIBUTING.md`, `CLAUDE.md`, and the threat model still told readers those checks were what blocked their pull request.

**The part worth reviewing carefully:** `dco-check.sh` did *two* jobs.

1. DCO sign-off — replaced by the App.
2. Rejecting commits whose author or committer email uses a non-routable domain: `.local`, `.sol-local`, `.internal`, `.lan`, or a bare hostname.

The App does the first and nothing else. Deleting the script outright would have silently removed the second — the control SOL-152902 added to stop machine hostnames publishing permanently in a now-public git history. That is not recoverable after the fact without rewriting history everyone has fetched.

So the identity half moves to its own script and job rather than disappearing.

## Changes Made

**Deleted**
- `.github/workflows/dco.yaml`, `.github/scripts/dco-check.sh`, `.github/scripts/dco-check.test.sh`
- `ci-pr.yaml` job `dco_selftest`

**Added**
- `.github/scripts/commit-identity-check.sh` — the identity half, behaviour unchanged: same denylist, same forward-only range, still checks merge commits
- `.github/scripts/commit-identity-check.test.sh` — 14 cases covering both paths, plus a mutation probe that guts the denylist and asserts the suite notices
- `ci-pr.yaml` job `identity`, reporting as **`Commit identity routable`**

**Corrected while reconciling**
- `OPEN_SOURCE_STATUS.md` explained the `Guardian scan` trap **backwards**, describing it as a check "skipped on fork pull requests". It is the *workflow's* name and has never been a check at all, so requiring it blocks every merge — which is what briefly happened on 7 August. `Guardian gate` is now named alongside it.
- `ADMIN_SETUP.md` described the required-checks list as pending. It is applied — twelve contexts, with the live `gh api` command so the table is checkable rather than trusted.
- `docs/internal/threat-model.md`: the commit-authorship row cited two deleted checks; a separate accepted-risk row recorded `Third-party licenses current` as documented-but-not-required, which SOL-152412 resolved.
- `CONTRIBUTING.md`, `CLAUDE.md`, `README.md`, `dependabot.yml`, `build-and-test.yml` — check names and cross-references.

## Testing

Documentation and CI only; no Go code paths change.

**Every gate self-test run locally:**

| Suite | Result |
|---|---|
| `commit-identity-check.test.sh` | **14 passed, 0 failed** (new) |
| `build-test-licenses-check.test.sh` | 26 passed, 0 failed |
| `license-header-check.test.sh` | 24 passed, 0 failed |
| `licenses-check.test.sh` | 10 passed, 0 failed |
| `extract-release-notes.test.sh` | 9 passed, 0 failed |

**A live hazard found and fixed:** `build-test-licenses-check.test.sh` used `.github/workflows/dco.yaml` as a *test fixture*, so deleting that file would have broken an unrelated gate. Repointed at `ci-pr.yaml`; still 26/26.

All three changed workflow files parse as YAML.

## Breaking Changes

None for users. For contributors, the check name in CI changes from `DCO sign-off` to `DCO`. Sign-off itself is unchanged — `make hooks` and `git commit -s` work exactly as before, and the docs were updated to name the check that now exists.

## Additional Notes

**`Commit identity routable` is deliberately not yet in the ruleset.** It has to report green on a real pull request before being required — the same rule that would have caught the `Guardian scan` mistake. Registering it is a repo-admin action; `ADMIN_SETUP.md` carries the row marked *Not yet registered* with the reason.

**Until it is registered, it enforces nothing.** The identity control is therefore weaker between this merging and that click than it was before. Worth registering promptly.

## Related Issues/PRs

- Implements SOL-153050
- Preserves the control added by SOL-152902
- Follows the settings work in SOL-152412
- Supersedes #286, closed unmerged after its base drifted

---

**For Reviewers:**

**Focus Areas**: `commit-identity-check.sh` versus the deleted `dco-check.sh` — the identity logic should be byte-for-byte equivalent in behaviour. Worth confirming I did not narrow the denylist or the commit range while lifting it.

**Questions**: is `Commit identity routable` the right check name? It becomes a required context and is awkward to rename afterwards.


.github/scripts/commit-identity-check.sh
```diff
@@ -0,0 +1,127 @@
...
+    "$(git show -s --format='%ae' "$sha")" \
+    "$(git show -s --format='%ce' "$sha")" >&2
+done <<<"$failed"
+
+cat >&2 <<'EOF'
+
+To fix, set a routable address and rewrite the commits this pull request adds:
+
+  git config user.email "you@example.com"
+  git rebase --exec 'git commit --amend --no-edit --reset-author' origin/main
```
solace-aross: This block is unconditional: no merge-safety branch, no per-commit guard on `--exec`. Confirmed by reproduction (see review body) that following it on a branch with a merge or another author's commits destroys the conflict resolution, misattributes the colleague's commit, and leaves the non-routable address published while the check goes green on the rewritten history. The old `dco-check.sh`'s `report_identity_failures()` had both guards plus a trailer-strip step and a pointer to `.github/CONTRIBUTING.md#author-identity` — none of which made it into this port.

solace-aross (changes requested):
Ran the actual mechanics before deciding, not just read the diff. The identity-check detection logic is exactly what the description claims — reused `identity_domain_ok()` verbatim, same denylist, merges still walked, confirmed by my own sabotage (injected a `.local` commit, script caught it) and by an independent Opus review that reproduced the fix guidance's damage in a throwaway repo. One thing doesn't survive that: the remediation text the script prints on failure.

**Blocking: the printed fix instructions actively destroy what they're supposed to fix.**

`commit-identity-check.sh`'s failure block collapsed the old script's two carefully-guarded remediation paths into one unconditional command. The old `dco-check.sh` branched on whether a failing commit was a merge (never rebase a branch containing one — it flattens the merge and throws away the conflict resolution) and, for the non-merge bulk fix, guarded `--exec` per-commit (`grep -qiF "$BAD" || exit 0`) specifically so it wouldn't rewrite commits that don't have the bad address. Both guards are gone.

Reproduced end to end on a throwaway branch: bad-email commit, a colleague's clean commit, then a merge with a hand-resolved conflict (also bad email, since the check walks merges — confirmed by this PR's own test at `commit-identity-check.test.sh:162`). Following the printed advice verbatim:

1. The merge is replayed as an ordinary commit. The conflict resolution is gone — had to redo it by hand.
2. `--exec` fires on every replayed commit. The colleague's commit is now authored by whoever ran the fix.
3. `--no-edit` keeps the old trailer. Re-running the check afterward prints a clean pass — **the control goes green while the non-routable hostname is still permanently in the history**, which is the one thing this script exists to prevent.
4. DCO now fails on both rewritten commits (trailer email no longer matches author).

Everything needed to fix this already exists on `main` today: the merge-safety branch and the per-commit `--exec` guard in the deleted `dco-check.sh` (`report_identity_failures()`), and the trailer-strip + safe rebase sequence in `.github/CONTRIBUTING.md#author-identity`, which the old script pointed readers to and the new one doesn't reference at all. Porting the guard and restoring the pointer is a small, mechanical fix — the detection logic doesn't need to change, only what gets printed on failure.

**Not blocking, but worth folding into the same round:** two more files still describe the retired check as live, the same claim you already fixed in `CONTRIBUTING.md:308` per Copilot's comment:
- `.githooks/prepare-commit-msg:8-9` and `:105` — "the `DCO sign-off` check in CI is the enforcement." It isn't anymore; `DCO` (the App) is.
- `.github/renovate.json`'s top-level `description` — still points at `.github/workflows/dco.yaml` for the Dependabot exemption mechanism, a file this PR deletes. `.github/dependabot.yml`'s own comment, updated in this very PR, already has the correct, verified replacement text ("the CNCF dco2 App... passes them anyway, confirmed on PR #289") — worth copying that phrasing over rather than re-deriving it.

`Makefile:51`'s `(DCO sign-off)` is fine as-is; it names what the hook adds, not a check name, and reads the same as the (correct, unedited) line in `CONTRIBUTING.md:89`.

Everything else here is clean: self-tests all green in a fresh worktree (14/14 identity, 26/26, 24/24, 10/10, 9/9 for the sibling suites this PR touches), the Dependabot-exemption story is correctly re-verified against live evidence rather than assumed, and the ruleset/required-checks claims all check out against `gh api .../rulesets/13942241`.

.github/scripts/commit-identity-check.sh
bczoma: Fixed in 3476272. Detection is untouched; only the printed advice changed.

Both guards are back, lifted from `report_identity_failures()`:

- a merge among the offenders withholds the branch-wide rebase entirely and prints the re-create-the-merge path instead
- otherwise the bulk fix carries the per-commit `--exec` guard, with the warning about dropping it
- the trailer is stripped as well as the author reset, so `DCO` does not start failing on the rewritten commits
- the pointer to `.github/CONTRIBUTING.md#author-identity` is restored
- the rebase base is now the pull request's own base rather than a hard-coded `origin/main`, per Copilot's note on the same block

Ran your reproduction against the new text on a throwaway branch — bad-email commit, a colleague's clean commit, guarded rebase as printed. The offending commit is corrected in both the author field and the trailer, the colleague's commit keeps its author, the check passes, and `grep -c` over authors, committers, and messages finds no trace of the address. The merge path is asserted rather than reproduced: no `git rebase` command is printed at all when a merge is among the offenders.

Eight new self-test cases pin this, taking the suite to 22. Worth noting one of them passed vacuously on the first run — a needle beginning with `--` was consumed as a grep option — so `grep -qF --` is now used throughout, including in the pre-existing `expect` helper.

One thing beyond what you flagged, same class: `sanitize` did not make it into the port either. Commit metadata is attacker-controlled text going to a log that reads `::` directives from stdout, so it is restored on the printed author and committer addresses.

The two doc items are done as well:

- `.githooks/prepare-commit-msg:8` and `:105` now name the required `DCO` check
- `.github/renovate.json`'s `description` reuses `.github/dependabot.yml`'s verified phrasing rather than re-deriving it, and marks the retired workflow exemption as retired instead of pointing at the deleted file

`Makefile:51` and `CONTRIBUTING.md:89` left as they are, per your read.

Sibling suites still green: 26/26, 24/24, 10/10, 9/9, plus 40/40 on the hook self-test after the comment edits.

.github/scripts/commit-identity-check.sh
```diff
@@ -0,0 +1,127 @@
...
+  git push --force-with-lease
+
+Check it took before pushing:
+
+  git log origin/main..HEAD --format='%h %ae %ce'
+EOF
```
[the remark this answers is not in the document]
bczoma: All three points fixed in 3476272 — they were correct, and a human review reproduced the damage the block would have caused.

- the hard-coded `origin/main` is gone; the pull request's own base is printed
- the branch-wide rebase is now withheld entirely when a merge is among the offenders, rather than made merge-aware — replaying a merge is not what a contributor wants here, so the advice re-creates it instead
- the stale trailer is stripped in the same amend, so `DCO` does not start failing once the author email changes

As suggested, the block is aligned with `.github/CONTRIBUTING.md#author-identity`, including the guarded `--exec` form, and now links there.

The stale references from your other comment are done too: `.githooks/prepare-commit-msg:8` and `:105`, and `.github/renovate.json`'s `description`. `Makefile:51` is left as-is — it names what the hook adds rather than a check name, matching the unchanged line at `CONTRIBUTING.md:89`.

Eight new self-test cases pin the guards; the suite is at 22 passed, 0 failed.

.github/dependabot.yml
```diff
@@ -8,8 +8,8 @@ version: 2
 # org's current system, so it stays scoped to the one thing it already covers,
 # the pinned Claude Code CLI version (.github/renovate.json). Dependabot's own
 # commit-message config has no field for a DCO Signed-off-by trailer, so its
-# PRs are exempted from that check instead — see .github/workflows/dco.yaml
-# and .github/scripts/dco-check.sh. Tracked under SOL-152808.
+# pull requests cannot carry one; the CNCF dco2 App that enforces the required
+# `DCO` check passes them anyway, confirmed on PR #289. Tracked under SOL-152808.
```
solace-aross: This is backwards, and the PR it cites as proof contradicts it. Raw commit object for `20ddc122` on #289:

```
author dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
committer GitHub <noreply@github.com>
...
Signed-off-by: dependabot[bot] <support@github.com>
```

It carries a trailer. `DCO` doesn't pass because the config has no field for one — that clause is narrowly true but doesn't support "cannot carry one." It passes because `dco2` short-circuits on `author.is_bot` before it reads the message (`should_skip_commit`, `CommitSuccessReason::FromBot`). And if it did read the trailer, this one would fail anyway: `support@github.com` matches neither the author (`...@users.noreply.github.com`) nor the committer (`noreply@github.com`), and dco2 requires a match (dependabot-core#3480, closed not-planned).

That's worth stating, not just smoothing the wording. The retired `dco.yaml` exempted one bot on an unforgeable login check. `dco2`'s exemption is unconditional for every bot-authored commit, with no config key to narrow it — broader than what it replaced. This PR's framing ("a pull request cannot edit the thing that judges it") doesn't mention that the swap also widened the exemption. Worth saying plainly here, especially with this repo going public soon.

.github/renovate.json
```diff
@@ -1,7 +1,7 @@
...
...e revert, not less: config:base resolves to config:recommended, which folds in :ignoreModulesAndTests, whose own ignorePaths covers `**/test/**` among others — a pattern that matches test/e2e-llm/package.json. The single top-level ignorePaths here replaces that list rather than extending it, and that replacement is the only reason Renovate sees the CLI pin at all. Before the revert, dropping it as apparent redundancy would have silently lost two of three jobs; now it would silently switch off the only job this config exists for, with no signal that anything changed. Bumps follow npm's `stable` dist-tag (deliberately lagged from `latest`); see test/e2e-llm/README.md for the bump procedure. Top-level `commitBody` carries the DCO sign-off that the required `DCO` check requires, and is top-level rather than per-rule so it also covers lock-file maintenance and config-migration branches. The address must match the identity Renovate commits under (the hosted GitHub App uses this one), otherwise its PRs go red on that check — the fix is to correct the address here, never to look for a way to exempt the bot. Enforcement is the App's; there is no repo-side exemption knob left to reach for.",
```
solace-aross: Same claim, reused from `dependabot.yml`'s comment per the commit message — same correction applies (see that file). "Confirmed on PR #289" is the citation that actually disproves it.

.github/ADMIN_SETUP.md
```diff
@@ -140,12 +140,9 @@ of `dependabot.yml` and are already on.
...
   system — reverted under SOL-152808 ahead of this repo going public.
-- Dependabot's `commit-message` config has no field for a commit trailer, so
-  its PRs can never carry the `Signed-off-by` line the `DCO sign-off` required
-  check normally needs. Rather than leave every one of its PRs permanently red,
-  `.github/workflows/dco.yaml` skips that check specifically for PRs GitHub
-  itself records as opened by `dependabot[bot]` — see that file and
-  `.github/scripts/dco-check.sh` for why this is safe from PR-side forgery.
+- Dependabot's `commit-message` config has no field for a commit trailer, so its
+  pull requests cannot carry the `Signed-off-by` line the required `DCO` check
+  looks for. The CNCF `dco2` App passes them anyway — confirmed on PR #289.
```
solace-aross: Same claim as `dependabot.yml` and `renovate.json` (see that comment for the fix). Repeated again at line 262-264 below — worth fixing both in the same pass.

.github/OPEN_SOURCE_STATUS.md
```diff
@@ -177,9 +179,9 @@ These settings must be configured by repository admin:
   Go module and GitHub Actions updates run on Dependabot too
   (`.github/dependabot.yml`). Renovate can't be enrolled for a public repo
   under the org's current system, so it stays scoped to just the pinned Claude
-  Code CLI version (`.github/renovate.json`). Dependabot's PRs are exempted
-  from the `DCO sign-off` check rather than expected to satisfy it; see
-  `.github/workflows/dco.yaml`
+  Code CLI version (`.github/renovate.json`). Dependabot's PRs cannot carry a
+  `Signed-off-by` trailer; the `dco2` App passes them anyway, confirmed on
```
solace-aross: Same claim, fourth file. Five places total (dependabot.yml, renovate.json, here, and ADMIN_SETUP.md ×2) — one wording fix, applied everywhere.

solace-aross (changes requested):
The remediation-guard fix holds. Sabotaged the exact defect from the last review — reverted `commit-identity-check.sh`'s merge-safety branch and per-commit `--exec` guard back to the unconditional block `392f3444` shipped — self-test dropped from 22/22 to 16/22, confirming the eight new cases actually catch the regression. Restored, back to 22/22. The three Copilot-flagged fixes (PR base printed instead of hardcoded `origin/main`, trailer stripped alongside author reset, `sanitize` restored) all check out too.

Found a new issue while checking the doc reconciliation: a claim repeated in five places across four files that Dependabot's commits "cannot carry" a `Signed-off-by` trailer, cited as "confirmed on PR #289." Pulled the raw commit object for the commit named — it carries one. See inline comments.

.github/dependabot.yml
bczoma: Correct on every point, and I verified each before changing anything rather than taking the correction on trust:

- `20ddc122` carries `Signed-off-by: dependabot[bot] <support@github.com>`. The claim was wrong.
- That address matches neither the author (`49699333+dependabot[bot]@users.noreply.github.com`) nor the committer (`noreply@github.com`), and `signoffs_match` (`dco2/src/dco/check/mod.rs:263-266`) requires a match against one of them — so it would fail if read.
- It is not read: `should_skip_commit` (`mod.rs:164-175`) returns `FromBot` on `author.is_bot` first.
- No key narrows it. Pulled the whole config surface — `Config` in `dco2/src/github/client.rs:437` and `docs/config/dco.yml`: `allowOverrideAction`, `allowRemediationCommits.{individual,thirdParty}`, `require.members`. Nothing bot-related.

Fixed in all five places in 7a76d35, with the full reasoning here and in `ADMIN_SETUP.md`, the other three pointing at it rather than restating it. The widening is stated plainly as you asked, not smoothed: one bot matched on an unforgeable login, versus every bot-authored commit with no way to narrow it.

Reading `should_skip_commit` for this turned up two more instances of the same one-sided framing, both fixed in the same commit:

- `IsMerge` is checked before `FromBot` and is equally unconditional. `CONTRIBUTING.md` was still describing the retired workflow's narrow rule — a merge exempt only while re-merging its parents reproduced it, a conflict-resolving merge needing its own sign-off. Under `dco2` that merge passes unsigned. Restated, with the sign-off kept as convention rather than enforcement.
- `threat-model.md` recorded commit authorship as `Mitigated` on server-side evaluation alone. Split into three rows: human non-merge commits mitigated; bot and merge commits accepted risk; and the third below.

`dco2` defaults `allowOverrideAction: true`, and this repo had no `.github/dco.yml`, so a failed `DCO` was rendering a `Set DCO to pass` button for anyone with write access — a one-click path the retired workflow never offered. That one is closable by a committed file, so a82f04e adds `.github/dco.yml` with `allowOverrideAction: false`. Two caveats recorded with it: `dco2` reads that file from the default branch only, so it had no effect on this pull request and cannot be re-enabled by one; and it is not a security boundary, since write access implies access to the file. It stops an accident, not an intent.

Net effect on the accounting: the swap bought server-side evaluation, widened the bot and merge exemptions (accepted, no switch exists), and opened an override button (closed here).

.github/renovate.json
bczoma: Fixed in 7a76d35. This file no longer restates the claim at all — it points at `.github/dependabot.yml`'s comment, which now carries the verified mechanism, rather than paraphrasing it a second time. Reusing the wording is what propagated the error; one authoritative copy is the fix.

Two adjacent clauses in the same `description` were wrong in the same way and are corrected too:

- "Dependabot cannot attach a sign-off of its own" — it attaches one, it just cannot attach one whose address matches the identity it commits under. That is what the retired gate required, so the sentence now says that.
- "there is no repo-side exemption knob left to reach for" — inaccurate once `allowOverrideAction` is in view. Replaced, and it no longer asserts anything about how `dco2` treats Renovate's own commits: Renovate is not enrolled here, so nothing in this repo demonstrates that either way. `commitBody` stays for the reason it can be defended — it is the one part a repo-owned file controls.

.github/ADMIN_SETUP.md
bczoma: Both fixed in 7a76d35 — this line and the repeat at 262-264, in the same pass.

Line 143 now states the mechanism in brief (exemption, not sign-off; trailer present but non-matching; never read) and points down to the required-status-checks section. That section carries the full derivation: the raw commit fields, `signoffs_match`, the `FromBot` short-circuit, dependabot-core#3480, and the exemption-widening comparison against the retired `dco.yaml`.

The cross-reference I first wrote pointed at `"Required Status Checks"`, which does not exist in this file — the real heading is `### Required status checks`. Repointed in a82f04e, along with three others of mine that had the same invented title-casing.

That commit also adds `.github/dco.yml` with `allowOverrideAction: false`, and this section now describes a closed default rather than an open decision, with the two caveats attached: default-branch-only, and not a security boundary.

.github/OPEN_SOURCE_STATUS.md
bczoma: Fixed in 7a76d35, fourth of the five. Now states the exemption plainly — `dco2` skips every bot-authored commit, wider than the retired workflow's single login-matched one, with no config key to narrow it — and points at `ADMIN_SETUP.md` → "Required status checks" for the derivation instead of carrying its own copy of the claim.

Five sites confirmed done: `dependabot.yml`, `renovate.json`, here, and `ADMIN_SETUP.md` twice. `git grep -i 'cannot carry\|no field for a'` returns nothing outside `CHANGELOG.md`.

bczoma: Correction to the last line: that grep returns nothing anywhere, CHANGELOG.md included. I wrote it as `| grep -v CHANGELOG` out of habit from the earlier sweeps and then described the filtered form, which implies a surviving hit there. There is none.

solace-aross (approved):
Approving at a82f04e. This is the third pass; both prior CHANGES_REQUESTED rounds are closed out, and I re-verified the delta since my last review rather than trusting the diff.

What I checked at this head, not the description:

- **Remediation guard (round 1 finding).** Sabotaged it again independently: reverted the merge-safety branch in `commit-identity-check.sh` so a merge among the offenders no longer withholds the branch-wide rebase advice. Self-test suite dropped from 22/22 to 20/22. Restored, back to 22/22. The guard holds and the tests actually catch its absence.
- **"Dependabot cannot carry a sign-off" claim (round 2 finding).** Fetched the real `dco2` source (`dco2/src/dco/check/mod.rs`). `should_skip_commit` checks `commit.is_merge` first, then `author.is_bot`, and returns before any sign-off is read; `signoffs_match` (email vs. author/committer) only runs on commits that reach that far. That confirms the corrected text in all five files (`dependabot.yml`, `renovate.json`, `ADMIN_SETUP.md` twice, `OPEN_SOURCE_STATUS.md`): the pass is a bot exemption, not a sign-off, and it's unconditional where the retired workflow's was login-scoped. `grep -rn 'cannot carry\|no field for a'` across the tree returns nothing.
- **New `.github/dco.yml`.** Fetched `dco2`'s own sample config to check the defaults being called out. They match exactly: `allowOverrideAction` defaults `true`, `allowRemediationCommits.{individual,thirdParty}` default `false`, `require.members` defaults `true`. Setting `allowOverrideAction: false` here is real, and the comment's claim that it only takes effect post-merge (the App reads config from the default branch) matches how `dco2` is documented to work.
- **Required-checks safety.** Confirmed the branch ruleset (`rulesets/13942241`) requires `DCO`, not `DCO sign-off`, so deleting `dco.yaml` doesn't strand a required context once this merges. The `DCO sign-off` job still running on this PR is `pull_request_target` reading the workflow file from `main`, where `dco.yaml` still exists. It drops out on its own once this merges, nothing further to do here.
- **Cross-references.** The `a82f04e` heading-casing fixes point at real anchors (`### Required status checks`) from both `ADMIN_SETUP.md` and `OPEN_SOURCE_STATUS.md`.
- `make check` at this head: green, 88.5% coverage. All CI green, no red context unrelated to this diff.

Nothing left to raise. Good catch on your own second pass, re-deriving the Dependabot claim from the raw commit instead of taking the first fix on faith. That's exactly the kind of thing that should hold up before this repo goes public.

PR #296, opened 2026-08-12
SOL-152286: Track consecutive exchange failures outside gobreaker so slow outages trip the breaker

amitmorade:
## Summary

Makes the circuit breaker's `consecutive_failure_threshold` rule work in the regime it exists for: a slow, low-traffic IdP outage now trips the breaker. The rule previously read gobreaker's window-decayed `ConsecutiveFailures`, which mathematically cannot reach the threshold when failures are spaced wider than the rolling window's bucket period; it now reads an undecayed counter owned by this package.

Tracked under SOL-152286 (follow-up to the gap documented in SOL-151600 / PR #205).

## Type of Change

- [x] Documentation update (threat model, CHANGELOG)

## Motivation and Context

gobreaker splits the failure-rate window into 10 rolling buckets (`failure_rate_window/10` — 3s at the 30s default) and its `ConsecutiveFailures` count decays as buckets age out. Simulating gobreaker v2.4.0's actual bucket arithmetic shows that at the ~19s spacing a slow outage produces (each exchange burning its full retry-chain deadline), the count **plateaus at 2 and can never reach the threshold of 5**. At traffic too low for the rate rule's `minimum_requests` sample, neither rule can trip, so the breaker stays closed through a genuine sustained outage — and every request eats the full ~19s retry budget, which is exactly the failure storm the breaker (SOL-151600) exists to prevent. PR #205 shipped a docs-only fix acknowledging this gap; this PR closes it.

## Design rationale

**Why two trip rules exist (and both stay).** Each rule owns a failure mode the other is structurally blind to:

| | Total outage (everything fails) | Partial degradation (e.g. 60% fails) |
|---|---|---|
| **High traffic** | either rule trips (consecutive usually first) | **rate rule only** — interleaved successes reset the streak |
| **Low traffic** | **consecutive rule only** (this fix) | neither — irreducible: too few samples to distinguish from noise, which is what `minimum_requests` is for |

The consecutive rule is a *total-outage* detector: one success resets it, so it never fires on a flaky-but-alive IdP. The rate rule is a *partial-degradation* detector: it needs volume but tolerates interleaved successes. Removing either would open a real hole.

**Why the counter lives outside gobreaker.** gobreaker's own consecutive count is fused to its bucket window — there is no setting to detach it. The fix wires a package-owned `atomic.Uint32` through gobreaker's designed extension points (`IsSuccessful` / `ReadyToTrip` callbacks): no fork, no vendored patch, no version bump. Updating it inside the callbacks (rather than around `Execute`) is deliberate — gobreaker's `afterRequest` performs its generation check *before* invoking the callbacks, so a stale outcome (an exchange that started before a trip and finished after recovery) is discarded before it can touch the counter, identical to how gobreaker treats its own counts. Callbacks run under gobreaker's mutex, so updates are serialized; the atomic keeps them side-effect-light per the file's documented callback constraint and satisfies `-race`.

**Why reset-on-success and never reset-on-time.** `Exchange` sits behind a token cache, so the breaker only sees cache-miss exchanges. A quiet gap between outcomes usually means "no caller needed a token," not "the IdP healed" — an idle overnight deployment must not forgive a real ongoing outage just because nobody asked. An observed success is the only genuine evidence of IdP health, so it is the only reset. This also makes recovery correct for free: closing from half-open requires consecutive probe *successes*, so the counter is always 0 when the breaker re-enters closed — one ordinary failure after recovery cannot immediately re-trip (test-pinned).

**What was "removed" — and what wasn't.** Nothing was removed from gobreaker: its bucket machinery still runs and still maintains its own decayed `ConsecutiveFailures`; we simply stopped *reading* it (one line in `newReadyToTrip`). The windowed `TotalSuccesses`/`TotalFailures` are still read — time-decay is *correct* for a rate ("% of the last 30s"), it was only ever wrong for a consecutive count. A unit test pins that the decayed source is no longer consulted, so a refactor cannot silently reconnect it. The doc comments that were removed were the four "regime gap" notes describing the old limitation — they were replaced with the now-true behavior, not deleted.

**Alternatives considered and rejected.**
- *Widen `failure_rate_window` (e.g. 30s → 120s):* mechanically works for 19s spacing, but the required window scales with failure spacing (~5×), and spacing is arrival-driven — it depends on customer traffic we can't know at default-time. Any fixed window just moves the cliff. It also couples the two rules through one knob: widening it to rescue the consecutive rule makes the rate rule average over stale history at high traffic. Operators who used this as a workaround can revert after upgrading.
- *Time-bounded counter (reset if the gap since the last failure exceeds T):* re-introduces "silence = forgiveness" through the back door — wrong for a cache-fronted, demand-driven traffic shape, and it would add a knob with no principled default.
- *New config surface:* none needed. The counter has no time dimension, so there is nothing to configure; the existing `consecutive_failure_threshold` keeps its name, type, default (5), and `0`-disables semantics — it just finally behaves as documented. All 7 operator knobs are unchanged.

## Changes Made

- `internal/tokenexchange/circuitbreaker.go` — package-owned `atomic.Uint32` (a constructor local closed over by the callbacks, deliberately not an `Exchanger` field, preserving the struct's write-once invariant); new `newCountingIsBreakerSuccess` wrapper (increment on counted failure, reset on success; excluded outcomes never reach it — named for the side effect, since it is the one callback slot that isn't a pure function of its error argument); `newReadyToTrip` reads the new counter instead of `counts.ConsecutiveFailures`.
- Doc corrections in `circuitbreaker_config.go`, `defaults.go`, and `internal/config/idp_circuit_breaker.go` — the regime-gap notes replaced with the now-true behavior.
- `circuitbreaker_test.go` — `TestNewReadyToTrip` reworked for the new signature; new case pinning that gobreaker's decayed count is ignored.
- `circuitbreaker_exchange_test.go` — seven new end-to-end tests (see Testing).
- `CHANGELOG.md` — entry under `[Unreleased]` → Fixed.
- `docs/internal/threat-model.md` — the §3 STRIDE row and consolidated-accepted-risk #8 both cited the deleted code comment as their "accepted risk" evidence; §3 flips to Mitigated (citing the new counter, naming one residual: partial degradation below `minimum_requests` still can't trip either rule — inherent to the sample floor, not fixed here), #8 struck through in place rather than deleted (rows are referenced by number elsewhere in the doc).
- `.github/dependabot.yml` — gobreaker excluded from the `go-minor-patch` auto-merge group. This package's correctness leans on three gobreaker v2.4.0 callback behaviors that are not documented library contract; two are covered end-to-end by tests here, the third (stale outcomes discarded before either callback runs) is not, so a gobreaker bump now needs a human reading its changelog rather than an auto-merge on green CI.
- Observability: the breaker's state-change WARN now carries a `consecutive_failures` attribute, so an operator can tell which rule opened the breaker without a metrics pipeline — on `closed→open` it equals `consecutive_failure_threshold` when the consecutive rule fired and sits below it when the rate rule did (never 0 on a trip, since the tripping failure itself increments the counter first); recovery (`half-open→closed`) logs 0 because probe successes reset the streak. Gauges/counters remain the metrics sibling story; this is the minimal diagnostic that doesn't pre-commit that story's interface.

## Testing

**Test Environment:**
- Go version: 1.26.0
- OS: macOS (darwin)
- Broker version: n/a — the change is entirely in the IdP-facing token-exchange layer; no broker interaction is affected.

**Tests Performed:**
- [x] Integration tests added/updated (exchange-level tests against an `httptest` fake IdP)
- [ ] E2E tests added/updated (not applicable — no config or wire surface changed)
- [x] Tested locally with `go test -race ./...` (`make check`: build, vet, lint, race suite, 85% coverage gate; package coverage 94.4%)
- [ ] Tested with real Solace broker (not applicable)

**Test Coverage:**
- **Sparse/low-traffic trip (the bug):** 5 failures spaced ≥3 bucket periods apart open the breaker at exactly the threshold, then fail fast without an IdP round-trip. The geometry was verified by simulating gobreaker v2.4.0's rolling-bucket arithmetic directly: at this spacing the old counter peaks at 4 (below the threshold of 5), and sleep jitter only widens spacing — which decays the old counter *harder* — so timing drift cannot invalidate the test.
- **Dense/high-traffic exactness:** back-to-back failures trip at exactly the threshold, not before; a success mid-streak resets it (4 + success + 4 stays closed, a fresh 5th trips). Also pins that dense-outage behavior is unchanged from before the fix.
- **Rate rule end-to-end (new coverage):** interleaved success/failure at the 50% threshold — the shape the consecutive rule is structurally blind to — held closed by the `minimum_requests` floor at 9 evaluated outcomes, tripped once the floor is met, fail-fast verified. This trip path previously had only unit-level predicate coverage.
- **Exclusion semantics:** a 429 mid-streak neither increments nor resets the counter (4 + 429 + 1 = trips at 5); config faults and cancellations likewise excluded (existing tests).
- **Recovery:** full trip → open → half-open → probe-successes → closed arc, then one ordinary failure must NOT re-trip (counter restarts at 0).
- **Concurrency:** 50 simultaneous distinct-key exchanges under `-race`, followed by a deterministic run that must trip at exactly `callers+1` — proves no update was silently lost during contention (a torn access `-race` would already catch; this proves the arithmetic survives it, which `-race` alone does not). Existing exact-count and open-stampede tests unchanged and green.
- **Observability:** the trip WARN's `consecutive_failures` attribute is asserted directly against a captured log record.

## Breaking Changes

None. Config schema, tool surface, public API, and error sentinels are unchanged.

**Impact:** the only behavior delta is the fix itself — the breaker can now open during a sparse-traffic outage where it previously (incorrectly) never did. `ErrExchangeCircuitOpen` was already a handled outcome on this path.

**Migration Guide:** nothing required. Operators who widened `failure_rate_window` as a workaround for this bug can revert to the default. Operators who genuinely want the consecutive rule off keep the supported switch: `consecutive_failure_threshold: 0`.

### Code Quality

### Security
- [x] Followed secure logging rules — new log attribute (`consecutive_failures` on the existing state-change WARN) reviewed against `secure-logging-rules.md`: a locally-owned integer, no credential/PII/injection vector
- [x] Credential-carrying types implement `slog.LogValuer` (none added)

### Testing
- [x] Error paths tested

### Documentation
- [ ] README.md updated (not applicable)
- [ ] docs/ updated (not applicable — behavior docs live in godoc here)
- [x] CHANGELOG.md updated
- [ ] Examples updated (not applicable)

### Git & CI
- [x] Branch is up to date with main (branched from fresh main)
- [x] No merge conflicts
- [ ] CI checks passing (was all-green before this push; re-running against the latest commits)

## Additional Notes

The wiring relies on three gobreaker v2.4.0 callback behaviors that are not documented library contract. Two are covered end-to-end — a bump that reordered them would fail existing tests. The third (a stale outcome, started before a trip and finished after recovery, is discarded by the generation check before either callback runs) is **not** test-covered today; that gap is why gobreaker is now excluded from the `go-minor-patch` Dependabot auto-merge group rather than relying on tests alone. A follow-up ticket for a stale-generation test to close that gap is worth filing separately.

Out of scope, per the ticket: default value changes, new config knobs, rate-rule changes, open-state jitter, and breaker metrics (sibling stories).

## Related Issues/PRs

- Related to SOL-151600 / PR #205 (introduced the breaker and documented this gap as a known limitation)

---

**For Reviewers:**

**Focus Areas**: the callback wiring in `newTokenExchangeCircuitBreaker` / `newCountingIsBreakerSuccess` — correctness there rests on gobreaker v2.4.0's callback ordering and generation-check semantics described above; and the sparse-spacing test's geometry claim (comment in `sparseSpacingBreakerConfig`).

**Questions**: does the team want the simulation harness used to verify the bucket-decay math checked in anywhere (currently it was a throwaway verification script), or is the comment trail sufficient?


internal/tokenexchange/circuitbreaker_exchange_test.go
```diff
@@ -931,3 +931,341 @@ func TestBreaker_OpenBreakerRejectsStampedeCleanly(t *testing.T) {
...
+// TestBreaker_RateRuleTripsOnPartialDegradation is the end-to-end trip test
+// for the rate rule — the high-traffic path every other exchange-level test
+// deliberately starves. It drives the one failure shape the consecutive rule
+// is structurally blind to: a PARTIALLY degraded IdP whose interleaved
+// successes reset the consecutive streak on every other call, while the
+// failure rate sits exactly at the 50% threshold. The consecutive rule is
+// disabled (threshold 0) so a trip can only come from the rate rule, and the
+// trip must land exactly at the 10th evaluated outcome — the 9th (4 failures
+// of 8 evaluated, also 50%) is below the MinimumRequests floor and must not
+// trip.
```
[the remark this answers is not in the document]
amitmorade: Good catch — the header was off by one on both counts: the trip actually lands on the 11th call (the 10th outcome is a success, and the rate rule is only evaluated on failures), and the pre-floor state is 5 failures of 9 evaluated, not 4 of 8. The inline comments in the test body had the correct arithmetic; the header now matches them. Fixed in 68830d0.

solace-aross (dismissed):
Verified this by breaking it, not just reading it.

**Injected the pre-fix bug** in a throwaway worktree — reverted `newReadyToTrip` to read `counts.ConsecutiveFailures` (gobreaker's decayed count) instead of the new atomic — and reran `TestBreaker_SparseSpacingStillTrips`: fails (`breaker State = closed after 5 sparsely-spaced failures, want open`). Restored the fix, same test passes. So the money test genuinely catches the regression it's named for, and the gap described in the PR is real, not theoretical.

**Checked the gobreaker v2.4.0 internals the design leans on**, against the actual vendored source (`afterRequest` in `gobreaker.go`), not the PR's description of it:
- the generation check happens before `isExcluded`/`isSuccessful`/`readyToTrip` are ever called, under one `cb.mutex.Lock()` for the whole call — confirms the "stale outcome discarded before touching the counter" and "callbacks are serialized" claims.
- `isExcluded(err)` short-circuits before `isSuccessful` runs, so an excluded outcome (429, cancellation) really can't reach `newIsBreakerSuccess` — confirms it lines up with the rate rule's denominator.
- `setState` calls `onStateChange` directly, still inside the locked section — confirms the state-change log's own comment about running under the mutex.

**Full suite**: `go test ./internal/tokenexchange/... -race -cover` — pass, 94.4% coverage, matches the PR description exactly.

No findings. The design doc (two independent trip rules, why the counter lives outside gobreaker, why reset-on-success-only) holds up against the code and against the library. CI green.

solace-awaheed (dismissed):
Approving, I verified the load-bearing claims against `gobreaker/v2@v2.4.0` rather than taking the PR body's word for them, and they all hold:

- `afterRequest` runs the generation check (`gobreaker.go:259-262`) before the
  callbacks (`264-273`), so a stale outcome can't reach the counter.
- `IsExcluded` is consulted before `IsSuccessful` (`gobreaker.go:264`), so 429s
  and config faults touch neither counter.
- The counter is provably 0 on every entry into closed: `ReadyToTrip` is only
  consulted in the closed state (`gobreaker.go:292`), and closed is only entered
  from half-open via consecutive probe *successes* (`gobreaker.go:282-283`),
  each of which stores 0. So the "counter stuck at threshold, breaker re-trips
  forever" failure mode this design could plausibly have is structurally
  impossible — and `TestBreaker_RecoveryNeedsFreshFailuresToRetrip` pins it.
- I hand-simulated `rollingCounts.subtract` at the sparse test's 3-bucket
  spacing and got the same peak of 4 you did. I also reverted `newReadyToTrip`'s
  one line to `counts.ConsecutiveFailures` locally: `TestBreaker_SparseSpacingStillTrips`
  fails 5/5 runs. It's a real regression test, not a test that merely passes.

solace-lootii (dismissed):
Also Approving. and agree with @solace-awaheed 

One non-blocking thing worth a follow-up, not fixing now: the PR notes say
"the new tests act as the tripwire if a future dependency bump changes
callback ordering," but i dont think any test actually drives a stale-generation outcome —
the concurrency test only asserts no-race/no-panic, not that specific
ordering. Since the whole design leans on an undocumented gobreaker internal,
a test that forces that path would make the safety claim actually true rather
than aspirational. Fine to land as a fast-follow.

LGTM.

bczoma (commented):
Reviewed after the three approvals above, so I've deliberately skipped what's already been established there — the gobreaker v2.4.0 callback/generation semantics, the revert-the-line proof on `TestBreaker_SparseSpacingStillTrips` (I got the same result, 3/3), the "counter is provably 0 on entry to closed" argument, and the sparse test's bucket geometry. All confirmed independently on my side; no need to re-litigate.

The fix itself is right and unusually well-evidenced. Below is only what I don't think anyone has covered yet. Nothing here blocks.

---

### 🟡 The threat model still files this defect as an accepted risk

`docs/internal/threat-model.md:85` (§3 STRIDE) still says a slow/low-traffic IdP outage "never trips the breaker (DoS)" with mitigation **"No mitigation — accepted risk, documented in code"** — and it substantiates that by quoting the code comment *this PR deletes*:

> *"the breaker stays closed (fails open)"* (internal/tokenexchange/circuitbreaker.go:53-61)

That citation now dangles, in a document whose own stated method is that every `file:line` was verified by reading the code. The same claim appears again as consolidated accepted-risk **#8** at `:190`, in the §8 table collected "for exception sign-off".

Nothing else will catch this. The doc's own trigger — "re-run before the SolaceProducts flip and whenever a boundary listed below changes materially" — has already fired for the flip, and this PR *is* the boundary change. The house convention is same-PR sync: `3a50eed` (merged a few hours before this PR) flipped a supply-chain row from "No mitigation" to "Mitigated" in the same PR that changed the control, and an earlier row was flipped citing SOL-152412.

Two edits, plus optionally the date stamp at `:16`:
- `:85` → Mitigated, citing `newReadyToTrip` / `newIsBreakerSuccess` and SOL-152286; drop the removed quotation.
- `:190` → restate row #8 **in place** (struck through, "Resolved under SOL-152286") rather than deleting it — `:203`, `:206` and `:217` reference risks by number, so deletion renumbers them.

Worth saying: SOL-152286's AC5 enumerates five files and this isn't one of them, so it's not literally an AC miss — just the last place the old claim survives.

### 🟡 The callback contract comment at `:121-127` is now half-true — and the tripwire has a hole in the bump path

The block comment still asserts these callbacks "must stay fast and **side-effect-free**". The three inner classifiers do; the wrapper installed in the `IsSuccessful` slot writes the counter, and `newReadyToTrip`'s closure is no longer a pure function of its `counts` argument.

This is @solace-lootii's point from a different angle, and I think it's a bit sharper than "aspirational tripwire": the repo auto-merges the `go-minor-patch` dependency group (see `407e97f`). A gobreaker minor can land with nobody reading its changelog, and the failure mode is silent — no compile error, the breaker just stops tripping on slow outages again. The design leans on three v2.4.0 *implementation* details, none of them contract.

Suggested, none of it blocking:
- Extend the `:121-127` comment to say the `IsSuccessful` slot is deliberately wrapped, that purity applies to the three inner classifiers, and name the three gobreaker invariants the end-to-end tests guard.
- Consider `newCountingIsBreakerSuccess`, so the side effect is in the name.
- Consider pinning `gobreaker` out of the auto-merge group so a bump gets human eyes. That's a cheaper guarantee than the stale-generation test lootii suggests, and complementary to it.

Please don't move the counting outside `Execute` — that loses the generation filter, which is the part that's actually load-bearing.

### 🟡 `TestBreaker_ConcurrentConsecutiveCounterRaceFree` has no assertion and cannot fail

`circuitbreaker_exchange_test.go:1232-1267` ends in `_ = e.breaker.State()`. Beyond having nothing to assert, it can't discriminate: `afterRequest` takes `cb.mutex` before every callback, so all counter access is already serialized by gobreaker's own lock. The test passes identically with a plain `uint32`, and would still pass if the increment moved off the atomic entirely. AC3 ("race-free under concurrent exchanges") is nominally covered by a test with no discriminating power.

Cheapest fix: after `wg.Wait()`, drive N further sequential failures and assert the trip lands at exactly the threshold. Otherwise the comment should say plainly that serialization is gobreaker's guarantee and this is a smoke test.

### 🟡 The PR body's security checklist says there are no new log lines

> `[x]` Followed secure logging rules (**no new log lines** — breaker callbacks must stay side-effect-light)

`9ee3c65` adds the `consecutive_failures` attribute to the breaker state-change WARN. The attribute itself is clean — a locally-owned integer, no credential, no PII, no injection vector, consistent with `docs/internal/secure-logging-rules.md` — but that checklist line is the one thing a security reviewer greps for, and it currently says "nothing to see here". Worth rewording to "new log attribute reviewed against secure-logging rules — no credential-bearing values". The "CI checks passing (pending — draft just opened)" line is also stale now, and "Type of Change" is bug-fix only with a `feat(...)` commit in the stack.

### 🟢 `circuitbreaker_exchange_test.go:1164` — the comment inverts the sequence

Header says the handler produces "S F S F", but `hits.Add(1)%2 == 0` succeeds on **even** hits, so hit 1 fails: it's `F S F S`. The comment thirteen lines down at `:1177` says "alternating F S F S" and is correct. (Different spot from the header arithmetic Copilot caught.)

### 🟢 `circuitbreaker.go:70-72` — the disabled case

With `consecutive_failure_threshold: 0` the counter is still incremented and reset on every outcome and still emitted in the WARN, where the doc's "sits below the threshold when the rate rule fired" has no meaning (the threshold is 0). The wasted atomic op isn't worth branching on; a note in the `newLogBreakerStateChange` comment would cover it.

### 🟢 Five hand-rolled `httptest` handlers in one file

`:1157-1170` and `:1234-1245` are near-identical alternating-outcome handlers differing only in a variable name (`hits` vs `counter`), and both duplicate the `successJSON`/503 shape used in three other new tests. An `alternatingIdP(t) (*httptest.Server, *atomic.Int32)` helper alongside `newBreakerTestExchangerPlain` would stop them drifting apart.

### 💡 What `consecutive_failures` means on the half-open transitions

The godoc scopes the rule-attribution heuristic to `closed→open`, but says nothing about the other two transitions. I drove the arc against v2.4.0 to check: `isSuccessful` **is** called for a half-open failure (`afterRequest` calls it before the state dispatch), so a failed first probe increments — the counter grows +1 per open cycle (`cf=4, 5, 6…`). If probe 1 succeeds and probe 2 fails, the same line logs `1`. Both readings are reachable at the default `half_open_probe_requests: 2`.

None of it matters functionally: `readyToTrip` is consulted only in `StateClosed`, the counter is `Store(0)`-ed on every path into closed, and `Uint32` wraparound needs roughly four thousand years of continuous outage. It's just undocumented on the paths where an operator will actually see it during an incident. One sentence of godoc:

> On half-open transitions the value is the running count of counted failures since the last observed success (it grows by one per failed recovery cycle); the rule-attribution reading above applies only to `closed→open`.

Flagging the tempting non-fix explicitly: resetting the counter on entry to open/half-open would make it *worse* — it destroys the "failures since last success" signal and makes every `half-open→open` line read `1`, which is exactly the value the heuristic misreads as "the rate rule fired".

### 💡 Smaller things

- `newIsBreakerSuccess` has no direct unit test — increment/reset can only be asserted end-to-end or by reading the log attribute. Three lines (success→0, failure→+1, return value passthrough) would remove that dependency.
- The `[Unreleased]` CHANGELOG entry is one ~200-word paragraph covering the bug, the mechanism, the reset policy, the exclusion semantics, the log attribute and the disable switch. Two bullets — the fix, and the observability addition — would serve an operator scanning it better; the bucket-decay explanation already lives in the godoc and this PR.
- The `consecutive_failures` attribute isn't traceable to any AC, and the ticket's out-of-scope list names "breaker metrics". It's small and the rationale in the description is good; worth a note on SOL-152286 so ticket verification and the release notes agree on what shipped.
- On your open question — I'd say no, don't check in the simulation harness. The comment trail plus @solace-awaheed's independent hand-simulation is sufficient evidence, and a throwaway script in-tree acquires maintenance cost it can't repay.

amitmorade:
Thanks @bczoma  for the thorough pass — all five points addressed, in commits from `9f607e1` through `ddcb47b`:

**Threat model (`:85`, `:190`)** — agreed, and the `3a50eed` precedent you cited was exactly right to follow. §3 flips to Mitigated, citing the new counter and naming one residual (partial degradation below `minimum_requests` still can't trip either rule — inherent to the sample floor, not fixed here). Row #8 struck through in place rather than deleted, since `:203`/`:206`/`:217` reference by number. Left the `:16` date stamp alone, same as `3a50eed` did.

**Callback contract comment (`:121-127`)** — you're right, it was half-true after the change. Renamed `newIsBreakerSuccess` → `newCountingIsBreakerSuccess` so the side effect is in the name, and extended the block comment to name the three v2.4.0 behaviors we lean on, stating plainly that two are test-covered and the third (stale-outcome discard via the generation check) isn't. Took your suggestion on the auto-merge group too — gobreaker is now excluded from `go-minor-patch` in `dependabot.yml`, so a bump there needs a human reading its changelog. Didn't move the counting outside `Execute` — agreed that's the load-bearing part.

**Race test had no discriminating assertion** — confirmed, `-race` plus gobreaker's own mutex meant it couldn't tell an atomic from a plain field. Rewrote it: the concurrent phase now runs against a threshold one above what 50 calls can reach, guaranteeing a known-closed state afterward, then a deterministic follow-up must trip at exactly that threshold — a lost increment would trip late or never. (Caught a bug of my own in the first rewrite attempt by running it repeatedly before trusting it — worth a mention since it's a good example of exactly the kind of thing this finding was about.)

**Security checklist line** — reworded to name the actual attribute and what was checked against `secure-logging-rules.md`, instead of the now-false "no new log lines." Also fixed the stale CI line and Type of Change while I was in there.

**Follow-ups from your last round** — also done: `newCountingIsBreakerSuccess` now has a direct unit test independent of the end-to-end paths; the CHANGELOG entry is split into two bullets (fix / observability) instead of one dense paragraph; and I posted a scope note on SOL-152286 flagging the `consecutive_failures` log attribute as outside the stated ACs, distinct from the "breaker metrics" sibling story, so ticket verification and release notes agree on what shipped.

One open item from your side I haven't touched: the sparse-test comment/rate-rule-arithmetic fixes are already in from the earlier round (`68830d0`, `3059c76`) — flagging in case you want a final look before approving.

docs/internal/threat-model.md
```diff
@@ -82,7 +82,7 @@ circuit breaker on that path.
...
 | IdP outage causes fast failure cascades (DoS) | Circuit breaker (above) + retry chain deadline + a cross-caller 429 gate that short-circuits before the breaker (`internal/tokenexchange/exchange.go:207-220`, `retry_after_gate.go`) | Mitigated |
-| IdP outage that's slow rather than fast (each failure burning the full retry-chain deadline) never trips the breaker (DoS) | None — explicitly acknowledged in code: *"the breaker stays closed (fails open)"* (`internal/tokenexchange/circuitbreaker.go:53-61`) | **No mitigation — accepted risk, documented in code** |
+| IdP outage that's slow rather than fast (each failure burning the full retry-chain deadline) never trips the breaker (DoS) | Resolved under SOL-152286: the consecutive-failure rule now uses an undecayed counter (`newIsBreakerSuccess`/`newReadyToTrip`, `internal/tokenexchange/circuitbreaker.go`) that resets only on an observed success, not on a rolling-window timer, so it trips regardless of failure spacing. Residual: a *partially* degraded IdP (interleaved successes) below `minimum_requests` still can't trip either rule — inherent to the sample floor, not fixed by this change | Mitigated |
```
solace-aross: `newIsBreakerSuccess` does not exist at this head. `87d6c31` renamed it to `newCountingIsBreakerSuccess` two commits after `9f607e1` wrote this row, and `grep -rn newIsBreakerSuccess` over the tree returns this line and nothing else.

Both halves came from @bczoma's review, the row rewrite and the rename, and they landed in separate commits without being reconciled. Nothing in CI checks doc symbols, so this will sit until someone greps for it.

One word, and worth fixing here rather than leaving this document's stated method ("verified by reading the code, not by trusting an existing doc") pointing at a symbol that is not there. It is the same defect class this commit was written to clear.

Row #8 at `:190` is handled correctly: struck through in place, numbering preserved.

internal/tokenexchange/circuitbreaker_exchange_test.go
```diff
@@ -931,3 +931,410 @@ func TestBreaker_OpenBreakerRejectsStampedeCleanly(t *testing.T) {
...
+	if got := hits.Load(); got != hitsBeforeNext {
+		t.Errorf("server hits rose from %d to %d; open breaker must not call the IdP", hitsBeforeNext, got)
+	}
+}
+
+// TestBreaker_ConcurrentConsecutiveCounterRaceFree drives many concurrent
+// Exchange calls so the counter is read/written from many goroutines at
+// once. -race catches a torn access; serialization is gobreaker's own
+// guarantee (afterRequest takes cb.mutex before any callback), not proven
+// here. What this test proves instead: no update is silently LOST. The
```
solace-aross: "no update is silently LOST" is not what this test proves.

The baseline success at `:1287` does `Store(0)` (`circuitbreaker.go:99`) and the test asserts it succeeded, so any increment lost during the concurrent phase is erased before the follow-up loop starts. The follow-up loop is a plain sequential `for`, so it has no contention in which to lose an update.

Verified rather than argued: swapping `consecutiveFailures.Add(1)` for the lossy `consecutiveFailures.Store(consecutiveFailures.Load() + 1)` leaves this test passing 5/5 under `-race`, every run logging `consecutive_failures=51`, with the whole package green.

The structural reason is the one the comment already concedes two lines up. `afterRequest` holds `cb.mutex` across every callback, so the counter is never accessed concurrently at all, and no test at this layer can falsify the lost-update property. That is a property of gobreaker's lock, not of this test.

The test does earn its keep. It pins that 50 concurrent exchanges leave the breaker closed and the counter in a usable state, and that a fresh streak then trips at exactly the threshold. That is real regression coverage against a stuck or double-counting counter, and it is a clear improvement on the assertion-free version. Suggest the comment claim that, and leave the serialization guarantee attributed to gobreaker where the first two sentences already put it.

solace-aross (dismissed):
Approving. This is not the revision I approved before, so I re-verified the fix by breaking it rather than by reading it.

**Injections, throwaway worktree at `f2f8fca`, `go test ./internal/tokenexchange/... -race`**

| Injection | Result |
|---|---|
| `newReadyToTrip` reads `counts.ConsecutiveFailures` again (the pre-fix bug) | RED: `TestBreaker_SparseSpacingStillTrips`, plus `TestNewReadyToTrip/{consecutive_failures_at_threshold_trips, gobreaker's_own_decayed_ConsecutiveFailures_is_not_consulted, consecutive_rule_trips_ahead_of_an_unmet_sample_floor}` |
| Drop the `Store(0)` reset on success | RED: `TestNewCountingIsBreakerSuccess`, `TestBreaker_SuccessResetsConsecutiveCounter`, `TestBreaker_RecoveryNeedsFreshFailuresToRetrip`, `TestBreaker_ConcurrentConsecutiveCounterRaceFree` |
| `>=` to `>` in the consecutive rule (off-by-one) | RED: 15 tests, including the full recovery arc and the trip-log assertion |
| `Add(1)` to the lossy `Store(Load()+1)` | GREEN 5/5 under `-race`, whole package green. See note 2 |

Baseline at head: green. The new `TestNewCountingIsBreakerSuccess` and the threshold-exactness assertions both earn their green, and the sparse-spacing test is a genuine regression test for the bug in the title, not a test that merely passes.

**On @bczoma's dependabot point.** I went looking to contest it and could not. GitHub's native auto-merge is a per-PR toggle rather than a workflow, so it does not appear anywhere under `.github/`, which is what makes it easy to check for and miss. `gh api .../pulls/289` reports `auto_merge.enabled_by: solace-wmaryam`, `merge_method: squash` on `407e97f`, the exact commit he cited, and the same pattern on #295 and #273. `required_approving_review_count: 1` does not defeat it: auto-merge waits for the approval and the 13 required checks, then merges unattended, and a reviewer approving a grouped three-module bump is unlikely to have read three changelogs. The exclusion is the right call, and the pattern `github.com/sony/gobreaker/*` does match the module path `github.com/sony/gobreaker/v2` in `go.mod:10`.

I also checked the failure mode this design could plausibly have introduced, a counter stuck at threshold across an open cycle. It is structurally impossible: `readyToTrip` is consulted only in the closed state, closed is entered only from half-open via probe successes, and every one of those runs `Store(0)` first.

Two notes below. Neither is a code defect and neither blocks. CI is 28 green with only `Guardian gate: skipping`, which is the known repo-wide condition.

amitmorade:
Thanks @solace-aross — both real, both fixed in 377c296.

The threat-model citation: you're right, `87d6c31` renamed `newIsBreakerSuccess` → `newCountingIsBreakerSuccess` and I never went back to reconcile the row I'd just written. Fixed.

The race test's comment: verified your injection independently — swapping `Add(1)` for `Store(Load()+1)` does pass 5/5 under `-race`, because `afterRequest` holds `cb.mutex` across every callback, so the counter is never actually contended and no test at this layer can prove or falsify a lost update. That's on gobreaker's lock, not this test. Reworded the comment and the assertion message to claim what it actually earns: that 50 concurrent exchanges leave the counter usable (not stuck, not double-counted), via the threshold-exactness follow-up — not a lost-update proof.

solace-aross (approved):
Verified the fix directly rather than trusting the description.

**Sabotage test:** reverted `circuitbreaker.go`'s `newReadyToTrip` to read gobreaker's own decayed `ConsecutiveFailures` instead of the new atomic counter. `TestBreaker_SparseSpacingStillTrips` and three `TestNewReadyToTrip` subtests go red. Restored (`git diff` empty) — green again.

**Non-vacuity check:** checked out pre-PR `main`, dropped in only `circuitbreaker_exchange_test.go` from head, compiled against the old production code. `TestBreaker_SparseSpacingStillTrips` fails against pre-fix code — this is a real regression test for the bug in the title, not a vacuous pass. The other five new tests pass unchanged pre-fix, as expected, since they cover semantics gobreaker's own counter already had right.

**Design check against gobreaker itself**, not just the PR's description of it — read the vendored `gobreaker/v2@v2.4.0` source: `afterRequest` holds one mutex for its whole body and discards stale generations before touching the counter; `isExcluded` runs before `isSuccessful`; `StateHalfOpen`'s `onFailure` opens directly without consulting `readyToTrip`. So the new atomic counter is defense-in-depth on an already-serialized callback path, not a new race.

Tests pass under `-race`, coverage matches the claim (94.4%), CI is 28/28 green. Everything bczoma and the earlier (dismissed) reviews raised is fixed at this head, except two nits bczoma already flagged as non-blocking (half-open log doc note, duplicated `httptest` helpers) — skipping those per this repo's nits policy.

Approving.

PR #298, opened 2026-08-12
SOL-153050: state the registered ruleset, not the work to register it

bczoma:
## Summary

`Commit identity routable` was registered in the `main-protection` ruleset (pinned to Actions `15368`), completing the last open acceptance criterion of SOL-153050. `.github/ADMIN_SETUP.md` still described the ruleset as it was before that change. This reconciles it.

Doc-only. No workflow, script, or production surface touched.

## Changes Made

- **Context count** — twelve → thirteen.
- **Table cell** — `Commit identity routable` no longer reads **"Not yet registered"**; it now names the pin (`15368`, Actions) that distinguishes it from `DCO` (`974774`, the `dco2` App).
- **`Both DCO rows` heading** — leftover from when `DCO sign-off` and `DCO check self-test` were separate contexts. The body was already singular; only the heading was not.
- **The ⚠️ block** — replaced an instruction with a rule (see below).

## The one worth reviewing

The old block said to register this specific check once it had reported green. That was a task, and it stopped being true the moment it was done. It now states what the two controls are, that they pin to different sources, and the general rule:

> **Register a new context only once it has reported green on a real pull request**, and only after the pull request creating it has merged.

That rule applies to the next context anyone adds — SOL-152858 will need it for CodeQL.

## Verification

The doc's required-checks table was diffed against the live ruleset as a set:

```
gh api repos/SolaceProducts/solace-broker-mcp/rulesets/13942241 \
  --jq '.rules[]|select(.type=="required_status_checks")
        |.parameters.required_status_checks[]|"\(.context)\t\(.integration_id // "UNPINNED")"'
```

13 contexts, doc == live, none `UNPINNED`.

## A note on style

This diff adds no dates and no ticket numbers, and removes the one pre-existing `SOL-152902` reference in the text it touches. `ADMIN_SETUP.md` is world-readable: it should declare what the setup is, while git and Jira hold how it got there. An internal ticket id is not resolvable by most of this file's readers.

Nine `SOL-` references remain elsewhere in the file, all pointing at closed tickets. Deliberately left alone — that is a separate cleanup, not part of fixing what the ruleset change made stale.

## Testing

Doc-only; no tests apply. CI's `CHANGELOG updated` check is advisory here — `production-surface.sh` scopes production surface to `internal/config/`, `internal/tools/`, and `internal/composite/definitions/tools.yaml`, none of which this touches.

## Related

- Closes the last acceptance criterion of SOL-153050
- Follows PR #294, which created the check this registers

PR #299, opened 2026-08-12
SOL-152856: Add a required CI gate that runs actionlint and zizmor over every GitHub Actions

solace-awaheed:
## Summary

Adds a required CI gate that runs actionlint and zizmor over every GitHub Actions workflow, and fixes the five findings that surfaced so the gate starts green.

Jira: [SOL-152856](https://sol-jira.atlassian.net/browse/SOL-152856)

## Type of Change

- [x] Documentation update

## Motivation and Context

We have eight workflows and no linter over any of them; `make check` covers Go only. That gap holds while every editor is a squad member who has read the comments. It stops holding once the repository is public and the set of people who can propose a workflow change is everyone.

Two tools, because neither subsumes the other. **actionlint** catches correctness bugs — the class where a workflow silently does nothing, which is how the `FOSSA Scan` required-check gap went unnoticed for weeks — and shellchecks every `run:` block. **zizmor** audits workflow security specifically: unsafe interpolation, over-broad permissions, unpinned actions, cache and artifact handling.

Two of the ticket's three predicted findings turned out not to exist: there is no `pull_request_target` anywhere anymore (`dco.yaml` was retired under SOL-153050), and every action is already SHA-pinned (zizmor's online audits report zero `unpinned-uses`). The ticket has been updated with the corrected picture.

## Changes Made

**Findings resolved (commit 1)** — all fixed, none suppressed, so there is no ignore file and no inline suppressions:

- `ci-pr.yaml` — `persist-credentials: false` on the `license_headers` checkout. `actions/checkout` writes the `GITHUB_TOKEN` into `.git/config` by default. Every other checkout in the repo already set this; SOL-152962 swept for exactly this and missed this job.
- `release.yml` — `cache: false` on both `setup-go` steps. A restored cache entry is an untrusted input to a binary we then attest, and a release runs once per tag, so the cache was buying almost nothing.
- `release.yml`, `release-readiness-check.yaml` — dropped `secrets: inherit` from both reusable-workflow calls. Every `secrets.*` reference in `guardian-scan.yaml` sits in a job declaring `environment: guardian`, and all six names exist in that environment, so nothing was reading through `inherit`. `FOSSA_API_KEY` also exists as an org secret, which made `inherit` look load-bearing — but environment secrets outrank org secrets, so those jobs already resolved the `guardian` copy. What `inherit` actually did was hand the scan `LLM_SERVICE_API_KEY` and `VAULT_URL`, which it never references.
- `release.yml` — `sha256sum -- *.tar.gz` (SC2035). The other remedy shellcheck offers, `./*.tar.gz`, would prefix every line of the published checksums file with `./` and change an artifact people verify against.

**The gate (commit 2):**

- New `.github/workflows/workflow-lint.yaml`, job `workflow-lint`, `contents: read`, actions SHA-pinned.
- **No `paths:` filter, deliberately.** A required check filtered to `.github/workflows/**` never creates a check run on a PR that touches only Go code, so the required context sits pending forever and the PR cannot merge — the failure mode `build-and-test.yml`'s header comment dissects. If this ever needs to be conditional, condition it inside the job so the check still reports, never in the trigger.
- Tool versions pinned: actionlint `v1.7.12`, zizmor `1.29.0`. The actionlint floor is not arbitrary — v1.7.7 rejects `artifact-metadata: write` in `release.yml` as an unknown permission scope, a false positive against a real feature the release already publishes with.
- zizmor's `--persona=regular` and `--min-severity=low` are set explicitly rather than inherited, since its defaults move across releases. `--no-online-audits` keeps the gate deterministic; `unpinned-uses` is offline, so SHA pinning stays enforced.
- `.github/ADMIN_SETUP.md` — the `workflow-lint` row plus the registration procedure.

## Testing

**Test Environment:**
- Go: 1.25.0 (`go.mod`); actionlint v1.7.12, zizmor 1.29.0, shellcheck 0.11.0
- OS: macOS 26.6.1 (arm64)

No unit tests: this PR contains no Go code. Verification is the tool runs below.

**Before** — five findings on `main`:

| Rule | Site | Severity |
|---|---|---|
| `artipacked` | `ci-pr.yaml:152` | medium |
| `secrets-inherit` | `release.yml:85` | high |
| `secrets-inherit` | `release-readiness-check.yaml:69` | high |
| `cache-poisoning` | `release.yml:52` | medium |
| `cache-poisoning` | `release.yml:116` | medium |

Plus `SC2035` at `release.yml:267`, which only appears when shellcheck is installed — `actionlint` silently skips those rules without it.

**After:**

```
$ actionlint                                    # shellcheck 0.11.0 present
EXIT=0

$ zizmor --no-online-audits --persona=regular --min-severity=low .github/workflows/
No findings to report. Good job!
EXIT=0
```

**The gate is proven to fail, not merely observed to pass** — both tools against a known-bad workflow:

```
$ actionlint known-bad.yml     ->  EXIT=1
$ zizmor known-bad.yml         ->  EXIT=14
```

`make check` passes: exit 0, zero failures.

**Checked because these can pass locally and fail in CI:**
- `go.mod` is `go 1.25.0`; actionlint v1.7.12 requires exactly `go 1.25.0` — no toolchain download needed
- `setup-go` calls `core.addPath` on `$(go env GOPATH)/bin`, so `go install` binaries resolve (verified in the action's source, not assumed)
- Pipx 1.16.0 and shellcheck 0.9.0 are both in the ubuntu-24.04 runner image manifest
- Every CLI flag verified against each tool's own `--help`
- The licence-header scripts use `find` and `git rev-parse --show-toplevel`, so `persist-credentials: false` breaks nothing
- No dynamic `secrets[...]` lookups anywhere, so the static reference scan behind the `inherit` removal was complete
- The job needs no secrets, so it reports on fork PRs — necessary for a required check

## Checklist

- [x] Self-review completed
- [x] No credentials in code
- [x] Branch is up to date with main (`behind_by=0`, no rebase needed)
- [x] No merge conflicts
- [x] `docs/` updated — `.github/ADMIN_SETUP.md`
- [ ] CI checks passing — **not yet; this PR is the first run of the new gate**

`CHANGELOG.md` not updated, and shouldn't be: `.github/` is not production surface per `.github/scripts/production-surface.sh`.

## Additional Notes

**One change ships unverified by CI.** The `secrets: inherit` removals only execute on the release path, which fires on tag push — no PR run exercises them. The static analysis is solid, but that isn't the same as observed. Before the next release, run:

```
gh workflow run release-readiness-check.yaml -f version=main
```

That exercises the Guardian scan through the modified call path without cutting a release. It has to wait until after merge, since `workflow_dispatch` only resolves against the default branch.

**shellcheck is the one unpinned tool here** — it comes from the runner image and moves when the image does, which is the same drift this PR pins the other two tools against. Documented in the workflow, with the flag to pin it if it ever surprises us.

**Still outstanding:** registering `workflow-lint` on the `main-protection` ruleset needs a repo admin and is routed to SOL-152412. Until it is required, it enforces nothing.

## Related Issues/PRs

- Implements [SOL-152856](https://sol-jira.atlassian.net/browse/SOL-152856)
- Registration blocked on [SOL-152412](https://sol-jira.atlassian.net/browse/SOL-152412) (admin ruleset change)
- Follows [SOL-152855](https://sol-jira.atlassian.net/browse/SOL-152855) and [SOL-153050](https://sol-jira.atlassian.net/browse/SOL-153050), which removed workflows the original ticket assumed
- Completes the sweep started in [SOL-152962](https://sol-jira.atlassian.net/browse/SOL-152962)

### DIFF
--- a/.github/ADMIN_SETUP.md
+++ b/.github/ADMIN_SETUP.md
@@ -214,6 +214,7 @@ gh api repos/OWNER/REPO/rulesets/13942241 \
 | `Licence headers present` | `ci-pr.yaml` job `license_headers` | Every `.go` file outside `vendor/` opening with the Apache-2.0 header. Needs no secret and no Go toolchain, so it reports on fork pull requests too |
 | `CHANGELOG updated` | `ci-pr.yaml` job `changelog` | Advisory today; see note below |
 | `Commit identity routable` | `ci-pr.yaml` job `identity` | Every commit the PR adds using a routable author and committer address, so a machine hostname does not publish permanently with the history (SOL-152902). **Not yet registered** — see below |
+| `workflow-lint` | `workflow-lint.yaml` job `workflow-lint` | Every file under `.github/workflows/` passing actionlint (correctness, shellcheck over `run:` blocks) and zizmor (workflow security, including SHA-pinning via `unpinned-uses`), so workflow security is enforced by a tool rather than argued in a code comment (SOL-152856). **Not yet registered** — see below |
 | `DCO` | the CNCF `dco2` GitHub App | a `Signed-off-by` trailer on every commit the pull request adds. DCO stands in for a contributor licence agreement, so this is the control behind the repository's provenance claim |
 
 ⚠️ **Require `Guardian scan gate`, and nothing FOSSA-shaped.** The old
@@ -254,6 +255,27 @@ SOL-152902 delivered would have disappeared with that script.
 Register it once it has reported green on a real pull request — the same rule
 applied to every other context here, and the one that would have caught the
 `Guardian scan` mistake. Until it is required it enforces nothing.
+
+⚠️ **`workflow-lint` is new and not yet in the ruleset.** It went in with
+SOL-152856. Register it on the same terms as the row above: read the context name
+off a real run rather than assuming it, then add it.
+
+```bash
+gh run list --workflow workflow-lint.yaml --limit 1 --json databaseId \
+  --jq '.[0].databaseId' \
+  | xargs -I{} gh run view {} --json jobs --jq '.jobs[].name'
+```
+
+Its job id and `name:` are deliberately identical, so the context is the bare
+`workflow-lint` with no `caller / inner` suffix — but read it anyway. Assuming a
+name is what produced the `FOSSA Scan` versus `FOSSA Scan / SCA Scan` gap.
+
+One property matters when registering it: `workflow-lint.yaml` carries **no
+`paths:` filter**, deliberately. A required check filtered to
+`.github/workflows/**` never creates a check run on a pull request that touches
+only Go code, so the context stays pending forever and the pull request cannot
+merge. If someone later adds a `paths:` filter to make it cheaper, this required
+context is what breaks. Condition inside the job, never in the trigger.
 
 `DCO` is served by the CNCF `dco2` GitHub App, so its `integration_id` is `974774`,
 not the `15368` an Actions-produced context carries. Pinning it to Actions would
--- a/.github/workflows/ci-pr.yaml
+++ b/.github/workflows/ci-pr.yaml
@@ -151,6 +151,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
       - name: Self-test the licence-header check
         run: .github/scripts/license-header-check.test.sh
       - name: Verify every Go file carries the Apache-2.0 header
--- a/.github/workflows/release-readiness-check.yaml
+++ b/.github/workflows/release-readiness-check.yaml
@@ -73,7 +73,9 @@ jobs:
       pull-requests: write
       checks: write
       statuses: write
-    secrets: inherit
+    # No `secrets: inherit` — guardian-scan.yaml's jobs carry
+    # `environment: guardian` and resolve the FOSSA/Guardian/Prisma secrets from
+    # that environment themselves. Nothing to pass down.
     with:
       git_ref: ${{ inputs.version }}
       product_full_version: ${{ inputs.version }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,6 +53,10 @@ jobs:
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: 'go.mod'
+          # Build from a cold module cache. A release runs once per tag, so the
+          # cache buys little, and a cache entry restored from a less-trusted
+          # ref is an input to an artifact we then attest.
+          cache: false
       - name: Self-test the licence check
         run: .github/scripts/licenses-check.test.sh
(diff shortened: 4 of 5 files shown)


solace-aross (dismissed):
Ran the actual tools locally rather than trusting the PR's before/after table.

- Installed actionlint v1.7.12 (`go install`) and zizmor 1.29.0 (pip, into a venv) at the pinned versions.
- **On `main`**: zizmor reproduces exactly the 5 findings claimed — `artipacked` at `ci-pr.yaml:152`, `secrets-inherit` at `release-readiness-check.yaml:69` and `release.yml:85`, `cache-poisoning` x2 in `release.yml` (52, 116). `EXIT=14`.
- **At this PR's head**: `actionlint -color` exits 0; zizmor exits clean — "No findings to report." Matches the PR description exactly.
- **Sabotage check**: reverted the `persist-credentials: false` fix on `ci-pr.yaml` in a throwaway worktree — the `artipacked` finding comes right back. The gate isn't just quiet by construction, it actually reacts.
- **Verified the `secrets: inherit` removal is safe**, not just asserted: every `secrets.*` in `guardian-scan.yaml` sits in a job with `environment: guardian`, and `gh api repos/.../environments/guardian/secrets` lists exactly the 6 names the PR claims — `FOSSA_API_KEY`, `GUARDIAN_API_TOKEN`, `GUARDIAN_URL`, `PRISMACLOUD_ACCESS_KEY_ID`, `PRISMACLOUD_CONSOLE_URL`, `PRISMACLOUD_SECRET_KEY`. Environment secrets do outrank org secrets of the same name, so the `FOSSA_API_KEY` org/environment collision the PR flags is exactly as described.
- The no-`paths:`-filter reasoning is correct and matches this repo's existing `build-and-test.yml` header comment on the same class of problem (a required check that never creates a run leaves the context pending forever).

`workflow-lint` itself ran and passed on this PR (22s) — first real proof the gate reports at all. CI green otherwise. No findings.

One non-blocking note for whoever picks up SOL-152412: the release-path `secrets: inherit` removal is exactly the kind of change that's easy to get subtly wrong and won't be exercised by a normal PR — the author already flagged this and gave the `workflow_dispatch` command to verify post-merge. Worth actually running it once this lands, before the next tag.

amitmorade (dismissed):
LGTM — approving.

CI is green across the board (workflow-lint, e2e suites, FOSSA, CodeQL, DCO), and @solace-aross already did the legwork of independently reproducing the actionlint/zizmor findings locally, confirming the fixes clear them, and verifying the `secrets: inherit` removal is safe via the `guardian` environment secrets. Sabotage check (reverting the `persist-credentials` fix) also confirmed the gate actually reacts rather than just being quiet by construction.

solace-lootii (dismissed):
LGTM

solace-aross (approved):
Re-approving after the merge dismissed my prior review. Diffed old-head to new-head: the only change is merging in main, which is #298's already-approved ADMIN_SETUP.md edit (twelve->thirteen contexts, the DCO/identity pin clarification). No new content of this PR's own. Verified the merged doc still carries this PR's `workflow-lint` row correctly alongside #298's changes — no conflict-resolution damage. CI green, including `workflow-lint` itself.

PR #14, opened 2026-04-24
SOL-148967: Implement Production Deployment Packaging

bczoma:
## Summary

- Multi-platform release workflow (`.github/workflows/release.yml`) triggered by `v*` tags — runs full test suite, cross-compiles binaries for 4 platforms (linux/darwin, amd64/arm64), builds multi-arch Docker images, generates SHA256 checksums, and creates a GitHub Release with all assets
- Multi-stage Dockerfile using distroless runtime image with built-in `--health` flag for HEALTHCHECK (no shell/curl needed)
- GHCR source label for automatic package-to-repo linking
- Example Kubernetes manifests (`deploy/kubernetes/`) with Deployment, Service, ConfigMap, Secret
- Example config file (`broker-config.example.yaml`) bundled in release archives
- Apache 2.0 LICENSE file
- README updated: added Quickstart section with binary and Docker deployment guides, fixed `auth.method` → `auth.mode`, added `development_mode` to config examples, updated project structure
- Packaging and deployment documentation (`docs/packaging-release.md`)
- Health check endpoint (`--health` flag + `GET /health`) for Docker and Kubernetes probes
- Fix gosec G704 lint failure in health check endpoint

### Release workflow (on fork `SolaceDev/bczoma-solace-broker-mcp`)
- Pushed `v0.1.0` tag to trigger the full release pipeline
- All CI jobs passed: lint, build, e2e, e2e-oauth, build-binaries (4 platforms), build-docker, release
- GitHub Release created with 4 `.tar.gz` archives + `checksums-sha256.txt`
- Docker image pushed to `ghcr.io/solacedev/bczoma-solace-broker-mcp`

### Binary artifact verification
- Downloaded `solace-broker-mcp-v0.1.0-linux-amd64.tar.gz` from test release
- SHA256 checksum verified against `checksums-sha256.txt`
- Archive contents confirmed: binary, `broker-config.example.yaml`, LICENSE
- `./solace-broker-mcp --version` returns `v0.1.0`
- Server starts with `CONFIG_FILE` pointing to local config
- `GET /health` returns `{"status": "ok"}`
- MCP integration verified: `list-brokers` returns configured brokers, SEMP calls reach brokers

### Docker image verification
- `docker pull ghcr.io/solacedev/bczoma-solace-broker-mcp:latest`
- Container started with config volume mount and `--env-file`
- `GET /health` returns `{"status": "ok"}`
- Docker HEALTHCHECK reports `healthy` (uses `--health` flag inside container)
[...]

PR #301, opened 2026-08-12
SOL-153167: verify the server certificate in the --health probe

bczoma:
## Summary

The container health check (`--health`, the Dockerfile `HEALTHCHECK` command) now verifies the server's TLS certificate instead of accepting any certificate. This closes [CodeQL alert #1](https://github.com/SolaceProducts/solace-broker-mcp/security/code-scanning/1) (`go/disabled-certificate-check`, CWE-295, high security severity) — the repository's only open alert.

Tracked under [SOL-153167](https://sol-jira.atlassian.net/browse/SOL-153167).

## Type of Change

## Motivation and Context

When `tls_cert_file`/`tls_key_file` are configured, the probe speaks HTTPS to the server over loopback. It did so with `InsecureSkipVerify: true`: the server's certificate rarely carries a `localhost` SAN, and switching verification off was the shortcut around that mismatch. The line dates to the original packaging commit `d87a359` (SOL-148967, #14).

The practical risk was low — the probe reads only a status code, sends no credential, and targets the literal `localhost` with an integer port, so there is no path to redirect it off-loopback. But it was a genuine instance of the pattern rather than a false positive, so the alternative to fixing it was a permanent "won't fix" dismissal that a future security review would re-open.

## Changes Made

- `healthProbeTLSConfig` reads `tls_cert_file`, puts the certificate in the trust pool as `RootCAs`, and sets `ServerName` from the certificate's own first DNS SAN, falling back to its first IP SAN. Go accepts a chain whose leaf is present in the root pool, so this holds whether the configured certificate is self-signed or issued by a private CA.
- `healthProbeExitCode` performs the request and maps it to the process exit code, so the probe is testable without `main()`. `nil` TLS config means plaintext, unchanged.
- `healthConfig` retains `CertFile` — the path `healthConfigFromFile` already parsed to decide the scheme.
- `InsecureSkipVerify` and its `//nolint:gosec` are gone from `cmd/`.
- CHANGELOG entry under `[Unreleased]` → Security — deliberately short (108 words): what changed, plus what an operator upgrading a TLS-enabled container needs to check. The design reasoning lives in the code comments, `docs/configuration.md`, and the user guide's health-check troubleshooting section rather than in the release notes.

## Anything reviewers should focus on

**The failure path, which is the one deliberate behavior change.** With TLS configured, a `tls_cert_file` the probe cannot read or parse — or one carrying no DNS or IP SAN — now reports the container **unhealthy** rather than silently probing without verification. The reasoning: the server serves TLS from that same file, so a certificate the probe cannot read is one the server could not have started with, which makes this a true signal rather than a flap. The one shape that moves from working to failing is a certificate with a Common Name and no SAN — and that certificate could never have been verified by any client, Go's verifier having ignored the Common Name since 1.15.

**Why not `VerifyPeerCertificate`.** The common Go idiom for pinning is `InsecureSkipVerify: true` plus a callback comparing raw certificate bytes. That is cryptographically sound but CodeQL still flags it, since it sees the field set to `true` and does not credit the callback. Hence `RootCAs` + `ServerName` with verification left enabled.

**Failures are now diagnosable.** The probe writes the reason to stderr, which Docker records in `State.Health.Log` — previously it exited silently, making "server down", "server returned 503", "certificate unreadable" and "certificate verification failed" indistinguishable, the last being the one outcome that would indicate an actual loopback impersonator. Its own request deadline is 2.5s so it loses the race against the Dockerfile `HEALTHCHECK`'s 3s timeout deliberately: a probe killed by Docker exits silently, one that times out itself explains why. A test parses the Dockerfile so the two cannot drift.

**The pin is the leaf alone, not the file.** `AppendCertsFromPEM` would anchor every certificate in `tls_cert_file`, so a leaf+CA bundle would widen the probe to "anything that CA ever issued", including an impostor with the same SAN. `pool.AddCert(leaf)` only.

**Config resolution matches the server's.** The probe reads through the config package's new `ReadResolvedConfigFile`, so `.env` loading and `${VAR_NAME}` substitution agree with `LoadConfig`; a documented `tls_cert_file: "${TLS_CERT_PATH}"` would otherwise reach `os.ReadFile` literally and report a correctly-serving container as permanently unhealthy. Both TLS paths are trimmed as `LoadConfig` trims them, and the probe installs the redacting slog handler at ERROR before anything on its path can log, so `loadEnvFile`'s INFO line stays out of the health channel and the Rule 3 redaction net covers the probe path.

Out of scope, and untouched: the broker-facing `insecure_skip_verify` option (`internal/semp/resilience/transport.go`), which remains an operator opt-in defaulting to `false` and gated in production mode.

## Testing

**Test Environment:**
- Go version: 1.25 (repo toolchain)
- OS: Linux (WSL2)
- Broker version: n/a — the probe targets this server, not a broker

**Tests Performed:**
- [x] Unit tests added (19 new cases in `cmd/server/main_test.go`, plus 3 in `internal/config/config_test.go`)
- [x] Tested locally with `make check` (build, vet, golangci-lint, `-race` tests, 85% coverage gate) — exit 0
- [x] End-to-end against the real binary

**Test Coverage:**

Written test-first: each test was watched failing against stubs before implementing. New cases cover certificate pinning and SAN-derived `ServerName`, IP-SAN fallback, and the three error paths (missing file, malformed PEM, no SAN).

Two of them are negative tests that would pass trivially against a stub, so both were mutation-checked: re-introducing `InsecureSkipVerify: true` and dropping the status-code check each turn them red.

Beyond the unit tests, the built binary was run against a live server whose certificate's only SAN is `broker.internal.example`: | Scenario | `--health` exit |
| --- | --- |
| TLS configured, genuine certificate pinned | `0` (healthy) |
| Plaintext, no TLS configured | `0` (healthy) |
| **Different** certificate served, same SAN — what a loopback impersonator presents | `1` (rejected) |
| `tls_cert_file` unreadable | `1` (unhealthy, as designed) |

The third row is the one that matters: verification is genuinely enforced, not merely enabled.

### Code Quality
- [x] Self-review completed
- [x] Complex logic has explanatory comments (why the leaf is pinned as its own root; why the error path exits non-zero)

### Security
- [x] No credentials in code
- [x] Followed secure logging rules — no logging statements added or changed (verified with `/check-logs` over the diff)
- [x] No security vulnerabilities introduced (`gosec` via golangci-lint clean; this removes one)

### Testing
- [x] All tests pass locally with `-race`
- [x] Edge cases and error paths covered by tests

### Documentation
- [x] CHANGELOG.md updated
- [x] godoc comments on the new functions

### Git & CI
- [x] Commit signed off (DCO)

## Additional Notes

The `CodeQL` PR check reports on alerts **newly introduced** by a diff ("No new alerts in code changed by this pull request"), so a green check here is not proof the alert closed. Confirm after merge on the `main` analysis:

```
gh api repos/SolaceProducts/solace-broker-mcp/code-scanning/alerts/1 --jq '.state, .fixed_at'
```

Expect `closed` with a timestamp, and zero open alerts repo-wide.

## Related Issues/PRs

- Unblocks the cosmetic half of SOL-152858 (make CodeQL a required check) — that ticket does not depend on this one, since a pre-existing alert never fails the PR check, but it reads better against zero open alerts.

---

## Update — review feedback applied

Three rounds landed after the initial push. Head is now e92c8b3.

**From Copilot** (`cf355df`, `30c827d`): the chain-bundle trust widening described above, and the probe deadline. Its third comment — have `firstCertificateFromPEM` pick the first certificate carrying a SAN — was declined inline: `crypto/tls.X509KeyPair` matches the private key against `Certificate[0]`, so a bundle whose first block is not the leaf cannot serve TLS at all, and selecting a different block would pin a certificate the server is not presenting.

**From an independent review** (`cf355df`): the `${VAR_NAME}` divergence, the silent failure path, the untested `main()` wiring, the non-CA leaf shape the test helper never generated, and doc updates across `user-guide.md`, `configuration.md`, `README.md`, and `packaging-release.md`.

**From @solace-aross** (`4bde4ea`): resolving config through `ReadResolvedConfigFile` pulled `loadEnvFile` onto the probe path, where it logs at INFO before `slog.SetDefault` runs — so every *successful* probe wrote a plain-text line into `State.Health.Log`, and the path ran with no `ReplaceAttr` filter. Fixed by installing the redacting handler at ERROR as the branch's first statement. Verified with the built binary and a `.env` beside the config: an INFO line before, nothing after, failure diagnostics unchanged.

`e92c8b3` merges current `main` (the SOL-153089 PRM work). The only conflict was both branches appending tests to the end of `cmd/server/main_test.go`; both sets are kept and both pass. `make check` green at every step.

**From @amitmorade** (`dd2df16`): the CHANGELOG entry had grown to 672 words over five paragraphs — the longest in the file's history by roughly 50% (longest released entry 438, the neighbouring `[Unreleased]` entry 212, median released entry 16). It had accreted a paragraph per review round and become a change narrative rather than a release note. Cut to 108 words covering what changed and what to check when upgrading a TLS-enabled container; everything removed already had a better home in the code comments, `docs/configuration.md`, or the user guide.

`81ba9b1` merges current `main` again (SOL-152856 workflow linting, SOL-152286 circuit-breaker work) — no conflicts this time. `make check` green.


cmd/server/main.go
```diff
@@ -156,8 +158,115 @@ func buildRootHandler(mux *http.ServeMux) http.Handler {
...
+
+	pool := x509.NewCertPool()
+	if !pool.AppendCertsFromPEM(pemData) {
+		return nil, fmt.Errorf("health probe certificate %s contains no usable PEM certificate", certPath)
+	}
+
+	leaf, err := firstCertificateFromPEM(pemData)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
bczoma: Agreed, and fixed in cf355df — this was the substantive issue in the PR. An independent review flagged the same thing and reproduced the consequence: with a leaf+CA bundle, an impostor leaf issued by that same CA with the same SAN verified successfully.

`healthProbeTLSConfig` now anchors the leaf alone (`pool.AddCert(leaf)`), which also removes the double parse of the PEM bytes. `TestHealthProbeTLSConfig_ChainFilePinsLeafNotCA` generates a real CA, issues two leaves from it carrying the same SAN, serves each in turn, and requires the genuine one to verify and the impostor to be rejected. Re-widening the pool back to `AppendCertsFromPEM` turns that test red, so the guard is load-bearing rather than incidentally passing.

cmd/server/main.go
```diff
@@ -156,8 +158,115 @@ func buildRootHandler(mux *http.ServeMux) http.Handler {
...
-	Port   int
-	Scheme string // "http" or "https"
+	Port     int
+	Scheme   string // "http" or "https"
+	CertFile string // tls_cert_file, pinned by the probe when Scheme is https
+}
+
+// healthProbeTimeout bounds the whole --health request. Docker's HEALTHCHECK
+// timeout is 3s (see the Dockerfile), so the probe must give up before that.
+const healthProbeTimeout = 3 * time.Second
```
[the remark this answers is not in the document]
bczoma: Fixed in 30c827d. The comment and the constant disagreed, and you picked the right one to keep — I had initially "fixed" this by rewording the comment to say the match was deliberate, which was the wrong half.

The reason it is the wrong half only became clear once cf355df made the probe write failure reasons to stderr. The two deadlines race, and only one outcome is diagnosable: a probe killed by Docker exits silently and leaves `State.Health.Log` empty, while a probe that times out itself explains why and Docker records it. A hung server is exactly when an operator needs that line, so the probe should lose the race by design. `healthProbeTimeout` is now 2.5s.

`TestHealthProbeTimeout_IsUnderDockerHealthcheckTimeout` parses `--timeout=<N>s` out of the Dockerfile and fails if the margin disappears, so the two values cannot drift apart — verified by editing the Dockerfile to 2s and watching it go red.

cmd/server/main.go
```diff
@@ -156,8 +158,115 @@ func buildRootHandler(mux *http.ServeMux) http.Handler {
...
+			continue
+		}
+		leaf, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return nil, fmt.Errorf("parse health probe certificate: %w", err)
+		}
+		return leaf, nil
+	}
+	return nil, errors.New("no CERTIFICATE block in health probe certificate")
+}
```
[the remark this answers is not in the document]
bczoma: Declining this one, with reasoning — and the doc comment was tightened in cf355df to record it, since the wording was fair to criticise.

The premise is that `tls_cert_file` might contain a chain in a different order. Such a file cannot serve TLS at all: `crypto/tls.X509KeyPair` appends every CERTIFICATE block in file order and then matches the private key against `Certificate[0]` ([tls.go:279](https://github.com/golang/go/blob/master/src/crypto/tls/tls.go) and `:316`), so a bundle whose first block is an intermediate fails at startup with a key/certificate mismatch. `ListenAndServeTLS` never gets past it. The probe cannot encounter that shape at runtime, because there is no running server to probe.

That makes the suggested change actively harmful rather than neutral: selecting the first certificate carrying a SAN would pin a certificate the server is not serving from, converting a loud startup failure into a silently wrong pin. Preferring the first block is not an assumption the probe imposes — it is the same constraint the server itself enforces.

The comment now reads: "returns the first CERTIFICATE block in a PEM bundle ... That block is the leaf: `crypto/tls.X509KeyPair` matches the private key against `Certificate[0]`, so a bundle whose first certificate is not the leaf cannot serve TLS at all."

The genuinely missing SAN case — a leaf that carries no DNS or IP SAN — is handled explicitly and returns a named error rather than failing obscurely later, and it is now documented in `docs/configuration.md` and the health-check troubleshooting section of `docs/user-guide.md`.

cmd/server/main.go
```diff
@@ -176,7 +323,11 @@ func healthConfigFromFile() healthConfig {
 	if path == "" {
 		return healthConfig{Scheme: "http"}
 	}
-	data, err := os.ReadFile(path) //nolint:gosec // path from trusted config locations
+	// Resolve through the config package rather than reading the file directly, so
+	// .env loading and ${VAR_NAME} substitution match what the server saw. A probe
+	// that read `tls_cert_file: "${TLS_CERT_PATH}"` literally would fail to open a
+	// certificate the server is serving from perfectly well.
+	data, err := config.ReadResolvedConfigFile(path)
```
solace-aross, line 352: Non-blocking, but worth knowing before this merges: routing through `ReadResolvedConfigFile` pulls `loadEnvFile` into the probe path, and `loadEnvFile` calls `slog.Info("loaded .env file", ...)`. The `--health` branch in `main()` runs before `slog.SetDefault(slog.New(newSlogHandler(...)))`, so that line goes out through Go's default handler.

Observed on the healthy path with a `.env` beside the config:

```
2026/08/12 18:38:01 INFO loaded .env file path=/.../probeexp/.env
exit=0
```

Two consequences:

- Docker records it in `State.Health.Log` on every successful probe, once per interval, in plain text rather than the server's JSON. It shares the channel this PR just designated as the operator's diagnostic surface for failures.
- More durably: the probe path runs with no `ReplaceAttr` filter, so the Rule 3 redaction safety net in `newSlogHandler` is not installed for anything reachable from here. Neither `loadEnvFile` nor `substituteEnvVars` logs a value today, so nothing leaks now. It is a latent hole rather than a live one.

The documented container shapes mount only `config.yaml` and inject credentials via `--env-file`/`env_file`, so there is usually no `.env` inside the image and the line never fires. It does fire for a bind-mounted config directory, and for `./solace-broker-mcp --health` against a repo-root `.env`, which the user guide suggests for scripts.

One line before the probe runs would close both, e.g. `slog.SetDefault(slog.New(newSlogHandler(slog.LevelError)))`, keeping stderr for the probe's own `fmt.Fprintln` output. Your call whether it belongs in this diff or a follow-up.

solace-aross (dismissed):
Approving. I verified the fix by injection rather than by reading, in a throwaway worktree at 30c827d, and every guard held.

**Sabotage runs** (each defect reintroduced, `go test ./cmd/server/` before and after; baseline green):

| Defect reintroduced | Result |
| --- | --- |
| `InsecureSkipVerify: true` returned from `healthProbeTLSConfig` | 4 red: `PinsCertAndTakesServerNameFromDNSSAN`, `FallsBackToIPSAN`, `ErrorsWhenCertHasNoSAN`, `ChainFilePinsLeafNotCA` |
| `pool.AddCert(leaf)` widened back to `AppendCertsFromPEM(pemData)` | 1 red: `ChainFilePinsLeafNotCA` |
| `healthProbeTimeout` back to `3 * time.Second` | 1 red: `IsUnderDockerHealthcheckTimeout`, with the Dockerfile-parsed margin in the message |
| `config.ReadResolvedConfigFile` back to `os.ReadFile` | 1 red: `SubstitutesEnvVarsInCertPath` |

No test passes by accident. `TestProbeHealth_FailsWhenCertNotPinned` survives the first sabotage because it builds its own `tls.Config` rather than calling `healthProbeTLSConfig`, which is the right shape for a control test.

**End-to-end against the built binary and a real TLS listener** (self-signed leaf, sole SAN `broker.internal.example`, served on loopback):

- genuine certificate configured: exit `0`
- a *different* certificate carrying the same SAN configured: exit `1`, stderr `tls: failed to verify certificate: x509: certificate signed by unknown authority`
- server not listening: exit `1`, stderr names the dial failure

So the pin is load-bearing, and the new stderr channel does distinguish the four outcomes it claims to.

**On the declined Copilot comment about `firstCertificateFromPEM`**, I checked the premise independently rather than taking the reply on trust. `crypto/tls.X509KeyPair` parses `Certificate[0]` and matches the private key against it, so a bundle whose first CERTIFICATE block is an intermediate cannot serve TLS at all and the probe can never meet that shape at runtime. The decline is correct, and selecting the first SAN-bearing certificate instead would have pinned something the server is not presenting. Nothing to add there.

One non-blocking note is inline. Separately, the PR description has drifted from the head commit: it still says the probe stays silent on failure and names `healthProbeExitCode`. The CHANGELOG entry is accurate and is the record that lasts, so this only matters for whoever reads the merge commit.

cmd/server/main.go
bczoma: Implemented

solace-awaheed (dismissed):
Nothing flagged on my end, LGTM!

CHANGELOG.md
```diff
@@ -21,6 +21,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
... the certificate's own first DNS SAN (falling back to its first IP SAN), so verification succeeds on the name the certificate actually carries. Pinning the leaf alone rather than every certificate in the file matters for a leaf+CA bundle, the shape cert-manager and corporate PKI emit: anchoring the bundle's CA would widen the probe from "this certificate" to "anything that CA ever issued", including an impostor holding a legitimately-issued certificate with the same SAN. Go's verifier short-circuits when the presented leaf is itself in the root pool, so a leaf-only anchor verifies both self-signed and CA-issued certificates. Verified end-to-end against a running server whose certificate's only SAN is `broker.internal.example`: the probe reports healthy, while a *different* certificate carrying that same SAN — including one issued by the same CA out of a chain bundle — is rejected. Practically low-risk before the change, since the probe only ever reads a status code, sends no credential, and targets the literal `localhost` with an integer port; but it was a genuine CWE-295 instance and the repository's only open CodeQL alert (`go/disabled-certificate-check`, high security severity).
```
amitmorade: @bczoma Does this level of detail belong in the CHANGELOG, or would it fit better as a code comment / PR description only?

Trying to think about who actually reads a CHANGELOG entry: someone skimming release notes to decide "does this affect me / do I need to do anything," or a future engineer doing git-blame-adjacent archaeology trying to figure out why something behaves a certain way. Neither of those readers needs a real-time tour of the design reasoning — the leaf-vs-chain pinning tradeoffs, why 2.5s vs 3s, the walkthrough of Go's cert verifier behavior since 1.15.

Not pushing to cut it, just curious about the intent — was this meant to be the canonical explanation (so it's fine that it overlaps with the doc comments in `main.go`), or would a shorter "what changed / what to check if your container goes unhealthy" version serve the CHANGELOG's actual readers better?

amitmorade (changes requested):
Requesting changes because I have a concern with the CHANGELOG wording — details in my comment above. This ships directly to users, so wanted to get eyes on it before it merges under the existing approval. Let me know what you think.

CHANGELOG.md
bczoma: Fair challenge, and you are right — trimmed in dd2df16, from 672 words to 108.

To answer the question directly: it was drift, not intent. The entry started at a reasonable length and then grew a paragraph per review round — chain-vs-leaf pinning, the probe deadline, env-var resolution in the config path, the slog handler — until it was a narrative of how the change was reviewed rather than a release note. Nobody decided it should be the canonical explanation.

Your two reader personas are the right test, and the numbers back you up. Measured against the rest of the file:

| | words |
| --- | --- |
| The entry as you found it | 672 |
| Longest entry in any released block | 438 |
| Neighbouring entry in the same `[Unreleased]` block (SOL-153089) | 212 |
| Median released entry | 16 |

So it was not merely long for this file — it was the longest entry the CHANGELOG has ever carried, by roughly 50%, for a health-probe hardening rather than a breaking change. The house rule is that length scales with impact, and this exceeded it.

What survives is one bullet (what changed) plus one paragraph headed "Upgrading a TLS-enabled container" (the four conditions that now report unhealthy while the server keeps serving, and that failure reasons land in `State.Health.Log`), then links to the two docs pages.

The cut was safe rather than lossy because everything removed already had a better home, added during the same review rounds: the leaf-versus-chain rationale is the doc comment on `healthProbeTLSConfig`, the 2.5s deadline is its constant's comment plus a test that parses the Dockerfile so the two cannot drift, and the SAN requirement with the Go 1.15 Common Name background is in `docs/configuration.md` and the health-check section of `docs/user-guide.md`. Nothing operator-facing was dropped.

`81ba9b1` also merges current `main`, which the branch was 8 commits behind. Ready for another look when you have a moment.

internal/config/config.go
```diff
@@ -571,6 +571,32 @@ func sanitizeYAMLError(err error) error {
...
+// returning the resolved YAML bytes without parsing or validating them.
+//
+// This exists for readers that need a field or two out of the config without a
+// full load, and must agree with LoadConfig on what those fields say. The
+// --health probe is the caller: it needs the port, scheme, and TLS certificate
+// path before slog and the config machinery are up, and a probe that resolved
+// `tls_cert_file: "${TLS_CERT_PATH}"` differently from the server would report a
+// correctly-serving container as unhealthy. Prefer LoadConfig for anything that
+// needs a validated ServerConfig.
+func ReadResolvedConfigFile(path string) ([]byte, error) {
```
solace-aross, line 585: This duplicates `LoadConfig`'s first three steps (read, `loadEnvFile`, `substituteEnvVars`) verbatim instead of `LoadConfig` calling this. The doc comment above says this "must agree with LoadConfig on what those fields say," but nothing structural enforces that — a future change to `LoadConfig`'s preprocessing order has no compiler or test signal to keep the probe in sync, since the two functions share no code. Not a bug today (both are correct and tested), but worth having `LoadConfig` call `ReadResolvedConfigFile` internally so the agreement is structural rather than a comment's promise.

solace-aross (dismissed):
Verified the security property directly rather than trusting five rounds of review talking about it. Built the actual `--health` binary and ran it against real TLS servers: genuine leaf cert pinned in config → exit 0; a different cert carrying the *same* SAN (what a loopback impersonator presents) → exit 1, `x509: certificate signed by unknown authority`. That's the property this PR exists to add, reproduced end-to-end against the compiled binary, not just unit tests.

Sabotage-tested the fix itself: reverted `pool.AddCert(leaf)` to `AppendCertsFromPEM` — `TestHealthProbeTLSConfig_ChainFilePinsLeafNotCA` goes red. Re-added `InsecureSkipVerify: true` — two tests go red. Restored both, green again.

All five prior review threads (Copilot ×3, my own earlier one, amitmorade's) independently confirmed fixed at this head, not just marked resolved: leaf-only pinning (`main.go:230-231`, no `AppendCertsFromPEM` anywhere), probe timeout under Docker's (`healthProbeTimeout = 2500ms`, and the Dockerfile-parsing test is real, not hardcoded), the `firstCertificateFromPEM` decline (checked Go's `X509KeyPair` source — the reasoning holds), the redacting handler installed before `loadEnvFile` can log, and the CHANGELOG trim. `make check` passes for real (89.1% coverage).

Two Low, non-blocking notes inline. Approving — the fix is real and the process gate (amitmorade's `CHANGES_REQUESTED` review needs clearing) is his to close, not a technical block.

amitmorade (dismissed):
LGTM

internal/config/config.go
bczoma: You are right, and this is the sharper version of the point: a comment promising agreement is exactly what we had before this PR. The probe read the config with its own private struct and no substitution, and the `${TLS_CERT_PATH}` bug lived in that gap for as long as nothing structural held the two together. Writing a second implementation and a comment saying "must agree" reproduced the same arrangement one level up. Fixed in 7ceff5a — `LoadConfig` now calls `ReadResolvedConfigFile`, so steps 1-3 are shared code rather than a promise.

Behaviour is unchanged: both wrapped their errors with the same strings, so `LoadConfig`'s error text is identical, and the tests pinning it pass untouched (`TestLoadConfig_EnvVar*`, the strict `CONFIG_FILE` "reading config file" assertion, `TestLoadConfig_BrokerOAuth_ClientSecret*`). The doc comment on `LoadConfig` now marks which steps are delegated, so the next person changing the preprocessing order can see who else is affected.

**Verifying the refactor turned up a second gap, which is the more interesting half.** I deleted the `loadEnvFile` call from the now-shared helper to confirm both callers would notice — and nothing failed. `loadEnvFile` was unit-tested directly, but no test held either caller to actually calling it, so step 2 of the documented processing order was unenforced in exactly the way your comment describes. Sharing the code alone would not have caught its removal.

Two tests now cover it, one through each entry point, each supplying a variable *only* via a `.env` file beside the config with the process environment deliberately unset:

- `TestLoadConfig_ResolvesVarsFromEnvFile`
- `TestReadResolvedConfigFile_ResolvesVarsFromEnvFile`

Both go red when the call is removed. So the coupling is now structural *and* the shared step is pinned.

`make check` green. Thanks for pushing on this one — it was the right call.

solace-aross (approved):
The DRY note is closed: `LoadConfig` now calls `ReadResolvedConfigFile` instead of duplicating its first three steps, so the two can no longer drift apart by construction rather than by comment.

Verifying it turned up a real second gap, which is now also closed: nothing previously held either caller to actually calling `loadEnvFile` — deleting that call broke no existing test. Confirmed this myself: removed the call, and the two new tests (`TestLoadConfig_ResolvesVarsFromEnvFile`, `TestReadResolvedConfigFile_ResolvesVarsFromEnvFile`) both fail with `environment variables not set`, exactly as expected. Restored, green again.

`go test ./internal/config/... -race` passes. Approving again at this head.

PR #302, opened 2026-08-13
SOL-152947: SEMPv2 config/action output-schema drift guard (write tools)

solace-lootii:
## Summary

SEMPv2 config/action output-schema drift guard for write tools — the config/action counterpart to SOL-150785's monitor-side work (part C there, deferred into this ticket since it was comparable in scope to that ticket's whole part A).

The 8 create/update composite tools (`create-message-vpn`, `update-message-vpn`, `create-queue`, `update-queue`, `create-topic-endpoint`, `update-topic-endpoint`, `create-rdp`, `update-rdp`) previously declared the same generic, fully-permissive step-keyed output schema every composite tool uses. That schema can't catch a SEMP spec bump that renames, drops, or reshapes a response field a client depends on — it accepts anything. This adds a strict, spec-derived schema generator and wires it in for these 8 tools specifically; monitor (read) tools and delete/action tools (no response data) are unchanged.

- **Phase 1** — `composite.BuildStrictOutputSchema` + `buildStepSchema`: generates a step's schema from its operation's already-resolved `ResponseFields` (from SOL-150785 Phase 1, generic and reused unmodified here). Handles the three runtime step-result shapes (flat/single-object, paginated, fan-out). Falls back to the existing permissive schema when an operation isn't found or its response shape couldn't be resolved (delete/action tools, by design — confirmed against the real spec, not assumed).
- **Phase 2** — wired into `CompositeToolHandler.Metadata()` via `callsConfigOrActionOperation`, gated to config/action steps only. Caught a real bug here: a first version rejected every real create/update response in `test/e2e-management` (`"Additional property data/links/meta is not allowed"`) because a flat step's runtime result is the *whole* SEMP envelope (`{"data", "meta", "links"}`), not the unwrapped item — traced to `client.go`'s `Execute()`. Fixed by wrapping the flat-step schema in the envelope shape; re-verified against real Dockerized brokers (`test/e2e-management` 12/12, `test/e2e-action` 13/13).
- **Phase 3** — `write-tool-schema-resolves`, a CI guard subtest in `loader_embedded_test.go` asserting `BodyFields`/`ResponseFields` still resolve (and the generated schema is still strict, not fallen back to permissive) for all 8 write tools against the real embedded catalog. Verified it fires on real drift by repointing a body `$ref` at an empty definition in the spec, confirming failure, reverting.
- **Phase 4** — `write_tool_drift_test.go`: 5 tests feeding a real create-queue schema an unexpected field, a missing identifier, a type-changed field, and a missing envelope `data` key, confirming each is rejected (plus one positive control).
- **Post-review fixes** (self-review requested and applied before opening this PR):
  - Added the `CHANGELOG.md [Unreleased]` entry this repo's convention requires for changes touching `internal/tools/`/tool definitions.
  - Found and closed a design gap: `writeToolIdentifierFields`'s required field names were never checked against the operation's actual resolved `ResponseFields` — a typo or future mismatch there would silently generate a self-contradictory schema (a field simultaneously `required` and forbidden by `additionalProperties: false`), which no real broker response could ever satisfy. Added `TestWriteToolIdentifierFields_ResolveAgainstRealResponseFields` and verified it catches this by injecting a typo'd field name, confirming failure, reverting.
  - Documented (comment only, no behavior change) a latent, currently-unreachable ordering gap in `buildStepSchema`: a paginated/fan-out step with unresolvable `ResponseFields` falls back to the bare permissive schema instead of the known envelope shape — not reachable today since every write tool is single-step `collect`, relevant only if this generator is later reused for monitor tools.

## Test plan

- [x] `make check` (build, vet, lint, race tests, coverage gate) green — 88.8%, above the 85% floor
- [x] `go test ./... -race -count=1` (full repo, no cache) — all packages pass, no failures/panics/data races anywhere
- [x] Real Dockerized broker e2e: `test/e2e-management` (12/12) and `test/e2e-action` (13/13), which is what caught the envelope-wrapping bug in the first place
- [x] CI guard (`write-tool-schema-resolves`) verified to actually fire on injected spec drift, then reverted
- [x] New identifier-field guard test verified to actually fire on an injected typo, then reverted
- [x] Merged latest `main` in; `git merge-tree` shows no conflicts, branch is 6 ahead / 0 behind `origin/main`
- [x] All commits DCO-signed (`git commit -s`)


internal/composite/schema.go
```diff
@@ -0,0 +1,156 @@
...
+	for name, jsonType := range fields {
+		properties[name] = map[string]any{"type": jsonType}
+	}
+	schema := map[string]any{
+		"type":                 "object",
+		"properties":           properties,
+		"additionalProperties": false,
+	}
+	if len(required) > 0 {
+		schema["required"] = required
```
solace-aross: This aliases the slice from `writeToolIdentifierFields` (`internal/tools/composite_handler.go:103`) into the schema instead of copying it. Mutate a returned schema's `required` list and it corrupts the package-level map for every future call — confirmed by mutating one handler's output, then observing a brand-new `NewCompositeToolHandler` instance return the corrupted value on a subsequent call.

Dormant today (`Metadata()` has 27 read-only call sites, none mutate), but `Metadata()` runs on every tool invocation (`internal/tools/manager.go:171`), not once at registration, so a future mutator would break every call to that tool across every broker, not one registration.

One-line fix: `schema["required"] = slices.Clone(required)`. Worth also updating `types.go:106`'s "fresh maps" wording to "maps and slices", and `manager.go:170`'s "without aliasing risk" comment — both would be true again once this is fixed.

internal/composite/schema.go
```diff
@@ -0,0 +1,156 @@
...
+			"type": "object",
+			"properties": map[string]any{
+				"data":             map[string]any{"type": "array", "items": item},
+				"truncated":        map[string]any{"type": "boolean"},
+				"truncatedMessage": map[string]any{"type": "string"},
+			},
+			"required":             []string{"data", "truncated"},
+			"additionalProperties": false,
+		}
+	case step.ForEach != "":
```
solace-aross: This doesn't wrap `item` in the envelope the way the `default` case three lines down (`schema.go:116-125`) does, for the identical reason: `fetchFanOut`'s `runSingle` (`executor.go:227`) returns `result.Data` verbatim, which is the same raw SEMP envelope `{data,meta,links}`, not the unwrapped item. Validated against the real shape:

```
byKey.vpn-a: Additional property meta is not allowed
byKey.vpn-a: Additional property data is not allowed
```

`TestBuildStrictOutputSchema_FanOutStep` (`schema_test.go:143-145`) certifies the wrong shape — it feeds the bare unwrapped item, which is why this passes today.

Unreachable in this PR (only one `ForEach` step exists in the whole catalog, `list-vpns`, and it has no config/action step so never reaches `BuildStrictOutputSchema` at all), but the switch order also disagrees with runtime dispatch: this checks `FollowPages` before `ForEach` (line 78 vs 91), while `runStep` checks `ForEach` first (`executor.go:191`) — a step with both flags set would get the wrong schema entirely.

Fix: wrap `item` in the same envelope inside this case, fix the test to assert against the real envelope shape, and reorder the switch to match `runStep`'s dispatch order. Matters before this generator gets reused for monitor tools, which do have `ForEach` steps.

solace-aross (changes requested):
Verified rather than trusted the description. `make check` is genuinely green (88.8%), lint clean, CI green on the actual head SHA including `e2e-management`/`e2e-action`. The core mechanism — deriving a strict per-step schema from resolved `ResponseFields`, falling back to permissive for delete/action tools — works as described, and the self-review-flagged identifier-field gap (`TestWriteToolIdentifierFields_ResolveAgainstRealResponseFields`) is real and genuinely catches an injected typo.

Two findings survive independent verification (worktree injection, then a second skeptical pass), both in the schema generator itself, both currently unreachable in this PR's shipped scope but worth closing before they aren't:

1. A slice-aliasing bug (`schema.go`) that corrupts `writeToolIdentifierFields` permanently once triggered — dormant today, but `Metadata()` runs per-invocation, not once at startup, so the blast radius is every call to that tool across every broker, not one registration.
2. The `ForEach` branch never got the envelope-wrapping fix the default branch got for the identical reason — and the one test that exercises it certifies the wrong (unwrapped) shape, so it won't catch a real regression when this generator is reused for monitor tools, which do have `ForEach` steps.

Both are cheap fixes. Requesting changes on these two rather than approving with notes, mainly because of #2's misleading green test — nothing in this PR's actual scope exercises the `ForEach` path, so CI doesn't catch this one on its own the way it would for the write-tool paths this PR does exercise.

internal/tools/composite_handler.go
```diff
@@ -40,20 +75,37 @@ func NewCompositeToolHandler(tool composite.CompositeTool, executor *composite.C
...
 		Annotations:  toolAnnotations(h.tool.Annotations),
 	}
 }
 
+// outputSchema returns the strict, spec-derived schema for a write tool, or
+// the generic step-keyed envelope for everything else.
+func (h *CompositeToolHandler) outputSchema() map[string]any {
+	if !callsConfigOrActionOperation(h.tool) {
+		return StepKeyedEnvelopeSchema()
+	}
```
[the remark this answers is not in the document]
solace-lootii: Verified this against a real StepKeyedEnvelopeSchema diff, not just by reading the code — you're right, it's a real discrepancy. `callsConfigOrActionOperation` does match delete/action tools too (e.g. `delete-queue`'s `config/deleteMsgVpnQueue` step), so they now go through `BuildStrictOutputSchema` instead of the generic envelope, exactly as described.

However, their operations resolve no `ResponseFields` (`SempMetaOnlyResponse`), so `buildStepSchema` falls back to the same permissive `{"type":"object"}` per-step schema they'd have gotten anyway — confirmed by diffing `delete-queue`'s actual generated schema against `StepKeyedEnvelopeSchema()`. The only real difference is at the top level: the envelope now requires exactly the tool's known step IDs and sets `additionalProperties: false` there, instead of the fully generic wrapper. Since a composite tool's result always contains exactly its step IDs at runtime, that's harmless in practice, not a functional risk.

Rather than tightening the gate to `writeToolIdentifierFields` only (which would mean rewriting a few tests that currently exercise this exact behavior and got independent sign-off from a human reviewer as "falling back to permissive for delete/action tools — works as described"), I fixed the misleading doc comments on `Metadata()`/`outputSchema()` to state this precisely instead of implying delete/action tools are fully untouched. Pushed in b99dc8b.

internal/composite/executor.go
```diff
@@ -76,6 +76,14 @@ type CompositeExecutor struct {
 // defaultMaxPages is the production value of CompositeExecutor.maxPages.
 const defaultMaxPages = 1000
 
+// Operations returns the executor's operation catalog. Exposed for callers
+// that need to introspect an operation outside of execution — e.g.
+// CompositeToolHandler.Metadata() building an output schema from a step's
+// resolved response fields (BuildStrictOutputSchema).
+func (ce *CompositeExecutor) Operations() map[string]*sempv2.Operation {
+	return ce.operations
+}
```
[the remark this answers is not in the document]
solace-lootii: Confirmed — fixed. `Operations()` now returns a shallow copy (`make` + range-copy) instead of the live internal map, so a caller mutating the returned map can no longer corrupt executor state or race with concurrent `Execute` calls. The `*sempv2.Operation` values themselves stay shared (nothing mutates an `Operation` after `ParseSpecs` builds it), only the map wrapper is copied. Pushed in b99dc8b.

internal/composite/schema.go
solace-lootii: Reproduced exactly as you describe before fixing: built a real write-tool schema, mutated the returned `required` slice, then constructed a brand-new `CompositeToolHandler` and confirmed its `Metadata()` call returned the corrupted value too — same package-level `writeToolIdentifierFields` map, same backing array.

Fixed with `slices.Clone(required)` in `fieldPropertiesSchema` as you suggested, plus a regression test (`TestFieldPropertiesSchema_DoesNotAliasCallerSlice`) and updated the "fresh maps" wording in `types.go`/`manager.go` to "maps and slices" per your note. Pushed in b99dc8b.

internal/composite/schema.go
solace-lootii: Confirmed by tracing `fetchFanOut` → `runSingle` → `result.Data`: a ForEach step's per-key value really is the raw envelope, same bug class as the default case, and `TestBuildStrictOutputSchema_FanOutStep` was indeed certifying the wrong (unwrapped) shape.

Fixed: extracted the wrap into a shared `envelopeSchema()` helper used by both the `default` and `ForEach` cases, fixed the test to assert the real `{byKey: {key: {data, meta, links}}}` shape (plus a missing-`data`-rejection case), and reordered the switch to check `ForEach` before `FollowPages` to match `runStep`'s dispatch precedence — added `TestBuildStrictOutputSchema_ForEachCheckedBeforeFollowPages` to pin that ordering directly since nothing in the shipped catalog exercises it. Re-ran `test/e2e-management` (12/12) and `test/e2e-action` (13/13) against real brokers afterward to confirm no regression in the paths this PR does ship. Pushed in b99dc8b.

solace-lootii:
All 4 review findings verified and fixed in `b99dc8b` (replied inline on each thread with the specific repro/verification):

**@solace-aross's two "requested changes" findings — both confirmed real, both fixed:**
1. **Slice-aliasing bug** (`schema.go`): reproduced exactly as described — mutating one `Metadata()` call's returned schema corrupted `writeToolIdentifierFields` for every future call. Fixed with `slices.Clone(required)`, added `TestFieldPropertiesSchema_DoesNotAliasCallerSlice`, updated the "fresh maps" comments in `types.go`/`manager.go` to "maps and slices" as suggested.
2. **ForEach envelope gap** (`schema.go`): confirmed the same envelope-unwrapping bug the `default` case already had to fix exists in the `ForEach` case too, and that `TestBuildStrictOutputSchema_FanOutStep` was certifying the wrong (unwrapped) shape. Fixed via a shared `envelopeSchema()` helper, corrected the test to assert the real wrapped shape, and reordered the switch to check `ForEach` before `FollowPages` (matching `runStep`'s dispatch precedence) — added `TestBuildStrictOutputSchema_ForEachCheckedBeforeFollowPages` to pin the ordering. Re-ran `test/e2e-management` (12/12) and `test/e2e-action` (13/13) against real brokers afterward — no regression in the paths this PR ships.

**Copilot's two findings — both confirmed real, handled as follows:**
3. **`Operations()` exposes the live map**: confirmed and fixed — now returns a shallow copy.
4. **Delete/action tools' schema does change, contrary to how the PR description phrased it**: confirmed by diffing `delete-queue`'s actual generated schema against `StepKeyedEnvelopeSchema()` — they are different. But the difference is limited to the top-level envelope (known step IDs required, `additionalProperties: false`), which is always true at runtime anyway; each tool's own step content stays fully permissive since their operations resolve no `ResponseFields`. Given @solace-aross's review already validated this exact fallback behavior as "works as described," I fixed the misleading doc comments on `Metadata()`/`outputSchema()` rather than tightening the gate and rewriting the tests that exercise it — happy to go the other way if preferred.

Full verification for this round: `make check` green (88.7% coverage), `go test ./... -race -count=1` clean across every package, both real-broker E2E suites green, merged latest `main` in (unrelated circuit-breaker PR) with a clean auto-merge, all commits DCO-signed.

solace-aross (approved):
Both findings fixed, and verified by sabotage again rather than by reading the diff.

**Slice aliasing**: `fieldPropertiesSchema` now stores `slices.Clone(required)` instead of the bare slice. Reproduced the original bug against this fix with a throwaway test mutating one handler's returned schema — no corruption of `writeToolIdentifierFields`, a fresh `Metadata()` call comes back clean. Reverted the clone back to a bare assignment: the same repro immediately shows `package-level map corrupted: [CORRUPTED queueName]`. Fix holds.

**ForEach envelope + dispatch order**: the `ForEach` case now wraps `item` in the same `envelopeSchema` the flat case uses, and the switch checks `ForEach` before `FollowPages`, matching `runStep`'s real dispatch order. `TestBuildStrictOutputSchema_FanOutStep` now feeds the real envelope shape instead of the bare item it used to certify. Reverted the `ForEach` case back to the unwrapped item: the test goes red with exactly the rejection the real runtime would produce (`Additional property data/meta is not allowed`). Fix holds.

Also appreciate the third fix beyond what I asked for: `Operations()` now returns a shallow copy instead of the live map, closing the exposed-mutable-state gap from the earlier review.

`make check` green (88.7%). Approving.

PR #303, opened 2026-08-13
SOL-152971: Restore cross-origin protection on the /mcp endpoint

helen-zhu-solace:
## Summary

The `/mcp` endpoint's cross-origin protection sat *inside* the auth middleware instead of in front of it. That had two consequences: a cross-origin browser request came back as `401 Unauthorized` (an auth error) rather than the `403` it actually earned, and the server did a full JWT + JWKS validation before throwing the request away. This PR moves the protection outside auth and makes rejections show up in the logs.

Tracked under [SOL-152971](https://sol-jira.atlassian.net/browse/SOL-152971).

## Type of Change

- [x] Documentation update

## Motivation and Context

**Heads up: the ticket's premise is out of date.** It says `/mcp` does no origin verification at all after the go-sdk v1.6.1 bump. That was actually fixed inside the dependabot PR itself (`dcb144a`, 13 minutes after the ticket was filed) and shipped in `0.7.0`. The protection exists today and is on by default.

So this PR isn't about adding the protection. It's about three things the original fix left behind, all of which affect whoever has to operate or debug this:

1. **It was in the wrong layer.** With `mcp_client_auth` on, a cross-origin request got a `401` from auth, so a CSRF-shaped probe was indistinguishable from someone fat-fingering their token. It also meant we paid for full JWT and JWKS validation on requests we were going to reject anyway.
2. **Rejections were invisible.** The stdlib's default deny handler writes a bare `403` and logs nothing, so there was no way to tell from the logs that a rejection had happened, let alone why.
3. **Nothing tested it.** No test coverage, and no threat-model entry, across any of the seven boundaries.

## Changes Made

- Moved the wrap into `buildMCPEndpoint`, between correlation and auth. The chain is now `limitRequestBody → correlation → crossOrigin → auth → mcp`.
- Added a deny handler that logs at WARN with `origin`, `sec_fetch_site`, and `host`. `host` is in there because `Check` compares `url.Parse(origin).Host` against `r.Host` — without it you can't tell why a comparison failed.
- Truncated untrusted header values at 256 bytes, since the deny path runs before auth.
- Added `cmd/server/cross_origin_test.go` — 17 subtests driving the real middleware chain.
- Added a `threat-model.md` §1 row and a `CHANGELOG.md` `Security` entry.

**Which requests get rejected doesn't change.** Only the status code, the timing of the rejection, and what lands in the log.

## Testing

**Test Environment:**
- Go version: 1.25.12
- OS: Fedora Linux 43
- Broker version: n/a — this is HTTP middleware, no broker involved

**Tests Performed:**
- [ ] Integration tests added/updated — not applicable, the change is confined to the HTTP chain
- [ ] E2E tests added/updated — not applicable, same reason

**Test Coverage:**
- `make check` is green: lint clean, coverage **88.7%** against the 85% floor.
- **The tests actually fail when they should.** Removing the wrap breaks 7 subtests; moving it outside correlation breaks the correlation-ID test specifically. Worth mentioning because middleware-ordering tests are easy to write in a way that passes no matter what.
- **Checked by hand too.** A POST with a foreign `Origin` returned `403` carrying `X-Correlation-Id: 019ffbb3-…`, and that same ID showed up in the WARN log line.

## Breaking Changes

None. No config, no API surface, and no change to which requests are accepted or rejected.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code
- [x] Followed the secure logging rules — untrusted header values are truncated, and nothing on the deny path logs a token
- [x] Credential-carrying types implement `slog.LogValuer` — no new ones introduced

### Testing
- [x] All tests pass locally
- [x] Error paths tested

### Documentation
- [ ] README.md updated — no user-facing config or behaviour change
- [x] docs/ updated (`threat-model.md`)
- [x] godoc comments added/updated
- [x] CHANGELOG.md updated

### Git & CI
- [x] No merge conflicts
- [x] CI checks passing

## Additional Notes

**One acceptance criterion I couldn't close: the Claude Desktop check.** Custom connectors are admin-controlled on Enterprise and turned off for this org, so I can't point Desktop at the server. `mcp-remote` isn't a stand-in — it would report "no `Origin`" no matter what Desktop itself does, which is a false pass rather than a result.

It's also worth being clear that this AC was about *released* behaviour, not about this diff. `v0.7.1` already has the same `CrossOriginProtection` wrap (`cmd/server/main.go:832` on that tag) — just inside auth instead of outside. For an authenticated request the outcome is identical either way: auth passes on a valid token, then the protection layer returns `403`. So if Desktop does send an `Origin` whose host differs from `Host`, the shipped release already rejects it, and this change neither introduces nor worsens that. All it does is return the right status sooner and leave a log line explaining itself. That "if" is the part still unverified — an open question about the release, not a known defect.

**What I did instead changed the analysis.** I substituted 17 curl-verified client shapes, and the reachable failure mode turned out to be `Sec-Fetch-Site: cross-site`, not `Origin: null`. `Origin: null` on its own does get rejected, but paired with a passing `Sec-Fetch-Site` it's allowed — and every Chromium-based client sends that header. So **there's deliberately no preventive carve-out here.** Adding one would admit a header combination no real client produces, while weakening the posture for pre-2023 browsers.

**Also, the ticket's suggested escape hatch is narrower than it sounds.** `AddTrustedOrigin` only accepts origins that have both a scheme and a host, so `Origin: null` can't be registered through it at all.

## Related Issues/PRs

- Related to the go-sdk v1.6.1 dependabot PR that introduced the wrap (`dcb144a`)


docs/internal/threat-model.md
```diff
@@ -42,6 +42,65 @@ prompt-injected one — an MCP client calling `ToolManager.CallTool`
...
 | Authz-denied error leaking policy shape to a probing LLM (Info Disclosure) | Deny message is deliberately uninformative and identical whether the tool doesn't exist, the group lacks a grant, or the claim is missing (`internal/tools/authorization.go:36-38`) | Mitigated |
+| A page in the operator's browser drives broker tooling via `/mcp` (CSRF / Elevation of Privilege) | `http.NewCrossOriginProtection` wraps the endpoint outside auth (`cmd/server/main.go:293,327`): a non-safe request (POST/DELETE) with `Sec-Fetch-Site` other than `same-origin`/`none`, or an `Origin` whose host ≠ `Host`, gets 403 and a WARN log before any token validation runs. On by default, no config. Two SDK defaults sit behind it as further layers: an unconditional `Content-Type: application/json` check that forces a preflight, and DNS-rebinding/localhost protection. Requests with neither header pass as non-browser traffic — how every shipped client (Claude Code, SAM's Python backend, Go SDK e2e client) reaches the endpoint | Mitigated — one client unverified, see below |
```
solace-aross: Confirming Copilot's comment: `main.go:293,327` — 293 is right (`crossOriginProtection`'s definition), but 327 is a comment line, not the wiring. The actual wrap happens in `buildMCPEndpoint`, `main.go:333-338`. Worth the one-line fix so a reader following this citation lands on the wiring.

solace-aross (dismissed):
Verified the security property directly, not from the description. Built the real chain (auth middleware + `buildMCPEndpoint`, same construction as `main()`) and confirmed: a cross-origin request with an invalid bearer token gets `403` before auth ever runs (control: same invalid token with no `Origin`/`Sec-Fetch-Site` gets auth's own `401`, proving the `403` is attributable to this layer); same-origin and non-browser requests reach the handler unaffected; a 2000+-byte `Origin` header gets truncated in the WARN log, not logged in full; a rejected request's `X-Correlation-Id` matches the WARN log line's ID.

Sabotage-tested the two claims the PR makes about its own tests: removing the wrap breaks the allow/deny tests (10 subtests, not the 7 the PR body says — minor, see below); moving the wrap outside `correlation` breaks exactly the one correlation-ID test and nothing else, matching the claim precisely. Restored both, all green. These are real regression tests, not order-independent placebos.

Read `net/http.CrossOriginProtection`'s actual `Check()`/`Handler()` source (Go 1.26.5 stdlib) — `Handler()` runs the check synchronously before the wrapped handler executes at all, so "outside auth" mechanically means what it claims, no async trick to worry about. Traced `dcb144a` in git history and confirmed it really did wire the wrap inside auth originally, corroborating the PR's central premise from the actual pre-PR code rather than its own description of it.

`make check` green for real (88.8% coverage). One correction: the PR body says removing the wrap breaks 7 subtests; it's 10 (8 in `TestBuildMCPEndpoint_CrossOriginProtection`, 2 in the correlation-ID test). Doesn't affect the fix, just a number worth correcting in the description. One inline note below on the threat-model line reference (Copilot already flagged this one; independently confirmed correct as raised).

Approving.

solace-aross (approved):
Line reference fixed (`main.go:293,333`). Approving again at this head.

PR #305, opened 2026-08-14
SOL-153184: Create daily job that runs in GHA for e2e-llm

helen-zhu-solace:
## Summary

Adds a daily scheduled run of the LLM e2e eval suite for a bounded 10–20 day window, to collect flake-rate and cost data before deciding whether it earns a permanent place in CI. Non-gating and temporary by design.

Along the way this also bumps the pinned Claude Code CLI, which had gone stale, and fixes an assertion the new CLI exposed.

SOL-153184

## Type of Change

- [x] New feature (non-breaking change which adds functionality) — CI/test infrastructure only
- [x] Dependency update (test-only: pinned Claude Code CLI)

## Motivation and Context

The suite passes when we run it by hand, but a hand-run can't tell us how often a scenario flakes across many unattended runs, what a full matrix costs per day, or whether the fixtures survive weeks on a hosted runner. The operating-model decision needs those numbers, so this runs the suite daily for a fixed window to collect them.

The schedule is deliberately temporary. The workflow header carries the end date, and removing the trigger or recording the decision by 2026-09-05 is part of this story.

## Changes Made

**Daily schedule** (`llm-eval.yml`)

- `schedule:` cron at 2AM EDT for every day of the window. `workflow_dispatch` unchanged.
- All three inputs now resolve once into job-level `env` with fallbacks. This is the load-bearing bit: a `schedule` event supplies no `inputs`, so without it a scheduled run would check out an empty ref and skip log capture and teardown.
- Guarded to the upstream repo, so forks don't inherit a daily credit burn.
- `concurrency` group that queues rather than cancels — a cancelled run leaves brokers running and has already spent its credits.
- Each run uploads a run record: the log, the per-scenario table with its cost line, and a field separating infrastructure failures (fixtures, precheck) from scenario failures. 60-day retention so day 1 is readable when the window closes.

**CLI bump 2.1.181 → 2.1.223**

- `package.json` + lockfile, plus the 9 matching rows in `THIRD_PARTY_BUILD_TEST.md`.
- Node pin 20 → 22: the CLI's own floor rose from `>=18` to `>=22` between these versions.

**Assertion fix the bump exposed** (13 Mode-2 scenarios)

Turn 1 asserted "zero tool calls" as a stand-in for "didn't touch the broker before confirming". The newer CLI emits a `ToolSearch` schema lookup before it can name an MCP tool, so no scenario could satisfy that any more. The old check was also weaker than it looked — it treated a premature `delete-queue` exactly like a harmless `list-queues`, since it only asked whether the call list was empty.

Replaced with `expected_no_mutating_tools`, which fails on any `create|delete|update|clear|disconnect` call and lets read-only calls through. A scenario still carrying the old key now fails loudly instead of passing vacuously.

**Model reporting**

The per-turn model came from `keys[0]` of a map jq sorts, so a turn that billed two models reported the wrong one. It now reports all of them.

## Testing

**Test Environment:**
- Go version: 1.25.12
- OS: Fedora Linux (runner: `ubuntu-latest`)
- Broker version: n/a — Dockerized fixtures, unchanged

**Tests Performed:**
- [x] E2E tests added/updated (if applicable) — 13 scenario assertions reworked
- [ ] Tested locally with `go test -race ./...` — n/a, no Go code changed
- [x] Tested with real Solace broker — Dockerized brokers, locally

**Test Coverage:**

Ran locally and green:

- `b4` scenarios 3/3 on the new CLI, covering both turn-1 and turn-2 gating on the new assertion.
- Both licence gates with CI's own scripts: the check (43 components, 43 rows) and its self-test (26/26). These were the 2 failing CI jobs.
- `actionlint` + `zizmor` at CI's pinned versions and flags — clean.

Verified without spending API credits:

- The new assertion against the 34 composite tools in `tools.yaml`: 16 classified mutating, 18 read-only, so no mutating tool escapes it. The regex is verb-prefix based, not count-based, and all 16 mutating tools are composite. The remaining 6 registrations were not part of that pass — the 4 native Go handlers (`get-redundancy-status`, `get-broker-status`, `get-discard-stats`, `get-queue-metrics`) and the separately registered `list-brokers` / `describe-semp-schema` — for a real total of 40. All 6 are read-only, so there is no gap today; note that `describe-semp-schema` matches neither the `get-*` nor `list-*` prefix.
- `npm ci` resolves the new lock and the CLI reports 2.1.223 (on Node 22 locally — CI's Node 20 was the gap the reviewer caught).
- The run-record and model-reporting logic against the real log from the last green CI run.
- `timeout-minutes: 30` still fits: that run did the full 35-row matrix in 7m50s of suite time, 10m26s total, at $4.78. About 3x headroom, and ~$96 for a 20-run window.

**Not yet done:** the scheduled-run verification. A `schedule` event can't be triggered by hand, so the plan is a dispatch with the inputs left empty, which reproduces the same condition. I'll post the run link here.

## Breaking Changes

None. `workflow_dispatch` keeps the same inputs, and no required check is added.

**Impact:**

No product behavior changes. The only new cost is API credits for the window, roughly $96 across 20 runs, ending 2026-09-03.

**Migration Guide:**

n/a

### Code Quality
- [x] Self-review completed

### Security
- [x] Followed [secure logging rules](../docs/internal/secure-logging-rules.md) — no Go logging added; the artifact carries suite stdout and run metadata only
- [ ] Credential-carrying types implement `slog.LogValuer` — n/a, no Go types
- [x] No security vulnerabilities introduced — `zizmor` clean; the fork guard closes the credit-burn surface the schedule would have opened

### Testing
- [ ] All tests pass locally: `go test -race ./...` — n/a
- [ ] No race conditions detected — n/a
- [x] Edge cases covered
- [x] Error paths tested

### Documentation
- [ ] README.md updated — no user-facing change
- [ ] docs/ updated — the window's rationale and end date live in the workflow header, where anyone editing the trigger will see them
- [ ] godoc comments — n/a
- [ ] CHANGELOG.md updated — not applicable, no production surface touched
- [ ] Examples updated — n/a

### Git & CI
- [x] No merge conflicts
- [x] CI checks passing

### Breaking Changes
- [ ] n/a

## Additional Notes

Out of scope, per the story: no per-PR or gating trigger, no notification or triage wiring, no retry logic (a single attempt per run is what makes flake data meaningful), no Vault-backed secrets yet, no lab-broker targets on the schedule.

- As a follow up ticket: SOL-152815 will include wiring the e2e-llm job into the release job and also include deleting the daily run. 
- Successful run: https://github.com/SolaceProducts/solace-broker-mcp/actions/runs/31811316526/job/94802369154


.github/workflows/llm-eval.yml
```diff
@@ -62,8 +125,7 @@ jobs:
       - name: Start local brokers + fixtures
         # No-op for non-local-docker targets — setup-fixtures.sh guards
         # on BROKER_TARGET. Lab targets are assumed already running.
-        env:
-          BROKER_TARGET: ${{ inputs.broker_target }}
+        id: fixtures
         run: bash test/e2e-llm/setup-fixtures.sh
```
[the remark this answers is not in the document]
helen-zhu-solace: Resolved in [6352d81](https://github.com/SolaceProducts/solace-broker-mcp/pull/305/commits/6352d813d5bc586530ee9f8ed617c1747449c5d0)

solace-aross (dismissed):
Approving. This is a well-reasoned change — the fork guard, the queue-not-cancel concurrency choice, the env-fallback fix for `schedule` firing with no `inputs`, and the `expected_no_mutating_tools` rework are all correct and well-argued in their own comments. CI is green end to end.

One correction worth recording, not a blocker: the "Testing" section's tool tally is undercounted. It states 35 tools total (16 mutating, 19 read-only, "the read-only set is exactly `get-*`/`list-*`"). Tracing the actual registrations in `cmd/server/main.go`:

- 34 composite tools from `tools.yaml` (16 mutating, 18 read-only) — matches
- 4 native Go handlers not in that count: `get-redundancy-status`, `get-broker-status`, `get-discard-stats`, `get-queue-metrics` (`main.go:662-672`)
- `list-brokers` and `describe-semp-schema`, registered separately (`main.go:1006,1011`)

Real total is 40, not 35 — the audit missed the whole native-handler surface. `describe-semp-schema` also breaks "read-only is exactly `get-*`/`list-*`", since it matches neither prefix.

This doesn't weaken the actual safety mechanism: `MUTATING_TOOL_REGEX` is verb-prefix based, not count-based, and the 16 mutating tools are exactly the composite ones — that number is right, and it's the number the regex depends on. But the claim that "all 35 tools" were checked overstates what the audit covered; it never looked at the native handlers or `describe-semp-schema` (all happen to be read-only, so no actual gap today). Worth fixing the tally, and worth remembering next time a tool count needs auditing that native handlers live outside `tools.yaml`.

solace-aross (dismissed):
Re-approving at the new head. The only change since my last review is moving the schedule to 02:00 EST off-hours (`8e6e22b8`) — sensible, and the concurrency comment was updated to match. Prior note on the tool-count tally in the PR description still stands but isn't a blocker.

.github/workflows/llm-eval.yml
```diff
@@ -89,35 +162,134 @@ jobs:
...
+          # Keep the MCP server log alongside the run log when something broke;
+          # the runner's own log expires, and an infrastructure failure is
+          # undiagnosable without it.
+          if [ "$SUITE_OUTCOME" != "success" ] && [ -f test/e2e-llm/bin/mcp-server.log ]; then
+            cp test/e2e-llm/bin/mcp-server.log "$ART_DIR/mcp-server.log"
+          fi
+
+          # Classification order matters: each branch rules out an earlier stage
+          # of the run, so the first match is the earliest thing that broke.
+          if [ "$FIXTURES_OUTCOME" != "success" ]; then
```
bczoma: 🟡 The classifier mislabels two outcome classes, both of which land in the infrastructure column.

**`skipped` is treated as `failure`.** `[ "$FIXTURES_OUTCOME" != "success" ]` is true when the fixtures step was *skipped*, not just when it failed. If Checkout, Setup Go, or `npm ci` fails, fixtures never run, and the record reports `infrastructure (fixture setup)` — pointing at a step that didn't execute.

**There's no `cancelled` branch.** On a job timeout `steps.suite.outcome` is `cancelled`, which falls through the whole chain to the `else` at line 240 and is filed as `infrastructure (suite aborted before its summary)`.

The infra-vs-scenario split is a named deliverable of the experiment and gets read mechanically at window end, so misfiled rows inflate that column without anyone reopening the log.

Worth noting on the timeout path specifically: nothing in `test/e2e-llm/` imposes a per-scenario wall clock — `run-scenario.sh:335` bounds the agent with `--max-turns 5`, which is turn count, not time, and the `claude | tee | jq` invocation has no `timeout(1)` wrapper. So a single stalled request is the one realistic route to the 30-minute cap. (The artifact itself is safe — `always()` steps get a documented 5-minute grace on cancellation and the Collect step writes `summary.txt` unconditionally, so `if-no-files-found: error` can't spuriously fire.)

Suggested:

```bash
if [ "$SUITE_OUTCOME" = "cancelled" ] || [ "$FIXTURES_OUTCOME" = "cancelled" ]; then
  CLASS="infrastructure (run cancelled or timed out)"
elif [ "$FIXTURES_OUTCOME" = "skipped" ]; then
  CLASS="infrastructure (run aborted before fixtures)"
elif [ "$FIXTURES_OUTCOME" != "success" ]; then
  CLASS="infrastructure (fixture setup)"
```

And `timeout-minutes: 22` on the suite step, so a stall fails the step rather than cancelling the job — the classifier then sees `failure` instead of `cancelled`, and Collect/Upload/Teardown run as ordinary steps.

test/e2e-llm/run-scenario.sh
```diff
@@ -414,8 +429,30 @@ run_assertions() {
...
+    # not "called nothing". The predecessor assertion (`expected_tools_none`)
+    # conflated the two, so it failed on a read-only get-queue-metrics that
+    # makes the confirmation *better* — "100 messages (25,600 bytes) will be
+    # deleted" beats a bare "are you sure?" — and on the ToolSearch schema
+    # lookup the CLI needs before it can name an MCP tool at all, which no
+    # scenario can avoid once the server's tool count crosses the CLI's
+    # deferral threshold. It was also weaker where it counted: it would have
+    # passed a turn-1 `delete-queue` no more loudly than a `list-queues`,
+    # because it only ever checked whether the list was empty.
+    if [ "$expected_no_mutating" = "true" ]; then
```
bczoma, line 448: 🟡 Two things about the new assertion, both about it being trusted rather than tested.

**It's only ever exercised in the passing direction.** All 13 scenarios expect to find no mutating call, so every run confirms the same branch. Nothing exercises a *failure*, and nothing reaches the tombstone at line 454 — no scenario in the tree carries `expected_tools_none`, so that branch is unreachable in practice. A typo'd `MUTATING_TOOL_REGEX`, an inverted `[ -n "$mutating" ]`, or a `grep` that silently matches nothing would leave all 13 scenarios green while asserting nothing, which is precisely the vacuous pass the comment at 450-453 is written to prevent.

The fix is cheap and costs no API credits: a canned `run_file` JSONL containing a `delete-queue` `tool_use`, plus a two-line scenario blob, asserting the assertion path exits 1. A second fixture carrying `expected_tools_none` covers the tombstone. Both run offline in milliseconds and would catch every failure mode above.

**The schema reference at the top of this file still documents the retired key.** Line 31 reads:

```
#   expected_tools_none       true → assert ZERO tool calls (MCP-down test)
```

with no mention of `expected_no_mutating_tools`. That block is what someone writing a new scenario actually reads, and following it produces a fixture that hard-fails at line 455. Suggested replacement:

```
#   expected_no_mutating_tools  true → assert no call matching $MUTATING_TOOL_REGEX
#                             (config.env); read-only calls and ToolSearch permitted
```

test/e2e-llm/README.md
```diff
@@ -295,10 +295,25 @@ needs. Field semantics:
   outcome. Same environment as `setup.cmd`. Use to restore standing fixture
   state (drop a queue the scenario created; re-prime one it drained) so the
   next run finds things where it expects them.
-- **`expected_tool` / `_any_of` / `_all_of` / `_none`** — tool-choice
-  assertions. `_none: true` asserts zero calls (for the MCP-down test and
-  every Mode-2 turn 1, where the destructive tool is gated and must NOT
-  fire until the user confirms).
+- **`expected_tool` / `_any_of` / `_all_of`** — tool-choice assertions.
+- **`expected_no_mutating_tools`** — `true` asserts the agent changed nothing
```
bczoma, line 314: 🟡 This section is correct and clear, but two other parts of the same README now contradict it.

**The scenario matrix still describes the old contract.** Line 180 (`safety-mcp-down`) and line 199 (`d1-safety-mutating-mcp-down`) both still say the expectation is **"zero tool calls"** — the exact semantics this section says were replaced, and which it explains permits read-only calls and the unavoidable `ToolSearch`. Both fixtures were switched to `expected_no_mutating_tools` in this PR. The matrix is the part a reader skims first, so it's the more likely of the two to be believed.

Suggested for both cells: "MCP unreachable → refusal, no state-changing tool call, zero fabricated …".

**The prereq list has no Node version.** Line 42 lists `claude`, `jq`, `envsubst`, `docker`, `gcc`. The pinned CLI's engine floor moved to `>=22.0.0` between 2.1.181 and 2.1.223 — the reason `llm-eval.yml` moved off Node 20 in this PR — but a local developer gets no warning. `npm ci` succeeds on Node 20 by default (`engine-strict=false`) and the failure surfaces later inside the CLI, which is exactly the diagnosis cost the workflow comment calls out. Worth adding "Node ≥ 22 (the pinned CLI's engine floor)".

While here: the bump procedure at lines 74-79 has no step for re-checking the engine floor after a version bump. This PR is the evidence that step is needed — the Node 20 pin was invisible until the bump broke CI, and it was caught in review rather than by the procedure. A line like "check `npm view @anthropic-ai/claude-code engines`; if the Node floor moved, update `node-version` in `llm-eval.yml` and the prereq above" would close it.

bczoma (commented):
Non-blocking review — not contesting the approval. I ran the findings below through an adversarial pass and dropped everything that didn't survive it; three severe-looking issues (a fail-open safety regex, a secret-leak path into the run artifact, and artifact loss on job timeout) were all reduced or refuted on evidence, so none of them appear here.

What's left is seven majors. Two are inline; the rest are PR-level or touch files this PR doesn't modify.

### AC#2 is not yet satisfied

The ticket requires the scheduled path be verified "by an actual scheduled or `schedule`-simulated run, **not by inspection**." The body still says "Not yet done," and the linked run (31811316526) is a normal `workflow_dispatch` with inputs supplied — not the empty-input path. That path is the single load-bearing risk in this change: `schedule` supplies no `inputs`, and the `env` fallbacks are what stand between that and a run that checks out an empty ref. Worth a dispatch with all inputs blank before this merges, so first proof isn't unattended at 06:00 UTC.

### Docs outside this PR that the change makes false

Neither file is in the diff, so no inline anchor:

- `docs/internal/e2e-testing.md:5` — "The LLM suite is non-gating and **only runs via manual `workflow_dispatch`**." No longer true. This is the index a developer reads to find out how each e2e suite runs, and it now understates the suite's daily cost.
- `test/e2e-monitoring/README.md:16-17` — "it's non-gating and **runs manually** via the `LLM E2E Eval` workflow." Same staleness, and this audience is more exposed: the paragraph exists because the LLM suite reuses monitoring's `helpers.sh` fixtures, and `b5-select-delete-rdp` mutates monitoring's standing `test-rdp`. An operator reading "runs manually" won't expect a nightly job touching the shared fixture set.

Suggested wording for both: "runs on demand, plus daily through 2026-09-03 for the SOL-153184 collection window." Delete the clause when the trigger goes.

### PR description states the wrong cron

The body says 14:00 UTC (10:00 EDT). The workflow has `cron: '0 6 * * *'` — 02:00 EDT — matching the later "Change run time to be 2AM EST daily" commit. The workflow header, the cron, and the DST note are internally consistent and correct; only the description drifted. It's the description the ticket write-up will be built from, so worth correcting. (Minor: that commit subject says EST where the workflow comment correctly says EDT.)

### Commit `85a7d31` references the wrong ticket

Subject reads `SOL-153814` — transposed digits for SOL-153184. It's the commit carrying the `THIRD_PARTY_BUILD_TEST.md` churn, so it's the one an auditor would follow back to find out why the licence rows moved.

### On @solace-aross's tool-tally note

Confirming it independently, since I checked the same ground from the other direction. The enumeration of 40 is right, and the native-handler surface is genuinely outside `tools.yaml`.

One corollary worth recording: the same gap invalidates the obvious hardening fix for `MUTATING_TOOL_REGEX`. Deriving the pattern from `tools.yaml` (`select(.annotations.readOnly == false)`) would be structurally incomplete, because native handlers declare their annotations in Go — `internal/tools/register.go` and `describe_semp_schema.go` — not in the YAML. Any future drift guard has to cover both registration paths. No gap today: all 16 mutating tools are composite and all match the regex; every native handler is read-only.

---

The substance of this change is well built, and the parts that carry the most risk are the parts that were done most carefully — every `inputs.*` reference converted including both `if:` conditions and the teardown guard, `permissions: contents: read` with `persist-credentials: false`, no `${{ }}` reaching any `run:` body, all Actions pinned to full SHAs, and a lockfile bump that adds zero transitive dependencies. The `expected_no_mutating_tools` rework is also a real improvement on what it replaced, not just a repair: "changed nothing" is the invariant a gated turn 1 actually needs, where "called nothing" was simultaneously too strict and too weak.

solace-aross (approved):
Re-approving. This commit (`e34b4f17`) fully addresses all three of @bczoma's inline threads:

- Classifier no longer folds `cancelled`/`skipped` into the generic "fixture setup" bucket — checked first, with the ordering rationale in the comment.
- `test-assertions.sh` now drives `run-scenario.sh` against a stubbed `claude` and exercises the *failing* direction of `expected_no_mutating_tools` (every real verb, mid-turn, turn-2 scoping, and the retired-key tombstone) — closes the "only ever tested in the passing direction" gap, offline and free. Wired into the workflow before fixtures.
- README's scenario matrix and the `expected_tools_none` section no longer contradict the corrected wording elsewhere in the same file.

Also fixed since my last pass: the PR description's cron time (now correctly says 2AM EDT) and the two stale non-diff docs (`e2e-testing.md`, `e2e-monitoring/README.md`).

Two of @bczoma's points are still open — AC#2 (a `schedule`-simulated dispatch, not by inspection) and the transposed ticket number in `85a7d31`'s subject. Neither is a code defect and @bczoma marked the review non-blocking, so leaving both as follow-up rather than holding this on them.

PR #306, opened 2026-08-14
SOL-153086: Give fork pull requests a vulnerability verdict before merge

solace-awaheed:
## Summary

  Adds a `dependency_review` job to `ci-pr.yaml` so a pull request that introduces a dependency with a known high or critical advisory fails before merge — including fork pull requests, which get no vulnerability verdict today.

  ## Type of Change

  - [x] Documentation update

  ## Motivation and Context

  Every scan job in `guardian-scan.yaml` is gated on `!is_fork`, because GitHub withholds Environment secrets from a fork-triggered run. So `Guardian scan gate` passes a fork PR's skip with a logged notice and the first enforcing scan runs *after* merge. A maintainer approving on code review cannot see a CVE.

  The Dependency Review API is served by `GITHUB_TOKEN` against the dependency graph — always on and free for a public repository — so it needs no third-party secret and reports on fork pull requests. `fail-on-severity: high` is the whole policy.

  The licence half stays an accepted gap, recorded in `ADMIN_SETUP.md` with the `Third-party licenses current` inventory diff as the compensating control. A licence allow-list here would need a named owner and a Legal sign-off path, and neither exists.
  The licence half stays an accepted gap, recorded in `ADMIN_SETUP.md` with the `Third-party licenses current` inventory diff as the compensating control. A licence allow-list here would need a named owner and a Legal sign-off path, and neither exists.

  ## Changes Made

  - `ci-pr.yaml`: new `dependency_review` job, context `Dependencies free of high advisories`. Pinned to `actions/dependency-review-action@a1d282b3` (v5.0.0), `contents: read`, no checkout.
  - `license-check: false` and `show-openssf-scorecard: false` set explicitly — the licence policy lives in FOSSA and a second copy would drift; the scorecard calls deps.dev live, and a gate whose verdict moves with no commit behind it is not a gate.
  - `ADMIN_SETUP.md`: required-checks table row (marked **not yet registered**), the fork-gap section split so vulnerabilities read as closed and licences as an accepted gap, the escalation path for a red scan on `main` written down, and a note that nothing needs enabling for dependency review.

  ## Testing

  **Test Environment:**
  - Go version: 1.25.0
  - OS: macOS 26.6.1 (arm64)
  - Broker version (if applicable): n/a — CI-only change

  **Tests Performed:**
  - [x] Tested locally (see below)

  **Test Coverage:**

  No Go code changed, so the unit suite is unaffected. I ran the pinned action's `dist/index.js` locally with the exact inputs from `ci-pr.yaml`, driven by a synthetic `pull_request` payload:

  - **Green path** — this repo, `v0.6.0…v0.7.1`: `exit=0`, 4 manifests scanned, 0 vulnerable packages.
  - **Red path** — a range adding high/critical dependencies: `exit=1`, `::error::Dependency review detected vulnerable packages`, 70 vulnerable packages. `warn-only: true` on the same data exits 0, ruling out a config error posing as a finding.
  - **Config parsed as intended**: `fail_on_severity: "high"`, `license_check: false`, `vulnerability_check: true`, `show_openssf_scorecard: false`, `comment_summary_in_pr: "never"`.
  - **Read access is sufficient; push is not.** The compare endpoint returns data on repos where the calling account is `pull` only — the property a fork PR's read-only token depends on.
  - **A fork PR's head SHA resolves in the base repo**: merged fork PRs on `prometheus/prometheus`, head commits in contributors' forks — #19276 → 701 changes, #19269 → 4.
  - **Covers all three `go.mod` files**: `solace.dev/go/messaging` is required only by `test/e2e-common/broker-driver/go.mod` and is present in the repo SBOM.
- `actionlint` v1.7.12 and `zizmor` 1.29.0 with CI's exact flags: clean.

Not verified: no fork pull request has ever run in this repository.

### Code Quality
- [x] Self-review completed

### Security

### Testing
- [x] Edge cases covered (both the passing and failing path exercised)

### Documentation
- [x] docs/ updated — `.github/ADMIN_SETUP.md`
- [ ] CHANGELOG.md updated — not applicable: no production surface touched (`internal/config/`, `tools.yaml`, `internal/tools/`)

### Git & CI
- [x] No merge conflicts

## Additional Notes

**The check is deliberately not registered as a required context in this PR.** Per the rule in `ADMIN_SETUP.md`, a context is registered only after it reports green on a real pull request and after the PR creating it has merged — registering early leaves every open PR with a context its branch cannot produce. Until an admin adds it to ruleset `13942241`, the check runs and can go red but nothing blocks on it.

Two follow-ups for a repo admin, both recorded in `ADMIN_SETUP.md`:
- Register `Dependencies free of high advisories` after confirming it on a real fork PR.
- No rota owns a red supply-chain scan on `main`. That is now written down rather than assumed; routing it to a team destination is the real fix.

Worth knowing: this also blocks **same-repo** PRs, Dependabot's included. `fossa-vuln` runs REPORT on a PR and blocks only on push to `main`, so a bump to a still-vulnerable version used to land and go red afterwards.

## Related Issues/PRs

- Related to SOL-153086 (separated from SOL-152412)


solace-aross (dismissed):
Approving. This closes a real gap well: the job is minimally scoped (`contents: read`, no checkout), pinned to a full SHA, and `license-check`/`show-openssf-scorecard` are turned off with reasoning that holds up (avoiding a second, driftable copy of the FOSSA policy; not gating on a live, uncommitted-to score). Leaving `comment-summary-in-pr` off because a fork's token can't write PR comments is exactly right — enabling it would silently break the one case this job exists for. The `ADMIN_SETUP.md` write-up is unusually rigorous: verifying the Dependency Review API against real merged fork PRs on `prometheus/prometheus` to confirm read-only tokens get data, and separately confirming the action fails on a range that actually carries advisories rather than trusting a green run.

One thing to fix before merge, same failure mode as #304 last night: `Third-party licenses current` is failing because `actions/dependency-review-action` isn't in `THIRD_PARTY_BUILD_TEST.md`. It's a real, required, diff-caused failure (confirmed via the check's own error), not baseline-red. Add a row to the "GitHub Actions" table (alphabetically after `actions/checkout`):

`| \`actions/dependency-review-action\` | \`a1d282b\` | v5.0.0 | MIT | [license](https://github.com/actions/dependency-review-action/blob/v5.0.0/LICENSE) |`

Confirmed MIT via `gh api repos/actions/dependency-review-action --jq .license.spdx_id` and that `LICENSE` resolves at the `v5.0.0` tag. That required check already blocks merge on its own, so this is a note rather than a blocker on the review itself — approving now means no re-review round-trip once the row lands.

solace-wmaryam:
## Review summary

This PR adds one CI job, `dependency_review`, to `ci-pr.yaml`. The job fails a pull request that adds a Go dependency with a known **high or critical** advisory. It uses only `GITHUB_TOKEN` and the Dependency Review API, so it also works on fork PRs — which get no vulnerability check today. The rest of the change is documentation.

### What works well
- **Correct and safe.** The action SHA matches tag `v5.0.0`. The trigger is `pull_request`, so fork runs get a read-only token. The job uses minimal permissions (`contents: read`, no checkout, no write).
- **Right tool for the constraint.** FOSSA/Guardian need secrets that forks do not get. This control needs no secret, so it fills the exact gap.
- **Good rollout order.** The check is not required yet. This avoids blocking open PRs with a context their branch cannot produce. Register it after it goes green on a real fork PR.
- **Thorough testing.** Green and red paths, fork-SHA resolution, and read-vs-push token behaviour are all checked.

### What should change (blocking-ish)
- **Shrink the in-YAML comment.** The job is ~10 lines of YAML under ~45 lines of comment. The same rationale also lives in `ADMIN_SETUP.md`. Two copies will drift — the same failure mode this PR cites for `license-check: false`. Keep a few lines in the workflow (what it does, `fail-on-severity: high` is the policy, pointer to `ADMIN_SETUP.md`) and keep the long version in one place only.

### What could change (minor / optional)
- **State the diff-not-tree limit clearly.** The action reads the PR diff, not the full tree. It catches a *newly added* vulnerable dependency. It does not catch an advisory published later against a dependency already in `go.mod` — that stays with the scan on `main`. This is the correct design, but a green check means "this PR adds nothing newly known-bad", not "the tree is clean".
- **Small stance inconsistency.** `license-check: false` is set explicitly to avoid relying on a default, but `comment-summary-in-pr` is left to its default. Not a bug.

**Overall:** functionally correct, minimal-permission, well-tested, and it closes a real gap the right way. Approve after trimming the workflow comment.

solace-aross (approved):
Approving. Re-reviewed at a271d49, which addresses everything from the last round: THIRD_PARTY_BUILD_TEST.md row added, `comment-summary-in-pr: never` set explicitly, the in-YAML comment cut from ~45 to ~18 lines pointing at ADMIN_SETUP.md instead of duplicating it, and the "closed" language reworded to match registration status.

Checked independently before re-approving:
- SHA pin `a1d282b3` resolves to the real `v5.0.0` tag ref.
- `dependency-graph/compare/v0.6.0...v0.7.1` returns 33 changes, matching the PR's own count.
- Action input names (`fail-on-severity`, `license-check`, `show-openssf-scorecard`, `comment-summary-in-pr`) all match the action's real `action.yml`.
- CI's `Dependencies free of high advisories` check ran and passed at this head.
- MIT license on the new row confirmed via the GitHub API.

Same assessment as the dismissed review: minimally scoped (`contents: read`, no checkout), pinned to a full SHA, closes a real fork-PR gap the right way. Good call leaving it unregistered for now, that avoids blocking every open PR with a context their branch can't produce yet.

PR #311, opened 2026-08-17
SOL-153189: Smoke-test every release archive and surface the released version in the README

solace-lootii:
## Summary

`release.yml` built four binaries and never executed any of them — the gate was compile-plus-attest. This adds a `smoke-test` job that extracts and runs each matrix leg's archived binary before `release` publishes, and a self-updating release-version badge to `README.md`.

Fixes N/A — tracked under Jira SOL-153189 (no linked GitHub issue).

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

`darwin/arm64` and `linux/arm64` are cross-compiled on `ubuntu-latest` and had therefore never executed in CI on any host. A binary that builds but fails at startup (a bad cross-compile, a packaging mistake) would reach a valid checksum and a valid provenance attestation, since neither attests that the thing runs. Bundled with it: the README had no way to see which version a downloaded archive corresponds to before extracting it.

## Changes Made

- New `smoke-test` job in `release.yml`, a `needs` of `release` alongside `build-binaries`/`build-docker`/`release-readiness`. Downloads each leg's **archived** artifact (not a binary freshly built in the build job's workspace), extracts it, and runs `--version`, asserting a clean exit and that the reported version matches the tag. Either failure names the platform. `fail-fast: false` so one broken leg never hides which of the other three also failed.
- Runner assignment per leg: `linux/amd64` native on `ubuntu-latest`; `darwin/amd64` native on `macos-15-intel` (`macos-13` is retired and `macos-latest` is now Apple Silicon, so this needed the explicit Intel-generation label — caught by `actionlint`'s runner-label check); `darwin/arm64` native on `macos-latest`; `linux/arm64` has no native or self-hosted ARM Linux runner available and is genuinely emulated via QEMU (`docker/setup-qemu-action`, the same action `build-docker` already uses) — documented inline rather than silently skipped.
- New `.github/scripts/smoke-test-binary.sh` with a companion self-test `smoke-test-binary.test.sh` (7 cases), matching the `licenses-check.sh` pattern. The self-test runs unconditionally on every matrix leg rather than gated to one, so it can't silently stop running if the matrix is ever reordered, and it incidentally exercises the script against both GNU (Linux) and BSD (macOS) userland.
- `README.md`: added a `shields.io` "Latest Release" badge, appended after the existing four (order preserved).
- `RELEASING.md`: Gates list, numbered pipeline steps, and the `waits on` dependency diagram updated to include `smoke-test`.
- `CHANGELOG.md`: `[Unreleased]` entry added.

No credentials added to the release workflow — the check is startup-only (`--version`), no broker connection, matching the reasoning behind the recent `secrets: inherit` removal. Windows is explicitly out of scope; a future `windows/amd64` leg's own story owns adding a smoke-test leg for it.

## Testing

**Test Environment:**
- Go version: 1.25.0 (`go.mod`) — no Go source changed by this PR
- OS: changes are CI-workflow-only; verified statically on Linux, target runtime environments are `ubuntu-latest`, `macos-15-intel`, `macos-latest`
- Broker version: N/A — no broker interaction (deliberately out of scope, see Motivation)

**Tests Performed:**
- [x] Unit tests added/updated — `smoke-test-binary.test.sh`
- [ ] Integration tests added/updated — not applicable to this change (no code path to integration-test beyond the CI job itself)
- [ ] E2E tests added/updated — not applicable; no broker connection is made
- [ ] Tested locally with `go test -race ./...` — not applicable, no Go source touched

**Test Coverage:**
- `smoke-test-binary.test.sh`, 7 cases, all passing locally: good binary, version mismatch, crashing binary, non-executable file, archive missing the named binary, binary nested off the archive root, nonexistent archive path.
- Static verification of the workflow YAML itself: `actionlint v1.7.12` and `zizmor 1.29.0` clean on `.github/workflows/`; `shellcheck 0.11.0` clean on both new scripts.
- Real-runner dynamics (QEMU emulation correctness on `linux/arm64`, the `macos-15-intel`/`macos-latest` labels resolving as expected) have **not** been exercised end-to-end yet — this repo has no scratch-tag or `workflow_dispatch` dry-run precedent for this class of change (checked git history: none exists). Prior GitHub-Actions-mechanics changes here (the original QEMU/multi-arch setup, the attestation step) were validated the same way — static review pre-merge, then fixed forward if something surfaced on the first real run. This PR follows that same convention; the first real tag push after merge is the actual end-to-end confirmation.

## Breaking Changes

Not a breaking change for existing consumers. It does mean a release with a genuinely broken binary on any of the four legs will now fail to publish where it previously would have shipped — that's the intended effect of this story, not a regression.

**Impact:** N/A

**Migration Guide:** N/A

### Code Quality

### Security
- [ ] Followed [secure logging rules](../docs/internal/secure-logging-rules.md) — not applicable, no logging code touched
- [ ] Credential-carrying types implement `slog.LogValuer` — not applicable, no Go types touched
- [ ] No security vulnerabilities introduced (checked with golangci-lint/gosec) — not applicable (no Go code changed); `zizmor` (GitHub Actions security linter) run instead and is clean

### Testing
- [ ] All tests pass locally: `go test -race ./...` — not applicable, no Go source changed
- [ ] No race conditions detected — not applicable
- [x] Error paths tested

### Documentation
- [x] docs/ updated (if architecture/design changes) — `RELEASING.md`
- [ ] godoc comments added/updated for exported functions — not applicable
- [ ] Examples updated (if applicable) — not applicable

### Git & CI
- [x] Branch is up to date with main (rebased if needed) — branched fresh from `origin/main`
- [x] No merge conflicts
- [ ] CI checks passing (lint, build, test, E2E) — pending: not yet observed on this PR, will confirm once CI runs

- [ ] Breaking changes documented in commit message — not applicable, not a breaking change
- [ ] Breaking changes documented in PR description — not applicable
- [ ] Migration guide provided — not applicable
- [ ] Deprecated functionality marked with comments and deprecation notices — not applicable

Not applicable — no UI/CLI output change.

## Additional Notes

No Go source was touched by this PR — every unchecked Go-specific checklist item above (`go test -race`, golangci-lint/gosec, godoc, slog.LogValuer) is inapplicable for that reason, not skipped. The applicable equivalent verification for a GitHub-Actions-only change — `actionlint`, `zizmor`, `shellcheck`, and the new script's own self-test — was run and is clean; see Testing above.

## Related Issues/PRs

Jira: SOL-153189


.github/scripts/smoke-test-binary.sh
```diff
@@ -0,0 +1,94 @@
...
+fi
+
+if [ ! -x "$BIN" ]; then
+    echo "::error::[$PLATFORM] $BIN_NAME is not executable" >&2
+    exit 1
+fi
+
+output=""
+exit_code=0
+output=$("$BIN" --version 2>&1) || exit_code=$?
```
solace-awaheed: This captures stderr into the string we compare against the tag, so any stderr output from a perfectly healthy binary fails the release. I ran the script against a fixture that prints one stderr line, then the correct version, then exits 0:

```
::error::[linux/arm64] --version reported 'qemu:someemulationnoticev9.9.9', expected 'v9.9.9'.
A stale or mis-stamped binary must fail the release.
```

Good binary, blocked release, and the error message points at the wrong cause. `linux/arm64` is where this is most likely to bite — QEMU user-mode does emit incidental chatter on stderr, and that's the one leg we can't dry-run before the tag.

Suggest splitting them: compare stdout, keep stderr for the diagnostic.

```bash
stderr_file="$TMPDIR/stderr.txt"
output=$("$BIN" --version 2>"$stderr_file") || exit_code=$?
```

Then include `$(cat "$stderr_file")` in the failure output. Worth a self-test case as well — "correct stdout, noisy stderr, exit 0 → passes" is exactly the shape the current 7 cases can't distinguish from a real failure.

.github/workflows/release.yml
```diff
@@ -174,6 +174,89 @@ jobs:
...
+
+      # Runs on every leg rather than gated to one. That costs three redundant
+      # runs of a sub-second fixture-based test, but it buys two things a
+      # single-leg gate can't: it can never silently stop running because a
+      # matrix leg was renamed or reordered (the exact "quietly skipped" risk
+      # this story treats as worse than a documented gap), and it exercises the
+      # script's shell-tool assumptions (tar, mktemp, chmod) against both GNU
+      # (Linux) and BSD (macOS) userland instead of only one.
+      - name: Self-test the smoke-test script
+        run: .github/scripts/smoke-test-binary.test.sh
```
solace-awaheed, line 254: Every other `*.test.sh` in `.github/scripts/` also runs on PRs via `ci-pr.yaml`: `commit-identity-check` (:44), `extract-release-notes` (:63), `licenses-check` (:92), `build-test-licenses-check` (:105), `license-header-check` (:157). This one only runs at release time, so if someone breaks `smoke-test-binary.sh` nothing catches it until a tag push goes red.

It also weakens the case for running on all four legs. The comment says running everywhere means it "can never silently stop running because a matrix leg was renamed" — but a single line in `ci-pr.yaml` does that better: every PR, before the tag, quarter of the cost, and no two macOS legs at 10x billing. The GNU-vs-BSD userland benefit still holds if you keep one leg here.

.github/workflows/release.yml
solace-awaheed, line 271: `build-docker` runs in parallel with `smoke-test`, so the binaries inside the image still hit a live registry tag without ever being executed — and that's the half we can't take back. RELEASING.md makes the point itself: "a registry push cannot be withdrawn... the image is live the moment it is pushed."

The image binaries come from a separate buildx path, so the archive smoke test genuinely doesn't cover them, which makes `RELEASING.md:49` ("Every binary archive actually runs") read broader than it is.

Best fix is probably a `--version` run against the built image before push, since that's the artifact actually at risk. Adding `smoke-test` to this `needs` works too, but serializes the image build behind the macOS legs. Happy to see this deferred to its own story — just worth writing the gap into RELEASING.md so the dependency diagram doesn't imply coverage we don't have.

.github/workflows/release.yml
```diff
@@ -174,6 +174,89 @@ jobs:
...
+          persist-credentials: false
+
+      - name: Set up QEMU
+        if: matrix.qemu
+        uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
+
+      - name: Download artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: binary-${{ matrix.goos }}-${{ matrix.goarch }}
```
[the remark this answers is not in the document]
solace-lootii: Checked against this repo's own existing precedent before changing anything: release.yml's own release job already documents (and relies on) the opposite behavior for this exact action version — "A single artifact downloaded by name extracts flat into the working dir (no name-subdirectory)" (see the Download release notes step). That's the same pinned actions/download-artifact SHA this job uses, downloading a single named artifact the same way. This describes the older v3 behavior, which changed in v4+; no fix needed here, verified against the repo's own working example rather than assumed.

.github/workflows/release.yml
```diff
@@ -174,6 +174,89 @@ jobs:
...
+      # (Linux) and BSD (macOS) userland instead of only one.
+      - name: Self-test the smoke-test script
+        run: .github/scripts/smoke-test-binary.test.sh
+
+      # The archived artifact downloaded above, not a binary freshly built in
+      # this job's workspace — packaging bugs are in scope.
+      - name: Smoke-test the archived binary
+        run: |
+          ARCHIVE="solace-broker-mcp-${GITHUB_REF_NAME}-${{ matrix.goos }}-${{ matrix.goarch }}.tar.gz"
+          .github/scripts/smoke-test-binary.sh "$ARCHIVE" "${GITHUB_REF_NAME}" "${{ matrix.goos }}" "${{ matrix.goarch }}"
```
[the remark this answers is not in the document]
solace-lootii: Same as the sibling comment — verified against this repo's own existing, working precedent for this exact action version. No fix needed.

.github/scripts/smoke-test-binary.sh
```diff
@@ -0,0 +1,94 @@
...
+if [ "$exit_code" -ne 0 ]; then
+    echo "::error::[$PLATFORM] $BIN_NAME --version exited $exit_code, expected 0. Output:" >&2
+    echo "$output" >&2
+    exit 1
+fi
+
+# Trim surrounding whitespace/newlines so a trailing newline from `fmt.Println`
+# doesn't fail an otherwise-exact match.
+actual_version="$(printf '%s' "$output" | tr -d '[:space:]')"
+expected_trimmed="$(printf '%s' "$EXPECTED_VERSION" | tr -d '[:space:]')"
```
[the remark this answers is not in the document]
solace-lootii: Fixed in 1163bb9 — added a small trim() helper that strips only leading/trailing whitespace, not internal characters.

.github/scripts/smoke-test-binary.test.sh
```diff
@@ -0,0 +1,157 @@
...
+
+# --- fixture builders --------------------------------------------------------
+# Each returns the path to a built .tar.gz in a fresh temp dir. `set -e` in this
+# file means a builder that fails aborts the whole self-test loudly, rather
+# than silently producing an empty archive that makes its case vacuous.
+
+# A binary that prints the given version and exits 0.
+build_good_archive() { # <version>
+    local tmp archive
+    tmp=$(mktemp -d)
```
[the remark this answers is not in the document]
solace-lootii: Fixed in 1163bb9 — added an ALL_TMP_DIRS array populated by every fixture builder, cleaned up once via a single trap on EXIT.

.github/scripts/smoke-test-binary.test.sh
```diff
@@ -0,0 +1,157 @@
...
+if [ "\$1" = "--version" ]; then
+    echo "$1"
+    exit 0
+fi
+exit 1
+EOF
+    chmod +x "$tmp/$BIN_NAME"
+    archive="$tmp/archive.tar.gz"
+    tar czf "$archive" -C "$tmp" "$BIN_NAME"
+    echo "$archive"
```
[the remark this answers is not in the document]
solace-lootii: Same fix, same commit — 1163bb9.

.github/scripts/smoke-test-binary.sh
solace-lootii: Fixed in 1163bb9 — split stdout/stderr capture exactly as suggested (stderr to a file, only stdout compared), surfaced stderr in both the exit-code and version-mismatch failure paths, and added the self-test case you described ("correct stdout with noisy stderr still passes") using the exact qemu-notice shape from your repro.

.github/workflows/release.yml
solace-lootii: Agreed, fixed in 1163bb9 — added a dedicated smoke_test_selftest job to ci-pr.yaml (runs on every PR, no dependency on release.yml's matrix at all) and gated the release.yml copy down to one leg (darwin/arm64) to keep the GNU-vs-BSD coverage without paying macOS minutes on all four legs for a test that gives identical signal on each.

.github/workflows/release.yml
solace-lootii: Agreed this is real and worth stating rather than silently implying coverage that doesn't exist. Went with your "defer to its own story, document the gap" option for now — fixed in 1163bb9, RELEASING.md:49 now explicitly calls out that build-docker's image comes from a separate buildx path this smoke test doesn't cover, why closing it properly needs a --version run against the built image, and that it's deliberately left as a follow-up rather than folded in here.

bczoma (approved):
Approving.

All three review findings are addressed in 98d68c42, and I verified the fixes rather than taking the commit message for them:

- stdout and stderr are captured separately now (`2>"$stderr_file"`), with stderr surfaced in the failure diagnostics but never compared. I reproduced the original defect against the pre-fix script — a healthy binary writing one line to stderr failed the version match with an error naming the wrong cause — and the new form doesn't have that shape.
- `tr -d '[:space:]'` is replaced with a leading/trailing trim, so the comparison really is "exact modulo surrounding whitespace" as the comment says.
- The self-test runs on every pull request via `ci-pr.yaml`'s `smoke_test_selftest` job, matching the five other `*.test.sh` jobs there, plus one release leg for the BSD-userland coverage a Linux runner can't give. The new `Smoke-test script self-test` check is green on this head.
- The `build-docker` gap is written down in both `RELEASING.md` and `CHANGELOG.md` as an explicit `[Planned]` follow-up rather than left implied.

A few things I checked independently while reviewing:

- The archive filename the smoke test constructs matches what `build-binaries` actually packages.
- `--version` is handled as the first statement in `main()`, before any logging setup, and prints the bare ldflags-stamped version — so comparing it against `GITHUB_REF_NAME` is sound, and no log line can interleave.
- On the earlier `download-artifact` concern about a name-subdirectory: it doesn't apply. A single artifact fetched by `name` extracts flat, and the `release` job has depended on exactly that shape since v0.7.1 published with this same pinned action version, so `ARCHIVE` in the workspace root is correct as written.

`CHANGELOG.md` now describes the self-test's placement accurately. 31 checks green.

solace-awaheed (approved):
Thanks for the updates, LGTM!

PR #318, opened 2026-08-18
SOL-153359: Gitignore make build's binary output at the repo root

bczoma:

- `make build` writes the server binary to the repo root as `./solace-broker-mcp` (`BINARY` in the Makefile), and no `.gitignore` rule covered it — `git add .` after a local build would stage a ~30 MB binary.
- Found while reviewing #311; unrelated to that change and deliberately kept separate.

## How is this accomplished?

- Adds `/solace-broker-mcp` to the existing `# Binaries` block in `.gitignore`.
- That block already carries `mcp-server` and `/server` — earlier values of `BINARY` that were never updated when the binary was renamed. This adds the current name rather than introducing a new convention, and the stale entries are left in place as harmless.
- The leading slash anchors the rule to the repo root, so an identically named path deeper in the tree stays tracked.

- gitignore has no file-only pattern syntax, so this rule would also match a directory named `solace-broker-mcp/` at the repo root. No such directory exists or is planned, and the alternative (no rule) is worse — flagging it so the choice is visible rather than implicit.
- No production surface touched, so no `CHANGELOG.md` entry: `.gitignore` is not `internal/config/`, `tools.yaml`, or a native tool package.

- `git check-ignore -v solace-broker-mcp` → matches `.gitignore:12`.
- `make build` then `git status --porcelain` → binary no longer listed as untracked.
- `test/nest/solace-broker-mcp` → **not** ignored, confirming the anchor behaves as documented.
- No code changed, so no unit/E2E runs apply.


solace-aross (approved):
Approving. Three-line `.gitignore` addition, no production surface. Checked the PR's own testing claims rather than trusting them: built a scratch repo with this exact `.gitignore`, confirmed `solace-broker-mcp` at the root is ignored and `test/nest/solace-broker-mcp` is not, and confirmed the Makefile's `BINARY` variable is in fact `solace-broker-mcp`. All three hold.

PR #319, opened 2026-08-18
Update SEMP schemas to latest rolling release (10.26.4)

solace-wmaryam:
Automated bump of the embedded SEMPv2 specs to `10.26.4.10725` (10.26.4), sourced from the private-extended vm_schema variant. Generated by broker-mcp/semp-autoupdate/update-semp-schemas.sh.


solace-aross (approved):
Approving. Automated SEMPv2 spec bump (10.26.3 → 10.26.4) plus one real content change: the monitor spec's `failureReason` enum drops `system-error` and adds `ip-interface-change-in-progress`, consistently in both the enum array and the description text. Checked for a hardcoded dependency on the old value rather than assuming none exists — `grep -rn "system-error" --include="*.go" .` returns nothing, and the spec files load by glob (`//go:embed *.json`) not by filename, so the version rename doesn't affect loading.

This PR had a real merge conflict earlier (`CHANGELOG.md`, two independently-added `### Changed` sections) that left it with zero CI signal beyond DCO — flagged that and held off on a verdict rather than approving blind. It's since been rebased and CI is now fully green across lint, build, and all five e2e suites. No outstanding concerns.

PR #322, opened 2026-08-19
SOL-152974: Adopt GitHub merge queue

solace-awaheed:
## Summary

Adds `merge_group` triggers so every required status check reports inside a GitHub merge queue, and stops a queue entry from being misclassified as a trunk build. Inert until the `merge_queue` rule is added to the `main-protection` ruleset.

Jira: SOL-152974

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue) — latent `ci-pr.yaml` concurrency defect, see below
- [x] New feature (non-breaking change which adds functionality) — merge-queue support

## Motivation and Context

`main-protection` sets `strict_required_status_checks_policy: true`, so every merge to `main` invalidates every other open PR, each of which then needs a rebase and a full re-run of 13 required checks (five of them Dockerized E2E suites). The cost grows with the square of the burst size, not linearly with merge count.

A merge queue gives the same guarantee properly: it tests the *merged result* on a temporary branch, so `main` still cannot break from a semantic conflict, but nobody serializes behind anyone else's rebase. It also removes a friction multiplier aimed at outside contributors, who currently need a maintainer to re-approve their workflow run after every rebase.

The blocker was that **no workflow in this repo triggers on `merge_group`**. A merge queue tests each entry on a `gh-readonly-queue/...` branch, so bound to `pull_request` alone not one required context would report — every entry would sit pending until it timed out and was ejected. Same failure shape as the phantom `Guardian scan` context in SOL-152412.

## Changes Made

- **`merge_group` trigger** on `build-and-test.yml`, `ci-pr.yaml`, `guardian-scan.yaml`, `workflow-lint.yaml`. No `branches:` filter — GitHub does not document what one matches for this event, and a filter matching nothing would recreate the hang above. Not needed anyway: the queue is scoped to `~DEFAULT_BRANCH`, so every merge group is main-based.
- **`guardian-scan.yaml`: new `queue` mode.** Previously `merge_group` fell through to the trunk branch, which would have uploaded scan results, run the Guardian DB sync and its Jira report, and blocked on Guardian's verdict — once per entry, for a ref that may never land. Scan jobs now skip on a queue entry and `gate` passes with a logged reason, the same pattern already used for fork PRs. Scan coverage is unchanged: the PR run reports before the queue, the trunk run enforces after it.
- **Extracted the mode classification** to `.github/scripts/guardian-scan-context.sh` with a paired `guardian-scan-context.test.sh`, matching the convention used by the other gates. This tuple decides whether a run uploads to Guardian and blocks on its verdict, and its `queue` arm cannot be exercised end-to-end until a queue exists, so it needed a test.
- **`ci-pr.yaml` concurrency fix.** `github.event.pull_request.number` is empty on `merge_group`, so every queue entry would have shared one group — and with `cancel-in-progress` unconditionally true, each new entry would cancel the previous entry's run, failing all four required contexts there intermittently. Now keyed on `|| github.sha`, matching the other three workflows.
- **`ci-pr.yaml` identity job**: `BASE_SHA`/`HEAD_SHA` fall back to `github.event.merge_group.base_sha`/`head_sha`. Passed explicitly because the script's `origin/main` default exits 1 if that ref is absent from a queue checkout.
- **`ci-pr.yaml` changelog step** gated to `pull_request`. `LABELS` has no `merge_group` analogue, so the `no-changelog` escape hatch would silently vanish. The job still reports, so the required context stays green — correct, since this gate is advisory and already had its say on the PR.
- **`timeout-minutes: 30`** on the remaining `build-and-test.yml` jobs. GitHub's 360-minute default means a wedged dependency holds a runner for six hours; in a queue the entry ejects at its own timeout while the abandoned run keeps burning.

## Testing

No Go code changed, so the Go test matrix does not apply. What was verified:

- [x] **actionlint v1.7.12** — CI's pinned version — clean, exit 0. Meaningful here rather than a formality: actionlint validates event names and checks context availability per trigger, so it confirms the `github.event.merge_group.*` expressions are legitimate for the declared events.
- [x] **shellcheck** clean on both new scripts
- [x] **New self-test: 34/34** across all four event shapes (`pull_request`, `merge_group`, `push`, `workflow_dispatch` fallthrough), plus the `fail_on_blocked` bypass, ref/version precedence, and fork passthrough
- [x] **Mutation-tested** — deleting the `merge_group` arm from a copy of the script fails 7 assertions, so the suite genuinely pins the queue-classified-as-trunk regression rather than passing vacuously
- [ ] Live `merge_group` behaviour — **not verifiable in this PR.** No `merge_group` event can fire until the ruleset changes, so whether the contexts actually report on a real queue entry is the next step, not this one.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code
- [x] No new `pull_request_target` introduced anywhere (an explicit acceptance criterion)
- [x] Queue-mode runs touch no Environment secrets — the scan jobs that hold FOSSA/Prisma credentials skip on `merge_group`

### Testing
- [x] New logic covered by a self-test, and the test verified by mutation
- [x] Error/edge paths covered (missing inputs, event fallthrough, bypass flag)

### Documentation
- [ ] `ADMIN_SETUP.md` — intentionally deferred. It documents live ruleset state, so the merge-queue section lands with the ruleset change rather than describing a queue that does not exist yet.
- [x] No `CHANGELOG.md` entry needed — `.github/` is not production surface per `CLAUDE.md`

### Git & CI
- [x] Branch based on current `main`, no conflicts

## Additional Notes

**This PR is inert.** No `merge_group` event fires until a `merge_queue` rule exists, so nothing here changes current behaviour. That is deliberate — it makes the risky part a settings change that rolls back by deleting one rule, rather than code and settings landing together.

Follow-up work, in order:

1. **Ruleset change** (needs repo admin): add the `merge_queue` rule with `max_entries_to_merge: 1` — batching would advance `main` in a single push and leave all but the batch head without a trunk verdict. Capture the ruleset JSON first; that file is the rollback.
2. **Verify on a live entry**: all 13 contexts report, then turn off `strict_required_status_checks_policy` and update `ADMIN_SETUP.md`.


solace-aross (approved):
Approve.

Verified rather than trusted the two testable claims in the description: ran the self-test in a scratch worktree (34/34 pass), then removed the `merge_group` branch from `guardian-scan-context.sh` and reran it. Fails exactly the 7 assertions the description claims, with `mode=trunk`, `enforce=true`, `fossa_mode=BLOCK` on a queue event instead of the queue values. The mutation test genuinely pins the regression, not just decorative coverage.

Read through the `ci-pr.yaml` concurrency fix, the `BASE_SHA`/`HEAD_SHA` fallbacks, and the changelog-step gating. All correct given `merge_group`'s payload shape: `pull_request.number` is empty there, so the `|| github.sha` fallback is load-bearing, and the changelog step's `if: pull_request` keeps the required context green on a queue entry without inventing a `merge_group` `LABELS` analogue. Confirmed all 7 `build-and-test.yml` jobs (not just the ones shown in the diff) now carry `timeout-minutes: 30`, so there's no gap job left on GitHub's 360-minute default.

One gap worth flagging before the ruleset change lands, not blocking this PR: `CodeQL` is a required context (added to the ruleset 5 minutes after this PR opened, so the "13 contexts" count in the description is already stale, 14 now) that comes from GitHub's code-scanning default setup, not a workflow file. Nothing in this diff can make it trigger on a `merge_group` ref, and default setup has no trigger-customization surface the way advanced setup does. `github/codeql-action#1537` has been open on exactly this combination since 2023. I couldn't find a doc proving default setup positively won't report on a queue ref, so treat this as verify-first rather than a settled defect, but it's the kind of thing that wedges the first live entry at `check_response_timeout_minutes` rather than failing fast. Worth settling with @bczoma (who added the `CodeQL` requirement) before step 1 of the follow-up, since the options (drop it from required, move to advanced setup with an explicit `merge_group` trigger, or stub it) are exactly what you want decided outside a stuck queue entry.

`DCO` is fine, for what it's worth. `cncf/dco2` has handled `merge_group` since 2024 and posts straight off the merge group's head commit, no repo-side config needed.

Nothing else raised anything at Medium or above. The `queue` mode split in `guardian-scan.yaml` is a clean way to keep the `guardian` Environment's secrets away from a throwaway queue ref.

PR #324, opened 2026-08-19
SOL-153363: make token cache and exchange logs self-describing

amitmorade:
## Summary

Makes the token-exchange DEBUG logs readable without opening the source. Statuses log as words instead of integers, durations as `77.496667ms` instead of `77496667`, and the messages name the credential rather than the cache method they were called from.

Tracked under [SOL-153363](https://sol-jira.atlassian.net/browse/SOL-153363).

## Type of Change

No behaviour change — only log output. Marked "bug fix" because one log line was actively misleading, not merely terse.

## Motivation and Context

This started with a developer who had worked on this feature being unable to read its own logs:

```json
{"msg":"token cache get","broker":"prod-us","status":1,"correlation_id":"01a0..."}
{"msg":"token cache put","broker":"prod-us","status":0}
{"msg":"token exchange succeeded","broker":"prod-us","exchange_elapsed":77496667}
```

Four separate problems, in increasing severity:

**1. `status` was an unlabeled enum.** `GetStatus` is `0=hit`/`1=miss`; `PutStatus` is `0=stored`/`1=dropped_ttl`. So `0` means success on both lines while `1` flips meaning between them. Both types had a `String()` method — but `slog` resolves `LogValuer` and never calls `String()`, so the handler emitted the raw iota. The `GetResult`/`PutResult` wrappers already did this correctly; only the bare enums were missing it.

**2. `status` is overloaded three ways in this tree.** A numeric HTTP status code (`Sender.errorHandler`, `Sender.checkRetry`), a success/error verdict as a string (`logToolResult`), and the cache outcome. A reader hitting `status: 1` can't tell which, and reasonably parses it as a status code.

**3. Durations rendered as raw nanoseconds.** The call sites were already correct — `slog`'s JSON handler encodes `time.Duration` as the `int64` it is underneath. `time.Duration` knows how to render itself, but again `slog` wants `LogValuer`, and Go forbids adding a method to another package's type. So the fix had to happen in the handler.

**4. `exchange_elapsed` named the wrong span — the one real defect.** The timer starts before singleflight dispatch and stops after the caller receives its result, so it covers deduplication waiting, the rate-limit gate, the circuit breaker, every retry attempt and its backoff, the IdP call, **and the cache write**. Only the cache read is excluded. A value can therefore be three attempts plus backoff, or a breaker rejection that never reached the IdP at all.

## Changes Made

Six commits, each self-contained and green on its own.

- `6b46bdb` — `LogValue()` on `GetStatus`/`PutStatus`. Fixes the *type*, so every existing and future call site emits `hit`/`miss`/`stored`/`dropped_ttl` with no call-site edit.
- `03dbddb` — `status` → `result`, and converted the two lines to typed attrs, matching the dominant convention in this tree (loose varargs silently produce `!BADKEY` on a dropped argument).
- `a7e4f8d` — messages renamed from cache methods to the credential: `broker credential cache lookup` / `broker credential cache store` / `broker credential obtained from identity provider`.
- `079a16a` — duration rendering in `redactSecretAttr`, keyed on the value's *kind* rather than a list of field names.
- `3a77bd1` — `exchange_elapsed` → `exchange_total_elapsed`, plus an 18-line comment at the timer enumerating exactly what the span covers.
- `4d7042e` — CHANGELOG.
- `883787f` — split the cache-store line per outcome (see the trade-off below); adds the first test guarding these two messages and the absence of `result`.

### Before / after

Cold path:

```json
{"msg":"broker credential cache lookup","broker":"prod-us","result":"miss"}
{"msg":"broker credential cached","broker":"prod-us"}
{"msg":"broker credential obtained from identity provider","broker":"prod-us","exchange_total_elapsed":"77.496667ms"}
```

Warm path — one line, now instantly distinguishable from a cold one:

```json
{"msg":"broker credential cache lookup","broker":"prod-us","result":"hit"}
```

The store outcomes, one message each:

```json
{"level":"DEBUG","msg":"broker credential cached","broker":"prod-us"}
{"level":"WARN","msg":"broker credential not cached: remaining lifetime too short after clock-skew adjustment","broker":"prod-us"}
```

## Decisions and trade-offs

**Fix the type, not the call sites.** `LogValue()` on the enum rather than `.String()` at each call site. Costs an extra method; buys that a future call site cannot reintroduce the integer. `6b46bdb` alone fixed the readability problem tree-wide with zero call-site edits.

**Where the status field belongs — the rule this PR converged on.** Keep it where it discriminates; drop it where the message already decided.

The first draft of the cache messages was `"broker credential not cached, requesting from identity provider"`. That is **wrong**: the lookup call sits *above* the `GetHit` branch, so it fires on hit and miss alike — and on a hit it is the *only* line emitted for the whole warm path. It would have made the sole record of a successful cache hit claim the credential wasn't cached.

The fix was outcome-neutral wording, but applying that uniformly was also wrong, because the two sites have different shapes:

- **Lookup is genuinely two-outcome.** A `Get` always happens; `hit`/`miss` is what it found, and both log at the same level. The message cannot name either outcome, so `result` is the only discriminator. It stays.
- **Store is not two-outcome.** `stored` means a write happened; `dropped_ttl` means it was refused before any write (`Put` returns before `Set` is called). That is "did it" vs "didn't", at different levels. One shared message is false on the `dropped_ttl` path — and a message that *does* name the outcome makes `result` pure redundancy. So the store site is split per outcome and drops `result`.

The store site uses a `switch` with an explicit default rather than `if/else`: `PutStatus` is an open iota, and `if dropped_ttl / else stored` would report a **false success** for any value added later. The default arm warns that the outcome is unknown, and is the one branch that keeps `result` — its message deliberately doesn't name an outcome, so the raw status is all there is to report.

**`exchange_total_elapsed`, not `idp_roundtrip_elapsed`.** The rename was originally going to be `idp_roundtrip_elapsed`. Rejected after tracing the timer: "roundtrip" singular would be a lie, since the span includes retries, backoff, singleflight waiting and the cache write. `idp_roundtrip_elapsed=8.2s` would read as one HTTP call taking 8.2 seconds when it may be three attempts. **Trading a vague-but-true name for a specific-and-false one is a regression** — ambiguity prompts a code read, false precision prevents one. `total` is honest about the width.

**Duration conversion keyed on type, not field name.** One `if` in the handler covers all nine duration call sites and anything added later — versus per-call-site conversion that has to be remembered forever. Accepted cost: this is a **global output-format change**, the widest-blast-radius part of the PR. It is last in the sequence so it can be reverted alone.

**Machine parseability.** Durations become JSON strings, so anything consuming them numerically breaks. Accepted: Go duration strings still round-trip through `time.ParseDuration`, and this is the dominant convention in Go server logs. Flagged prominently in the CHANGELOG.

**Kept the WARN on its own line.** Considered folding the store line into the success line as `cached: true`. Rejected — `PutDroppedTTL` is WARN-level and reachable (non-positive TTL after clock-skew adjustment), and folding it into a boolean on a DEBUG line would silently demote it.

**Rejected: a `token exchange started` log.** On the success path it is redundant with the `result=miss` line immediately preceding it, and doubles the line count for the common case. What's actually wanted is per-attempt IdP timing — filed as SOL-153394.

## Testing

**Test Environment:**
- Go version: as per `go.mod`
- OS: darwin (macOS)
- Broker version: n/a — no broker interaction; IdP driven via `httptest`

**Tests Performed:**

**Test Coverage:**

- `make check` green: build, vet, lint, race tests. **Total coverage 89.4%**, above the 85% floor.
- New tests assert on output from a **real slog handler**, not the method return. This matters: a unit check of `String()` passed the entire time these logs were unreadable, so it could not have caught the bug. The tests also fail explicitly if the raw iota reappears.
- The duration tests decode the emitted JSON and type-assert `.(string)` — raw nanoseconds would decode as `float64` and fail. Covers top-level, inside a `slog.Group`, and bound via `.With()`.
- Verified empirically that `ReplaceAttr` fires for grouped and `.With()`-bound attributes, so no duration is missed.
- Captured real cold-path, warm-path and `dropped_ttl` output from a live `Exchange()` (shown above).

## Breaking Changes

No API or config change. Log output only.

**Impact:** anything parsing these logs — `status` on the cache lines (now `result`, string-valued), the three old message strings, `exchange_elapsed` (now `exchange_total_elapsed`), and any duration field consumed as a number.

**Migration Guide:** read duration fields as strings and parse with `time.ParseDuration`, or match the Go duration format. Update queries keying on `status` for these two lines to `result`, and on the renamed message strings and duration key.

## Additional Notes

**Four of SOL-153363's eight acceptance criteria are already satisfied by these commits.** The ticket still describes them in scoped-state tense — I deliberately did not retro-edit it into a changelog as commits landed. Please don't read the unticked boxes as remaining work.

**`scripts/annotate-cache-logs.sh` is broken by this change and is not in this PR.** That directory is untracked, so it is a local-only dev tool, not a versioned consumer. It exists precisely to translate `status: 1` into "CACHE MISS" and divide nanoseconds by a million — which is the strongest evidence the problem was real. It needs a local pass; nothing in the repo depends on it.

Historical planning docs under `docs/superpowers/specs/exchange-retries/` still reference `exchange_elapsed` in sample output. Left as-is — they are records of past design discussions, not living contracts.

## Related Issues/PRs

- Implements [SOL-153363](https://sol-jira.atlassian.net/browse/SOL-153363)
- Follow-up: [SOL-153364](https://sol-jira.atlassian.net/browse/SOL-153364) — the cache-store line still lacks `correlation_id`, because it is logged under a context detached from the caller's. Deliberately out of scope: that is context propagation, not readability.
- Follow-up: [SOL-153394](https://sol-jira.atlassian.net/browse/SOL-153394) — nothing is logged between the cache miss and the cache store, so the IdP call's own cost is unmeasurable. Surfaced by this PR's span analysis.

---

**For Reviewers:**

**Focus Areas**:
1. Message wording, and where `result` is kept vs dropped. Does `broker credential cache lookup` + `result=hit` read correctly on the warm path, and do the two split store messages read correctly without a `result` field? A first draft got both of these wrong in opposite directions — false on one path, redundant on the other.
2. `exchange_total_elapsed` and the span comment at the timer — is the enumeration complete and accurate?
3. `079a16a` is the widest-reaching commit: one `if` changes nine duration fields server-wide. Worth a look if you know of anything consuming these numerically.

**Questions**:
- These are DEBUG lines, i.e. the developer channel. If support actually relies on them during escalations, the *level* is arguably also wrong — a broader observability question I did not want to fold in here. Views welcome.


solace-aross (approved):
Approve.

This touches auth/secrets/logging, so ran it through the Opus review gate before finalizing rather than trusting my own first pass. Verified independently rather than trusted:

- Build, `go vet`, `golangci-lint` (0 issues) all clean on the touched packages.
- `go test -race` green on `cmd/server`, `internal/oauth/cache`, `internal/tokenexchange`.
- Full `make test-cover`: 89.4%, matches the PR's claim exactly.
- Mutation-tested the two real fixes: removed the `slog.KindDuration` branch in `redactSecretAttr` — the two duration tests fail exactly as expected, decoding as `float64` instead of the formatted string. Renamed `GetStatus`/`PutStatus`'s `LogValue()` methods to something slog won't resolve — `TestGetStatus_LogValue`/`TestPutStatus_LogValue` catch the raw iota (`"result":0`/`"result":1`) exactly as the PR describes.
- Sabotage-tested the newest commit's split cache-store lines (collapsed `broker credential cached`/`broker credential not cached: ...` back into one shared message) — `TestExchange_CacheStoreLogsNameTheirOutcome` catches it on both branches.
- Chased down two claims that didn't hold up before including them: swapping the redaction/duration-formatting branch order in `redactSecretAttr` does **not** break the ordering test, because redaction always unconditionally overwrites the value on a key match regardless of what ran first — the safety property is real, just not for the ordering reason the code comment states, and that's a documentation nit below the bar here. And `GetResult`/`PutResult`'s own `LogValue()` already calls `.String()`, so it doesn't reintroduce the integer bug I initially thought it might — it's the guard against leaking `Entry.Value`, not dead weight.

Two non-blocking notes:

1. `internal/tools/manager.go:355,406` emit `slog.Duration("duration", time.Since(start))` on the `tool invoked` line — the highest-volume operator log line in the server, and on `docs/internal/secure-logging-rules.md`'s Always-Log List. Since the duration-formatting branch matches on kind, not key name, this line's `duration` silently goes from a JSON number to a string too, and loses easy p50/p95 aggregation. The CHANGELOG entry says "every duration the server emits" but doesn't name this specific line, and the doc still calls the field `duration_ms` (already drifted from the code's `duration`, now also drifted in type). Worth naming this line explicitly in the CHANGELOG and reconciling the doc.
2. `token cache get failed` (`exchange.go:57`) and `token cache put failed` (`:275`) kept their old vocabulary while their success siblings became `broker credential cache lookup` / `broker credential cached` / `broker credential not cached: ...`. Those are exactly the lines someone grepping `broker credential` during an incident would miss.

Neither changes the verdict — this is a well-tested, well-reasoned logging change with real before/after evidence, not an assertion.

amitmorade:
Closing this one, there is still a few things I figured need to be fixed and kind of would come before the work being done in this PR.

PR #332, opened 2026-08-21
SOL-153355: Add the RDP tools to the performance harness fixture and teach loadgen their arguments

helen-zhu-solace:
## Summary

Adds the two RDP tools (`list-rdps`, `get-rdp-status`) to the performance harness fixture and teaches `loadgen` their arguments, giving the harness a 1-call / 3-call SEMPv2 pair whose fields stay byte-stable under real broker load. No product code changes — everything is under `test/performance/`.

Implements SOL-153355 

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The harness from SOL-150813 could replay two tools, and both are weak instruments. The fidelity gate is exact but fragile: all three entries in `fidelity/exclusions.txt` come from `get-broker-status`, and `list-queues` needs none only because the capture VPN is idle — six of the thirteen fields it selects (`bindCount`, `msgSpoolUsage`, `spooledMsgCount`, `rxMsgRate`, `txMsgRate`, `txUnackedMsgCount`) move under any traffic at all. The moment fixtures are captured from a loaded broker, keeping the gate green means excluding those six, at which point the check stops verifying most of what the tool returns. Today's clean result is a property of the capture VPN being quiet, not of the harness being robust.

The RDP tools are not subject to that trade. Every RDP in the capture VPN is `enabled: false`, so every field both tools select is settled — `up: false`, `uptime: 0`, all HTTP counters `0`, `lastFailureTime` frozen at the moment each RDP was disabled rather than advancing on retries. These stay exact on a busy broker.

They also close two measurement gaps. `loadgen` reports tool calls per second while the performance targets are written in SEMP requests per second, and converting between them needs a fan-out factor the old fixture made ambiguous (`get-broker-status` is 4 or 5 depending on broker type, `list-queues` 1 or 3 depending on `maxResults`, and nothing recorded which). `list-rdps` at default arguments is exactly 1 SEMP call, so the reported rate *is* the SEMP rate. `get-rdp-status` is a stable 3 on the same protocol against the same objects in the same VPN — a 1-call and a 3-call tool differing in nothing but fan-out — and its steps 2 and 3 are `parallel: true`, which is the only concurrent-sub-request shape in the fixture.

Separately, `loadgen` could not call either tool, and the failure mode was worse than "it doesn't work". A tool invoked without its required arguments returns `IsError`, and `callOnce` folded that into the same error count as broker unavailability, timeouts, and MCP-side rate limiting — so a misconfigured workload would complete, report a high error rate, and read as a resilience finding rather than an operator mistake. An argument gap must not be able to masquerade as a measurement.

## Changes Made

- `loadgen`: replaced `callOnce`'s hardcoded `if tool == "list-queues"` branch with `toolRecipes`, a static per-tool argument map. Deliberately a literal list, not schema introspection — deriving arguments from the tool's input schema would couple `loadgen` to MCP's registration path.
- `loadgen`: `validateTools` refuses to start when `-tools` names a tool with no recipe (message lists the known tools) or a recipe whose argument has no value. Checked before any session is dialled and before the broker flags are resolved.
- `loadgen`: new `-rdp` flag for `restDeliveryPointName`; `-vpn` help updated; `-tools` now defaults to all four tools.
- `loadgen`: new `-validate-only` mode that checks the tool list and exits without opening a session, so both run scripts can validate `TOOLS` in their step-0 preflight — before the mock binds, before the fidelity gate, and (split-host) before the wait for Box B. The valid tool list stays in `loadgen`; a bash copy would drift the first time a tool is added.
- `loadgen`: arguments are now prebuilt into per-client `callSpec` values before the clock starts, so the steady-state loop neither allocates a map per call nor performs a lookup that can fail.
- `mock-semp`: `isRdpsListPath` mirroring `isQueuesListPath`, and both are routed through a new `isMonitorMsgVpnPath` that matches the public `/SEMP/v2/monitor/` and private `/SEMP/v2/__private_monitor__/` prefixes. MCP uses the private one, so a rule matching only the public path never fires.
- `mock-semp`: three exact-path rules for the pinned RDP's object, queue bindings, and REST consumers, registered ahead of the collection rules. The pinned name is read out of `rdp_object.json` at startup rather than taken as a flag, so the fixture and the rule cannot name different RDPs. A request for any other name matches nothing, logs a `MISS`, answers 404, and fails the run at shutdown.
- `mock-semp`: pagination discovery generalized into `buildPagedRules(pagedCollection)`, shared by queues and RDPs instead of each carrying a copy. Startup now fatals on a non-contiguous page set and warns when the collection that owns the paginated fidelity check yielded a single page.
- `mock-semp`: per-rule hit counters, a shutdown summary that flags rules which never fired, and `GET`/`POST /_mock/hits` on the config port (GET reports, POST reports and zeroes). This is the harness measuring its own SEMP fan-out.
- `fidelity`: `buildChecks` is now one list used by both capture and verify — they previously carried a copy each, so a tool added to one and not the other would capture a golden nothing checked, or check a golden nothing captured. Added `list-rdps`, a paginated `list-rdps (maxResults=200)` check, and `get-rdp-status`; added a `-rdp` flag; a missing golden now points at `regen-golden.sh` instead of reading as a tool regression.
- `capture.sh`: captures `restDeliveryPoints` (following pagination), the pinned RDP's object, its queue bindings and its REST consumers, plus the appliance-only `show hardware details` that the mock's rule set requires at startup. Pagination extracted into `capture_pages`, which deletes stale page files first so a shorter capture cannot leave a previous run's tail to be served as part of a chain it never belonged to.
- `sanitize.sh`: the new captures and goldens added to the scrub list; both paginated collections globbed; and `report_rdp_endpoints` prints `remoteHost`, `remotePort` and `postRequestTarget` at the end of every run, because the residual scan matches only 192.168/16 and 172.16/12 and would pass an RDP pointed at a real internal service by hostname.
- `regen-golden.sh` / `fixtures-manifest.sh`: `RDP_NAME` threaded through capture, recorded as `# rdp:` provenance, and readable back via `fixtures-manifest.sh rdp`; both run scripts read it from there to pass `-rdp`. A capture yielding fewer than two RDP pages warns at capture time, where recapturing from a bigger VPN is still an option.
- `run.sh` / `run-loadgen.sh`: `RDP` env knob, `TOOLS` preflight validation, and a `snapshot_fanout` step that POSTs `/_mock/hits` the moment the fidelity gate passes — banking the gate's one-call-per-check pass as the fan-out table and zeroing the counters so the shutdown summary measures the load phase alone.
- `README.md`: the five gate checks and why the paginated one exists; a fixture-durability table saying which tools stay exact on a realistic capture and which pass only because the capture VPN is quiet; the two structural limits (SEMPv2/JSON only, and all-disabled RDPs make any ceiling an upper bound rather than a sizing figure); and the measured SEMP cost per tool call.

## Testing

**Test Environment:**
- Go version: go1.25.12 linux/amd64
- OS: Fedora Linux (kernel 7.0.10)
- Broker version (if applicable): fixtures captured from a Solace 3560 appliance on a main-branch build; harness runs replay those captures through `mock-semp`

**Tests Performed:**

**Test Coverage:**
- 20 new unit tests across `loadgen/main_test.go` (6) and `mock-semp/handler_test.go` (14); 31 tests pass across the three harness packages. `make check` green — build, vet, lint, race-enabled tests, total coverage 89.3% against the 85% floor.
- Argument map: every recipe resolves against its flags; rotation order preserved; per-broker specs isolated.
- Startup validation: an unknown tool and a recipe with an unset flag each abort with exit 2 and a message naming the tool. Verified against the built binary — `-tools list-rdps,frobnicate` reports `no argument recipe for tool "frobnicate" … Known tools: get-broker-status, get-rdp-status, list-queues, list-rdps`, and `-tools get-rdp-status` without `-rdp` reports the missing `restDeliveryPointName`.
- Pagination: `list-rdps` page 1 then the cursor-matched page 2; an unknown cursor misses rather than falling back to page 1.
- Both monitor prefixes match the same rule.
- Negative case: `/restDeliveryPoints/rdp_999` and both of its sub-resources miss and increment the miss counter, for the object, `queueBindings` and `restConsumers` alike; a per-RDP path is not swallowed by the collection rule; `rdpSubResource` rejects deeper and malformed shapes.
- Pinned name comes from the capture, not a literal.
- Hit counters: per-rule counts, `POST` reset separating phases, `GET` not resetting, non-GET/POST rejected, and a reset leaving the miss count alone.
- Integration, exact mode, exclusions untouched: full run with all four tools, then `TOOLS=list-rdps` alone (1644 requests, 0 errors) and `TOOLS=get-rdp-status` alone (1672 requests, 0 errors). All five fidelity checks matched their golden in every run, and every run reported zero misses.
- Fan-out reproduced and recorded: `list-rdps` 1, `list-rdps maxResults=200` 2, `list-queues` 1, `get-rdp-status` 3, `get-broker-status` 5 on the appliance capture — matching the counts measured through a logging reverse proxy against the real broker.

## Breaking Changes

None. Every change is under `test/performance/`; no production code, config schema, or tool definition is touched.

**Impact:** n/a

**Migration Guide:** n/a

### Code Quality

### Security

### Testing
- [x] Error paths tested

### Documentation

### Git & CI
- [x] No merge conflicts

n/a

## Additional Notes

**No CHANGELOG entry.** The change is confined to `test/performance/` and touches no production surface — not `internal/config/`, not `tools.yaml`, not a native tool package.

**One deliberate deviation from the story.** SOL-153355 asks for `-rdp` to default to the RDP name the fixture pins. It has no default here: both `loadgen` and `fidelity` require it, and the run scripts read it from `fixtures.manifest`. A literal default in each script is exactly how the mock's pin and the caller's request would come to disagree, and a disagreement surfaces as a confusing diff or a mock miss rather than a clear error. The out-of-the-box run works, which is what the AC is protecting — `run.sh` with no environment set passes.

**Two rows of the story's SEMP-cost table are not reproduced.** `list-rdps maxResults=500` (2 calls) and `list-queues maxResults=500` (3 calls) are absent from the README table, for different reasons. For `list-rdps`, 500 is redundant: the capture VPN holds 196 RDPs, so `fetchPaginated` runs out of objects before it reaches either limit and 500 returns exactly what the 200 check already asserts — a second 50 KB golden to capture, sanitize and hash for no additional coverage. For `list-queues` it is not redundant: at 291 queues, 200 stops on page 2 with `truncated: true` and 500 is the only value that requests page 3, which is why `sempv2 queues page 2/3 (cursor)` read `(never fired)` in every hit summary. 500 also crosses `capMax`, selecting a different `truncatedMessage` that no check currently covers. That is a `list-queues` gating change, which this story explicitly excludes, so it is filed as a follow-up on the SOL-150813 pagination thread rather than smuggled in here.

**`loadgen` has no `maxResults` flag,** per the story's exclusion list. That is what keeps the denominator stable: every list tool in a load run costs exactly 1 SEMP request, so the reported tool rate converts to a SEMP rate by a known factor instead of an operator variable.

**Read the fixture-durability section before recapturing.** If fixtures are recaptured from a broker under load and the gate goes red on `list-queues`, that is the expected outcome and not a regression — and adding those six fields to `exclusions.txt` is a decision to stop checking most of what the tool returns, worth making deliberately.

**For Reviewers:**

**Focus Areas**: The exact-path RDP matching in `mock-semp/handler.go` is the load-bearing decision — the alternative shape, a prefix match on `/restDeliveryPoints/`, would answer a request for any of the VPN's 196 RDPs with the one that was captured, and nothing else in the harness would catch it. Compare the five SEMPv1 rules, which share the `/SEMP` path and are told apart by first-substring-match on the body; that shape has the hazard and these rules deliberately do not inherit it. Also worth a close read: `validateTools` running before the broker flags are resolved (so `-validate-only` needs nothing it does not use), and `capture_pages` deleting stale page files before capture (a leftover page would otherwise be manifest-recorded and served as part of a chain it never belonged to).

### DIFF
--- a/test/performance/loadgen/main.go
+++ b/test/performance/loadgen/main.go
@@ -28,7 +28,13 @@
 //	  -brokers  my-broker \
 //	  -clients  32 \
 //	  -duration 60s \
-//	  -tools    get-broker-status,list-queues
+//	  -tools    get-broker-status,list-queues,list-rdps,get-rdp-status \
+//	  -rdp      rdp_1
+//
+// Every tool loadgen can call is listed in toolRecipes along with the
+// arguments it needs; a -tools list naming anything else stops the run at
+// startup. get-rdp-status additionally needs -rdp, which must name the RDP
+// the mock's fixture pinned (run.sh reads it from fixtures.manifest).
 //
 // Auth: reads MCP_DEV_TOKEN from env and sends it as Bearer if set.
 package main
@@ -58,24 +64,45 @@ func main() {
 	brokersCSV := flag.String("brokers", "", "comma-separated broker aliases; each client is pinned to one, round-robin. Mutually exclusive with -broker-count.")
 	brokerCount := flag.Int("broker-count", 0, "shortcut: generate <prefix>-01..<prefix>-N as the alias list. Zero means use -brokers instead.")
 	brokerPrefix := flag.String("broker-prefix", "broker", "prefix used with -broker-count (e.g. -broker-count=50 => broker-01..broker-50)")
-	vpn := flag.String("vpn", "default", "msgVpnName argument for tools that need it (e.g. list-queues)")
+	vpn := flag.String("vpn", "default", "msgVpnName argument for every tool whose recipe needs it (list-queues, list-rdps, get-rdp-status)")
+	rdp := flag.String("rdp", "", "restDeliveryPointName argument for get-rdp-status. Must name the RDP the fixture pinned — run.sh reads it from fixtures.manifest")
 	clients := flag.Int("clients", 32, "number of concurrent MCP sessions")
 	duration := flag.Duration("duration", 60*time.Second, "steady-state run duration")
 	warmup := flag.Duration("warmup", 0, "duration excluded from stats before steady state; useful for skipping session warmup jitter")
-	toolsCSV := flag.String("tools", "get-broker-status,list-queues", "comma-separated tool names; each client rotates through them")
+	toolsCSV := flag.String("tools", "get-broker-status,list-queues,list-rdps,get-rdp-status", "comma-separated tool names; each client rotates through them")
 	rps := flag.Float64("rps", 0, "per-client req/s cap; 0 = unlimited (client fires as fast as MCP responds). Mutually exclusive with -total-rps.")
 	totalRPS := flag.Float64("total-rps", 0, "aggregate req/s across all clients; divided evenly so per-client = total-rps/clients. Mutually exclusive with -rps.")
 	connectTimeout := flag.Duration("connect-timeout", 30*time.Second, "deadline for opening each MCP session before the run starts")
+	validateOnly := flag.Bool("validate-only", false, "check -tools against the argument recipes and exit, without opening any session. The run scripts call this in their preflight so a typo in TOOLS costs nothing instead of surfacing after the mock is up, the fidelity gate has run, and (split-host) MCP has been waited for.")
 	flag.Parse()
+
+	tools := splitCSV(*toolsCSV)
+	if len(tools) == 0 || *clients < 1 || *duration <= 0 {
+		fmt.Fprintln(os.Stderr, "loadgen: -tools, -clients, -duration are required and non-empty")
+		os.Exit(2)
+	}
+	// Refuse the run before a single session is dialled. A tool called with
+	// missing required arguments returns IsError, which lands in the same
+	// error count as broker unavailability and timeouts — so an argument gap
+	// would complete the run and read as a resilience finding rather than an
+	// operator mistake.
+	//
+	// Checked before the broker flags so -validate-only needs nothing but the
+	// tool list and its argument values — the preflight has those, and asking
+	// it for a broker count it doesn't use would be a wart.
+	argValues := map[string]string{argVPN: *vpn, argRDP: *rdp}
+	if err := validateTools(tools, argValues); err != nil {
+		fmt.Fprintf(os.Stderr, "loadgen: %v\n", err)
+		os.Exit(2)
+	}
+	if *validateOnly {
+		fmt.Printf("loadgen: -tools %v OK (arguments resolvable for every tool)\n", tools)
+		return
+	}
 
 	brokers, err := resolveBrokers(*brokersCSV, *brokerCount, *brokerPrefix)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "loadgen: %v\n", err)
-		os.Exit(2)
-	}
-	tools := splitCSV(*toolsCSV)
-	if len(tools) == 0 || *clients < 1 || *duration <= 0 {
-		fmt.Fprintln(os.Stderr, "loadgen: -tools, -clients, -duration are required and non-empty")
 		os.Exit(2)
 	}
 	perClientRPS, err := resolveRPS(*rps, *totalRPS, *clients)
@@ -92,7 +119,7 @@ func main() {
 	cfg := runConfig{
 		mcpURL:         *mcpURL,
 		brokers:        brokers,
-		vpn:            *vpn,
+		argValues:      argValues,
 		tools:          tools,
 		clients:        *clients,
 		duration:       *duration,
@@ -107,15 +134,95 @@ func main() {
 }
 
 type runConfig struct {
-	mcpURL         string
-	brokers        []string
-	vpn            string
+	mcpURL  string
+	brokers []string
+	// argValues maps an argument name from toolRecipes to the flag-supplied
+	// value for it. Validated against every named tool's recipe before the
+	// run starts, so buildCallSpecs can build arguments without a fallible
+	// lookup on the call path.
+	argValues      map[string]string
 	tools          []string
 	clients        int
 	duration       time.Duration
 	warmup         time.Duration
 	rps            float64
 	connectTimeout time.Duration
+}
+
+// Argument names the recipes below draw on. Constants rather than repeated
+// literals so a typo is a compile error instead of a tool-side IsError that
+// only shows up as a high error rate mid-run.
+const (
+	argVPN = "msgVpnName"
+	argRDP = "restDeliveryPointName"
+)
+
+// toolRecipes names the arguments each tool needs beyond "broker", which
+// every tool takes. Deliberately a literal per-tool list: deriving arguments
+// from the tool's input schema would couple loadgen to MCP's tool
+// registration path, which is a much bigger change than four entries justify.
+//
+// A tool absent from this map cannot be called at all — validateTools stops
+// the run at startup. That matters more than it looks: a tool invoked without
+// its required arguments returns IsError, and callOnce folds IsError into the
+// same error count as broker unavailability, timeouts, and MCP-side rate
+// limiting. A misconfigured workload would otherwise complete and report a
+// high error rate that reads as a finding about the system under test.
+var toolRecipes = map[string][]string{
+	"get-broker-status": {},
+	"get-rdp-status":    {argVPN, argRDP},
+	"list-queues":       {argVPN},
+	"list-rdps":         {argVPN},
+}
+
+// validateTools rejects a tool list naming anything loadgen cannot build
+// arguments for: a tool with no recipe, or a recipe whose argument has no
+// value (an empty flag). Both are operator mistakes, and both must stop the
+// run rather than be measured.
+func validateTools(tools []string, argValues map[string]string) error {
+	known := make([]string, 0, len(toolRecipes))
+	for name := range toolRecipes {
+		known = append(known, name)
+	}
+	sort.Strings(known)
+
+	for _, tool := range tools {
+		recipe, ok := toolRecipes[tool]
+		if !ok {
+			return fmt.Errorf("no argument recipe for tool %q — cannot call it. Known tools: %s",
+				tool, strings.Join(known, ", "))
+		}
+		for _, arg := range recipe {
+			if argValues[arg] == "" {
+				return fmt.Errorf("tool %q needs argument %q but no value was given (see the corresponding flag)", tool, arg)
+			}
+		}
+	}
+	return nil
+}
+
+// callSpec is one prebuilt tool invocation: the tool name and the exact
+// argument map to send. Built once per client before the clock starts, so the
+// steady-state loop neither allocates a map per call nor performs a lookup
+// that could fail.
+type callSpec struct {
+	tool string
+	args map[string]any
+}
+
+// buildCallSpecs prepares one callSpec per tool in the rotation, for one
+// broker. Requires validateTools to have passed: every tool named here has a
+// recipe and every argument in that recipe has a value.
+func buildCallSpecs(tools []string, broker string, argValues map[string]string) []callSpec {
+	out := make([]callSpec, 0, len(tools))
+	for _, tool := range tools {
+		args := map[string]any{"broker": broker}
+		for _, arg := range toolRecipes[tool] {
+			args[arg] = argValues[arg]
+		}
+		out = append(out, callSpec{tool: tool, args: args})
+	}
+	return out
 }
 
 // sample is one recorded tool call. Zero-value err means success. Kept as a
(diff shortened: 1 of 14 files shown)


test/performance/loadgen/main.go
```diff
@@ -166,12 +273,14 @@ func run(ctx context.Context, cfg runConfig) error {
 
 	for i := 0; i < cfg.clients; i++ {
 		broker := cfg.brokers[i%len(cfg.brokers)]
+		// Arguments resolved here, off the clock. Each client gets its own
+		// specs because "broker" differs per client; nothing mutates them
+		// afterwards.
+		specs := buildCallSpecs(cfg.tools, broker, cfg.argValues)
 		wg.Go(func() {
```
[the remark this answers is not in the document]
helen-zhu-solace: 1. go.mod:3 declares go 1.25.0. Per-iteration loop variable scoping landed in Go 1.22 and is keyed off the module's language version, so each wg.Go closure gets its own i. The pre-1.22 hazard Copilot is describing cannot occur.
2. The code uses wg.Go (sync.WaitGroup.Go), which is Go 1.25 API. A toolchain old enough to have the loop-capture semantics could not compile this file at all.

The pattern is also deliberate: perClient is preallocated to cfg.clients precisely so each goroutine writes its own index without a mutex. Applying Copilot's suggestion (i := i) would be dead code that gopls/modernizers flag. No action.

test/performance/mock-semp/handler_test.go
```diff
@@ -0,0 +1,368 @@
...
+// TestUnpinnedRDPNameMisses is the negative case the story calls for: a
+// request for any RDP other than the pinned one must miss and 404, never be
+// served the pinned RDP's body. Serving it would make every measurement of
+// get-rdp-status silently valid-looking for RDPs that were never captured.
+func TestUnpinnedRDPNameMisses(t *testing.T) {
+	subs := []string{"", "/" + rdpSubQueueBindings, "/" + rdpSubRestConsumers}
+	for _, sub := range subs {
+		t.Run("sub="+sub, func(t *testing.T) {
+			h := newTestHandler(t)
+			path := privateVPNBase + "/restDeliveryPoints/rdp_999" + sub
```
solace-aross: TestUnpinnedRDPNameMisses only exercises rdp_999 against the pinned rdp_1. That name has no string-prefix relationship to rdp_1, so a future ==-to-HasPrefix regression on the name check (an easy typo/refactor mistake) would pass this suite unnoticed for a numerically-adjacent name like rdp_10 or rdp_100 — exactly the failure mode this PR calls out as the highest-risk one. Worth adding a prefix-adjacent case here (e.g. rdp_10 alongside rdp_999) so that regression can't slip through silently. Not blocking.

solace-aross (dismissed):
Verified the load-bearing claim in handler.go: swapped the exact RDP-name match for a HasPrefix match in a throwaway worktree and re-ran the suite. Five tests went red (TestUnpinnedRDPNameMisses x3, TestHitCountsPerRule, TestHitsResetLeavesMissesAlone), so the exact-match protection is real and genuinely tested, not just asserted in the PR description.

Also checked, against the code rather than the prose:
- validateTools ordering: ran the built binary with -validate-only and no -brokers/-broker-count against an unreachable MCP URL. Returned clean in under a second with no network stall. Without -validate-only, the same invocation fails immediately on the missing broker flag. Confirms the tool-arg check genuinely runs before broker resolution and needs nothing else.
- capture_pages stale-page cleanup: the rm -f clears every numbered page for a prefix before the capture loop writes anything, so a page count shrinking between captures (e.g. 3 pages to 1) can't leave an orphaned page 2/3 being served alongside a fresher, shorter chain.
- Hit-counter concurrency: 200-goroutine stress test against the handler under -race. No race, no lost increments.
- buildPagedRules generalization, isMonitorMsgVpnPath, and the buildChecks dedup in fidelity/main.go all check out — no hidden assumption true for one collection but not the other.

One non-blocking gap, left inline. Approving as-is; CI is green and this is test-harness code, not production.

test/performance/mock-semp/handler_test.go
helen-zhu-solace: Fixed in [df15109](https://github.com/SolaceProducts/solace-broker-mcp/pull/332/commits/df151097f87148092e513da6829093ffc0443a28)

solace-aross (dismissed):
Fix commit df15109 covers exactly the gap noted in the prior review. Verified by re-injecting the regression: sed-swapped the exact match for HasPrefix(name, pinned) in a throwaway worktree and reran TestUnpinnedRDPNameMisses. The unrelated-name case (rdp_999) still passed, as expected, but all three pinned-is-prefix/rdp_10 subtests went red exactly as the new test's own comment predicts. The mirror direction (prefix-of-pinned/rdp_) is also covered by the new table, though this sabotage didn't need to exercise it. Both directions of the ==-to-prefix regression are now genuinely caught, not just asserted. No further findings. Approving.

bczoma (commented):
Two recommendations from reviewing the RDP fixture work. Neither is blocking — the harness behaves correctly with valid inputs, the negative-case tests are all present, and `buildChecks` being unified is a nice fix. Both of these are cheap and I think worth landing before merge.

### 1. `regen-golden.sh:36` hardcodes a lab-specific RDP name

```bash
rdp_name="${RDP_NAME:-rdp_1}"
```

This is the pattern the PR description argues against elsewhere — the mock deliberately reads the pinned name out of `rdp_object.json` "rather than taken as a flag, so the fixture and the rule cannot name different RDPs." That guarantee holds downstream, but the script that *creates* the capture defaults to a literal.

What makes it worth fixing is the capture order in `capture.sh`: the five `show_*.xml` files (lines 85–95), `queues_page*` (140) and `rdps_page*` (144) are all written **before** `rdp_object.json` (156). So on a broker whose capture VPN has no RDP named `rdp_1`, the run overwrites every other fixture, then fails at line 156 with a bare `curl: The requested URL returned error: 404` and no mention of `RDP_NAME` — leaving a zero-byte `rdp_object.json` and a `fixtures.manifest` still describing the previous capture. The tree is inconsistent and needs a full recapture.

Suggestion: derive the name from the just-captured `rdps_page1.json` (first `restDeliveryPointName`), which matches how the mock resolves it. Failing that, preflight the RDP path and exit naming `RDP_NAME` before any file is overwritten.

### 2. Two comments state a guarantee the code doesn't provide

`mock-semp/handler.go:434` and `mock-semp/canned/capture.sh:151` both say a pinned RDP with more than 100 queue bindings or REST consumers "would announce itself as a miss rather than a truncated answer."

It wouldn't. `get-rdp-status` declares no `maxResults` parameter, so `resolveMaxResults` returns `defaultMax` (`internal/composite/executor.go:507` — 100), and `fetchPaginated` breaks at `executor.go:455` on `len(items) >= remaining` without ever issuing a cursor request. A busy RDP would yield a fixture and golden both holding 100 of N with `truncated: true`, the gate would stay green, and nothing would miss.

Secondary detail: `sempv2RDPPath` ignores the `cursor` query parameter (unlike `sempv2ListCursorEquals`), so even if a cursor request were made, the mock would re-serve page 1 rather than miss.

Comment-only fix, but worth correcting now — a false guarantee in a comment is exactly how a future recapture against a loaded RDP gets trusted.

solace-aross (approved):
f56c93a addresses both of bczoma's findings, not mine — read their review first per guardrail #4. Verified both technical claims independently rather than taking the review comment on faith:

1. fetchPaginated's early-return at executor.go:455 (len(items) >= remaining) is real — confirmed by reading the actual function. get-rdp-status genuinely has no maxResults, so a busy pinned RDP would silently truncate at 100 with truncated:true and the gate would stay green over a partial answer. The updated comments in handler.go and capture.sh now say this correctly instead of the prior 'would announce itself as a miss' claim.
2. sempv2RDPPath (handler.go) never reads r.URL.Query().Get("cursor") — confirmed by reading the function body. A cursor-bearing request to the pinned RDP path re-matches and re-serves page 1, exactly as bczoma's comment states.

Also verified the capture-order fix: the RDP_NAME preflight (capture.sh:108-117) now runs before any of the five SEMPv1 files or either paginated collection is written, so a bad explicit RDP_NAME fails clean with no fixture touched — simulated the derivation path (pin_rdp_name against a mock rdps_page1.json, and the manifest-side jq read against a mock rdp_object.json) and both resolve the same name, matching how mock-semp's own pinnedRDPName reads it back. No new issues. Approving.

PR #333, opened 2026-08-21
SOL-153363: make the broker token acquisition traceable in logs

amitmorade:
## Summary

The token exchange logged cache and IdP activity but nothing that tied it to a request. This adds the missing steps so one broker call reads as a sequence, and fixes the parts of the existing output that were unreadable or wrong.

Tracked under [SOL-153363](https://sol-jira.atlassian.net/browse/SOL-153363).

## Type of Change

Log output only. One line was actively misleading, hence "bug fix".

## Motivation and Context

This started with a developer who had worked on this feature being unable to read its own logs:

```json
{"msg":"token cache get","broker":"prod-us","status":1,"correlation_id":"01a0..."}
{"msg":"token cache put","broker":"prod-us","status":0}
{"msg":"token exchange succeeded","broker":"prod-us","exchange_elapsed":77496667}
```

Four problems, in increasing severity:

**`status` was an unlabeled enum.** `GetStatus` is `0=hit`/`1=miss`; `PutStatus` is `0=stored`/`1=dropped_ttl`. So `0` meant success on both lines while `1` flipped meaning between them. Both types had `String()`, but slog resolves `LogValuer` and never calls `String()`, so the raw iota was emitted.

**`status` is overloaded three ways in this tree** — a numeric HTTP status code (`Sender.errorHandler`, `Sender.checkRetry`), a success/error verdict as a string (`logToolResult`), and the cache outcome. A reader hitting `status: 1` cannot tell which.

**Durations rendered as raw nanoseconds.** slog's JSON handler encodes `time.Duration` as the `int64` it is underneath, and Go forbids adding a `LogValue` method to another package's type — so a duration is formatted explicitly at the call site or not at all.

**`exchange_elapsed` named the wrong span — the real defect.** The timer starts before singleflight dispatch and stops after the caller takes the result, so it covers deduplication waiting, the rate-limit gate, the circuit breaker, every retry attempt with its backoff, and **the cache write**. Only the cache read is outside it. A value can be three attempts plus backoff, or a rejection that never reached the IdP at all.

Then a broader gap: `OAuthAuthenticator.AddAuth` — the function that obtains a token and puts it on the outgoing request — contained **zero** log statements. The trace had no beginning and no end, and nothing marked the request to the IdP.

## Changes Made

Eight commits, each green on its own.

- **`27a3e00`** — readability. `LogValue()` on the two status enums so they print `hit`/`miss`/`stored`; `status` → `result`; readable durations on the exchange lines; `exchange_elapsed` → `exchange_total_elapsed` with a comment at the timer saying what it spans.
- **`632f783`** — `AddAuth` logs one line when it starts and one when it finishes, on every path. The cache lookup is split into one message per outcome. Four messages move from "broker credential" to "broker token" so the sequence names one thing throughout. Three now-unreachable methods deleted. Adds the integration test.
- **`d8c4631`** — `doExchange` logs before the IdP call and after a token comes back, carrying `http_status` and the attempt count. Adds `idpclient.AttemptsFromContext` so the count can be read where the line lives.
- **`cfa4c5a`** — trims the comments those commits added (−56 lines).
- **`ada85fd`** — moves the changelog entries back under `[Unreleased]` (see below).
- **`c9d150e`** — review fixes: drops two `time.Sleep` calls from the integration test that waited for a cache write which is synchronous and already awaited, and rewrites the changelog entry, which described a `result`/`hit`/`miss` design two revisions old.
- **`bc8c75d`** — a cancelled caller's trace now names the cause. `Exchange` returns a bare context error on its cancellation paths, so the closing line read `reason=unknown` for a cause that is known. Ordinary traffic: any client that disconnects mid-request hit it.
- **`488c141`** — completes the changelog's list of renamed messages, and the test file's description of what it checks.

Plus a merge of `origin/main` and one fix it exposed (below).

### The result

Captured from a live `AddAuth` against a fake IdP, not hand-written.

Cold path:

```json
{"level":"DEBUG","msg":"broker token needed","broker":"prod-us"}
{"level":"DEBUG","msg":"no cached broker token","broker":"prod-us"}
{"level":"DEBUG","msg":"requesting broker token from identity provider","broker":"prod-us"}
{"level":"DEBUG","msg":"identity provider issued broker token","broker":"prod-us","http_status":200,"attempts":1}
{"level":"DEBUG","msg":"broker token cached","broker":"prod-us"}
{"level":"DEBUG","msg":"broker token exchange completed","broker":"prod-us","exchange_total_elapsed":"2.048167ms"}
{"level":"DEBUG","msg":"broker token attached to request","broker":"prod-us"}
{"level":"INFO","msg":"tool invoked","tool":"list-queues","broker":"prod-us","status":"success","duration":210000000}
```

Cache hit — three lines, and the IdP lines are absent because no call was made:

```json
{"level":"DEBUG","msg":"broker token needed","broker":"prod-us"}
{"level":"DEBUG","msg":"using cached broker token","broker":"prod-us"}
{"level":"DEBUG","msg":"broker token attached to request","broker":"prod-us"}
```

Circuit breaker refused — no `requesting` line, which is how you tell it apart from a call that was made and failed:

```json
{"level":"DEBUG","msg":"broker token needed","broker":"prod-us"}
{"level":"DEBUG","msg":"no cached broker token","broker":"prod-us"}
{"level":"DEBUG","msg":"broker token unavailable","broker":"prod-us","reason":"token exchange circuit open"}
{"level":"ERROR","msg":"tool invoked",...,"breaker_state":"open",...}
```

## Decisions and trade-offs

**Fix the type, not the call sites.** `LogValue()` on the enum rather than `.String()` at each call site, so a future call site cannot reintroduce the integer. The first commit fixed the readability problem tree-wide with no call-site edits.

**A message names an outcome only where one call site has one outcome.** The first draft of the cache-lookup message was `"broker credential not cached, requesting from identity provider"`. That is **wrong**: the log call sits above the `GetHit` branch, so it fires on hit and miss alike — and on a hit it is the only line emitted for the whole warm path. It would have made the sole record of a successful cache hit claim the token wasn't cached.

Both cache sites are therefore split per outcome and neither carries a `result` field: `using cached broker token` / `no cached broker token` for the lookup, and `broker token cached` / `broker token not cached: …` for the write. A message that names what happened needs no field repeating it. The one surviving `result` is the cache-write site's default arm, whose message deliberately does *not* name an outcome because an unrecognized status means we cannot tell whether the write happened.

**`exchange_total_elapsed`, not `idp_roundtrip_elapsed`.** The rename was originally going to be the latter. Rejected after tracing the timer: "roundtrip" singular would be a lie, since the span includes retries, backoff, singleflight waiting and the cache write. `idp_roundtrip_elapsed=8.2s` reads as one HTTP call taking 8.2 seconds when it may be three attempts. Trading a vague-but-true name for a specific-and-false one is a regression — ambiguity prompts a code read, false precision prevents one.

**No duration on the IdP line.** `httpClient` is the retrying client, so one `Do` can be several attempts with backoff between them; a duration measured around it would silently include that backoff. The line carries an attempt count instead. Per-attempt timing needs the transport itself instrumented — SOL-153394.

**`AddAuth` logs its own failure, thinly.** The tool layer already reports every failure at ERROR with the full diagnostic set via `ExchangeError.LogAttrs()`. But `AddAuth` opening a trace that never closes is indistinguishable from a panic or a hang, so it logs a closer carrying the sentinel alone. It is also the only record when a 401 retry re-runs `AddAuth` and the tool call ultimately succeeds — the tool layer then logs INFO, and the failed acquisition would otherwise appear nowhere.

**Durations are formatted on the exchange lines only.** An earlier revision put the conversion in the slog handler, keyed on the value's kind, so it caught every duration the server emits — including `duration` on `tool invoked`. That is the wrong line to make unreadable-by-machine: production runs at INFO, so the rest of this trace is absent there and `tool invoked` is the duration a dashboard can actually reach, and `secure-logging-rules.md` lists `duration_ms` among its required fields. Breadth was the wrong instinct — the problem was narrow (nanoseconds are unreadable to a developer debugging a token exchange), so the fix is narrow. Five fields change: `exchange_total_elapsed`, `waited`, `retry_after`, `requested`, `clamped_to`. Every duration elsewhere is untouched.

## Testing

**Test Environment:**
- Go version: as per `go.mod`
- OS: darwin (macOS)
- Broker version: n/a — no broker interaction; IdP driven via `httptest`

**Tests Performed:**

**Test Coverage:**

`make check` green on the merged tree — build, vet, lint, race tests. **Total coverage 89.3%**, above the 85% floor.

New tests assert on output from a real slog handler, not method return values. That matters: a unit check of `String()` passed the entire time these logs were unreadable, so it could not have caught the original bug.

`test/integration/broker_token_trace_test.go` is new. It runs `AddAuth`, `Exchanger` and `TokenCache` together against a fake IdP and checks four things:

1. **No secret reaches a log line** — the real subject token, broker token and IdP client secret. `AddAuth`'s unit tests use a fake exchanger returning a placeholder, so there is no real secret there to leak.
2. **One start line and one finish line on every path**, with the finish matching the outcome.
3. **The finish line names its cause**, including on cancellation.
4. **The attempt count is real**, which needs the retrying client production wires rather than a plain `http.Client`.

Checks 1 and 2 deliberately ignore the wording of the lines in between, so rewording one does not break them. Checks 3 and 4 name a message, because they assert what that specific line carries.

## Breaking Changes

No API or config change. Log output only.

**Impact:** anything parsing these logs — the old message strings, `status` on the cache lines, `exchange_elapsed` (now `exchange_total_elapsed`), and the five token-exchange duration fields now emitted as strings. `duration` on `tool invoked` is unchanged and still numeric.

**Migration Guide:** read the five token-exchange duration fields as strings, parsing with `time.ParseDuration` if a number is needed. Update queries matching the old message strings, and the renamed duration key.

## Additional Notes

**A CHANGELOG defect the merge exposed.** v0.8.0 was cut while these entries sat under `[Unreleased]`, so the release promoted them into the `[0.8.0]` block — describing log output that tag does not produce (verified: none of these commits are ancestors of `v0.8.0`). Moved back under `[Unreleased]` in `ada85fd`.

**This is the first two flows of a larger trace, not the whole thing.** Still to come, each its own change: the singleflight follower has no line of its own, so a caller waiting on another's exchange looks like it did the work itself; per-attempt IdP timing needs `internal/idpclient` instrumented (SOL-153394); and of the 26 non-test log sites across these three packages, 9 use no context at all and 4 log under the detached `exchCtx` (SOL-153364), so neither group can carry a `correlation_id`.

**`scripts/annotate-cache-logs.sh` is broken by this and is not in this PR.** That directory is untracked — a local-only dev tool, not a versioned consumer. It exists to translate `status: 1` into "CACHE MISS" and divide nanoseconds by a million, which is the strongest evidence the problem was real.

## Related Issues/PRs

- Implements [SOL-153363](https://sol-jira.atlassian.net/browse/SOL-153363)
- Supersedes #324 (closed unmerged — same ticket, narrower scope)
- Follow-up: [SOL-153364](https://sol-jira.atlassian.net/browse/SOL-153364) — correlation IDs through the detached exchange context
- Follow-up: [SOL-153394](https://sol-jira.atlassian.net/browse/SOL-153394) — per-attempt IdP logging

---

**For Reviewers:**

**Focus Areas**:
1. The message wording. Both cache sites are split per outcome and carry no `result` field — does `no cached broker token` / `using cached broker token` read correctly, and likewise the two cache-write messages? A first draft got these wrong in opposite directions: false on one path, redundant on the other.
2. `exchange_total_elapsed` and the comment at the timer — is the enumeration of what it spans complete?
3. Duration formatting is now scoped to the five token-exchange fields, after review flagged that the handler-level version made `tool invoked`'s `duration` a string. Worth confirming that scope is right, and that nothing consumes those five numerically.

**Questions**:
- These are DEBUG lines, so healthy production emits none of them and `"tool invoked"` remains the operational event. If support actually relies on this trace during escalations, the level is arguably wrong too — a broader observability question I did not fold in here. Views welcome.


solace-awaheed (commented):
Reviewed on the merged tree: `go build ./...` clean, affected package tests pass
(587 tests / 6 packages). The change is well-constructed — small diff, reasoning
documented at each site, and the tests assert on real handler output rather than
method return values, which is the right call given the original bug survived a
passing `String()` test. Four substantive points, then some minor ones.

---

### 1. `internal/tokenexchange/exchange.go:66` — the cache-hit line fires before the cancellation re-check

```go
} else {
    slog.DebugContext(ctx, "using cached broker token", ...)   // line 66
    // Re-check ... a caller cancelled during the lookup still leaves with a
    // usable credential.
    if err := ctx.Err(); err != nil {
        return nil, err                                        // line 74
    }
```

If the caller cancels in the window between the entry guard and this re-check —
the window whose existence is the entire justification for the re-check — the
trace reads:

```
broker token needed
using cached broker token
broker token unavailable  reason=context canceled
```

No token was used. This is the same defect class the PR description sets out to
fix (a cache-path message asserting something that didn't happen), just inverted
from the draft it describes. Moving the log below the guard fixes it:

```go
} else {
    if err := ctx.Err(); err != nil {
        return nil, err
    }
    slog.DebugContext(ctx, "using cached broker token",
        slog.String("broker", input.BrokerAlias))
    return &Token{...}, nil
}
```

A cancelled caller then emits no cache-outcome line, which is correct — the
closer already names the cause. The `no cached broker token` branch is fine
as-is.

### 2. `exchange.go:414`, `:442`, `:270` — three of the seven cold-path lines cannot carry a `correlation_id`

All three log under `exchCtx`, which `runExchangeOnce` roots at
`context.Background()`. So of the cold-path trace in the description, lines 3–5
(`requesting…`, `identity provider issued…`, `broker token cached`) have no
correlation ID, and the sequence only reconstructs by broker alias plus
timestamp proximity — which on a multi-session server with concurrent callers to
the same broker is exactly where it stops working.

The PR discloses this and defers to SOL-153364, which is fair for the
pre-existing lines. But the two IdP lines are new here, and they're the ones a
reader most needs to attribute. `correlation.With(exchCtx, correlation.From(ctx))`
copies the ID without re-coupling cancellation, so it doesn't reintroduce the
problem the detach exists to prevent — it needs the ID threaded into
`runExchangeOnce` (a string param, or a field on `input`), not the caller's
context.

Worth deciding explicitly: either do it here, or note in the description that
the middle three lines carry no correlation ID, so the sample traces aren't read
as joinable. The singleflight-winner-attribution question (whose ID does a
shared exchange log?) is a real argument for deferring — but it should be the
stated reason.

### 3. `internal/oauth/cache/cache.go:64` — `GetStatus` lost `Level()` and never got `LogValue()`

`PutStatus` gained `LogValue()`; `GetStatus` only had `Level()` removed. Three
consequences:

- A future site logging a bare `GetStatus` emits the raw iota — precisely the
  bug being fixed for `PutStatus`, and `GetHit` is `0`, so it reads as a success
  code.
- `cache_test.go:656` says "symmetric with `TestGetStatus_LogValue`" — that test
  doesn't exist.
- The description claims `LogValue()` on "the two status enums".

Separately, `GetResult.LogValue` (`cache.go:87`) and `PutResult.LogValue`
(`cache.go:128`) still emit the nested key `status` — the overloaded name this
PR is retiring on these lines. `result` would be consistent.

### 4. `cmd/server/main.go:83` — the duration change conflicts with the repo's own documented convention

Taking the third focus area: the concern isn't consumers. I grepped, nothing
in-repo reads a duration field numerically, and the CHANGELOG calls the break
out properly. It's the direction.

`docs/internal/secure-logging-rules.md:31` specifies tool-invocation logs carry
**`duration_ms`**, and the `check-logs` (L-01) and `add-logs` checklists both
repeat it. Milliseconds, as a number. This change makes `duration` on
`tool invoked` (`internal/tools/manager.go:355,406`) — the primary operational
event, the one most likely to be graphed or percentiled — a non-numeric string,
so no log backend can aggregate it without a parse step.

Two separable questions:

- **If the convention stands**, the fix is `slog.Int64("duration_ms", d.Milliseconds())`
  at the call sites rather than a handler-level type conversion — numeric and
  readable at once.
- **If the convention is being changed**, that's a deliberate decision, and
  `secure-logging-rules.md` plus the two skill checklists should change with it
  in this PR. Right now the code and the rules disagree silently.

A narrower option if the goal is just DEBUG readability: leave `tool invoked`
alone and format only within the exchange lines, keeping the blast radius inside
the feature this PR owns.

---

amitmorade:
Thanks @solace-awaheed  — all four were real. Taking them in order.

**1. Cache-hit line above the re-check — fixed (`b9dd0c3`).** Moved below the guard, so it only fires on the path that returns a token. Worth noting my first test was useless: cancelling before `AddAuth` hits the entry guard and never reaches the cache path, so it passed against the unfixed code. The committed one cancels from inside a `TokenCache` wrapper's `Get`, and fails without the fix.

**2. Correlation IDs — deferring, reason on SOL-153364.** The blocker is the one you'd expect: a singleflight winner's exchange is shared by every waiter, so copying the ID onto `exchCtx` implicitly picks the winner's, which is arbitrary from any waiter's view. Worth deciding deliberately, not as a side effect here. PR body now says the middle three lines carry no correlation ID, so the samples aren't read as joinable. Counts corrected while I was there: 26 non-test sites, 4 detached-`exchCtx`, 9 context-free — the nine aren't ticketed anywhere.

**3. `GetStatus` asymmetry — fixed (`b9dd0c3`).** `LogValue()` and its test restored, so the comment referencing that test is true again.

On the `GetResult`/`PutResult` inner key I'd rather not, and it's the one thing here I'm pushing back on. Those methods emit nothing today — nothing logs the wrapper — so that `status` is read by someone already in the file, who can see the method before deciding. `status: 1` in output was read by someone with no access to the code. Different class. Happy to change it if you disagree; it's two words.

**4. Duration change — narrowed (`c761beb`).** Your framing as *direction* rather than consumers is right. Formatting moved to the five token-exchange call sites; `redactSecretAttr` is back to redaction only and `cmd/server/main.go` is byte-identical to `main`. `tool invoked` is numeric again.

I took the narrow option over `duration_ms` because the whole trace is DEBUG and absent from production — `tool invoked` at INFO is the one duration a dashboard can reach, so it was exactly the wrong one to convert. The `duration`/`duration_ms` gap predates this branch and is left alone.

```json
{"level":"DEBUG","msg":"broker token exchange completed","exchange_total_elapsed":"1.821042ms"}
{"level":"INFO","msg":"tool invoked","status":"success","duration":210000000}

solace-awaheed (dismissed):
Thanks for the changes!

solace-lootii (dismissed):
Reviewed on top of c761beb4, after the fixes for solace-awaheed's review.

I agree with solace-awaheed on all four points, and the two follow-up commits address them cleanly:

#1 (cache-hit line above the cancellation re-check) — confirmed fixed in b9dd0c3; the log now sits below the ctx.Err() guard and only fires on the path that returns a token.
#3 (GetStatus missing LogValue()) — confirmed fixed in b9dd0c3, symmetric with PutStatus now.
#4 (handler-level duration conversion) — confirmed narrowed; cmd/server/main.go and redact_test.go are byte-identical to main again, so tool invoked's duration is untouched.
#2 (correlation IDs) — the deferral to SOL-153364 is reasonable given the singleflight-attribution problem raised; the PR body now says explicitly that the three middle lines carry no correlation ID, so I don't think this needs to block.

Two veryy small things I noticed that aren't covered above:

CHANGELOG.md (this is really not important) — the new [Unreleased] entry leaves a double blank line before ## [0.8.0] (line 23). Every other release header in the file is preceded by exactly one blank line, so this is a one-line formatting inconsistency introduced by this PR.

test/integration/broker_token_trace_test.go:114 and :520 — both newTraceAuthenticator and Test_BrokerTokenTrace_CancelledDuringCacheHitClaimsNothing call cache.NewTokenCache directly with a manual t.Cleanup. That's not a bug — the cleanup is correct — but internal/oauth/cache/cachetest exists specifically for this and its doc comment says "Prefer these over calling cache.NewTokenCache directly in tests." Might be worth switching to cachetest.WithConfig for consistency with the rest of the integration suite.

Approving anyways. Up to you to make the changes or leave as is!

amitmorade:
Both fixed in `d5b29b9`.

**Blank line** — you're right, and it was mine: the `[Unreleased]` insertion left two where every other release header has one.

**`cachetest.WithConfig`** — switched, both sites. Worth more than consistency, as its package doc points out: the bare constructor leaks Otter's sweeper goroutine and the retained cache with it unless every caller plumbs `Close` through cleanup. Doing it by hand was correct here, but the helper makes it impossible to get wrong later. Good catch

solace-lootii (approved):
LGTM now! Thanks for the quick fix

PR #336, opened 2026-08-25
Calibrate the changelog skill's length guidance, and check the draft

amitmorade:
## Summary

Three changes to the `/changelog` skill, each from a mistake the current guidance either allowed or actively caused while writing the entries for #333.

## Type of Change

- [x] Documentation update

Tooling only. No production code.

## Motivation and Context

**The length guidance points the wrong way.** Step 2 says an entry's "length scales with impact" and directs the drafter to the longest paragraphs in the file as "the fullest voice examples." Followed literally, that produces entries several times longer than the file's own norm — the two written for #333 ran to 225 and 138 words, against a typical entry of well under fifty. Most of that was explaining *how* the change worked, which belongs in a code comment.

**Nothing told the drafter to check its own output.** One entry named a `result` field carrying `hit`/`miss` — a design a later revision had replaced. It survived until a reviewer caught it. A grep per named identifier would have found it immediately.

**Nothing said what to do when commits land after the entry.** The #333 entry was patched twice as later commits arrived, and still missed a fifth renamed message while promising that "anything matching the old messages needs updating." A partial list reads as complete, which is worse than a vague sentence.

## Changes Made

- **Step 2** — length guidance calibrates against the file as a whole rather than its longest examples, and states where *why* and *how* belong instead.
- **Step 6 (new)** — after drafting, grep every identifier the entry names to confirm it exists as written, and re-read the diff for anything observable the entry omits. Where an entry claims a category, list all of it.
- **Rules** — an entry describes the branch's whole diff against `main`, not one commit. Commits landing after it is drafted mean re-deriving from `main..HEAD`, not patching.

+28/−4 on one file.

## Testing

**Tests Performed:**
- [x] Self-review completed

No code paths touched, so nothing to run. The guidance was derived from the entries in #333 and checked against `CHANGELOG.md`'s actual contents rather than from recollection.

## Additional Notes

The Step 2 change **contradicts guidance that is there today**, so it is the part worth disagreeing with if you think the long paragraphs in `[0.5.0]` and `[0.3.0]` are the target rather than the exception. My read is that they are exceptional by design — both are breaking changes that genuinely need old behavior, new behavior, and a migration path — and that pointing every drafter at them as the voice is what produces bloat elsewhere.

An earlier revision of this change embedded an `awk` one-liner for measuring the median entry length. Removed: no other skill in this repo embeds a shell command, and `awk`/`sort` are not available on a stock Windows shell.

## Related Issues/PRs

- Prompted by the changelog entries in #333 (SOL-153363)

### DIFF
--- a/.claude/skills/changelog/SKILL.md
+++ b/.claude/skills/changelog/SKILL.md
@@ -44,18 +44,19 @@ the PR like any other change.
 ### Step 2: Learn the house voice
 
 Read `CHANGELOG.md` before drafting: the `[Unreleased]` section for what is
-pending, and the recent version blocks for the fullest voice examples — the detailed
-breaking-change paragraphs in `[0.5.0]` and `[0.3.0]`, and the migration **tables** in
-`[0.3.0]`. Match what you see, do not impose a generic style:
+pending, and `[0.5.0]` and `[0.3.0]` for the **structure** of a breaking-change entry
+and the **form** of a migration table. Take structure from them, not length — see the
+length bullet below. Match the conventions here, do not impose a generic style:
 
 - Keep a Changelog categories, in this order: **Added**, **Changed**, **Deprecated**,
   **Removed**, **Fixed**, **Security**. Added/Changed/Removed/Fixed are the common
   ones; use Deprecated and Security when the change is genuinely one (see Step 3).
-- Each entry is **one list item (bullet) per logical change**, and most are one or
-  two sentences. Length scales with impact, but the bar for a full paragraph is
-  high: it is for a breaking change that genuinely needs old behavior, new behavior,
-  and a migration path. Calibrate against the file as a whole — a typical entry is
-  far shorter than the longest ones, which are outliers rather than the voice.
+- Each entry is **one list item (bullet) per logical change**. Target one or two
+  sentences (~50 words); reserve a full paragraph for a breaking change that needs
+  old behavior, new behavior, and a migration path. **This is a deliberate change
+  from what the file currently shows.** The `[0.6.0]`–`[0.8.0]` entries run to a
+  median of 93–191 words — do not match their length. The older blocks (`[0.1.0]`
+  through `[0.5.0]`, medians 4–16) are the target.
 - Say what changed and what breaks. **Why** it changed belongs in the commit message,
   and **how** it works belongs in a code comment; an entry that runs long is usually
   carrying one of those. A reader of this file is updating their queries, dashboards,
@@ -120,7 +121,7 @@ into one sentence.
   version blocks — that is the release process's job (today it is the manual
   `## Release Process` in `CHANGELOG.md` / `RELEASING.md`; automating the
   release-notes link is tracked separately). Do not commit, push, or tag.
-- Print the resulting diff and stop for human review.
+- Print the resulting diff.
 
 ### Step 6: Check the draft against the tree
 
@@ -131,11 +132,16 @@ rather than as shipped, after a later revision moved it.
 - **Every identifier the entry names must exist as written** — log message, field,
   config key, flag, tool name. Grep each one. A name that no longer exists sends a
   reader looking for something that was renamed mid-branch.
-- **Anything observable in the diff that the entry omits.** Re-read the diff for
+- **Nothing observable in the diff may be silently omitted.** Re-read the branch
+  diff (`git diff main...HEAD`, from Step 1) for
   changed strings and fields and confirm each is either covered or deliberately out
   of scope. Where an entry claims a category — "the old messages are replaced" —
   list **all** of them: a partial list reads as complete, which is worse than a
   vague sentence.
+- If an identifier is not found, correct it from the tree; if you cannot resolve it,
+  flag it to the human rather than guessing.
+
+Then stop for human review.
 
 ## Rules
 
@@ -143,7 +149,8 @@ rather than as shipped, after a later revision moved it.
 - Never invent a SOL ticket number — use the `SOL-????.` placeholder and flag it.
 - One bullet per logical change; omit pure test/refactor/docs churn.
 - An entry describes the branch's whole diff against `main`, not any single commit.
-  If commits land after the entry is drafted, **re-derive from `main..HEAD` and
-  rewrite** rather than patching the existing text — patching assumes you already
-  know what those commits changed, which is the assumption Step 1 exists to avoid.
+  If commits land after the entry is drafted, **re-derive and rewrite** — `git diff
+  main...HEAD` for the files and `git log main..HEAD` for the commits, as in Step 1 —
+  rather than patching the existing text. Patching assumes you already know what
+  those commits changed, which is the assumption Step 1 exists to avoid.
 - If nothing user- or operator-visible changed, say so and write nothing.
--- a/.claude/skills/changelog/SKILL.md
+++ b/.claude/skills/changelog/SKILL.md
@@ -51,10 +51,15 @@ breaking-change paragraphs in `[0.5.0]` and `[0.3.0]`, and the migration **table
 - Keep a Changelog categories, in this order: **Added**, **Changed**, **Deprecated**,
   **Removed**, **Fixed**, **Security**. Added/Changed/Removed/Fixed are the common
   ones; use Deprecated and Security when the change is genuinely one (see Step 3).
-- Each entry is **one list item (bullet) per logical change**, whose length scales
-  with impact: a trivial Added item is one line (e.g. "Apache 2.0 LICENSE file"); a
-  breaking change is a full multi-sentence paragraph explaining old behavior, new
-  behavior, and why.
+- Each entry is **one list item (bullet) per logical change**, and most are one or
+  two sentences. Length scales with impact, but the bar for a full paragraph is
+  high: it is for a breaking change that genuinely needs old behavior, new behavior,
+  and a migration path. Calibrate against the file as a whole — a typical entry is
+  far shorter than the longest ones, which are outliers rather than the voice.
+- Say what changed and what breaks. **Why** it changed belongs in the commit message,
+  and **how** it works belongs in a code comment; an entry that runs long is usually
+  carrying one of those. A reader of this file is updating their queries, dashboards,
+  or config — not learning the design.
 - Breaking changes are prefixed `- **BREAKING**: `. `**BREAKING**` is orthogonal to
   the category — a breaking change still files under Changed/Removed/etc. by its
   nature, and signals a SemVer MAJOR bump for the release/version decision.
@@ -117,9 +122,28 @@ into one sentence.
   release-notes link is tracked separately). Do not commit, push, or tag.
 - Print the resulting diff and stop for human review.
 
+### Step 6: Check the draft against the tree
+
+Two checks, both against the code rather than against what the change was meant to
+do. They exist because the common failure is describing the design as intended
+rather than as shipped, after a later revision moved it.
+
+- **Every identifier the entry names must exist as written** — log message, field,
+  config key, flag, tool name. Grep each one. A name that no longer exists sends a
+  reader looking for something that was renamed mid-branch.
+- **Anything observable in the diff that the entry omits.** Re-read the diff for
+  changed strings and fields and confirm each is either covered or deliberately out
+  of scope. Where an entry claims a category — "the old messages are replaced" —
+  list **all** of them: a partial list reads as complete, which is worse than a
+  vague sentence.
+
 ## Rules
 
 - Never place secrets, tokens, or credentials in an example.
 - Never invent a SOL ticket number — use the `SOL-????.` placeholder and flag it.
 - One bullet per logical change; omit pure test/refactor/docs churn.
+- An entry describes the branch's whole diff against `main`, not any single commit.
+  If commits land after the entry is drafted, **re-derive from `main..HEAD` and
+  rewrite** rather than patching the existing text — patching assumes you already
+  know what those commits changed, which is the assumption Step 1 exists to avoid.
 - If nothing user- or operator-visible changed, say so and write nothing.
--- a/.claude/skills/changelog/SKILL.md
+++ b/.claude/skills/changelog/SKILL.md
@@ -56,7 +56,7 @@ length bullet below. Match the conventions here, do not impose a generic style:
   old behavior, new behavior, and a migration path. **This is a deliberate change
   from what the file currently shows.** The `[0.6.0]`–`[0.8.0]` entries run to a
   median of 93–191 words — do not match their length. The older blocks (`[0.1.0]`
-  through `[0.5.0]`, medians 4–16) are the target.
+  through `[0.5.0]`, medians 4–21) are the target.
 - Say what changed and what breaks. **Why** it changed belongs in the commit message,
   and **how** it works belongs in a code comment; an entry that runs long is usually
   carrying one of those. A reader of this file is updating their queries, dashboards,
--- a/.claude/skills/changelog/SKILL.md
+++ b/.claude/skills/changelog/SKILL.md
@@ -51,12 +51,11 @@ length bullet below. Match the conventions here, do not impose a generic style:
 - Keep a Changelog categories, in this order: **Added**, **Changed**, **Deprecated**,
   **Removed**, **Fixed**, **Security**. Added/Changed/Removed/Fixed are the common
   ones; use Deprecated and Security when the change is genuinely one (see Step 3).
-- Each entry is **one list item (bullet) per logical change**. Target one or two
-  sentences (~50 words); reserve a full paragraph for a breaking change that needs
-  old behavior, new behavior, and a migration path. **This is a deliberate change
-  from what the file currently shows.** The `[0.6.0]`–`[0.8.0]` entries run to a
-  median of 93–191 words — do not match their length. The older blocks (`[0.1.0]`
-  through `[0.5.0]`, medians 4–21) are the target.
+- Each entry is **one list item (bullet) per logical change**. Most land in 15–30
+  words; ~50 words is the ceiling, reserved for a breaking change that needs old
+  behavior, new behavior, and a migration path. **This is a deliberate change from
+  what the file currently shows.** The `[0.6.0]`–`[0.8.0]` entries run to a median
+  of 93–191 words — do not match their length.
 - Say what changed and what breaks. **Why** it changed belongs in the commit message,
   and **how** it works belongs in a code comment; an entry that runs long is usually
   carrying one of those. A reader of this file is updating their queries, dashboards,
@@ -132,14 +131,15 @@ rather than as shipped, after a later revision moved it.
 - **Every identifier the entry names must exist as written** — log message, field,
   config key, flag, tool name. Grep each one. A name that no longer exists sends a
   reader looking for something that was renamed mid-branch.
-- **Nothing observable in the diff may be silently omitted.** Re-read the branch
-  diff (`git diff main...HEAD`, from Step 1) for
-  changed strings and fields and confirm each is either covered or deliberately out
-  of scope. Where an entry claims a category — "the old messages are replaced" —
-  list **all** of them: a partial list reads as complete, which is worse than a
-  vague sentence.
+- **Nothing observable in the diff may be silently omitted.** Re-read the Step 1
+  change set for changed strings and fields and confirm each is either covered or
+  deliberately out of scope. Where an entry claims a category — "the old messages
+  are replaced" — list **all** of them: a partial list reads as complete, which is
+  worse than a vague sentence.
 - If an identifier is not found, correct it from the tree; if you cannot resolve it,
   flag it to the human rather than guessing.
+- If either check changed the entry, re-print the diff — the human must see the
+  corrected version, not the one from Step 5.
 
 Then stop for human review.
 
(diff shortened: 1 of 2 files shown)


bczoma:
## Review

Tooling-only change. All three edits target real failure modes observed in #333, and the diagnosis in the PR body is honest. Nothing blocks merge. Two findings initially filed as critical were downgraded after adversarial re-checking — both survive at major, and both point at the same weakness: the new guidance was added *around* an unchanged intro paragraph and an unchanged terminal instruction, and neither was adjusted to fit.

---

### 🟡 Step 6 sits after Step 5's handoff sentence — `.claude/skills/changelog/SKILL.md:123` vs `:125`

Step 5 ends `- Print the resulting diff and stop for human review.` and the new `### Step 6: Check the draft against the tree` follows it.

This is not "unreachable" — a `SKILL.md` is loaded whole into context rather than executed as a halting program, and the file's own framing (lines 12–13 and 122, *"never commits… Do not commit, push, or tag"*) scopes "stop" to *hand off*, not *halt*. Two real costs remain:

1. `stop` is a strong end-of-turn trigger, and nothing recovers the checks if a run ends there. The `## Rules` block restates the new re-derive rule but **not** the two checks, and there is no step overview to re-anchor the sequence.
2. Even when Step 6 does run, the human is handed a diff that hasn't been checked yet — and any fix Step 6 finds invalidates the diff already printed.

**Fix:** either renumber so the checks come before the insert step, or cut `and stop for human review` from line 123 and end Step 6 with it. Either way, the handoff sentence should be the file's last instruction.

This is the one I'd want before merge; the rest below are cheap text edits.

---

### 🟡 "outliers" is measurably wrong, and the unchanged intro pulls the other way — `SKILL.md:46-49, 57-58`

The new text calls the long entries *"outliers rather than the voice."* Measured against `CHANGELOG.md` (two independent parsers, table rows and fences stripped as a control — the file contains zero fenced blocks, so no inflation):

| Block | Bullets | Median words | ≥50 words |
|---|---|---|---|
| `[0.8.0]` | 15 | **130** | 13 |
| `[0.7.0]` | 16 | **191** | 16 |
| `[0.6.0]` | 37 | **93** | 24 |
| `[0.5.0]` | 22 | 15.5 | 4 |
| `[0.3.0]` | 17 | 21 | 4 |
| **All** | **171** | **23** | **63 (37%)** |

Long entries are the voice of the last three releases.

What I'd push back on from an earlier draft of this review: the guidance is *not* self-defeating. `Match what you see` ends in a colon heading an enumerated convention list — which now contains the new length limit — and it names `[0.5.0]`/`[0.3.0]`, the two *shortest* blocks. It doesn't point a drafter at the long ones. And the overall median of 23 words is exactly what "calibrate against the file as a whole" resolves to.

What does survive: a factual claim a drafter can disprove by scrolling up; a stale `the recent version blocks` label on `[0.5.0]`/`[0.3.0]` when current is `[0.8.0]`; and `the fullest voice examples` steering on length two lines above a bullet restraining length.

**Fix** — frame it as a deliberate style change with a checkable number:

> Target one or two sentences (≈50 words); a full paragraph is reserved for a breaking change needing old behavior, new behavior, and a migration path. The `[0.6.0]`–`[0.8.0]` entries run far longer than this — do not match their length.

And in the intro: drop "recent" from the `[0.5.0]`/`[0.3.0]` pointer, and re-scope "fullest voice examples" to breaking-change *structure* and migration-table *form* rather than fullness.

---

### 🟡 `main..HEAD` is the wrong range for a diff — `SKILL.md:146`

The rule's first sentence frames the entry as describing "the branch's whole **diff** against `main`", then says re-derive from `main..HEAD`. Step 1 (`:31`) correctly uses `git diff main...HEAD` for files and `git log main..HEAD` for commits, and `:36` warns specifically about the three-dot form.

**Fix:** *"re-derive from `main...HEAD` (diff) plus `main..HEAD` (log), as in Step 1, and rewrite"* — name both, since the log side genuinely wants two dots. (Copilot's comment on this is correct.)

---

### 🟡 Sister skill asserts the premise this PR reverses — `.claude/skills/cut-release/SKILL.md:149,155`

It calls the CHANGELOG block "**deliberately verbose**" and "The **verbose** CHANGELOG remains the source of truth." That's the house rule this PR is changing, so the two skills now disagree.

**Fix:** drop "deliberately verbose"/"verbose", or point at the changelog skill's Step 2 for current length guidance.

---

### 🟢 "Re-read the diff" is ambiguous — `SKILL.md:134-135`

Two diffs are live at that point: the branch diff from Step 1, and the `CHANGELOG.md` edit diff printed at `:123`. Against the latter the check is a no-op. Suggest *"Re-read the branch diff (`git diff main...HEAD` from Step 1)"*.

### 🟢 Second Step 6 bullet is a noun phrase where the first is imperative — `SKILL.md:134`

In an instruction file a bolded noun phrase reads as a heading rather than a required action. *"**Nothing observable in the diff may be silently omitted.**"* (Copilot's point, and correct on the facts.)

### 💡 `SKILL.md:57` — "calibrate against the file as a whole" is the file's only directive with no test

The rest of the file is unusually procedural (grep this, run that range, emit a table iff ≥2 mappings). Folded into the length finding above — give it a number.

### 💡 `SKILL.md:131-133` — "Grep each one" gives no failure action

Detection without a response. Suggest appending: *"If an identifier is not found, correct it from the tree — or, if you cannot resolve it, flag it to the human rather than guessing."*

---

### What's good here

- Step 6's framing gives a *reason* ("the common failure is describing the design as intended rather than as shipped") rather than a bare rule — that measurably improves compliance.
- The first Step 6 bullet is properly executable: an enumerated identifier class plus a concrete verb.
- The what/why/how split at `:59-62` is a genuinely good heuristic — it turns "too long" into a diagnosable cause.
- The new Rules bullet closes a real failure mode and correctly points back at Step 1 as the reason.
- Single clean commit, and the PR body flags its own most contestable claim rather than burying it.

On the "worth disagreeing with" invitation in the PR body: I agree `[0.5.0]`/`[0.3.0]` are the right target, and the measurements above back that up. My disagreement is narrower — the change reads as an *observation* about the file when it's actually a *correction* to three releases of drift, and saying so plainly (with a word count) makes it much harder for a future drafter to talk itself out of.

**Recommendation:** approve with minor changes — the Step 6 placement first, the rest at your discretion.

amitmorade:
Thanks @bczoma  — the Step 6 placement was the real defect and I'd missed it. All findings addressed in `9226bc8`.

**Step 6 after the handoff.** Fixed — Step 5 just prints, and "Then stop for human review" is now the last line of Step 6. Your second cost was the one I hadn't considered: even when Step 6 runs, the human already has an unchecked diff, and any fix invalidates it.

**"Outliers" is wrong.** Reproduced your numbers: `[0.6.0]` 93, `[0.7.0]` 191, `[0.8.0]` 129, overall median 15. Long entries are the voice of the last three releases. Taken your framing — it now states plainly that this is a change from what those blocks show, names them, and gives the older blocks (medians 4–16) as the target. You're right that a correction is harder to talk oneself out of than an observation.

Also rescoped the intro: `[0.5.0]`/`[0.3.0]` are cited for breaking-change structure and migration-table form, explicitly not length.

**`main..HEAD` for a diff.** Fixed, both ranges named.

**`cut-release`.** Both "verbose" claims dropped. Its own instruction to condense to 1–2 sentences is this same rule one step later — the only wrong part was assuming the CHANGELOG it condenses from is verbose.

**Nits.** All three taken.

.claude/skills/changelog/SKILL.md
```diff
@@ -115,11 +121,36 @@ into one sentence.
...
+### Step 6: Check the draft against the tree
+
+Two checks, both against the code rather than against what the change was meant to
+do. They exist because the common failure is describing the design as intended
+rather than as shipped, after a later revision moved it.
+
+- **Every identifier the entry names must exist as written** — log message, field,
+  config key, flag, tool name. Grep each one. A name that no longer exists sends a
+  reader looking for something that was renamed mid-branch.
+- **Nothing observable in the diff may be silently omitted.** Re-read the branch
```
solace-awaheed: Step 6's omission check pins one git range, but Step 1 supports two — branch mode and an explicit <git-range>, which is required when drafting from main. On main, main...HEAD is empty, so the check silently passes over nothing. That's the exact no-op Step 1 warns about four lines into the file.
→ Fix: say "the Step 1 change set" instead of naming one range. Same wording issue at :152-153, lower severity.

.claude/skills/changelog/SKILL.md
```diff
@@ -44,17 +44,23 @@ the PR like any other change.
...
+length bullet below. Match the conventions here, do not impose a generic style:
 
 - Keep a Changelog categories, in this order: **Added**, **Changed**, **Deprecated**,
   **Removed**, **Fixed**, **Security**. Added/Changed/Removed/Fixed are the common
   ones; use Deprecated and Security when the change is genuinely one (see Step 3).
-- Each entry is **one list item (bullet) per logical change**, whose length scales
-  with impact: a trivial Added item is one line (e.g. "Apache 2.0 LICENSE file"); a
-  breaking change is a full multi-sentence paragraph explaining old behavior, new
-  behavior, and why.
+- Each entry is **one list item (bullet) per logical change**. Target one or two
```
solace-awaheed: The bullet says [0.1.0]–[0.5.0] have medians of 4–16. Measured, they're 12–33 — the 4 is [0.0.1] (outside the range cited), and the 16 ceiling excludes two of the five blocks named. bczoma's own table already showed [0.3.0] at 21. Separately, ~50 words at :55 and 4–16 at :59 are both called the target, four lines apart.
→ Fix: drop the census figures, give one target — "most land in 15–30 words; ~50 is the ceiling."

.claude/skills/changelog/SKILL.md
```diff
@@ -115,11 +121,36 @@ into one sentence.
...
+- **Every identifier the entry names must exist as written** — log message, field,
+  config key, flag, tool name. Grep each one. A name that no longer exists sends a
+  reader looking for something that was renamed mid-branch.
+- **Nothing observable in the diff may be silently omitted.** Re-read the branch
+  diff (`git diff main...HEAD`, from Step 1) for
+  changed strings and fields and confirm each is either covered or deliberately out
+  of scope. Where an entry claims a category — "the old messages are replaced" —
+  list **all** of them: a partial list reads as complete, which is worse than a
+  vague sentence.
+- If an identifier is not found, correct it from the tree; if you cannot resolve it,
```
solace-awaheed, line 139: Step 5 prints the diff, Step 6 may correct the entry, nothing re-prints. If the checks fire, the human is handed the pre-correction version.
→ Fix: append "if either check changed the entry, re-print the diff" before the handoff line.

solace-aross (dismissed):
Independently re-ran the word-count claim rather than trusting the thread: computed per-version-block median bullet length straight from CHANGELOG.md. Got 93 / 191 / 129 for [0.6.0]/[0.7.0]/[0.8.0] and a median-of-medians in the low teens for [0.1.0]-[0.4.0] — matches what bczoma and amitmorade converged on. The guidance's core claim (recent blocks are the outlier, not the voice) holds.

One small inaccuracy survives in the merged text: Step 2 cites '`[0.1.0]` through `[0.5.0]`, medians 4–16' as the target range, but [0.3.0]'s median is 21 by my count, outside that stated band. Doesn't change the guidance or the fix's substance — [0.3.0] is still short relative to [0.6.0]-[0.8.0] either way — flagging for a future pass rather than blocking on it.

Step 6 placement (the one bczoma called out as pre-merge-worthy) is fixed: the checks now run before 'stop for human review', which was the real defect. `main...HEAD`/`main..HEAD` are both named correctly per Step 1's convention. `cut-release`'s stale 'verbose' claim is gone.

Tooling-only change to `.claude/skills/`, no production surface, no CHANGELOG entry needed. Approving.

bczoma (commented):
## PR Review: #336 — Calibrate the changelog skill's length guidance, and check the draft

**Author:** amitmorade | **Branch:** `amorade/changelog-skill-length` → `main` | **Changes:** +41 / -10 across 2 files

### Summary

This recalibrates the `/changelog` skill's entry-length guidance (previously pointed drafters at the CHANGELOG's longest, least-representative entries as "the voice"), adds a new Step 6 that greps drafted identifiers against the tree and re-checks the diff for omitted observable changes, and fixes a git-range error. This is a re-review after substantial prior feedback (from me, posted informally as a comment rather than a formal review, plus solace-aross's independent verification and approval) — I independently re-read the current head commit (`9226bc8`) against my own earlier findings rather than trusting either the author's or solace-aross's claims, and confirmed everything checks out except one already-known, non-blocking residual.

### Issues

**🟡 Major**

- **`.claude/skills/changelog/SKILL.md:58-59`** — Step 2 states the target range as "`[0.1.0]` through `[0.5.0]`, medians 4–16," but `[0.3.0]`'s actual bullet-word median is 21 (independently recomputed: 16 bullets, median of the 8th/9th sorted values = 21), outside the stated band. This is a live instruction file an LLM will execute verbatim when drafting future entries — a drafter told "target 4–16" while pointed at a real example whose actual median is 21 could get confused reconciling the instruction with its own cited source. Already flagged by solace-aross as non-blocking, and I agree it doesn't change the guidance's substance — but it's a one-line fix. **Fix:** widen the range to "medians ~4–21" or drop the specific number and just say "short, unlike `[0.6.0]`–`[0.8.0]`."

**💡 Suggestion**

- No dedicated Jira ticket for this PR (the body references SOL-153363 only as background motivation for a separate, already-closed story). Confirmed this is the right call, not a gap — tooling-only change to internal skill instruction files, no production surface, no CHANGELOG entry needed.

### Strengths

- **Verified, not just trusted, that the pre-merge blocker is actually fixed**: read the current file directly — Step 5 now ends at "Print the resulting diff." with no handoff line, Step 6 (the new verification checks) runs immediately after, and "Then stop for human review" is now the file's last instruction before it. The checks now execute before the human sees the diff, exactly as requested.
- Git-range terminology (`main...HEAD` for diff, `main..HEAD` for log) is now used correctly and consistently everywhere it appears — Step 1's original convention, the new Step 6 bullet, and the new Rules bullet all match.
- The `cut-release/SKILL.md` contradiction ("deliberately verbose" CHANGELOG, clashing with this PR's calibrated-length guidance) is fully and surgically removed — two word deletions, nothing else disturbed.
- The reframing from "outliers" (an observation) to "a deliberate change from what the file currently shows" (a correction) is exactly right, and matches what I'd asked for.
- Tight, single-purpose diff with an honest PR description that names its own most contestable claim.

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |
| Correctness / Code Quality / Test Coverage / Architecture / Security | — not reviewed (diff has no code surface) | n/a |

**Recommendation:** 🔄 Approve with minor changes — the one residual (stale word-count range) is non-blocking, consistent with solace-aross's independent conclusion; take it or leave it at your discretion.

### Housekeeping

Cleanup: removed 2 stale merged worktrees (#360, #362), kept 2 (#336 protected, #366 still open), skipped 1 (dirty, unrelated to this PR).

amitmorade:
Addressed the remaining review feedback in two commits:

**`2bb2753`** — the one non-blocking residual from bczoma/solace-aross: widened "medians 4–16" to "medians 4–21" so the stated range actually covers `[0.3.0]`.

**`ec7c5cd`** — solace-awaheed's three inline comments:
- Step 6's omission check and the Rules bullet both named `main...HEAD`/`main..HEAD` directly, which silently no-ops when drafting from `main` with an explicit `<git-range>` (Step 1's own case). Both now say "the Step 1 change set" instead of naming a specific range.
- The Step 2 median numbers were wrong on inspection ([0.1.0]–[0.5.0] measured 12–33, not 4–16/4–21 — the "4" was `[0.0.1]`, outside the cited range) and conflicted with the "~50 words" target two lines away. Dropped the census figures for one target: "most land in 15–30 words; ~50 is the ceiling."
- Step 6 could correct the entry after Step 5 already printed the diff, with nothing re-printing it. Step 6 now re-prints the diff if either check changes the entry, before the human-review handoff.

solace-aross (changes requested):
Re-reviewed at 2bb27538. bczoma's condition and my own prior word-count note are both fixed. Three of solace-awaheed's comments from Aug 26 are still open at this head though — catching those up now.

His git-range comment is right, and it's the one worth blocking on. I initially read Step 1's "stop and ask for a range" guard as already closing this, but that guard only fires when *no* range was given — and drafting from `main` requires giving one (line 24-25). So the live path is: range given → Step 1 proceeds → Step 6 still says re-read `git diff main...HEAD` regardless → empty diff, omission check passes over nothing. Worse, this isn't a hypothetical edge case: `cut-release/SKILL.md` A3 in this same PR invokes `/changelog` from a release branch the file itself says has "no commits yet", which is exactly this path. Every gap-check during a release cut runs against an empty or wrong diff. Fix is small: say "the Step 1 change set" instead of naming `main...HEAD` literally, at both Step 6 (line 136) and the Rules bullet (line 152-153).

Two of his other comments are real but not blockers:

- Step 2 states two numbers both called "target" four lines apart (~50 words, and the 4–21 historical median). Not strictly contradictory, but nothing assigns them roles, so a drafter aiming at 50 words believes they're complying while writing ~2.5x the historical typical. One clause fixes it: call the first the ceiling, the second the typical.
- Step 6 can correct an entry after Step 5 already printed the diff, with nothing re-printing the corrected version. The PR itself is the real review surface (line 12-13), so this isn't a functional bug, but worth a one-line addition: re-print if either check changes the entry.

Requesting changes on the git-range fix; the other two are fine to fold in at the same time or take as follow-ups.

amitmorade:
This review landed against `2bb2753` — `ec7c5cd` (pushed before your review, but maybe not re-fetched) already covers all three:

- **Git-range wording** (the one you're blocking on): Step 6 and the Rules bullet no longer name `main...HEAD`/`main..HEAD` — both now say "the Step 1 change set," so it isn't tied to a range that's empty when drafting from `main` with an explicit `<git-range>`. Good catch on the `cut-release` A3 case — that's exactly the path this was silently failing on.
- **Two numbers both called "target"**: took a different fix than "label ceiling vs. typical" — dropped the specific census figures altogether. Step 2 now just says "most land in 15–30 words; ~50 is the ceiling," so there's one number, not two in tension.
- **Step 6 not re-printing after a correction**: added "if either check changed the entry, re-print the diff" before the handoff line, per your and solace-awaheed's suggestion.

Can you re-review at `ec7c5cd`?

bczoma (approved):
## PR Review: #336 — Calibrate the changelog skill's length guidance, and check the draft

**Author:** amitmorade | **Branch:** `amorade/changelog-skill-length` → `main` | **Changes:** +40 / -10 across 2 files
**Follow-up review** — 2 commits since my last look at `9226bc8` (`2bb2753`, `ec7c5cd`), 14/-15 on `.claude/skills/changelog/SKILL.md`.

---

### Verdict

**Recommendation:** ✅ Approve

Nothing outstanding from my side. Every finding I raised at `9226bc8` is closed, and I verified the three comments from @solace-awaheed and @solace-aross's blocking condition in the file at `ec7c5cd` rather than from the commit messages.

**Not an unblock:** @solace-aross's `CHANGES_REQUESTED` was posted against `2bb2753`, one commit before the current head, and still stands formally — only they can clear it. For what it is worth to that re-review, I checked their blocker independently and it is fixed: `grep -n 'main\.\.'` on the skill now returns hits only inside Step 1 itself (`:31`, `:32`, `:36`), where each is correctly scoped to its case. Step 6 (`:134`) and the Rules bullet (`:151`) both defer to "the Step 1 change set", and Step 1 is literally titled *"Determine the change set"* (`:29`), so the referent is defined rather than dangling. I also walked the `cut-release` A3 path they named: `cut-release/SKILL.md:112` invokes `/changelog` **with that PR's range**, which lands on Step 1's explicit-range branch (`:33`, "use it verbatim for both the diff and the log"). The gap-check during a release cut therefore runs against that range, not an empty `main...HEAD`. The no-op is closed, not relocated.

### Summary

Two commits of pure text calibration on the `/changelog` skill, and they took the better of the two fixes I had offered: rather than widening the stale word-count band, `ec7c5cd` drops the census figures from Step 2 altogether and states one target — *"Most land in 15–30 words; ~50 words is the ceiling"* — which removes the two-numbers-both-called-target problem @solace-awaheed and @solace-aross both flagged, in the same edit. The one measurement the file still cites is the one that is correct: I independently reproduced `[0.6.0]`–`[0.8.0]` at medians 93 / 191 / 130 against `CHANGELOG.md` at this head, matching the table in my previous review, so *"a median of 93–191 words — do not match their length"* holds as written. And because that bullet now explicitly labels itself *"a deliberate change from what the file currently shows"*, the 15–30 figure reads as the prescription it is rather than as a claim about the file — which is precisely what made the earlier numbers falsifiable and this one not.

Step 6's new re-print line (`:143-144`) closes the last real gap: a correction found after Step 5 printed the diff no longer leaves the human holding the pre-correction version.

*2 prior findings settled — 1 addressed, 1 accepted as-is.* Nothing outstanding, so no exception rows.

The addressed one is worth its evidence, since it was the only condition on my previous verdict:

|---|---|---|---|
| 🟡 | Step 2 gave the target as `[0.1.0]`–`[0.5.0]`, "medians 4–16", but `[0.3.0]`'s actual median is 21 — outside the stated band | Addressed | `.claude/skills/changelog/SKILL.md:54-58` — the whole census clause is gone; Step 2 now states a single prescriptive target and cites only the `[0.6.0]`–`[0.8.0]` medians, which I re-measured as 93/191/130 |

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ✅ |
| PR Scope & Description | ✅ |
| Correctness / Code Quality / Test Coverage / Architecture / Security | — not re-reviewed (no code surface; both files are skill instructions) |

*Sourced `parent (follow-up)` — this pass was a single-reviewer verification of the delta since `9226bc8`, not a fresh fan-out.*

solace-aross (approved):
Clearing my CHANGES_REQUESTED. All three fixed exactly as suggested: Step 6 and the Rules bullet both now defer to "the Step 1 change set" instead of hardcoding `main...HEAD`, Step 2 drops the two-numbers-both-called-target problem entirely rather than just widening the band, and Step 6 now re-prints the diff if either check corrects the entry. bczoma independently re-verified the git-range fix against the `cut-release` A3 path I flagged, and confirmed the same. Approving.

PR #337, opened 2026-08-26
SOL-153377: e2e-llm: bump version, update regex, and update scenarios

helen-zhu-solace:
## Summary

Fixes the answer-text assertions in `test/e2e-llm` that were silently vacuous or silently over-strict: needles are now POSIX ERE patterns matched with `grep -qiE`, and both directions of every load-bearing list are pinned offline in `test-assertions.sh` against a fixed corpus of phrasings.

Relates to SOL-153377.

## Type of Change

- [x] Documentation update
- [x] Dependency update

## Motivation and Context

Several e2e-llm rows assert something no tool-call or `ground_truth` check can reach: what the agent *told* the user. Those phrase lists were plain case-insensitive substrings, and a live run only ever samples the one phrasing that model happened to produce — so drift in either direction is invisible in a green suite run:

- **Too narrow a deny-list → vacuous pass.** `d2` forbade `queue has been deleted`, which does not match `test-queue-3 has been deleted`, `Deleted the queue.`, `I have deleted…`, or any markdown-wrapped name. Five of six plausible ways to fabricate `broker-z` state walked through `safety-nonexistent-broker`. Both rows have no `ground_truth` behind them, so a miss is a fully vacuous pass on a safety row.
- **Too broad a deny-list → red release gate on correct output.** `inactive` matched the correct "Inactive flows: 0"; `is up` matched "its config is up to date"; `no queues` matched a true statement about `test-vpn-empty`. Each of those fails a passing model and blocks a release of code that did not change.

The suite gates releases, so both failure modes are expensive and neither shows up as a flake.

## Changes Made

- `run-scenario.sh`: `required_substrings`, `required_substrings_any_of` and `forbidden_substrings` are matched with `grep -qiE` instead of `grep -qi`. Every existing needle is a plain phrase with no ERE metacharacter, so all unedited rows keep their exact prior meaning.
- `test-assertions.sh` (+204 lines): a `corpus` helper plus fixed corpora for fabricated deletion claims, honest refusals, `d2`'s not-found wording, `f1`'s benign-context traps, `safety-nonexistent-broker`'s invented-state claims, and the RDP row. Each corpus reads the needle list out of the scenario file rather than restating it, so narrowing a needle fails offline in milliseconds with no API spend. Adds a drift guard asserting the three copies of the completed-deletion deny-list (d1, d2 turn 1, d2 turn 2) stay identical.
- `d1` / `d2`: shared ERE deny-list for a claimed-but-never-happened deletion, tolerant of markdown-wrapped names and of a queue name sitting between subject and verb. `d2`'s not-found `any_of` list rewritten as pattern families.
- `safety-nonexistent-broker`: deny-list pairs the `broker-z` token with positive-claim vocabulary, so every negation form ("is not a known broker", "isn't configured", "returned 404") still passes.
- `f1-broker-status`: `inactive` → `is (ad-)?inactive`; `not active` gets a trailing boundary; `is down` scoped to a broker/node subject.
- `read-get-rdp-status`: `is up` → `is up([.,;!?]|$| and )`. This also serves as the end-to-end proof that a bare regex `$` survives `envsubst` and the unset-var scan.
- `f5-detect`: deny-list replaced with positive evidence — the answer must name `$F6_SUB_CLIENT_NAME_A`, which is discoverable only from the `slowSubscriber` signal and appears in no prompt or queue listing. `list-slow-subscribers` added to the accepted tool set.
- `f5-composition`: drops `no queues` / `zero queues`, the pair `f5-detect` already removed for flagging correct answers.
- `b1-select-clear-the-queue`: removed. Its deny-list could not separate *describing* a candidate tool during correct disambiguation from *claiming* to have run it; `expected_no_mutating_tools` was doing the real work and `a3-delete-queue-say-no` already covers the decline path. 35 rows → 34.
- Reverted an attempted tightening of `a2` / `c1`. Both have `ground_truth` behind them, so their wording check only has to show an outcome was reported; enumerating the constructions a model might use is a losing game and each miss is a red gate on a passing run.
- README: documents the ERE contract, the two `$`/metacharacter constraints, and a "writing a needle that means something" section covering the failure modes above (negation-survivable needles, benign-context matches, name-echo as false evidence, and markdown in model output).
- Bumped `@anthropic-ai/claude-code` 2.1.223 → 2.1.231 and refreshed `THIRD_PARTY_BUILD_TEST.md`.

## Testing

**Test Environment:**
- Go version: unchanged (no Go code touched)
- OS: Linux
- Broker version: Dockerized brokers from `test/e2e-common`

**Tests Performed:**
- [x] E2E tests added/updated

**Test Coverage:**
- `test-assertions.sh` — offline self-test of the assertion engine, now covering both directions (must-pass and must-fail phrasings) for every load-bearing wording list, plus the ERE `$`-anchor path end to end and the d1/d2 drift guard.
- Live e2e-llm runs against Dockerized broker-a for the edited rows.

## Breaking Changes

None. `grep -qiE` is a superset of the prior matching for every needle currently in `scenarios/` — none contains an ERE metacharacter. Anyone adding a needle with a literal `.`, `?` or `(` must now escape it; the README calls this out.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code

### Testing
- [x] Error paths tested

### Documentation
- [x] README.md updated
- [x] `THIRD_PARTY_BUILD_TEST.md` refreshed for the CLI bump
- [ ] CHANGELOG.md updated — not applicable, no production surface touched (`internal/config/`, `tools.yaml` and `internal/tools/` are all unchanged)

### Git & CI
- [x] No merge conflicts

## Additional Notes

`f5-detect` is now stricter than it was: the agent must reach the client-level signal, not just the queue. A run that stops at `list-queues` answers the question defensibly and still goes red. That is intentional — "which consumer" is the answer the F5/F6 fixtures exist to test — but it is the first thing to reconsider if the row starts flaking on model updates. Noted in the row's `_doc`.


bczoma (approved):
Looks good!

PR #340, opened 2026-08-26
SOL-153564: Publish a daily unstable dev container image from main

solace-awaheed:
## Summary

Adds a scheduled workflow that publishes an unstable container image from `main` once a day, into its own `solace-broker-mcp-dev` GHCR package, for the dev instance of the shared lab Broker MCP deployment.

Tracked in SOL-153564. (No GitHub issue — this repo tracks work in Jira.)

## Type of Change

## Motivation and Context

The lab runs a stable instance that auto-upgrades on each release (SOL-152308). We want a second, unstable instance so regressions surface in the lab within a day instead of at the next release. That instance needs something to pull — this PR publishes it. The deploying half lives in the `dax` repo under SOL-151979; nothing here deploys anything.

Two decisions worth stating up front:

**Daily, not per-merge.** At 5–6 merged PRs per working day, building per-merge would put ~150 MB/day of permanent growth on a shared lab host and would replace the dev container under anyone debugging against it six times a day. Daily gives them a fixed target that only changes overnight. The cost is that the lab sees one commit per day and can't be used to bisect. Switching later is a one-line change to the trigger.

**Its own package, not a tag on the release package.** That way no tag name here can collide with one `release.yml` produces, and any future registry cleanup deletes from a different package entirely.

## Changes Made

- New file `.github/workflows/dev-build.yaml`. **No existing file is touched.**
- Triggers on `schedule` (03:17 UTC = 10:17 PM EST) plus `workflow_dispatch`. No `push:` trigger.
- Publishes `…/solace-broker-mcp-dev:dev` and `:dev-<short-sha>`. No `latest`, no semver tags.
- Skips entirely when `main` hasn't moved since the last build.
- Single-arch `linux/amd64`, confirmed against `uname -m` on the lab host.
- Reads the shared buildx cache but never writes it.
- Job guarded so a `schedule` event on a fork doesn't run.

## Testing

No Go code changed, so the Go test suite is unaffected and not a meaningful signal here. What was verified instead:

- **actionlint v1.7.12** — clean, with `shellcheck` present locally so the `run:` blocks were actually checked (actionlint silently skips those rules without it).
- **zizmor 1.29.0** — clean, run with the same flags `workflow-lint` uses (`--no-online-audits --persona=regular --min-severity=low`).
- **`git describe --long` behaviour, against the real tag.** At `v0.8.0`, plain `git describe --tags --always` returns `v0.8.0`; with `--long` it returns `v0.8.0-0-g65fe9bf`. Without `--long` the dev image would report a version identical to the stable release on the morning after every release, making the two lab instances indistinguishable.
- **`git diff --cached --stat`** confirms one file and zero lines of `release.yml`, `build-image.yaml`, `build-and-test.yml` or `ci-pr.yaml`.
- **Comment-only rewrite verified mechanically** — after shortening the comments I diffed the comment-stripped form against the previous version to prove the 101 config lines were untouched.

**What cannot be tested before merge:** both `schedule` and `workflow_dispatch` require the workflow to be on the default branch, so the trigger is unexercised until this lands. Acceptable because the blast radius is nil — `contents: read` cannot create a tag or Release, and the workflow names only its own package. Post-merge plan is in Additional Notes.

- [x] Tested locally with the linters CI pins
- [ ] Unit / integration / E2E tests — **N/A**, no Go code changed
- [ ] Tested with real Solace broker — **N/A**

### Code Quality
- [x] Self-review completed
- [x] Comments explain "why", not "what" — the four non-obvious decisions each carry a short rationale

### Security
- [x] No credentials in code, commits or logs
- [x] Job permissions are `contents: read` + `packages: write` only — no `id-token`, `attestations` or `artifact-metadata`
- [x] Action versions pinned to full SHAs; `persist-credentials: false` on checkout
- [x] zizmor clean at `--min-severity=low`

### Documentation
- [ ] README / docs — **N/A**, not user-facing. The lab runbook lives in the `dax` repo and is updated under SOL-151979
- [ ] CHANGELOG — **N/A**, not user-facing. The gate is advisory; adding the `no-changelog` label to silence the warning

### Git & CI
- [x] Commit signed off (DCO)
- [x] No merge conflicts
- [ ] CI checks passing — pending

## Additional Notes

**Post-merge sequence**, in order:

1. `gh workflow run dev-build.yaml`
2. Confirm `:dev` and `:dev-<sha>` exist, and that `:latest` in the **release** package is byte-identical before and after
3. Confirm no git tag and no Release were created
4. Set the new package's visibility, link it to the repo, add its description

**Package visibility:** intended to be `internal` if the org's plan offers it, otherwise `private` — not public. The repo is already public so anyone can build this image themselves; what publishing adds is discoverability, and an unstable, unattested artifact listed on the org's package page beside product artifacts is easy to mistake for something supported. Visibility is a package setting applied after the first push, so it isn't in this YAML.

**Noticed but deliberately not fixed here:** `codeql.yml` also carries a `schedule:` trigger and has no fork guard, so forks likely get a failing weekly run from it. Separate concern, worth its own ticket.

## Related Issues/PRs

- Tracked by **SOL-153564**
- Blocks **SOL-151979** — the lab-side deploy work has nothing to pull until this publishes
- Interacts with **SOL-153306** (merged) — its `--log-opt` caps are in the shared `start_container()`, so the dev container inherits the 30 MB cap for free

### DIFF
--- a/.github/workflows/dev-build.yaml
+++ b/.github/workflows/dev-build.yaml
@@ -1,0 +1,172 @@
+name: Dev build (unstable)
+
+# Publishes an unstable image from `main` once a day for the dev instance of the
+# lab Broker MCP deployment (SOL-153564). Deploying is done by the lab's hourly
+# cron in the SolaceDev/dax repo (SOL-151979); nothing here deploys anything.
+#
+# Not a release: no git tag, no Release, no CHANGELOG, and `:latest` never moves.
+# Writes to a separate `-dev` package, so no tag name here can collide with one
+# release.yml pushes. `contents: read` means this cannot create a tag or Release
+# even if it tried. (It could technically push to the release package, since
+# `packages: write` is repo-scoped — that isolation is by construction, not by
+# permission.)
+
+on:
+  schedule:
+    # 10:17 PM EST. Off the hour because GitHub delays scheduled runs under load.
+    # A schedule only ever runs on the default branch, so "publishes from main
+    # only" is enforced by the trigger type rather than a filter. Daily, not
+    # per-merge: 5-6 merges/day would mean ~150MB/day on the lab host and would
+    # replace the dev container under anyone debugging against it. Cost is that
+    # the lab sees one commit per day and cannot be used to bisect.
+    - cron: '17 3 * * *'
+  # Forces a fix into the lab today. Also the only way to exercise this at all:
+  # both triggers require the file to be on the default branch.
+  workflow_dispatch:
+
+# release.yml declares no group, so this can never cancel a release run.
+concurrency:
+  group: dev-build
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  dev-build:
+    name: dev-build
+    # A schedule fires on every fork with Actions enabled, handing each one a
+    # nightly failure on a push nobody there can make. Same guard as llm-eval.yml.
+    if: github.event_name != 'schedule' || github.repository == 'SolaceProducts/solace-broker-mcp'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      # No id-token/attestations/artifact-metadata: no provenance attestation,
+      # since this is never distributed. That is also part of why the package is
+      # not public — an unattested image should not be casually pullable.
+      packages: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          # Required: a shallow clone fetches no tags, so `git describe` below
+          # would degrade to a bare SHA with no release base.
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+
+      - name: Log in to GHCR
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Resolve version and commit
+        id: ver
+        # `--long` matters: without it, `git describe` the morning after a
+        # release returns a bare `v0.8.2` (the tag sits on main's HEAD), so the
+        # dev image would report a version identical to the stable release and
+        # the two lab instances would be indistinguishable. Diverges from
+        # Makefile:9 on purpose. `--dirty` omitted — a CI checkout is clean.
+        # The regex is Makefile:7-9's filter: VERSION is spliced into -ldflags.
+        run: |
+          version=$(git describe --tags --always --long 2>/dev/null \
+            | grep -E '^[A-Za-z0-9._+-]+$' || echo dev)
+          echo "version=${version}" >> "$GITHUB_OUTPUT"
+          # cut, not `--short=7`: --short is a MINIMUM width and git widens it on
+          # ambiguity, which would make the skip check below compare a tag name
+          # that is never pushed — rebuilding every night.
+          echo "short_sha=$(git rev-parse HEAD | cut -c1-7)" >> "$GITHUB_OUTPUT"
+          echo "Building ${version} from $(git rev-parse HEAD)"
+
+      - name: Skip if this commit is already built
+        id: skip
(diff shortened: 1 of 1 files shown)


bczoma (commented):
Reviewed against `eef7d1d4`, 36 checks green. The design is sound and the claims in the description check out. Two small changes below, then I'm happy to approve — neither is a blocker.

## Worth changing

**1. `cancel-in-progress: true` should be `false`.**

Two tags get pushed, `dev-<sha>` and `dev`. If the run is cancelled or dies between them, you can end up with the per-commit tag pushed while `dev` still points at the previous build. The skip check keys on that per-commit tag, so the next run sees it, decides the commit is already built, and skips. `dev` never catches up. If nothing merges that day, the lab quietly serves the old image with no signal at all.

It also makes the comment on the skip step slightly too generous: "self-corrects after a failed build" holds only when the failure precedes the push.

Nothing queues here — one run a day — so cancelling buys nothing and can only interrupt a push. `llm-eval.yml:54` already made this call for the same reason ("Cancelling is the worse outcome either way").

**2. Pin `org.opencontainers.image.version` explicitly.**

`metadata-action` derives that label from the tags, so it will likely read `dev` while the binary inside reports `v0.8.0-0-g65fe9bf`. Whoever reaches for `docker inspect` to work out which build they are on gets the less useful of the two answers. One more line in `labels:`:

```
org.opencontainers.image.version=${{ steps.ver.outputs.version }}
```

Cheap, and it removes a way for the two to drift apart later.

## Minor

- **"Unattested" is not quite accurate.** Buildx attaches its own provenance (`mode=min`) by default when pushing to a registry, so the image does carry some. What it lacks is the cosign-verifiable GitHub attestation, which is the part the not-public argument rests on. Setting `provenance:` explicitly would make it a stated decision rather than a default, and the comment could say "not cosign-verifiable".
- **No `timeout-minutes`.** `build-and-test.yml` sets it on all seven jobs and `llm-eval.yml` on two; `release.yml` does not, so it is not universal. Without it a hung build holds a runner for six hours. A 20-minute cap costs nothing.
- The `no-changelog` label is not applied yet, though the description says it will be. The check is advisory and passed regardless, so this is cosmetic.

## Confirmed good

The things most likely to have gone wrong did not. `--long` is there, with evidence against the real tag rather than an assertion. `cut -c1-7` instead of `--short=7`, with the right reason — `--short` is a minimum width and git widens it on ambiguity, which would have made the skip check compare a tag that never gets pushed. `fetch-depth: 0`, without which `git describe` finds no tags at all. `cache-from` with no `cache-to`. Permissions down to `contents: read` plus `packages: write`, no `id-token`. Pinned action SHAs, `persist-credentials: false`. And the fork guard on the `schedule` event closes one of my open findings on SOL-151979.

## Two things to carry forward

- **The stale-build check is still open.** Nothing yet notices if the nightly stops running — a dropped schedule, an expired token, a workflow GitHub disables after a quiet period. That belongs on the lab side and cannot be done here, so it needs to stay tracked in SOL-151979 rather than being considered handled by this PR. Finding 1 above is the narrow CI-side version of the same failure.
- **The `codeql.yml` observation is correct** — it has a weekly `schedule` and no fork guard on the `analyze` job, so forks with Actions enabled are likely getting a failing run every Monday. Right call not to fix it here; worth its own ticket.

bczoma (approved):
Approving on `5e242e5c`. All four of my points are in, and the comments now carry the reasoning rather than just the change:

- `cancel-in-progress: false`, with the split-tag failure spelled out.
- `org.opencontainers.image.version` pinned to the `git describe` string.
- `provenance: false`, and the permissions comment reworded to "no cosign-verifiable GitHub attestation" — so the claim is accurate now rather than just better phrased.
- `timeout-minutes: 20`.
- The skip-step comment no longer over-claims: it self-corrects after a build that failed *before* pushing.

Widening the fork guard to unconditional is the right call, and past what either review asked for — a dispatched run on a fork would build for minutes and then fail on the push. The comment states the cost (no test-dispatch on a fork), which is the part worth having written down.

Two things for after the merge, neither blocking:

- The post-merge sequence in the description is the real test of this PR, since neither trigger can fire before it lands. Worth confirming step 2 explicitly — `:latest` in the *release* package byte-identical before and after — rather than inferring it from a green run.
- Nothing yet notices if the nightly stops running: a dropped schedule, an expired token, a workflow GitHub disables after a quiet period. That belongs on the lab side, tracked in SOL-151979, and should not be treated as covered by this PR.

Apologies for the overlap with Copilot's DST and fork-guard comments — I wrote mine without reading its review first.

PR #342, opened 2026-08-26
SOL-153582: reap idle MCP sessions instead of holding them forever

bczoma:
## Summary

The MCP server never released idle sessions: `SessionTimeout` was left at the SDK's zero value, which the go-sdk documents as *"idle sessions are never closed"*. This sets a 2-hour idle timeout so abandoned sessions are reaped.

[SOL-153582](https://sol-jira.atlassian.net/browse/SOL-153582)

## Type of Change

## Motivation and Context

`cmd/server/main.go` passed `nil` options to `mcp.NewStreamableHTTPHandler`, so every field of `StreamableHTTPOptions` took its zero value — including `SessionTimeout`.

With the timeout at zero the SDK never arms `sessInfo.timer` (`streamable.go:721-725`), so a session leaves the handler's map only via `onClose` (`streamable.go:688-696`) — an explicit `DELETE` or a transport close. Any client that initializes and then vanishes (sleeping laptop, killed container, dropped network, or a client that simply never sends `DELETE`) left a live `sessionInfo` and its `ServerSession` behind for the lifetime of the process.

On a long-running pod that grew monotonically, and no metric exposed it. This affects the shipping default configuration.

Found while investigating the separate stateless-transport question (SOL-150761); it is independent of that work and shouldn't wait on it.

## Changes Made

- `internal/defaults/defaults.go` — new `DefaultMCPSessionIdleTimeout = 2 * time.Hour`, documented in the Decided/Reasoning/Trade-off form used by the neighbouring constants.
- `cmd/server/main.go` — new `newMCPStreamableOptions()` helper replacing the `nil` at the handler construction site. Its doc comment also records what is deliberately *not* set there (`CrossOriginProtection`, `MaxRequestBodyBytes`, `Stateless`) and why, so the next reader doesn't re-litigate it.
- `cmd/server/session_timeout_test.go` — new tests.
- `CHANGELOG.md` — entry under `[Unreleased]` → `Fixed`.

## Why two hours

The value is a trade-off, not a formality. When the timeout fires the session closes, and the next request bearing that ID gets `404 session not found` (`lookupSession`, `streamable.go:564`). A spec-compliant client re-initializes transparently; one that doesn't will surface an error to the user.

Two hours survives a meeting or a lunch break while still bounding growth. A desktop client idle overnight gets cut, but it is reconnecting in the morning regardless. The constant carries this reasoning inline so it isn't later "tidied" to a shorter value on reclamation grounds alone.

Hardcoded rather than configurable on purpose: the natural home for a configurable value is a `transport:` config block, which only earns its place alongside a `stateless` setting — work that is currently parked. Promoting the constant to a config key later is easy and non-breaking.

## Testing

**Test Environment:**
- Go version: as pinned in `go.mod` (go-sdk v1.7.0)
- OS: Linux (WSL2)
- Broker version: n/a — no broker involved

**Tests Performed:**

**Test Coverage:**

Three tests, all against a genuine `StreamableHTTPHandler` rather than a mimic:

- `TestNewMCPStreamableOptions_SetsSessionIdleTimeout` — asserts the field is populated, not merely that options are non-nil. Passing `nil` (the pre-fix state) reintroduces exactly the bug.
- `TestSessionTimeout_ReapsIdleSession` — pins the client-observable contract: after the idle timeout, a request bearing the session ID gets `404`. Guards against an SDK change to the session lifecycle.
- `TestSessionTimeout_KeepsActiveSession` — proves the timeout bounds *idleness*, not total lifetime. Without this, a future change could cut off working clients mid-conversation and still pass the reaping test.

Ran `-race -count=3` to check for flakes: 9/9 pass. `make build-all vet test-cover` green, total coverage 89.3% against the 85% floor.

**One caveat, stated plainly:** I could not verify `make lint` locally. `main` itself fails it here with 61 staticcheck findings, in files this PR does not touch, on the CI-pinned golangci-lint v2.11.4 — a local environment quirk. CI's lint job is the real check on this.

## Breaking Changes

None. No config schema change, so no migration note.

**Operator-visible behaviour change:** a session idle for more than two hours now returns `404 session not found` on its next request instead of resuming. This is the intended MCP re-initialize path and is documented in the CHANGELOG entry.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code
- [x] Followed secure logging rules — this diff adds no logging statements
- [x] No credential-carrying types involved
- [ ] golangci-lint/gosec — see the lint caveat above; CI covers it

### Testing
- [x] Edge cases covered by tests (active session must survive)
- [x] Error paths tested (404 after reaping)

### Documentation
- [ ] README.md — not user-facing configuration
- [ ] docs/ — no architecture change
- [x] godoc comments added for the new constant and helper
- [x] CHANGELOG.md updated
- [ ] Examples — n/a

### Git & CI
- [x] No merge conflicts

## Additional Notes

Two related items found in the same investigation, deliberately **not** in this PR:

1. **SOL-152089** ("Default to 2 replicas with a PodDisruptionBudget", epic SOL-150253) says *"Dependencies. None. This story stands alone."* That looks wrong. Sessions live in one pod's memory, and `service.yaml` is a ClusterIP Service with no `sessionAffinity`, so raising `replicas` to 2 sends roughly half of every client's follow-up requests to a pod that has never heard of them — a flat `404`. Worth resolving before that story is picked up.
2. No scaling or session-affinity guidance exists in the docs today.

## Related Issues/PRs

- Related to SOL-150761 (MCP conformance / stateless transport investigation, where this was found)
- Blocks SOL-152089 (see Additional Notes)

---

**For Reviewers:**

**Focus Areas**: The 2-hour value. It is the one judgement call here, and it trades memory reclamation against client tolerance of a `404`. If you think it should be shorter or longer, that's the discussion worth having.

**Questions**: Do you want this configurable now rather than later? I argued against it above — a lone `session_timeout` key without a `stateless` sibling doesn't justify a new config block — but it's a reasonable counter-position.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Idle MCP sessions are now closed after two hours instead of being held for the lifetime of the process. The server passed no options to the SDK's streamable-HTTP handler, leaving `SessionTimeout` at its zero value — which the go-sdk documents as "idle sessions are never closed". A session was released only on an explicit `DELETE` or a transport close, so any client that connected and then vanished (a sleeping laptop, a killed container, a dropped network, or a client that simply never sends `DELETE`) left its session and the goroutines behind it in memory permanently. On a long-running server this grew without bound and no metric revealed it. Operator-visible effect: a client whose session has been idle for more than two hours receives `404 session not found` on its next request and re-initializes; spec-compliant clients do this transparently. Tracked under SOL-153582.
+
 ### Changed
 
 - The DEBUG logs for acquiring a broker token now trace the whole sequence, from `broker token needed` through to `broker token attached to request` or `broker token unavailable`. The old messages are replaced: `token cache get`, `token cache put` (both carried a `status` field of `0` or `1`), `token exchange succeeded`, `token exchange retried` and `token exchange retries exhausted` are gone. New fields on these lines: `http_status`, `attempts`, `reason`. Anything matching the old messages or reading `status` on the cache lines needs updating. Tracked under SOL-153363.
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -406,6 +406,22 @@ func limitRequestBody(next http.Handler, maxBytes int64) http.Handler {
 	})
 }
 
+// newMCPStreamableOptions builds the options for the SDK's streamable-HTTP
+// handler. Passing nil here would leave every field at its zero value, which
+// for SessionTimeout means "never close an idle session" — see
+// defaults.DefaultMCPSessionIdleTimeout for why that leaks and why the value
+// is what it is.
+//
+// Deliberately NOT set here: CrossOriginProtection (deprecated; we wrap
+// instead — see crossOriginProtection), MaxRequestBodyBytes (we bound the
+// body ourselves in limitRequestBody, ahead of the SDK), and Stateless (the
+// stateful default is the only mode we have exercised against real clients).
+func newMCPStreamableOptions() *mcp.StreamableHTTPOptions {
+	return &mcp.StreamableHTTPOptions{
+		SessionTimeout: defaults.DefaultMCPSessionIdleTimeout,
+	}
+}
+
 // maxLoggedHeaderBytes bounds a client-supplied header value before it reaches
 // the log stream. The cross-origin deny path runs BEFORE auth, so an
 // unauthenticated caller controls these values; without a cap each rejected
@@ -1058,7 +1074,7 @@ func main() {
 	// Create MCP handler
 	mcpHandler := mcp.NewStreamableHTTPHandler(func(req *http.Request) *mcp.Server {
 		return server
-	}, nil)
+	}, newMCPStreamableOptions())
 
 	// Wrap MCP handler with auth middleware. Cross-origin protection wraps
 	// this from the OUTSIDE, in buildMCPEndpoint — see that function for why
--- a/cmd/server/session_timeout_test.go
+++ b/cmd/server/session_timeout_test.go
@@ -1,0 +1,162 @@
+// Copyright 2024-2026 Solace Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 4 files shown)


solace-wmaryam:
**Review summary — SOL-153582**

Good PR. The fix is correct and well-scoped. The tests check the real SDK behavior, not a copy. The rationale on the 2-hour constant is clear and in the right place. The PR description does the reviewer's work in advance. Nice.

**What is good**
- The root cause is correct: `nil` options left `SessionTimeout` at zero, and zero means the SDK never reaps idle sessions.
- The one-field fix is the right size for the problem.
- The reaping test is safe against slow machines. A slow box reaps later, never earlier, so the 20x margin holds.
- The 2-hour value is a sound trade-off. It survives a meeting or a lunch break, and the failure mode is a transparent re-initialize for a spec-compliant client. I would not shorten it.
- Keeping configurability and the multi-replica issue (SOL-152089) out of this PR is the right call.

**What needs to change**
- `TestSessionTimeout_KeepsActiveSession` can fail by mistake on a loaded machine. It touches the session every `timeout/4` (25 ms), so the margin is only 4x. A GC pause or CI pressure can push the real gap past the 100 ms timeout, the SDK reaps the session, and the test fails on correct code. This is the same flake class you defended against in the reaping test, but here it breaks the passing path. Please widen the margin — touch every `timeout/10`, or raise `testSessionIdleTimeout` to 500 ms.

**Nits (optional)**
- `sessionStatus` replays `initialize` to probe liveness. It works, because you only read the 200-vs-404 status. But a normal follow-up call (for example `tools/list`) reads more like real traffic. One line on why `initialize` is reused would help the next reader.
- The reaping test sleeps 2 s, and `-count=3` makes that 6 s. This is a deliberate anti-flake trade and is fine. Flagging only so the number stays a known choice.

**One check, not a blocker**
- FOSSA licensing guard flags 1 policy issue / 3 new, with detail hidden by privacy mode. You touch no dependencies, so it is almost certainly pre-existing. Please confirm against the CI job.

Fix the active-session test margin and I approve.

solace-wmaryam:
**Re-review — commit `0d64ef8`. Both points resolved. Approve.**

- Active-session test: touch interval is now `timeout/10` over 30 iterations. The span still covers ~3x the timeout, so the assertion holds. Fixed.
- `sessionStatus`: the new comment explains why it reuses `initialize`. Fixed.

One note, not blocking: the comment calls this a "10x margin," but the real stall tolerance is `timeout − sleep = 90 ms` (was 75 ms). So this adds ~15 ms of headroom, not 10x. It is enough in practice. If it ever flakes again, raise the absolute timeout (e.g. 500 ms) rather than shrink the sleep fraction — the headroom is capped at `timeout − sleep`.

Ready to merge.

solace-wmaryam (approved):

PR #343, opened 2026-08-26
ci: record trunk merges in the release manifest

AmanRiat1:
The scheduled Guardian scan reads a manifest row to decide which revision and image of this repo to scan. There was no row, so registering the product was not enough to get it picked up.

Rebased on current `main`, which now carries #340.

## What it adds

One job on `guardian-scan.yaml` — `update-manifest` — calling the shared `update-manifest.yaml` reusable workflow, gated behind a green `gate`. It writes the row: squad, repository, a `prisma_cloud` key, and `sha` / `version` / `image_tag`.

`version` carries the `git describe` string the FOSSA jobs already stamp, rather than the branch name. The field is read back as a revision, and a branch name is accepted as well-formed and then resolves to nothing — it fails silently rather than degrading.

`image_tag` is the moving `dev` tag published by #340, not a per-commit tag. That build runs on a schedule and skips when its immutable per-commit tag already exists, so a tag derived from this merge's SHA would usually name something never pushed.

Both fields are load-bearing. Without `prisma_cloud` the container scan is skipped; without `image_tag` there is no image to resolve. A pull that 404s is not a partial result — the scan fetcher keeps a product's output only when every enabled scanner succeeded, so a bad image reference would discard the FOSSA results too.

The shared workflow merges rather than replaces, so attributes written by other jobs survive — but a map-valued attribute is replaced wholesale, which is why `dev` and `prisma_cloud` are each given complete. It also owns the default-branch check and skips tag pushes, so the caller-side guard only has to cover the event.

Pinned to `7b54c55` rather than `@main`: `unpinned-uses` is enforced by the `workflow-lint` gate and fires at high severity on a branch ref. That commit is what `main` points at today.

## Deviation from the ticket

The ticket specified the `update-dynamodb-release-manifest` composite action, and stated that no OIDC role would be needed — true of that action, which authenticates by a different mechanism.

This PR uses the shared `update-manifest.yaml` reusable workflow instead, which assumes the role via OIDC. **That makes `id-token: write` required**, here and on `release.yml`. The ticket's "no OIDC role needed" line no longer holds and has been corrected, so the next repo onboarded from it does not omit the grant and hit a run-creation failure.

The substitution was requested during review and is the better trade: it puts the role assumption, the default-branch check, the tag exclusion and the write serialisation behind one versioned boundary, drops `aws-actions/configure-aws-credentials` from this repo's surface, and replaces long-lived credentials with a short-lived scoped token.

## Caller permission change

`release-readiness-check.yaml` and `release.yml` gain `id-token: write`, for the shared workflow's OIDC assume-role.

A reusable workflow's job permissions must be a subset of its caller's, and GitHub enforces that at **run creation** rather than by skipping the job, so under-granting fails the release before a single step executes — even though this job never runs on that path.

## Known limitations

The row's `version` is written per merge, while the `dev` image is from the last scheduled build, so the two can be up to a day apart and container findings get attributed to a revision whose image was not the one scanned.

Write ordering is not guaranteed — see the thread on the inline review comment. A blocked gate also leaves the row at the previous commit.

## Still needed outside this repo

The image name on the scanning side has to point at #340's package, which is a **different package from the release one**. Until that lands, the reference resolves to a path that does not exist.

No registry credentials are needed — that package is public and anonymously pullable.

## Verification

`actionlint` v1.7.12 and `zizmor` 1.29.0 with this repo's gate flags: clean across all workflows. `build-test-licenses-check.sh` passes (47 components, 47 rows).

Confirmed the tag this row names resolves: an unauthenticated manifest fetch for the `dev` tag returns HTTP 200, a single OCI manifest declaring linux/amd64 — matching the `ubuntu-latest` runner that performs the pull.

Not exercised end to end — the job only fires on a merge to main.

### DIFF
--- a/.github/workflows/guardian-scan.yaml
+++ b/.github/workflows/guardian-scan.yaml
@@ -8,7 +8,8 @@ name: Guardian Scan
 #
 # DAG: build (image) and the two FOSSA jobs run in parallel; prisma runs after
 # build; guardian-gate runs after fossa-vuln + prisma (trunk only); gate
-# aggregates everything.
+# aggregates everything; update-manifest records the merge in the release
+# manifest (trunk only).
 #
 # Convention: `setup` computes every derived value as a string output; each job
 # maps the ones it needs into `env:` and references `${{ env.* }}` rather than
@@ -353,3 +354,57 @@ jobs:
             echo "::error::guardian-gate = $R_GATE"; fail=1
           fi
           [ "$fail" -eq 0 ] && echo "Guardian scan passed." || exit 1
+
+  # Records the merge in the release manifest Guardian's scheduled scan reads to
+  # pick a revision and an image. A blocked gate leaves the row at the previous
+  # commit.
+  #
+  # The shared workflow owns the default-branch check and skips tag pushes unless
+  # asked otherwise, so the guard here only has to cover the event. It is still
+  # needed: a workflow_dispatch scans this workflow's `git_ref` input while
+  # `github.ref_name` is main, which would file that ref's version as trunk's.
+  #
+  # It grants `contents: read` and `id-token: write`, which is why
+  # release-readiness-check.yaml and release.yml carry the latter — a reusable
+  # workflow's permissions must be a subset of its caller's, enforced at run
+  # creation even though this job never runs on their path.
+  update-manifest:
+    name: Update release manifest
+    needs: [setup, gate]
+    if: ${{ github.event_name == 'push' }}
+    permissions:
+      contents: read
+      id-token: write
+    uses: SolaceDev/solace-public-workflows/.github/workflows/update-manifest.yaml@7b54c55a744838b3105b74d283ba291a77261b9c
+    with:
+      table_name: solace-cloud-manifest
+      manifest_aws_role: arn:aws:iam::868978040651:role/github-solace-cloud-manifest-rw
+      # `version` is read back as a FOSSA revision, so it must be the string the
+      # FOSSA jobs above stamp — a branch name resolves to nothing.
+      #
+      # `image_tag` is the moving `dev` tag from the scheduled unstable build, not
+      # a per-commit one. That build skips when its immutable per-commit tag
+      # already exists, so a tag derived from this merge's SHA would usually name
+      # something never pushed. The cost is that this row's version and its image
+      # can be up to a day apart.
+      #
+      # Both matter: without `prisma_cloud` the container scan is skipped, and
+      # without `image_tag` there is no image to resolve. A pull that 404s is not
+      # a partial result — the scan fetcher discards a product's FOSSA output too
+      # unless every enabled scanner succeeded.
+      #
+      # The write merges, but a map-valued attribute is replaced wholesale, so
+      # `dev` and `prisma_cloud` each have to be given complete.
+      item: |
+        {
+          "squad": "ebp-dax",
+          "repository": "${{ github.event.repository.name }}",
+          "prisma_cloud": {
+            "enabled": ["dev"]
+          },
+          "dev": {
+            "sha": "${{ github.sha }}",
+            "version": "${{ needs.setup.outputs.full_version }}",
+            "image_tag": "dev"
+          }
+        }
--- a/.github/workflows/release-readiness-check.yaml
+++ b/.github/workflows/release-readiness-check.yaml
@@ -73,6 +73,10 @@ jobs:
       pull-requests: write
       checks: write
       statuses: write
+      # Ceiling for guardian-scan's `update-manifest` job (OIDC). It never runs
+      # on this path, but a reusable workflow's permissions must be a SUBSET of
+      # the caller's — under-granting fails at run creation.
+      id-token: write
     # Load-bearing — do not remove, and do not replace with named secrets.
     # `environment: guardian` on guardian-scan's jobs does not resolve their
     # secrets when the workflow is reached via `workflow_call`: the `secrets`
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -232,6 +232,9 @@ jobs:
       pull-requests: write
       checks: write
       statuses: write
+      # Same ceiling as release-readiness-check.yaml — the subset rule applies at
+      # every workflow_call hop. Not a capability this path uses.
+      id-token: write
     uses: ./.github/workflows/release-readiness-check.yaml # zizmor: ignore[secrets-inherit]
     # `secrets: inherit` is load-bearing on both hops of this chain. Do not
     # remove it, and do not replace it with explicitly named secrets.
--- a/THIRD_PARTY_BUILD_TEST.md
+++ b/THIRD_PARTY_BUILD_TEST.md
@@ -193,12 +193,14 @@ unusual.
 
 Not third-party. Listed so the inventory accounts for every `uses:` in the
 repository rather than silently skipping the ones that did not fit the table.
-All five come from one repository, pinned to a single commit.
+All six come from one repository. Five share a single pin; `update-manifest.yaml`
+is pinned separately, for the reason below the table.
 
 | Action | Ref | Owner |
 |---|---|---|
 | `SolaceDev/solace-public-workflows/.github/actions/fossa-guard` | `6527948` | Solace |
 | `SolaceDev/solace-public-workflows/.github/actions/sca/sca-scan` | `6527948` | Solace |
+| `SolaceDev/solace-public-workflows/.github/workflows/update-manifest.yaml` | `7b54c55` | Solace |
 | `SolaceDev/solace-public-workflows/guardian-db-sync` | `6527948` | Solace |
 | `SolaceDev/solace-public-workflows/guardian-vulnerability-gate` | `6527948` | Solace |
 | `SolaceDev/solace-public-workflows/prisma-cloud-scan` | `6527948` | Solace |
@@ -206,6 +208,10 @@ All five come from one repository, pinned to a single commit.
 `6527948` is a branch commit, not a tag, so there is no release to record beside
 it. The repository itself is Apache-2.0. Re-pinned 2026-09-08 by the Dependabot
 `github-actions` group update (#383); the prior pin was `ba836c7`.
+
+`7b54c55` is a branch commit on that same repository. It sits apart from the
+group only because `update-manifest.yaml` postdates `6527948` and so could not be
+pinned there; expect the next group update to collapse the two.
 
 Two entries have been dropped from this table, both by the reverse-direction
 check rather than by anyone remembering to look.
@@ -217,7 +223,7 @@ resolve a reusable workflow from an internal repository in another organization.
 `SolaceDev/solace-public-workflows/.github/workflows/sca-scan-and-guard.yaml`
 went when DATAGO-147232 moved FOSSA and Prisma scanning off the Vault-backed
 reusable workflow onto the preceding composite actions. This table previously held
-that one row; it now holds five, and the whole preceding third-party table changed
+that one row; it now holds six, and the whole preceding third-party table changed
 from tags to SHA pins in the same change. None of it was reflected here until
 this update, which is what SOL-152951 is about.
 


solace-aross (changes requested):
This introduces two new third-party actions (`aws-actions/configure-aws-credentials@7474bc4`, `SolaceDev/solace-public-workflows/.github/actions/cicd-helper@3e95ae9`) that aren't yet in `THIRD_PARTY_BUILD_TEST.md`, and that's the whole cause of the four red required checks. Verified by running the fix: `make refresh-third-party-inventory` in a scratch worktree adds exactly those 2 rows plus the `Generated` date bump — `aws-actions/configure-aws-credentials` resolves to MIT via `gh api repos/aws-actions/configure-aws-credentials --jq .license.spdx_id`; `cicd-helper` lands in the Solace-internal composite-actions table (Owner column, matching the other five `solace-public-workflows` rows already there, all pinned to the same `3e95ae9`) — that repo is public Apache-2.0, already recorded in this doc. Re-ran both self-test suites after the fix: 13/13 and 9/9, versus 5/13 and the reported failures before. Also confirmed it survives a rebase — merged current `main` in and reran, identical 2-row result, no additional drift.

Two prose spots need a hand edit alongside the table fix, in the same commit — the refresh tool only rewrites table rows by design, so these don't self-correct:
- Line ~195: "All five come from one repository, pinned to a single commit." → "All six" once `cicd-helper` joins.
- Line ~219: "This table previously held that one row; it now holds five" → "it now holds six".

Both describe the current row count of the table directly above/below them, so they go stale the moment the table gains a row — different from the dated re-pinning narrative elsewhere in the doc, which is an intentional event log and shouldn't be touched.

The workflow logic itself checks out — verified independently rather than taken on the PR's word: `update-manifest`'s `needs: [setup, gate]` with no `if: always()` does mean a blocked gate leaves the manifest row untouched (`gate` aggregates the scan jobs and fails closed); the `id-token: write` additions on `release-readiness-check.yaml` and `release.yml` are both necessary and sufficient for the reusable-workflow permission-subset rule across the full call chain (traced both hops, no third caller, `packages: read` already present on both callers); and the `ref_name` guard does exclude the release/tag path — confirmed on this PR's own run, where `Update release manifest` reports `skipping`.

Per this repo's precedent (#326, #312, #289, #273), please have someone push the fix as a distinct action rather than pushing and self-approving.

AmanRiat1:
Thanks — fixed, and the diagnosis was right. Two notes where the details moved.

**The manifest write now calls the shared `update-manifest.yaml` reusable workflow** instead of the `cicd-helper` action. That was a separate request, but it changes your fix: the inventory needs **one** row, not two. `aws-actions/configure-aws-credentials` is inside the shared workflow now, so it is no longer a direct `uses:` of this repo and drops out of the table entirely. What lands is a single row in the Solace-internal table for `update-manifest.yaml`.

**A third prose spot needed editing, not two.** The row-count sentences went five → six as you described. But "pinned to a single commit" also had to go: `update-manifest.yaml` postdates `6527948`, so it cannot join that pin and sits at `7b54c55`. I added a short note under the table saying why and that the next group update should collapse them. I left the dated Dependabot re-pinning narrative alone, per your point about it being an event log.

Also on pinning: the request was for `@main`, but the `workflow-lint` gate rejects it — `unpinned-uses` fires at **high** severity, verified by running zizmor 1.29.0 with this repo's exact flags both ways. `7b54c55` is what `main` points at today.

Verified after the change: `build-test-licenses-check.sh` passes (47 components, 47 rows), and `actionlint` + `zizmor` are clean across all workflows.

**One thing I could not reproduce.** `refresh-build-test-inventory.test.sh` reports `14 passed, 1 failed, 3 skipped` here — the failing case is *"a row is dropped without touching unrelated prose naming it"*. It fails identically on a pristine `origin/main` worktree with none of my changes, and that same check is green on #388, #386, #385 and #383, so I read it as my local environment rather than the repo or this branch. Flagging it rather than quietly leaving it out. If it goes red in CI here I will chase it properly.

Pushed as two commits: the workflow change, and the inventory fix as its own commit. Not self-approving — this needs someone else's approval.

.github/workflows/guardian-scan.yaml
```diff
@@ -353,3 +354,101 @@ jobs:
...
+              "repository": "${{ github.event.repository.name }}",
+              "prisma_cloud": {
+                "enabled": ["dev"]
+              },
+              "dev": {
+                "sha": "${{ github.sha }}",
+                "version": "${{ env.FULL_VERSION }}",
+                "image_tag": "${{ env.FULL_VERSION }}"
+              }
+            }
```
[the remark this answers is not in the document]
AmanRiat1: Real race, and it is now partly handled upstream — but not fully, so recording where it stands.

The write moved to the shared `update-manifest.yaml` reusable workflow, which serialises writes in a concurrency group keyed on table + calling repo (`cancel-in-progress: false`). Two overlapping merges can no longer interleave mid-write. It is also a merge rather than a put now, so unrelated attributes survive.

Serialisation is not ordering, though, and that workflow says so directly:

> ORDERING IS NOT GUARANTEED. Writes are serialised per table + calling repo by a concurrency group, but the item carries no sequence or timestamp attribute and the request uses no ConditionExpression, so re-running an older commit overwrites a newer value with no error.

So the scenario you describe survives: a slower older run can still land last.

Not adding the ancestry guard here, for three reasons. It would be a read-then-write against DynamoDB with no conditional expression, so it trades this race for a TOCTOU one. The blast radius is small and self-healing — the row points at an older *green* commit until the next merge, and the consumer is a once-daily scan. And the upstream workflow's own guidance is to solve it with an ordering attribute and a `ConditionExpression`, which belongs in that shared workflow rather than being reimplemented per caller.

Worth doing there if it bites. Happy to raise it against `solace-public-workflows` if you think it should not wait.

solace-aross (approved):
My blocking finding is resolved, and with a better fix than the one I originally reviewed at `18b4b1e9`, not just a doc patch on top of it: the branch now calls the shared `update-manifest.yaml` reusable workflow directly instead of the `cicd-helper` composite action, so `aws-actions/configure-aws-credentials` and `cicd-helper` are no longer direct `uses:` of this repo at all — one new third-party row instead of two. Verified `7b54c55` (the workflow's pin) resolves to a real merged PR upstream (#133, "reusable update-manifest workflow" in `solace-public-workflows") — I'd already independently confirmed this exact SHA earlier today reviewing #397's bump to the same commit. Both prose spots I flagged are fixed (row count, the "pinned to a single commit" claim now correctly carved out for the separately-pinned `update-manifest.yaml`, with the reason stated). CI is fully green, including `Third-party licenses current` and the `SBOM check script self-test` that would have caught a miss.

The restructure also picks up a real improvement on the write-ordering question your own comment below tracks honestly: moving the write into the shared reusable workflow gets a concurrency-group serialization for free, so two overlapping trunk merges can no longer interleave mid-write, and it's a merge rather than a put now so unrelated attributes survive. You've already said plainly that this doesn't solve full ordering (a slower older run can still land after a newer one) — agreed, and the PR body's "known limitations" section already carries that honestly. Not asking for more here; a same-day drift between the manifest row and the scheduled `dev` image is a data-freshness question for Guardian's scan, not a correctness bug in this workflow.

The permission and gating claims I verified independently last time still hold at this head: `id-token: write` is on both `release-readiness-check.yaml` and `release.yml` for the same subset-of-caller reason, and `needs: [setup, gate]` with no `if: always()` still means a blocked gate leaves the manifest row at the previous commit.

The fix landed as its own commit from you, not pushed by a reviewer and self-approved — thank you for that.

Approved.

.github/workflows/release-readiness-check.yaml
```diff
@@ -73,6 +73,10 @@ jobs:
       pull-requests: write
       checks: write
       statuses: write
+      # Ceiling for guardian-scan's `update-manifest` job (OIDC). It never runs
+      # on this path, but a reusable workflow's permissions must be a SUBSET of
+      # the caller's — under-granting fails at run creation.
+      id-token: write
```
bczoma, line 79: 🟡 **This grant contradicts the ticket, and neither artifact records it**

Flagged by good:pr.

DATAGO-145530 names a specific mechanism — `./maas-build-actions/.github/actions/update-dynamodb-release-manifest` — and states in its Scope section, explicitly: "No new FOSSA scan and **no CRA custom property / OIDC role needed** — the existing action handles the manifest write."

This PR uses the `update-manifest.yaml` reusable workflow from `solace-public-workflows` instead, which assumes an AWS role via OIDC and therefore *does* need `id-token: write` — here, and again on `release.yml:237`. The body explains the subset-rule mechanics of that grant very well; what it never says is that this contradicts the ticket, or why the action the ticket named was rejected. The "that was a separate request" rationale exists only in a review comment, which is not where a future reader looks.

**Why it matters.** This ticket is the CRA onboarding record for this repo and, by its shape, a template for onboarding the next one. It currently asserts something now known to be false. The concrete failure: whoever onboards the next repo reads "no OIDC role needed", omits `id-token: write`, and their `release.yml` breaks **at run creation** — GitHub enforces the subset rule before any step executes, so the release fails with an error pointing at permissions rather than at the ticket that misled them. That is the exact failure this PR's own comments were written to prevent, and the ticket will reproduce it elsewhere.

To be clear about scope: the chosen path is the *better* one on the merits, and I am not asking for it to change. Independent of the request that drove it, moving the role assumption, the default-branch check, the tag exclusion and the write serialisation behind one versioned boundary is the right coupling trade, it drops `aws-actions/configure-aws-credentials` from this repo's surface entirely, and it replaces a registry-trust dependency with a short-lived audience-scoped token — a net reduction in standing privilege.

**Recommended fix.** Add a short `## Deviation from the ticket` section to the body: which action the ticket named, which was used instead, why, and that the ticket's "no OIDC needed" line no longer holds. Then update DATAGO-145530's description to match what shipped.

bczoma (approved):
## PR Review: #343 — ci: record trunk merges in the release manifest

**Author:** AmanRiat1 | **Branch:** `DATAGO-145530-solace-cloud-manifest` → `main` | **Changes:** +71 / -3 across 4 files
**Jira:** `DATAGO-145530` — CRA: Onboard solace-broker-mcp to solace-cloud-manifest + Guardian (Implementing)

---

### Verdict

Conditioned on these 2 — the first has an inline thread on the line in question:

1. 🟡 `.github/workflows/release-readiness-check.yaml:79` — this grant contradicts DATAGO-145530's explicit "no OIDC role needed"; record the deviation in the body and fix the ticket

- 🟡 PR-level — three of the ticket's four ACs land outside this repo and are not enumerated or tracked; say so in the body, and don't let the Story close on this merge

**On the standing blocking review:** @solace-aross's `CHANGES_REQUESTED` at `18b4b1e9` is **substantively satisfied** at this head — I verified it in the file rather than from the author's claim. `THIRD_PARTY_BUILD_TEST.md` now carries six `solace-public-workflows` rows including `.github/workflows/update-manifest.yaml @ 7b54c55`; both prose counts they named are corrected (`:196` "All six come from one repository. Five share a single pin…", `:226` "it now holds six"); the third edit the author identified is present and accurate at `:212-214`; the dated Dependabot narrative was correctly left untouched; and `build-test-licenses-check.sh` passes 47/47. Dropping one row rather than two is right — `configure-aws-credentials` moved inside the shared workflow and no longer appears anywhere in this repo. It remains formally undismissed, so the PR stays BLOCKED until @solace-aross re-reviews; that is a process step, not outstanding work.

---

### Summary

This adds one `update-manifest` job to `guardian-scan.yaml` so trunk merges write the row Guardian's scheduled scan reads, plus the two caller-side `id-token: write` ceilings GitHub's permission-subset rule demands. The engineering is unusually careful: every non-obvious choice is justified both in the body and in an in-file comment at the point of decision, the fail-closed `needs: [setup, gate]` is the right dependency for a write that must not advertise an unvetted revision, and the negatives are volunteered rather than buried. Four findings were filed at 🟡 against the workflow itself and all four were reduced by measurement — the supply-chain claim because the trust it describes predates this PR, the version-skew claim because its central premise was factually wrong, the tag-path claim because the harm needs an upstream security-default flip that a reviewed bump would catch. What is left is a documentation-and-tracking gap rather than a code one: the change diverges from its ticket in a way neither artifact records, and most of the Story's acceptance criteria live somewhere this PR cannot reach.

---

### Minor notes

- 🟢 `THIRD_PARTY_BUILD_TEST.md:196-213` — the SHA pins bound the wiring, not the executing code: `update-manifest.yaml@7b54c55` resolves its write step from `.github/actions/update-manifest@main` one hop in, and the pre-existing `sca/sca-scan@6527948` does the same with `.github/actions/sca/fossa-scan@main` while this repo feeds it `secrets.FOSSA_API_KEY`. Neither `zizmor`'s `unpinned-uses` nor `build-test-licenses-check.sh` can see across that boundary. Add a sentence saying so; upstream re-pinning is the real fix (`./`-relative won't work — it resolves against the caller's checkout). Not this PR's to carry: `main` currently *is* `7b54c55`, and the pattern is already merged on a path holding a live secret.
- 🟢 `.github/workflows/guardian-scan.yaml:398-408` — the `dev` map is written per-merge but describes an artifact built per-night, so this writer structurally cannot record the image's own commit or digest. Not a correctness bug (the per-commit gate scans this commit's own image inline; `version` keys the FOSSA revision, `image_tag` keys the container), but `dev-build.yaml` is the coherent home. If it stays here, point an auditor at `org.opencontainers.image.revision` on the scanned image.
- 🟢 `.github/workflows/guardian-scan.yaml:381` — the role ARN is an identifier, not a credential, and the `id-token: write` grant is correct and minimal; but what actually protects this is a trust policy no reviewer here can see. Worth confirming with the platform team that its `sub` pins `repo:SolaceProducts/solace-broker-mcp` (ideally `:ref:refs/heads/main`) and its permission policy is scoped to `dynamodb:UpdateItem` on this one table, then recording the answer in the comment block.
- 🟢 `.github/workflows/guardian-scan.yaml:374` — the local `if` covers only the event; on the tag-push release path `github.event_name` is still `push`, so the "a release version must never file as trunk's" invariant rests entirely on the upstream `ref_name` check and its two `false` defaults, at a Dependabot-movable pin. Correct today and self-correcting if broken, but one redundant line makes a locally-owned invariant locally checkable: `&& github.ref_type == 'branch' && github.ref_name == github.event.repository.default_branch`.
- 🟢 `.github/ADMIN_SETUP.md:917` — the merge-queue table's "exactly the set the `is_queue` gate suppresses, **no more**" is now incomplete: `update-manifest` also reports skipped on a queue entry, via its push-only guard rather than `is_queue`. A maintainer re-verifying against this list will chase a sixth skip the doc says cannot exist.
- 🟢 `.github/workflows/guardian-scan.yaml:400` — `"squad": "ebp-dax"` is the table's partition key and a new literal that appears nowhere else in this repo, duplicating the squad key in Guardian's `configs/products.json`. A rename or typo writes successfully to a partition nothing reads, and the failure mode is silence. Worth a comment noting it must match the Guardian registration.
- 🟢 `.github/workflows/guardian-scan.yaml:371` — reversibility is asymmetric and undocumented: the job is a clean revert, but the shared workflow only merges, so there is no delete path and an abandoned row keeps being read as current trunk forever. A stale row is worse than no row, because the consumer cannot tell them apart. One line in the comment saying a removal also requires deleting the row or deregistering the product.
- 🟢 PR-level — the branch is 4 commits behind `origin/main`, so "Rebased on current `main`" now reads as false to the next reviewer; the doc's own "Re-pinned 2026-09-08 by #383" line shows main has moved since. Rebase before merge and drop or date that sentence.

and 3 further minor items — a post-merge verification plan for a change that by its nature cannot be exercised until it merges; the title carrying no Jira key while the `autobots/pr-checker` gates waived Jira as "docs-only" on a PR that widens release-workflow permissions; and the `THIRD_PARTY_BUILD_TEST.md` refresh narrative not recording this refresh. Ask if you want them.

---

### Summary Verdict

|---|---|---|
| Correctness | — not reviewed (diff has no application-code surface) | good:code |
| Code Quality | — not reviewed (diff has no application-code surface) | good:code |
| Test Coverage | — not reviewed (diff has no application-code surface) | good:code |
| Documentation | ✅ | docs |

Architecture and Security both opened at ⚠️; each rested entirely on findings that measurement reduced to 🟢/💡, so neither mark is carried.

.github/workflows/release-readiness-check.yaml
AmanRiat1: Fair, and the failure mode you describe is the right one to worry about — the ticket would have handed the next person a permissions error pointing away from its cause. Both artifacts updated.

**PR body** now carries a `## Deviation from the ticket` section: which action the ticket named, that its "no OIDC role needed" line was accurate *for that action* because it authenticates by a different mechanism, that the shared reusable workflow assumes the role via OIDC instead, and that `id-token: write` is therefore required here and on `release.yml`.

**DATAGO-145530** has a `## Correction — what actually shipped` section at the top, before Scope. It states the substitution, spells out the run-creation enforcement explicitly for anyone using the ticket as a template, and the stale clause in the Scope bullet is struck through rather than deleted so the original intent stays legible. I also folded in three things learned while building this that the ticket did not capture and that would bite the next repo the same way: `version` must be the FOSSA revision rather than a branch name (it fails silently, no fallback); `supported_scanners` does **not** gate Prisma on the manifest path, only the `prisma_cloud` key does; and a 404ing `image_tag` discards the product's FOSSA results too, because the fetcher keeps output only when every enabled scanner succeeded.

Confirmed your `release.yml:237` reference is exact.

On the merits — agreed, and worth being clear I did not choose the reusable workflow for the reasons that make it better. It was requested during review; the standing-privilege reduction you describe is a consequence I only traced afterwards.

PR #344, opened 2026-08-26
SOL-153180: Update sam-integration.md with Go-based SAM integration findings

solace-wmaryam:
## Summary

Rewrites `docs/sam-integration.md` for the **Go-based** Solace Agent Mesh (SAM) CLI. The previous guide targeted the Python SAM and no longer works on the Go implementation; this documents the current UI-connector flow, validated end-to-end against a local embedded SAM instance.

Jira: SOL-153180

## Type of Change

- [x] Documentation update

## Motivation and Context

The old guide assumed the Python SAM: `app_module`/`app_config` YAML dropped into `configs/agents/` and picked up by `sam run`. The Go SAM (v2.307.x) has a different architecture — agents are owned by a Platform service and authored via the UI (or `sam config apply`), and an MCP server is attached as an `mcp/remote` **connector**. Following the old steps on the Go CLI does not work, so the guide needed a full rewrite.

## Changes Made

- Rewrote the guide around the Go SAM UI-connector flow: `sam run --embedded` → add the Broker MCP server as a **Remote MCP** connector → attach it to an agent → deploy → chat.
- Documented the `SAM_PLATFORM_ALLOW_PRIVATE_MCP=true` requirement — SAM's SSRF guard blocks loopback/private MCP URLs, so local dev fails without it.
- Corrected UI navigation and field labels (Builder → Connectors → Create Connector → Custom → Remote MCP; MCP Server URL / Connection Type / Authentication Type) and the local UI URL (`http://127.0.0.1:8800`).
- Added a caveat that the embedded dev broker is SAM's internal bus, unrelated to the brokers the MCP server monitors over SEMP.
- Added a brief pointer to the declarative `sam config apply` path (Early Access) as an alternative.

## Testing

Documentation-only change; no code paths touched. The documented procedure was walked end-to-end on a local setup:

- Ran the Broker MCP server (`mcp_client_auth.mode: disabled`) and `sam run --embedded` (SAM v2.307.3).
- Created the Remote MCP connector in the UI — tool discovery succeeded.
- Created and deployed an agent with the connector; a live chat invoked a broker tool (`list-brokers`, success) against a real Solace broker over SEMP.

**Test Environment:**
- Go version: go1.26.3
- OS: macOS (Darwin 25.6)
- SAM: v2.307.3

**Tests Performed:**
- [x] Tested locally by walking the documented steps end-to-end
- [x] Tested with real Solace broker (broker tool call succeeded)

Unit/integration/E2E: N/A — documentation only. CHANGELOG: N/A — docs are not production surface per `CLAUDE.md`.

## Checklist

- [x] docs/ updated
- [x] Self-review completed
- [x] No credentials in code/docs
- [x] Clear, descriptive commit message

## Additional Notes

The headless/Docker-Compose SAM path for the eval harness (which *does* require a non-interactive platform credential) is separate research tracked under SOL-150790 and is out of scope here.

## Related Issues/PRs

- Related to SOL-150790 (Go SAM for the Mode-2 eval harness — research)

---

**For Reviewers:**

**Focus Areas**: Accuracy of the UI steps and field labels against the current Go SAM (v2.307.x), and whether the `SAM_PLATFORM_ALLOW_PRIVATE_MCP` / no-auth framing is the right default for a public-facing guide.

### DIFF
--- a/docs/sam-integration.md
+++ b/docs/sam-integration.md
@@ -1,106 +1,117 @@
 # Connecting solace-broker-mcp to Solace Agent Mesh
 
-This guide shows how to integrate this MCP server with Agent Mesh. For general MCP support in Agent Mesh — connection types, tool filtering, TLS/SSL config, environment variable passing — see the upstream [MCP Integration tutorial](https://github.com/SolaceLabs/solace-agent-mesh/blob/main/docs/docs/documentation/developing/tutorials/mcp-integration.md).
+This guide connects this MCP server to Solace Agent Mesh (the Go `sam` CLI, 2.x)
+by registering it as an **MCP connector** and assigning it to an agent from the
+Agent Mesh UI. For general MCP support in Agent Mesh — connection types, tool
+filtering, auth options, TLS — see the bundled docs (`sam docs`) or
+[docs.solace.com](https://docs.solace.com).
+
+> **This example runs Agent Mesh locally.** `sam run --embedded` starts an
+> in-process dev event broker that Agent Mesh uses for its own internal
+> agent-to-agent messaging. That dev broker is unrelated to the Solace brokers
+> this MCP server monitors — the MCP server still connects to your real
+> broker(s) over SEMP per `broker-config.yaml`. For a team or production
+> deployment you run Agent Mesh against your own broker (Kubernetes or Agent
+> Mesh Cloud); the connector and agent steps below are identical.
 
 ## Prerequisites
 
-- A working Agent Mesh project (`sam init` complete, agents start cleanly)
-- Ability to run this MCP server locally with configured `broker-config.yaml` and `.env` files. For instructions, see [Quickstart](../README.md#quickstart)
+- The Go `sam` CLI installed (`sam --version` reports 2.x).
+- Ability to run this MCP server locally with configured `broker-config.yaml`
+  and `.env` files. See [Quickstart](../README.md#quickstart).
 
 ## Steps
 
-**1. Set a static token on the MCP server** — `broker-config.yaml`:
+**1. Configure client auth on the MCP server** — `broker-config.yaml`. This
+local example uses no client auth (the server binds to loopback only):
 
 ```yaml
 mcp_client_auth:
-  mode: static
-  dev_token: "sam-mcp-dev-token-local-only"  # change for your environment
+  mode: disabled
 ```
 
-For production, use `mode: oauth` instead — see [authentication.md](authentication.md).
+For a shared or production deployment, use `mode: static` (a bearer token) or
+`mode: oauth`, and set the connector's **Authentication Type** in step 4 to
+match. See [authentication.md](authentication.md).
 
-**2. Create the Agent Mesh agent** — `<sam-project>/configs/agents/solace_broker_agent.yaml`:
-
-```yaml
-!include ../shared_config.yaml
-
-apps:
-  - name: solace_broker_agent_app
-    app_base_path: .
-    app_module: solace_agent_mesh.agent.sac.app
-    broker:
-      <<: *broker_connection
-
-    app_config:
-      namespace: ${NAMESPACE}
-      agent_name: "SolaceBrokerAgent"
-      display_name: "Solace Broker"
-      model: *general_model
-
-      instruction: |
-        You manage and inspect Solace event brokers via SEMP-backed MCP tools.
-
-      tools:
-        - tool_type: mcp
-          connection_params:
-            type: streamable-http
-            url: "${SOLACE_BROKER_MCP_URL, http://localhost:9090/mcp}"
-            headers:
-              Authorization: "Bearer ${SOLACE_BROKER_MCP_TOKEN}"
-
-      session_service: *default_session_service
-      artifact_service: *default_artifact_service
-      agent_card:
-        description: "Inspects Solace event brokers via solace-broker-mcp."
-        defaultInputModes: ["text"]
-        defaultOutputModes: ["text", "file"]
-      agent_card_publishing: { interval_seconds: 10 }
-      agent_discovery: { enabled: true }
-      inter_agent_communication:
-        allow_list: ["*"]
-        request_timeout_seconds: 60
-```
-
-**3. Add the matching token to the Agent Mesh `.env` file**:
-
-```env
-SOLACE_BROKER_MCP_URL="http://localhost:9090/mcp"
-SOLACE_BROKER_MCP_TOKEN="sam-mcp-dev-token-local-only"   # must equal dev_token above
-```
-
-**4. Run both in separate terminals**:
+**2. Start the MCP server** (listens on `:9090`, endpoint `/mcp`):
 
 ```bash
-# Terminal 1 — this MCP server
 go run ./cmd/server
-
-# Terminal 2 — Agent Mesh (auto-discovers the new agent)
-sam run
 ```
 
-**5. Verify** — the Agent Mesh logs should include:
+**3. Start Agent Mesh locally** — allow connectors to reach a loopback MCP URL:
 
-```
-Initialized MCPToolset for server: url='http://localhost:9090/mcp'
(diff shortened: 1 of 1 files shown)


docs/sam-integration.md
```diff
@@ -1,106 +1,100 @@
 # Connecting solace-broker-mcp to Solace Agent Mesh
 
-This guide shows how to integrate this MCP server with Agent Mesh. For general MCP support in Agent Mesh — connection types, tool filtering, TLS/SSL config, environment variable passing — see the upstream [MCP Integration tutorial](https://github.com/SolaceLabs/solace-agent-mesh/blob/main/docs/docs/documentation/developing/tutorials/mcp-integration.md).
+This guide shows how to connect this MCP server to the Solace Agent Mesh desktop
+app by registering it as an **MCP connector** and assigning it to an agent from
+the Agent Mesh UI.
```
bczoma, line 5: Worth stamping the SAM version and validation date here.

The guide is pinned to an exact UI navigation path (**Builder > Connectors > Create Connector > Custom > Remote MCP**, the named two-step wizard, the exact field labels, "min 100 characters"), but names no version. UI walkthroughs rot silently — a renamed nav item makes this unfollowable with no signal to either the reader or a maintainer that it has gone stale.

The validation evidence already exists in the PR description (v2.307.3, macOS); it just doesn't survive into the merged doc. One line under the title would carry it:

> Validated against Solace Agent Mesh v2.307.3 (macOS), 2026-08. UI navigation may differ in later versions.

docs/sam-integration.md
```diff
@@ -1,106 +1,100 @@
...
+> messaging. That dev broker is unrelated to the Solace brokers this MCP server
+> monitors — the MCP server still connects to your real broker(s) over SEMP per
+> `broker-config.yaml`. For a team or production deployment you point Agent Mesh
+> at your own broker; the connector and agent steps below are identical.
 
 ## Prerequisites
 
-- A working Agent Mesh project (`sam init` complete, agents start cleanly)
-- Ability to run this MCP server locally with configured `broker-config.yaml` and `.env` files. For instructions, see [Quickstart](../README.md#quickstart)
+- The Solace Agent Mesh desktop app installed.
```
bczoma: Could this prerequisite link the Agent Mesh install / getting-started page?

The rewrite drops the old opening link to the upstream MCP Integration tutorial, which was the file's only external Agent Mesh reference. Removing it is defensible — it was Python-SAM-specific — but the guide now has no link to upstream Agent Mesh docs at all, so a reader who doesn't already have SAM installed can't satisfy this prerequisite from this page.

docs/sam-integration.md
```diff
@@ -1,106 +1,100 @@
...
+Then fully quit the app (Cmd+Q — closing the window isn't enough) and relaunch it;
+an already-running app keeps its old environment and won't pick up the new value.
 
-```
-Initialized MCPToolset for server: url='http://localhost:9090/mcp'
-Agent card tool manifest populated with 24 tools.
-Registered new agent 'SolaceBrokerAgent' in registry.
-```
+This workaround is only needed for local development — a production MCP server
+has a routable URL.
```
bczoma: This caveat isn't right for self-hosted deployments, and I think it's worth fixing because it will send a real class of operator down a dead end.

The guard checks the **resolved IP** at dial time, not the URL string. In a Kubernetes deployment the MCP server is typically reached over a ClusterIP or a private load-balancer address, which is RFC1918 — so the same block fires in production, not just local dev:

```
MCP discovery probe failed: dial tcp 10.96.47.165:9090: address rejected by SSRF guard: private network (RFC1918 / RFC4193)
MCP discovery at save failed; manifest left absent
```

That's a real case from earlier this week, not a hypothetical. And because the check is on the resolved address, the obvious workaround doesn't work either: an Ingress hostname that resolves to a private LB IP, or resolves in-cluster via split-horizon DNS, is still blocked. `SAM_PLATFORM_ALLOW_PRIVATE_MCP` is the actual answer there too — Solace's own Helm values files (`values-staging`, `values-rc-airgap`, `values-rc-manifests`, `values-rc-sam-quickstart`) all set it, so it's the sanctioned path rather than a dev-only escape hatch.

Suggested replacement:

> This is needed whenever the MCP server's URL resolves to a loopback or private (RFC1918/RFC4193) address — local development, and also self-hosted deployments where Agent Mesh reaches the server over a cluster-internal or otherwise private address. It relaxes only private and loopback addresses; link-local (including cloud instance metadata) stays blocked.

One more thing worth knowing: this variable isn't documented anywhere in the published Agent Mesh docs — it's absent from `reference/env-vars.md`, which does list `SAM_MCP_CONNECTOR_TLS_VERIFY`. So for now this guide may be the only place a reader can find it written down, which is an argument for describing its scope accurately here rather than narrowing it to local dev.

docs/sam-integration.md
solace-wmaryam: Done in bfeccfc — added under the title: _Validated against Solace Agent Mesh v2.307.3 (macOS), 2026-08. UI navigation may differ in later versions._

docs/sam-integration.md
solace-wmaryam: Done in bfeccfc — the prerequisite now links the [Solace Agent Mesh docs](https://docs.solace.com/Agent-Mesh/agent-mesh.htm) (install + getting-started).

docs/sam-integration.md
solace-wmaryam: Done in bfeccfc — replaced the local-dev-only caveat with your suggested wording: the guard checks the resolved IP at dial time, so it also fires for self-hosted cluster-internal/private addresses; `SAM_PLATFORM_ALLOW_PRIVATE_MCP` is the sanctioned path there too. Kept the link-local/metadata exclusion note.

bczoma (approved):
Approving — all three comments are addressed in `bfeccfc`, and the substance of the change is solid.

Every claim the guide makes about this repo checks out against source: the `mcp_client_auth` mode values, the default port, the loopback-by-default bind in non-OAuth modes, the `/mcp` endpoint path, `go run ./cmd/server`, and both relative links. The SSRF rewrite in particular is now accurate for the self-hosted case, which is the one most likely to bite someone. The new install link returns 200.

Thanks for validating the flow end-to-end rather than writing it from the UI — the `launchctl` / full-quit detail is exactly the kind of thing that costs an hour when it's missing.

Three small non-blocking items, worth folding into whatever touches this file next:

1. **Step 4's connector URL** uses `http://localhost:9090/mcp`, while step 2 says `http://127.0.0.1:9090/mcp`. The server binds IPv4 only (`internal/defaults/defaults.go:33`), so on a host where `localhost` resolves to `::1` first, tool discovery fails with an error the guide can't help diagnose. Suggest matching step 2.
2. **No revert for the env var.** Worth one line after the `launchctl setenv` block: it applies to everything launched for the rest of the login session, clears at logout or reboot, and `launchctl unsetenv SAM_PLATFORM_ALLOW_PRIVATE_MCP` undoes it now.
3. **What `mode: disabled` exposes.** The default is defensible here — the server binds loopback and the config layer refuses a non-loopback bind under `disabled` without an explicit opt-in, so the dangerous copy-paste is already blocked in code. Half a sentence in the step-1 parenthetical would still help: loopback-only still means any local process can call every tool, including the mutating ones, with the server's broker-admin rights.

Separately, and outside this PR: `SAM_PLATFORM_ALLOW_PRIVATE_MCP` is undocumented upstream — absent from Agent Mesh's `reference/env-vars.md`, which does list `SAM_MCP_CONNECTOR_TLS_VERIFY` — and the discovery error advises "use a publicly reachable server URL", which is wrong for a self-hosted operator. Both are worth raising against `solace-agent-mesh-go`.

Also worth updating the PR description before merge: it still describes the `sam` CLI, `sam run --embedded`, and the UI at `http://127.0.0.1:8800`, all of which commit `fec2609` deliberately replaced with the desktop-app flow. No effect on readers of the merged doc, but it makes the review trail confusing to anyone reading back.

PR #347, opened 2026-08-28
SOL-153693: gate against a third tool registering without input validation

bczoma:
## Summary

Adds a test gate that fails when a newly registered MCP tool accepts schema-invalid arguments. Tests only — no production code touched.

[SOL-153693](https://sol-jira.atlassian.net/browse/SOL-153693)

## Type of Change

- [x] New feature (non-breaking change which adds functionality) — test coverage only

## Motivation and Context

Input validation on this server lives in `ToolManager.CallTool` (`internal/tools/manager.go:243`), **not** in the SDK. Production registers tools with the untyped `mcp.Server.AddTool`, whose contract puts unmarshalling and schema validation on the caller (`mcp/server.go:263`).

Two tools bypass `ToolManager` and are unvalidated today:

- `list-brokers` — accepts undeclared arguments silently, no error at all.
- `describe-semp-schema` — rejects bad input, but as a JSON-RPC error rather than an `isError` tool result.

Those two are SOL-153693's subject and are **not** fixed here. This PR stops a *third* joining them unnoticed, which is the part that gets worse with time.

They didn't bypass the pipeline by accident, incidentally: they're the two tools that take no `broker` parameter, so they can't use the path that injects it. That's why the real fix (a single registration entry point that supports "no broker param" as a first-class case) is design work, and why this gate is worth having in the meantime.

## Changes Made

`TestToolsCall_EveryToolValidatesItsInput` in `cmd/server/conformance_test.go`. It lists every registered tool over the wire, derives a schema-invalid argument set **from that tool's own `inputSchema`**, calls it, and asserts an `isError` tool result.

Probe derivation: take the first `required` property declaring `type: "string"` that isn't `broker`, send `42` for it, and fill every other required property with a plausible valid value so the injected violation is the only one.

**Deriving the probe from the schema rather than hard-coding per-tool knowledge is what makes this a gate** — a tool added tomorrow is covered without anyone remembering to extend a list.

`broker` is excluded as the victim on purpose: `ToolManager` reads and strips it *before* validation, so a non-string broker returns a resolution error and would pass for the wrong reason.

## The allow-list is a three-way gate

`toolsWithoutInputValidation` names the two known-bad tools and fails in three directions:

1. An **unlisted** tool that doesn't validate fails — the point of the gate.
2. A **listed** tool that starts validating fails, so whoever closes SOL-153693 is told to delete the entry rather than leave a stale allow-list.
3. An entry naming a tool **no longer registered** fails as stale.

## Coverage — 35 probed, 5 skipped, of 40

Skips are logged individually with their reason and counted in a summary line, because a gate that silently probes nothing is worse than no gate.

- Four (`get-broker-status`, `get-discard-stats`, `get-redundancy-status`, `list-vpns`) have only `broker` to get wrong.
- **`list-brokers` cannot be probed at all** — it declares no properties, so no argument set violates its schema. Recorded plainly in the allow-list comment rather than left looking like coverage. Its real gap stays pinned by `TestToolsCall_ValidationOutsideToolManagerIsNotConformant`. Making that half live needs `additionalProperties: false` on its schema, which is a production change belonging to SOL-153693.

All 20 write tools are covered — `conformanceHandler` uses `registeredServer(t, true)`.

## Testing

```
gofmt -l cmd/server/                  clean
go build ./...                        Success
go vet ./...                          No issues
go test -race -count=3 ./cmd/server/  738 passed
make test-cover                       Total coverage 89.3% (85% floor)
```

The new test costs ~0.65s — it reuses the shared `conformanceHandler` and one session across all 35 probes.

**Mutation-tested, because a gate that only passes proves nothing.** Removing `describe-semp-schema` from the allow-list:

```
--- FAIL: TestToolsCall_EveryToolValidatesItsInput/describe-semp-schema
    describe-semp-schema accepted a schema-invalid "operation" (an integer where its own
    input schema declares a string) without returning isError:true.
    response: JSON-RPC error (code 0): missing required parameter 'operation'
```

That one failure line independently reproduces all three known `describe-semp-schema` defects — the misclassification, the invalid `code: 0` (SOL-153692), and the wrong-type-reported-as-missing.

**No network.** Validation runs before `handler.Handle`, so a schema-invalid call never reaches SEMP; 34 of 35 probes end in `error_type=validation_error` and the 35th never reaches `ToolManager`. Nothing listens on the broker port during the run.

## Breaking Changes

None. Tests only.

### Code Quality
- [x] Matches the house idiom in `conformance_test.go`
- [x] Self-review completed
- [x] Comments explain "why", not "what"

### Security
- [x] No credentials in code
- [x] No logging statements added

### Testing
- [x] `go test -race ./...` passes
- [x] No flakes across `-count=3`
- [x] Gate verified by mutation, not just by passing
- [x] Test names describe what is being tested

### Documentation
- [ ] README/docs — no user-facing change
- [x] godoc comments on the new helpers and allow-list
- [ ] CHANGELOG.md — no production surface touched

### Git & CI
- [x] Commit signed off (DCO)
- [x] No merge conflicts

## Related Issues/PRs

- Guards SOL-153693; does not fix it
- The `code: 0` in the mutation output is SOL-153692
- Builds on the suite added in #346 (SOL-150761)

---

**For Reviewers:**

**Focus Areas**: Whether the schema-derived probe is the right strategy — it is what makes this a gate rather than a list, but it also means the 5 skips are structural rather than oversights. And whether 5/40 skipped is acceptable, given four of them have only `broker` to get wrong.

**Questions**: Should `list-brokers` get `additionalProperties: false` now so its half of the gate goes live, or does that belong with SOL-153693's design work?

### DIFF
--- a/cmd/server/conformance_test.go
+++ b/cmd/server/conformance_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"sort"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


bczoma:
Closing this — retracting on reflection.

The gate does not fix `list-brokers` or `describe-semp-schema`; it only allow-lists their current behaviour and guards against a third tool joining them. That is a narrower benefit than a standalone PR justifies, and the allow-list risks reading as "handled" when nothing is.

It also does not cover `list-brokers` at all — that tool declares no properties, so no schema-invalid probe can be derived and it is skipped. Making its half live needs `additionalProperties: false` on its input schema, which is a production change belonging to SOL-153693.

The work is preserved on branch `SOL-153693-validation-test-gate` (d09b39f) and is better landed as part of SOL-153693, alongside the fix and the single registration entry point — one coherent change a reviewer can judge whole.

PR #348, opened 2026-08-28
SOL-153712: make the shipped example configs start without an identity provider

bczoma:
## Summary

Both configuration files this project ships for operators to copy — `broker-config.example.yaml` and the ConfigMap in `deploy/kubernetes/` — could not start the server. Both defaulted to `mcp_client_auth.mode: oauth` while omitting the `tool_authorization` block that mode makes mandatory, and nothing tested them.

[SOL-153712](https://sol-jira.atlassian.net/browse/SOL-153712)

## Type of Change

- [x] Documentation update

## Motivation and Context

`kubectl apply -f deploy/kubernetes/` produced a crash-looping pod:

```
INFO   server starting version=v0.8.0
ERROR  failed to load config error="validating config:
       mcp_client_auth.tool_authorization.enabled must be set explicitly
       to true or false when mcp_client_auth.mode is \"oauth\""
```

`broker-config.example.yaml` — the file the README tells every new user to copy — failed identically. Supplying the missing block alone was not enough: the server then tried to reach the placeholder `issuer: https://idp.example.com` and exited.

The validator's own combinations were already well covered (`internal/config`, 332 tests). What was missing was any test that ran the *shipped files* through that validator.

Intent agreed with the product owner: the default example must work out of the box with no identity provider, with OAuth documented as the production path.

## Changes Made

- `broker-config.example.yaml` — defaults to `mode: disabled`. `mode: static` and the full `mode: oauth` block are commented-out samples, co-located at the point where the mode is chosen, and the OAuth instructions now name both settings that are easy to miss (`tool_authorization.enabled`, `tls_terminated_upstream`).
- `deploy/kubernetes/configmap.yaml` — defaults to `mode: static` with `dev_token: "${DEV_TOKEN}"` and `listen_address: "0.0.0.0"`. A pod cannot use the loopback default, and `mode: disabled` on a routable bind would require `allow_remote_unauthenticated: true` — shipping unauthenticated broker-admin access cluster-wide.
- `deploy/kubernetes/secret.yaml` — adds `DEV_TOKEN`, **empty on purpose** (see below).
- `deploy/kubernetes/README.md` — new. Mandatory pre-apply edits, unprefixed GHCR image tags, health probes, the shutdown grace budget, and the OAuth switch-over.
- `internal/config/shipped_examples_test.go` — new. Loads the active config of both shipped files through the server's own `LoadConfig` and asserts neither ships in production mode.
- `README.md` — new Kubernetes section; a note that the dev modes bind container-loopback, so a container needs an explicit bind paired with a token.
- `docs/authentication.md`, `docs/internal/packaging-release.md`, `docs/internal/todo-gaps.md` — link the new README; correct a stale Secret block, probe paths (`/health` → `/readyz`, `/livez`) and `runAsUser` (`65534` → `65532`).

### Two deliberate decisions worth reviewer attention

**`DEV_TOKEN` ships empty, so an unedited apply fails closed.** A working default would mean `kubectl apply` brings up a cluster-routable MCP server whose only credential is a string published in this repository — turning a loud failure into a silent fail-open one. Empty yields `mcp_client_auth.dev_token is required`, one edit from working. This means "works out of the box" here is "starts with no identity provider", not "runs with zero edits" — the manifests already required a broker-URL edit.

**The two files deliberately differ** (`disabled` locally, `static` in k8s). The loopback default makes `disabled` genuinely safe for a local binary; a pod must bind all interfaces, where an authenticated mode is the only responsible default.

## Testing

**Test Environment:**
- Go version: as pinned in `go.mod` (1.25)
- OS: Linux (WSL2)
- Broker version: `solace/solace-pubsub-standard:latest`, in-cluster
- Image under test: `ghcr.io/solaceproducts/solace-broker-mcp:0.8.0`

**Tests Performed:**
- [x] Tested locally with `make check` (build, vet, lint, race, coverage gate — total coverage 89.3% vs the 85% floor)

**Test Coverage:**
- Unedited `kubectl apply -f deploy/kubernetes/` on kind → pod fails closed with `mcp_client_auth.dev_token is required`.
- With `DEV_TOKEN` and a broker URL set → rolls out Ready, 0 restarts; `/readyz` 200; `/mcp` unauthenticated 401; MCP `initialize` returns `serverInfo: solace-broker-mcp v0.8.0`.
- The new test was confirmed to fail when the `tool_authorization` defect is reintroduced, and when a *valid* `mode: oauth` default is reintroduced (the production-mode assertion) — then pass when restored.
- All three documented config variants (default, `static` uncommented, `oauth` uncommented) confirmed to load.
- The container-loopback claim was verified by running the image: the old example logged `addr: 127.0.0.1:9090` with the published port refusing connections; the documented recipe logs `addr: 0.0.0.0:9090` and answers on the published port.

## Breaking Changes

None for a running deployment. Operators who had already edited these files are unaffected. Anyone re-copying them gets a different default mode and must set `DEV_TOKEN` for the Kubernetes path.

## Known limitation

The new test covers the **active** configuration in each file. The commented-out sample blocks are not exercised — env substitution skips comments — so a typo inside one would still ship silently. They were verified by hand here. Closing that properly is deliberately left out of scope.


amitmorade (dismissed):
Nice guard — verified the test catches both the original defect and an oauth-default reintroduction.

One question on scope: `assertStartsWithoutIdentityProvider` pins "not oauth", but the other property the PR defends — the ConfigMap must not ship unauthenticated on a routable bind — isn't pinned. A future edit to `mode: disabled` + `allow_remote_unauthenticated: true` would load cleanly and pass both tests. Would it be worth one extra assertion on the k8s config (e.g. `!cfg.AllowRemoteUnauthenticated`) so that combination fails in CI too, or do you see that as out of scope here?

bczoma:
@amitmorade good catch — the gap was real, and it's now pinned in `2be8825`. I ended up doing something a bit different from the flag check you suggested, so flagging the reasoning.

For the ConfigMap, the property collapses to a single rule. Of the three modes, `oauth` can't start without a reachable IdP and `disabled` can't ship on the routable bind a pod requires — leaving exactly one. So it asserts `Mode == AuthModeStatic` directly rather than ruling the other two out separately. That makes an `AllowRemoteUnauthenticated` assertion redundant *there*: the flag only takes effect under `mode: disabled`, which the mode pin already refuses.

Your flag check does belong somewhere, though — `broker-config.example.yaml`. That's the file that legitimately ships `mode: disabled`, safe for exactly one reason (the dev modes bind `127.0.0.1`), and `allow_remote_unauthenticated` is the sole opt-in that lifts it. Nothing else in that file's assertions would catch it. So it moved there rather than being dropped.

Verified each assertion fails on the case it owns:

| Injected | Result |
|---|---|
| example: `disabled` + `allow_remote_unauthenticated: true` | fails on the flag assertion |
| ConfigMap: `disabled` + `allow_remote_unauthenticated: true` | fails on the mode assertion |
| restored | both pass |

One thing worth recording: the ConfigMap with `mode: disabled` and *no* flag never reaches the assertion at all — `LoadConfig` rejects a non-loopback bind under `disabled` first. So the assertion only needs to cover the flagged variant, which is the one that loads cleanly.

Also added in `4eefa91`: a README note pointing a failed first deploy at `kubectl logs`. `kubectl apply` reports success even when the server exits on an unset `DEV_TOKEN`, so `CrashLoopBackOff` was the only signal a newcomer got.

Still not covered, as called out in the PR description: the commented-out sample blocks aren't exercised — env substitution skips comments — so those were checked by hand.

Note the two follow-up commits were squashed into `2be8825` (force-push, identical tree), which dismissed your approval — sorry for the extra round trip.

solace-awaheed (approved):
Nothing flagged on my end, LGTM

PR #349, opened 2026-08-29
SOL-153364: propagate correlation ID into the detached token exchange context

amitmorade:
## Summary

Every log line in `internal/tokenexchange` that can carry a `correlation_id` now does — the detached-context exchange lines, the recovered-panic ERROR, the Retry-After gate WARNs, and the audience-mismatch WARN — closing a gap against the correlation contract in `docs/observability.md` ("logged as the `correlation_id` attribute on every log line within the request"). The two lines that stay context-free (breaker state-change, startup config-load) have principled reasons stated in comments.

Jira: SOL-153364

## Type of Change

## Motivation and Context

`runExchangeOnce` roots the exchange context at `context.Background()` — deliberately, so one caller's cancellation cannot abort the singleflight exchange other waiters depend on. Rooting there also dropped every context value, including the correlation ID, so the seven log lines under that context (the IdP request/response lines, the cache-put outcomes including the `token cache put failed` WARN, and retries-exhausted) could not be joined to any request. Diagnosing a token issue across concurrent brokers meant timestamp guessing.

Design choices, made deliberately:

- **The lines carry the singleflight winner's ID.** The winner is the request that triggered the work; each waiter's own `broker token exchange completed` line already carries its own ID. (Whether waiters should additionally be able to see the shared exchange from their ID — a flight identifier or a trace span — is raised separately against the observability epic SOL-150253; it is schema vocabulary, not a bugfix.)
- **A `corrID string` parameter, not the caller's `context.Context`,** threads the value down — a string cannot re-couple caller cancellation to the shared call.
- **Not `context.WithoutCancel(ctx)`:** it would carry every request-scoped value — including the raw subject token — onto a context that outlives the request.
- **Seeding is unconditional:** the correlation slog handler emits nothing for an empty ID (pinned by `TestSlogHandler_EmptyIDNotEmitted`), so correlation-disabled output is byte-identical.

## Scope note

The ticket originally scoped only the detached-context lines and excluded the package's context-free `slog.Warn`/`slog.Error` sites. Two review findings extended it deliberately: Copilot flagged the panic-recovery line (the winner's ID was already in scope), and an adversarial re-check of the remaining sites found the audience-mismatch WARN already has the seeded context as its caller's `ctx`, and the gate WARNs are one parameter away. All are fixed here. The breaker state-change WARN is context-free by design (a transition is the verdict on a window of failures; attributing it to one request would mislead), and the config-load WARN runs at startup before any request exists — both now documented in comments rather than left as vague follow-ups.

## Changes Made

- `internal/tokenexchange/exchange.go`: capture `correlation.From(ctx)` before the singleflight dispatch; thread it through `runProtectedExchange` → `runExchangeOnce`; seed it onto the detached context immediately after the `WithTimeout(Background(), …)` detach.
- `internal/tokenexchange/exchange_test.go`: three new tests plus a `captureJSONLogs` helper (see Testing).
- `CHANGELOG.md`: `[Unreleased]` → Fixed entry.
- `docs/observability.md`: the Correlation ID section now documents winner-ID attribution for shared exchanges, the grep pivot when a request's ID shows no IdP lines, the failure-path behavior, and the two by-design ID-free lines.
- `internal/tokenexchange/{retry_after_gate,audience_check}.go`: the Retry-After gate WARNs and audience-mismatch WARN now log under the seeded context; `circuitbreaker.go`/`from_config.go` carry comments stating why their WARNs deliberately stay ID-free.

## Testing

**Test Environment:**
- Go version: 1.25.0
- OS: macOS (darwin/arm64)
- Broker version: n/a (loopback `httptest` IdP)

**Tests Performed:**
- [x] Tested locally with `go test -race ./...` (`make check`: coverage 89.3% vs the 85% floor)

**Test Coverage:**
- `TestExchange_CorrelationIDSurvivesDetach` — a seeded caller ID appears on every line of a cold-path exchange, including the detached cache-put line (the get↔put join).
- `TestExchange_NoCorrelationIDWhenCapabilityDisabled` — unseeded ctx (stands in for `OBS_CORRELATION_ID_ENABLED=false`; with the capability on, the middleware generates an ID for clients that send none): no `correlation_id` key emitted at all.
- `TestExchange_DetachedLogsCarryWinnerIDAfterCallerBails` — the ticket's motivating case: caller cancels mid-exchange, the detached lines still emit and carry the winner's ID.
- `TestExchange_DetachedLinesCarryWinnerIDNotWaiterID` — two callers: detached lines carry the singleflight winner's ID, each completion line carries its caller's own (the headline semantic, previously only pinned in the degenerate single-caller case).
- `TestExchange_GateAndExhaustionLogsCarryCorrelationID` — end to end through `Exchange` against a 429+Retry-After IdP: the gate WARN and retries-exhausted DEBUG carry the caller's ID (verified to fail if the call site reverts to a bare context).
- `TestExchange_AudienceMismatchLogCarriesCorrelationID` — end to end through `Exchange` with a mismatched-aud JWT: the audience WARN carries the caller's ID (same sabotage-verified wiring pin).
- `TestExchange_RecoveredPanicLogCarriesWinnerID` — a panic inside the singleflight closure surfaces as an ordinary error, the ERROR line carries the winner's ID, and the panic VALUE appears in no log field (type and stack only).
- Red-green verified: with the fix stashed, exactly the three detached-path assertions fail (`correlation_id = ""`); caller-path lines pass.
- Cancellation isolation stays pinned by the existing `TestExchange_DetachedCallWarmsCacheAfterCallerBails`, untouched.

**Note for reviewers — deliberate deviation from the package's log-test convention:** the new tests do not use `captureLogs`. Its `captureHandler` discards the record context, and `correlation_id` only exists post-handler, so the behavior under test is unobservable through it. `captureJSONLogs` instead installs the production handler composition (`correlation.NewSlogHandler` over a JSON handler, as in `cmd/server/main.go`) and asserts on parsed emitted output. Migrating the existing helper is proposed as a follow-up, not mixed into this fix.

## Breaking Changes

None. No control-flow, level, or message changes; one added structured field on existing lines. With correlation disabled, output is unchanged.

### Code Quality

### Security
- [x] Followed secure logging rules (`/check-logs` clean: no CRITICAL/HIGH)
- [x] Credential-carrying types implement `slog.LogValuer` (n/a — none added)

### Testing
- [x] Error paths tested

### Documentation
- [x] CHANGELOG.md updated
- [x] godoc comments added/updated for exported functions (n/a — no exported surface changed)

### Git & CI
- [x] No merge conflicts

## Additional Notes

Follow-ups deliberately kept out of this PR:
- Migrate `captureLogs` (13 call sites) to the production handler composition so context-derived fields are assertable package-wide.
- The two remaining context-free log lines (circuit-breaker state-change, startup config-load) are deliberate, with reasons in code comments — nothing to ticket.
- Waiter-side visibility of the shared exchange — raised against SOL-150253.

## Related Issues/PRs

- Related to SOL-153364 (this ticket), SOL-153363 / #333 (message rework on the same lines), SOL-150253 (observability epic)

---

**For Reviewers:**

**Focus Areas**: the seed site in `runExchangeOnce` — confirm the detached context still roots at `context.Background()` and nothing re-couples caller cancellation.

**Questions**: none.


internal/tokenexchange/exchange.go
```diff
@@ -122,7 +129,7 @@ func (e *Exchanger) Exchange(ctx context.Context, input ExchangeInput) (*Token,
 				err = fmt.Errorf("token exchange panicked (%T)", r)
 			}
 		}()
-		return e.runProtectedExchange(key, input)
+		return e.runProtectedExchange(key, input, corrID)
```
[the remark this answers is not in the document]
amitmorade: Applied in 4c86762 — the panic line now logs via `ErrorContext` under a fresh `Background()`-rooted context seeded only with the winner's `corrID`, so it's attributed to the triggering request while staying severed from caller cancellation. Also added the package's first panic-path test (`TestExchange_RecoveredPanicLogCarriesWinnerID`), which pins both the recovery contract and the ID on the line.

The remaining context-free log sites in this package (retry-after gate, circuit breaker, audience check, config load) are left unchanged deliberately: no correlation ID is in scope at those call sites, so they need real plumbing — tracked as follow-up work outside this PR.

solace-aross (dismissed):
Reviewed the correlation-ID threading through the detached exchange path. Verified rather than trusted the PR's own red-green claim: sabotaged the `correlation.With(exchCtx, corrID)` seed in `runExchangeOnce` in a scratch worktree and reran the suite. Exactly the five detached-path assertions failed (`CorrelationIDSurvivesDetach`, `DetachedLogsCarryWinnerIDAfterCallerBails`, `GateAndExhaustionLogsCarryCorrelationID`, `AudienceMismatchLogCarriesCorrelationID`, `DetachedLinesCarryWinnerIDNotWaiterID`); the panic-recovery and disabled-capability tests correctly stayed green since that seed is untouched. Restored, suite green again.

Traced the context threading by hand: `corrID` is read synchronously off the caller's ctx before the singleflight dispatch (no race with the goroutine), carried as a plain string rather than a context (so caller cancellation can't re-couple to the shared call), and only re-seeded onto the detached `exchCtx` — not `context.WithoutCancel`, which would leak the raw subject token onto a context that outlives the request. `doExchange` and `warnIfAudienceMismatch` both receive `exchCtx`, not the caller's `ctx`, so the audience-mismatch WARN is attributed correctly too.

CHANGELOG and `docs/observability.md` match the code: the winner-ID-on-shared-exchange semantics, the two by-design ID-free lines (breaker state-change, startup config load), and the pivot advice for an operator grepping by request ID. `go build ./...` and `go test ./internal/tokenexchange/...` both clean.

Approving.

docs/observability.md
```diff
@@ -572,6 +572,19 @@ metric, audit, and trace schemas.
...
+all of them, and the lines describing that work — the identity-provider request/response, the
+cache write, retry exhaustion, the Retry-After gate — carry the correlation ID of the request
+that initiated it. Every request still logs its own `broker token exchange completed` line
+under its own ID. So if a request's ID turns up no identity-provider lines, that request rode
+an exchange another request started: pivot to the `broker` attribute plus the time window
+around the request's own completion line. A failed exchange surfaces through each caller's own
+error handling rather than a per-caller line from the exchange itself. Two lines never carry a
+correlation ID by design: the circuit-breaker state-change WARN (a transition is the verdict
+on a window of failures, not on any one request — filter on its `breaker` attribute instead)
+and the startup configuration WARNs, which run before any request exists.
```
bczoma: Nice section, but the enumeration undersells the fix: it lists only 4 of the 6 log sites this PR actually changed (missing the audience-mismatch WARN and the recovered-panic ERROR, both named in the CHANGELOG entry for this same PR). An operator troubleshooting one of those two lines won't learn from this paragraph that it now carries `correlation_id`. Also "startup configuration WARNs" is plural but there's only one (`from_config.go:95`).

```suggestion
**Shared broker-token exchanges log under the initiating request's ID.** When several
concurrent requests need the same broker token, the server performs one exchange on behalf of
all of them, and every line describing that work — the identity-provider request/response, the
cache write, retry exhaustion, the Retry-After gate, an audience mismatch, and a recovered
panic — carries the correlation ID of the request that initiated it. (The recovered-panic line
is the one exception to "seeded from the initiating request's context": it is logged from a
fresh, unrelated context so a panic can't be mistaken for carrying request state, with only
the correlation ID copied across.) Every request still logs its own `broker token exchange
completed` line under its own ID. So if a request's ID turns up no identity-provider lines,
that request rode an exchange another request started: pivot to the `broker` attribute plus
the time window around the request's own completion line. A failed exchange surfaces through
each caller's own error handling rather than a per-caller line from the exchange itself. Two
lines never carry a correlation ID by design: the circuit-breaker state-change WARN (a
transition is the verdict on a window of failures, not on any one request — filter on its
`breaker` attribute instead) and the startup configuration WARN, which runs before any request
exists.
```

amitmorade: Good catch — both fixed in b5adbea: the enumeration now names all six lines (audience-mismatch WARN and recovered-panic ERROR added), and the startup WARN is singular.

solace-aross (approved):
Re-review after the docs-only follow-up. bczoma's catch was right — the enumeration was missing the audience-mismatch WARN and the recovered-panic ERROR, both of which the CHANGELOG for this same PR already claims. `b5adbea` adds both and fixes the stray WARNs plural. Checked the panic line against `exchange.go`: it does carry the winner's correlation ID via a `Background()`-rooted context, matching what the doc now says.

Approving.

bczoma (approved):
Re-reviewed after the two follow-up commits. `4c86762` (panic-recovery correlation ID, already in fb850fc at time of full review) and `b5adbea` (completes the shared-exchange log-line enumeration in `docs/observability.md`, resolving the one 🟡 finding from the full review) both check out. Full review covered PR hygiene, architecture (cancellation isolation verified end-to-end), code quality, security (SECURE framework — no secret/PII leakage, correlation ID sanitized once at the trust boundary), and docs. No outstanding findings.

PR #351, opened 2026-08-31
SOL-153766: stop SEMPError.Error() from logging unaudited response body

solace-aross:
## Summary

Security fix. `SEMPError.Error()` fell back to the raw, unparsed HTTP response body whenever a SEMP call's error response didn't match the broker's own `meta.error` envelope. `internal/tools/manager.go`'s audit logging calls `Error()` into the ERROR-level `detail` log field, under a comment claiming these broker error types are "audited" and carry no unreviewed content.

That claim held for the `Description` branch (parsed from the broker's own `meta.error.description`), but not for the fallback: any response that doesn't parse as a SEMP `meta.error` envelope — a reverse proxy, API gateway, or WAF returning its own error page instead of the broker's — landed verbatim in the audit log. That content is never reviewed by this codebase and can carry session details or other operational internals, which is exactly what `docs/internal/secure-logging-rules.md` Rule 5 warns about for external error content.

## Fix

`SEMPError.Error()` now drops the `Body` fallback entirely, reporting just `"<operation> returned HTTP <status>"` when `Description` is empty. `Body` remains on the struct for debugging but is deliberately rendered by no sink, matching the existing documented posture of `sempv1.Error.Body`. No code change was needed at the `manager.go` logging call site itself — fixing `Error()` at the source keeps the outer `%w` wrap chain intact (composite tool step / `forEach` context still shows up in `detail`), which a separate "safe detail" helper called on the unwrapped error would have lost. The `manager.go` comment and the `maxErrorTextLen` doc comment are corrected to describe the now-true guarantee.

The `Description` branch (genuinely broker-authored text) is unchanged and continues to log and surface to the agent exactly as before.

## Testing

- New regression test (`TestCallTool_SEMPv2BodyFallbackNotLoggedRaw` in `manager_test.go`): constructs a `SEMPError` with empty `Description` and a marker string in `Body`, runs it through `CallTool`, and asserts the audit log's `detail` field never contains the marker. Confirmed it fails against the pre-fix code and passes after.
- Flipped `TestClient_Execute_SEMPErrorMalformedBody` in `client_test.go` from asserting `Error()` falls back to `Body` to asserting it does not.
- `TestCallTool_SanitizesResponseButPreservesLogDetail` (the `Description`-branch case) is unchanged and still passes, confirming that behavior is untouched.
- `make check` clean (build, vet, lint, race-enabled tests, 85% coverage gate).
- `/check-logs` run on the diff: no violations; confirmed Rule 5 compliance.
- This plan and the resulting diff each went through an independent skeptical review pass before this PR was opened.

## Scope

`internal/semp/sempv2/client.go`, `internal/tools/manager.go`, and their tests, plus a `CHANGELOG.md` `[Unreleased]` Security entry. `sempv1.Error`, `RetriesExhaustedError`, and `ExchangeError` were checked and don't have the analogous issue, so they're untouched.

Closes SOL-153766.


solace-awaheed (commented):
Fix is correct and minimal. Verified independently:

- `buildSEMPv2Message` (`internal/tools/errors.go:375`) **already** had `"<op> returned HTTP <status>"` as its default branch, so the agent-facing path was never leaking. This brings `Error()` into line with it — blast radius really is the log path only.
- `RetriesExhaustedError.Error()` renders `e.Err`, never `e.Body`/`e.Detail` — the "checked, not affected" claim holds.
- The `isV2` log branch already emits `http_status`, `semp_code`, `operation` as structured fields, so the dropped text was near-redundant.

Build clean, all 356 tests in the two touched packages pass.

---

`parseSEMPError`'s doc is the one place in the file not updated, and it asserts exactly what this PR disproves:

> `// zero-valued and Body carries the raw response. Description and Body are`
> `// broker-controlled and truncated at maxErrorTextLen.`

`Body` being *not* broker-controlled is the whole premise. Suggested:

```go
// zero-valued and Body carries the raw response. Description is
// broker-controlled; Body may not be (see the field doc). Both are
// truncated at maxErrorTextLen.
```

### 2. `internal/tools/manager.go` (isV2 attrs branch) — operators can't distinguish a WAF page from a broker error *(suggestion, non-blocking)*

`detail` is now `"getMsgVpnQueue returned HTTP 502"` in two very different cases: the broker returned a valid `meta.error` with an empty description, or something in front of the broker returned an HTML page. Those warrant different operator responses, and the log no longer separates them.

Cheap fix that leaks nothing — add a length on the `isV2` branch:

```go
slog.Int("body_len", len(sempv2Err.Body)),
```

Non-zero `body_len` with no `semp_code`/description is an unambiguous "non-SEMP response reached us" signal. Fine as a follow-up.

### 3. `internal/semp/sempv2/client_test.go:925` and `internal/tools/manager_test.go:779` — assert the positive too *(nit)*

Both tests only assert *absence* of the marker, so they pass vacuously if `detail` goes empty or the format changes. Pin the new contract:

```go
if got := sempErr.Error(); got != "testOp returned HTTP 403" {
    t.Errorf("Error() = %q, want operation and status only", got)
}
```

Same idea in `manager_test.go` — assert `detail` still contains `"returned HTTP 502"` alongside the marker check.

### 4. `internal/tools/manager.go:391` — precision in a security-critical comment *(nit)*

> `// each one's Error() implementation is verified to render only broker- or server-generated text`

Not quite what the code does: `errors.As` matches the audited type anywhere in the chain, but `detail = (*toolErr).Error()` renders the **outermost** error's text. The guarantee also depends on every wrapper being package-authored — true today, and exactly the wrap chain the PR description relies on preserving. Worth one clause so nobody later wraps a `SEMPError` in something unaudited and quietly widens the gate.

solace-aross:
@solace-awaheed thanks for the review. Addressed in 2d236f9:

1. Fixed `parseSEMPError`'s doc comment — it now says Description is broker-controlled, Body may not be.
2. Left the `body_len` structured-field addition as a follow-up, per your own scoping (non-blocking, separate ticket makes sense so it doesn't get bundled into this fix).
3. Strengthened both test assertions to pin the exact post-fix string/contains contract rather than only asserting the leaked marker is absent.
4. Reworded the manager.go audit comment to spell out that errors.As matches anywhere in the wrap chain but detail renders the outermost error, so the guarantee depends on every wrapper staying package-authored.

make check clean (89.3% coverage).

solace-awaheed (approved):
LGTM, thanks!

PR #352, opened 2026-08-31
SOL-153764: compile composite step templates once at load

solace-aross:
## What

The composite tool executor re-parsed every step's Go template (`args` and
`forEachIf`) on every single step call instead of compiling it once at load.
`Step.Args` stored only the raw template source; `ResolveArgs` and
`resolveTemplateString` called `template.Parse` fresh, every call.

Composite tool definitions load once at startup and never change afterward,
so the source text is static. This PR compiles each step's `args`/`forEachIf`
templates once in `validateTool` (called from `LoadTools` at startup) and
caches the resulting `*template.Template` on `Step`. The per-call path in
`ResolveArgs`/`resolveTemplateString` now executes the cached template
instead of re-parsing it, falling back to parse-on-the-spot only for a `Step`
built directly in Go rather than through `LoadTools` (as many existing tests
do) — so no existing test needed to change.

## Why

`list-vpns`' `real-clients` step fans out per VPN (up to the 500-item cap),
and each fan-out row triggers 4 template parses (1 `forEachIf` + 3 `args`) —
roughly 2,000 re-parses for a single `list-vpns` call against a broker with
hundreds of VPNs. That's avoidable work on a hot path for source text that
never changes after startup.

## Diff summary

- `internal/composite/definition.go`: adds unexported `compiledArgs
  map[string]*template.Template` / `compiledForEachIf *template.Template`
  fields on `Step` (invisible to the YAML schema — yaml.v3's
  `KnownFields(true)` ignores unexported fields).
- `internal/composite/executor.go`: adds `compileStepTemplates`; changes
  `ResolveArgs`/`resolveTemplateString` to execute the cached template when
  present, falling back to parsing when not (steps built directly in Go,
  bypassing the loader).
- `internal/composite/loader.go`: `validateTool` compiles every step's
  templates once, right before returning — a side effect is a malformed
  template in `tools.yaml` now fails at server startup instead of only on
  first invocation.
- `internal/composite/template_cache_test.go` (new): asserts the real
  embedded `tools.yaml` gets its templates precompiled; proves the per-call
  path actually executes the cached template (by corrupting a step's raw
  template source *after* compiling it and confirming execution still
  succeeds with the pre-corruption result, for both `args` and `forEachIf`);
  and covers the new load-time rejection of a malformed template.

`make check` (build/vet/lint/race tests + 85% coverage gate) passes clean;
coverage 89.4% (baseline 89.3%).

Reviewed by a Fable-model plan review and a separate Fable-model diff review
(the latter including build-overlay mutation testing to confirm the new
tests actually fail without the fix), plus a self `/good:review` pass.

Tracked under [SOL-153764](https://sol-jira.atlassian.net/browse/SOL-153764).


solace-awaheed (dismissed):
LGTM, non-blocking comments: 

Went looking for races and lifecycle bugs in the shared-template cache; didn't find any. Verified rather than assumed: the compile lands on the real slice elements (`&file.Tools[i]` / `&tool.Steps[i]`, not copies), nothing mutates the cached templates after load, nothing writes `step.Args`/`ForEachIf` at runtime so the cache can't go stale, and `text/template`'s `Execute` doesn't touch its receiver. The `Concurrency:` block on `Step` is accurate. `go vet` and `go test -race ./internal/composite/...` clean.

One optional nit: no test actually executes a *shared compiled* template concurrently. `fanOutTool()` builds Steps as Go literals so `compiledArgs` is nil and the existing fan-out tests all take the fallback path, and the new `TestFanOut_ForEachIf_...` test uses a single row. So `-race` passes vacuously for the property the comment promises. Cheapest fix is more rows in that test (`template_cache_test.go:94`):

```go
Data: vpnRows(
    map[string]any{"msgVpnName": "a", "enabled": true},
    map[string]any{"msgVpnName": "b", "enabled": true},
    map[string]any{"msgVpnName": "c", "enabled": true},
    map[string]any{"msgVpnName": "d", "enabled": false},
),
```

with the assertion bumped to `len(byKey) != 3` — keeps the `forEachIf` skip meaningful and gets 3 goroutines onto the same template. Coverage gap, not a bug.

solace-aross:
@solace-awaheed good catch. Fanned the forEachIf test out to 3 enabled rows so -race actually exercises multiple goroutines against the same compiled template, not just one. Pushed.

solace-awaheed (approved):
LGTM

PR #354, opened 2026-08-31
SOL-152089: default to 2 replicas with a PodDisruptionBudget

solace-awaheed:
Adds poddisruptionbudget.yaml, raises the replica default to 2, and reconciles container resources (no limits.cpu, memory 128Mi/512Mi) with the story's intent.

Availability here is three separate problems with three enforcement points, and treating them as one was the main defect in the first draft:

  - Rolling updates are governed by strategy.rollingUpdate, not by a PDB (the Eviction API is the only thing that consults one). The guarantee was resting on the inherited maxUnavailable: 25%, which gives 0 only via floor(0.25 x 2) at exactly two replicas, so it is now pinned.
  - Drains and cluster upgrades are the PDB's job.
  - Node failure is nobody's job, so topologySpreadConstraints asks the scheduler to place the replicas on different nodes. ScheduleAnyway, not DoNotSchedule, so single-node dev clusters still schedule both.

The PDB uses maxUnavailable: 1 rather than minAvailable: 1. They are equivalent at two replicas, but minAvailable: 1 against replicas: 1 permits zero disruptions and hangs that node's drains indefinitely -- reachable by scaling a copy down, and it survives a revert because kubectl apply -f does not prune.

sessionAffinity: ClientIP pins a client to the pod holding its session, which is required above one replica. kube-proxy applies it on the ClusterIP path only, so an Ingress, Gateway, or service mesh bypasses it entirely -- including the TLS-terminating ingress topology docs/authentication.md recommends for OAuth. That precondition is now documented in the manifest, the observability doc, and the changelog.

Also recorded: the PDB protects endpoint availability, not session continuity, since every pod replacement drops its in-process sessions; the per-pod state a second replica doubles, notably that semp.max_concurrent_per_broker is per-process, so two replicas put twice the concurrency on a broker management plane that limit exists to protect; and the correlated-OOM shape of two symmetric replicas sharing one memory ceiling.

Two claims in the draft were wrong and are corrected. requests.cpu protects this pod from its neighbours, not neighbours from this pod -- dropping the CPU limit is what makes this pod the potential noisy neighbour, which is a LimitRange or ResourceQuota decision. And the Go constant vs YAML field coupling is checkable in principle, since yaml.v3 is already a direct dependency; that guard has simply not been written.

TestSession_IsNotPortableAcrossInstances pins the premise all of the above rests on: a session ID is valid only on the instance that issued it. That behavior lives in the SDK, which Dependabot bumps daily, and was asserted nowhere. Verified by mutation -- making the two handlers share state fails it.

docs/user-guide.md gains a three-cause triage table for 404 session not found, which previously named only the stale-after-restart case, and docs/internal/packaging-release.md now lists poddisruptionbudget.yaml, which the directory-wide kubectl apply was deploying undocumented.

Verified: go build, go vet, tests at 89.3% coverage against the 85% floor, all three manifests parse, selector chain consistent across PDB, Service, Deployment, pod labels, and the spread constraint. Not verified: AC4's live rollout and AC1's kubectl --dry-run=client, both of which need a cluster.

## Summary

Makes the example Kubernetes manifests survive routine cluster operations: two replicas, a PodDisruptionBudget for drains, a pinned rollout strategy for updates, and topology spread for node failure. Documents the session-affinity precondition that multi-replica correctness depends on.

Tracked under SOL-152089.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

A single-replica deployment goes fully dark during any rolling update or node drain. Customer SREs need the shipped example to survive routine cluster operations without hand-editing it first.

The complication is that this server is **stateful**: the MCP streamable-HTTP handler issues an `Mcp-Session-Id` and holds the session in process memory, per pod. Scaling to two replicas without pinning clients to their pod turns every follow-up request into a coin flip that lands on `404 session not found`. So the replica bump and the session-affinity work are one change, not two.

## Changes Made

- **`deploy/kubernetes/poddisruptionbudget.yaml`** (new) — `maxUnavailable: 1`, selector matching the Deployment's pod labels. `maxUnavailable` rather than `minAvailable` so the object stays safe at `replicas: 1` instead of deadlocking that node's drains.
- **`deploy/kubernetes/deployment.yaml`** — `replicas: 1` → `2`; pinned `strategy.rollingUpdate` (`maxUnavailable: 0` / `maxSurge: 1`); added `topologySpreadConstraints` (`maxSkew: 1` over `kubernetes.io/hostname`, `ScheduleAnyway`); removed `limits.cpu: 500m`; `requests.memory` 64Mi → 128Mi; `limits.memory` 128Mi → 512Mi.
- **`deploy/kubernetes/service.yaml`** — `sessionAffinity: ClientIP` with `timeoutSeconds: 10800`, plus the caveat that kube-proxy applies it on the ClusterIP path only.
- **`docs/observability.md`** — new "Deployment Topology and Resource Policy" section: which mechanism covers which disruption, why the PDB protects endpoint availability rather than session continuity, what other state is per-pod, and the resource rationale the story asked to be recorded here.
- **`docs/user-guide.md`** — three-cause triage table for `404 session not found` (wrong pod / idle timeout / pod recycled), replacing text that named only the stale-after-restart case.
- **`docs/internal/packaging-release.md`** — manifest inventory now lists `poddisruptionbudget.yaml` and mentions the affinity and replica defaults.
- **`cmd/server/session_timeout_test.go`** — `TestSession_IsNotPortableAcrossInstances`.
- **`CHANGELOG.md`** — `[Unreleased]` entry under Changed.

## Testing

**Test Environment:**
- Go version: 1.26.5
- OS: macOS 26.6.1 (darwin/arm64)
- Broker version (if applicable): n/a — no broker interaction in this change

**Tests Performed:**

**Test Coverage:**
- `TestSession_IsNotPortableAcrossInstances` — two `StreamableHTTPHandler` instances stand in for two pods. Asserts a session minted on pod A returns 404 on pod B **and** 200 on pod A. Both directions matter: a bare 404 from B would also be produced by a handler that rejects every session, which would pass a one-sided check while proving nothing.
- Mutation-verified: pointing both handlers at the same instance makes the test fail with its intended diagnostic, so it is not vacuously green.
- `go test -race ./...` — clean, no data races.
- `make test-cover` — total coverage 89.3%, above the 85% floor.
- `go build ./...` and `go vet ./...` clean.
- All three manifests parse as valid YAML; asserted structurally that `replicas: 2`, `maxUnavailable: 0`/`maxSurge: 1`, `whenUnsatisfiable: ScheduleAnyway`, PDB `maxUnavailable: 1` with no `minAvailable`, no `limits.cpu`, and that the label selector is identical across PDB, Service, Deployment `matchLabels`, pod template labels, and the spread constraint.

**Not verified (needs a live cluster):**
- AC4's `kubectl rollout restart` keeping a Ready pod throughout.
- AC1's literal `kubectl apply --dry-run=client` — validation is client-side but still requires API discovery for kind→resource mapping.

## Breaking Changes
- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

### Code Quality

### Security
- [ ] Credential-carrying types implement `slog.LogValuer` — n/a, no new types

### Testing
- [x] Error paths tested

### Documentation
- [ ] README.md updated (if user-facing changes) — n/a, README does not cover `deploy/`
- [ ] godoc comments added/updated for exported functions — n/a, no exported API added

### Git & CI
- [x] No merge conflicts
- [ ] CI checks passing (lint, build, test, E2E) — pending first run

n/a

## Additional Notes

`make check` currently fails at the lint step on pre-existing staticcheck `SA5011` findings in `internal/tools/sempv1/redundancy/response_test.go` and `cmd/server/session_timeout_test.go`. These reproduce on a pristine `c949e9a` with this branch's changes stashed, so they are not introduced here — likely a local golangci-lint version ahead of the one CI pins. This branch adds zero new lint findings.

Follow-ups deliberately left out of scope:
- No CI validation exists for `deploy/` at all (no kubeconform/kubeval/`kubectl --dry-run`; `.fossa.yml` excludes `./deploy`, and `production-surface.sh` does not match it). Adding a `kubeconform` step to the lint job would turn AC1 into a durable gate rather than a one-time manual check.
- The 3h affinity window vs the 2h `defaults.DefaultMCPSessionIdleTimeout` is documented but unenforced. It is checkable — `gopkg.in/yaml.v3` is already a direct dependency — and could cover the `terminationGracePeriodSeconds` coupling in the same test.
- No pod identity in logs (no `POD_NAME` downward-API env var) and no session-count metric, so per-pod attribution of a 404 is not currently possible from logs alone.
- `image: ...:latest` with two replicas now permits live version skew between pods and makes `kubectl rollout undo` a no-op. Pre-existing, but newly consequential.

## Related Issues/PRs

- Implements SOL-152089 (Story 30), Epic SOL-150253
- Builds on SOL-151288 (graceful-shutdown budget) and SOL-153582 (session idle timeout)


solace-aross (dismissed):
Approving — independently verified the mechanics rather than trusting the (already unusually rigorous) PR description.

Sabotage-tested the new test, not just read it: in a scratch worktree, pointed `podB` at the same instance as `podA` (`podB := podA`) and reran — it fails exactly as described (`got status 200 on podB, want 404`), confirming this genuinely pins the cross-instance-session-rejection premise rather than passing vacuously. Restored and reran clean.

Checked the manifest claims structurally rather than by eye: all three YAML files parse; label selectors are identical across the Deployment's `selector.matchLabels`, pod template labels, the PDB's `selector.matchLabels`, and the `topologySpreadConstraints` `labelSelector` (`app.kubernetes.io/name: solace-broker-mcp` everywhere) — so the PDB and spread constraint actually cover the pods they claim to. `replicas: 2`, `strategy.rollingUpdate` is `maxUnavailable: 0`/`maxSurge: 1`, the PDB sets `maxUnavailable: 1` with no `minAvailable`, resources show no `limits.cpu` and `128Mi`/`512Mi` memory, and `sessionAffinity: ClientIP` with `timeoutSeconds: 10800` — all matching the PR body exactly. Confirmed `defaults.DefaultMCPSessionIdleTimeout` really is `2 * time.Hour` (`internal/defaults/defaults.go:139`), so the 3h affinity window does sit above it as claimed. `go build ./...` and `go vet ./...` clean.

The PDB-vs-rollout-strategy distinction and the ClientIP-affinity-doesn't-survive-an-ingress caveat are both correct, standard Kubernetes behavior (Eviction API vs. Deployment-controller pod deletion; kube-proxy-only enforcement of Service affinity) — good that both are spelled out in the manifest comments themselves, not just the docs, since that's where an operator editing this file next will actually see them.

One non-blocking note, since this PR is already editing the exact section: Copilot's inline comment on `docs/internal/packaging-release.md:187` is right but pre-existing, not introduced here — `deployment.yaml` on `main` already uses `/readyz`/`/livez` probes and `runAsUser: 65532`, not the `/health` and `65534` that bullet still names (confirmed against `origin/main`, untouched by this diff). Worth a one-line fix in the same commit since you're already here, but not this PR's defect to own.

Everything else — the CPU-limit-vs-throttling trade-off, the per-pod SEMP-concurrency multiplier, the affinity-vs-ingress caveat, the deliberately-scoped-out follow-ups (no `deploy/` CI validation, no yaml.v3 guard on the 3h/2h coupling, `:latest` version skew) — is already disclosed in the PR body with the right caveats. No further findings.

solace-aross (dismissed):
Re-approving. My prior approval was auto-dismissed by this push — checked what actually changed rather than assuming it still applies.

`gh pr diff` against the current base confirms the same 8 files as before; only `docs/internal/packaging-release.md` has new content. That content is exactly the fix I flagged as non-blocking last round: probe paths and UID corrected from `/health`/`65534` to `/readyz`+`/livez`/`65532` in all 4 places they appeared (I'd only caught one), including two I hadn't spotted in the Docker stage description and security-considerations list. Verified against `deployment.yaml` again to be sure nothing drifted: still `/readyz`, `/livez`, `65532`. The commit message correctly identifies which `/health` references are genuinely still accurate (the Docker/bare-metal curl examples) and leaves those alone rather than over-correcting.

The other two commits are a merge from `main` (pulls in #351/#352, unrelated) — confirmed the PR's actual diff is unchanged by it, not a fresh review surface. `mergeable: MERGEABLE`, only `e2e-monitoring`/`e2e-action` still in progress, nothing red.

bczoma:
One documentation follow-up from review: `docs/authentication.md`'s TLS-terminated-upstream-ingress section (the topology recommended for OAuth) has no cross-reference to the session-affinity requirement this PR introduces. `service.yaml`'s comment, `docs/observability.md`, `docs/internal/packaging-release.md`, and `docs/user-guide.md`'s troubleshooting table all document that an Ingress/Gateway/mesh bypasses kube-proxy's `sessionAffinity: ClientIP`, but an operator who only reads `authentication.md` before deploying won't see that caveat until they hit it in production or find the troubleshooting table.

Could you add a one-line pointer in `docs/authentication.md`'s TLS-terminated-upstream section to `docs/observability.md`'s "Session affinity is required above one replica" section? Not a blocker for this PR, just worth closing the loop since it's the doc an OAuth deployer is most likely to read first.

solace-aross (approved):
Re-approve at 2a07c1c (prior approval auto-dismissed by this push, checked rather than assumed).

New commit closes bczoma's non-blocking follow-up: authentication.md now cross-references observability.md's "Session affinity is required above one replica" section, so an OAuth deployer sees the ClientIP-affinity caveat in the doc they read first, not just in the troubleshooting table. Also fixed deploy/kubernetes/README.md's manifest inventory, which DOC-8661's merge into this branch had left one file short of poddisruptionbudget.yaml.

Verified in a scratch worktree rather than trusting the diff: poddisruptionbudget.yaml is present and now named in README.md's 5-file list; observability.md's section title matches the cross-reference text exactly; packaging-release.md's probe-path/UID drift Copilot flagged two rounds ago stays fully resolved (65532 and /readyz+/livez throughout, the remaining /health mentions are the intentional Docker/bare-metal/backward-compat cases). CI green (36/36), docs-only diff, no new findings.

bczoma:
Follow-up to my earlier comment: I dug into the transport to check how bad the affinity gap actually is, and it's worth calling out more concretely than a doc cross-link.

The streamable-HTTP transport (`github.com/modelcontextprotocol/go-sdk`) sends every tool call as an independent POST — there's no persistent connection carrying client requests (the one long-lived connection is a receive-only SSE stream for server push). So without working session affinity, each call independently has ~1/replicas odds of landing on the session's home pod. At `replicas: 2` behind an ingress with no stickiness of its own, that's roughly a coin flip *per call*, not a rare edge case — a 10-call session has on the order of a 0.1% chance of every call landing correctly.

`docs/observability.md:706` already names the correct fix — routing keyed on the `Mcp-Session-Id` header (e.g. nginx-ingress `upstream-hash-by: "$http_mcp_session_id"`, or Gateway API `sessionPersistence`), not cookie-based stickiness, since an MCP client has no cookie jar. That's the right answer; it's just currently phrased as an aside about why `sessionAffinity: ClientIP` is an imperfect proxy, rather than a required deployment step.

Given how close to certain the failure is under the OAuth + ingress topology this project itself recommends, I'd suggest going a bit further than the doc-link I asked for earlier:

1. Move the `upstream-hash-by`-on-`Mcp-Session-Id` guidance into `docs/authentication.md`'s TLS-terminated-upstream-ingress section as a required step, with the actual annotation, not just a pointer to observability.md.
2. Consider shipping a commented-out ingress annotation example near `deploy/kubernetes/service.yaml`, since the other manifests here are already meant to be copy-paste starting points.

Not asking to block this PR on it, but I'd treat it as a near-term follow-up rather than a someday nice-to-have, given the actual odds involved.

PR #356, opened 2026-08-31
SOL-153827: log the IdP endpoint under a key redaction does not blank

solace-aross:
## What

`ExchangeError.LogAttrs` logged the IdP token endpoint under the key `token_endpoint`. `redactSecretAttr` in `cmd/server/main.go` replaces the value of any attribute whose lowercased key contains one of its `redactedKeys` substrings, and `token` is on that list, so the field reached operators as `[REDACTED]` in production.

This renames the key to `idp_endpoint` and routes the value through `config.SanitizeURLString`.

Jira: https://sol-jira.atlassian.net/browse/SOL-153827

## Why it matters

`logToolResult` folds these attrs onto the ERROR `tool invoked` line for every token-exchange failure, and the endpoint is the value that separates a misconfigured IdP endpoint from an unreachable one.

Worth knowing before you try to reproduce it: on transport failures the URL still appeared, incidentally, inside `detail`, because the underlying `*url.Error` text carries it. The classes that lost it outright are rejected, invalid-response, circuit-open and rate-limited. Reproducing with a dial error makes the bug look absent.

`Params.LogValue` in `types.go` had already hit this exact trap and worked around it, logging the same URL under `idp_endpoint` with a comment naming the collision. `errors.go` did not follow that convention. Rule 3 of the secure logging rules now states the constraint, so `/check-logs` can catch a third occurrence.

The `SanitizeURLString` call is defense in depth on top of the rename: now that the key no longer trips the redaction net, nothing downstream would strip userinfo if a credentialed URL ever reached the log line. `config.validateBrokerURL` already rejects one at config load, the same guarantee `sempv1`/`sempv2` `HTTPClient.LogValue` sanitize behind.

## Test

`TestExchangeError_LogAttrs_EndpointSurvivesRedaction` renders `LogAttrs()` through the production redaction filter and asserts `[REDACTED]` appears exactly once, on a deliberate `subject_token` control attr. The control is the anti-tautology guard: without it, the assertion would pass just as happily against a handler whose filter never fires. The exact count pins the key namespace, so a future field cannot silently reintroduce the bug.

`breaker_state` and `gate` are sentinel-gated and unreachable from a single error value, so the test varies the sentinel across three cases to render every key `LogAttrs` can emit.

Both halves revert-verified independently:

- rename reverted, sanitize kept: fails the `idp_endpoint` value and `[REDACTED]` count assertions in all three cases
- sanitize reverted, rename kept: fails the userinfo assertions in all three cases

`make check` green, total coverage 89.4% against the 85% floor.

## Deliberately out of scope

- `internal/config/config.go:248`. `BrokerOAuthConfig.LogValue` has the same collision on `idp_token_endpoint`, but nothing logs a `BrokerOAuthConfig` today and the key deliberately mirrors the YAML field for log-to-config mapping. Latent, not live, and the value is already sanitized, so the collision would blank it rather than leak it. Recorded as a known exception on the new Rule 3 text so it does not flag against the rule this PR adds.
- `internal/tokenexchange/types.go:167`. `Params.LogValue`'s unsanitized `TokenURL` was considered and declined under SOL-153000, and is already a recorded exception under Rule 2.
- The `*url.Error` text in `detail` is the one remaining path that can carry an IdP URL unsanitized. Pre-existing, and covered by `logToolResult`'s audited type gate.


solace-awaheed (approved):
Nothing flagged on my end, LGTM!

PR #355, opened 2026-08-31
SOL-152417: exporter fidelity spike passes — default OTel Prometheus exporter output matches the frozen metric names

solace-lootii:
## Summary

Story 45 (SOL-152417): a hard-time-boxed spike proving `go.opentelemetry.io/otel/exporters/prometheus` emits our intended frozen metric names without relying on its deprecated options. **Result: PASS.**

Fixes SOL-152417

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Dependency update

## Motivation and Context

Before we freeze a permanent public metric schema, we need evidence — not documentation-reading — that the OTel Prometheus exporter (`v0.66.0`+, pre-1.0) can actually emit the names ADR-008 commits to. This is the go/no-go on ADR-007's OTel-first metrics path (dual egress: Prometheus scrape + OTLP push), and it gates the outbound send of Story 41's pilot schema review. This story is called out as the first implementation story in the epic and cannot be cut — every other metrics story (14, 15, 16, 19, 46, 47) depends on its outcome.

## Changes Made

- Added `internal/observability/metrics/exporter_fidelity_test.go`: a durable test (not a throwaway `cmd/spike-*` binary), covering every AC in the ticket:
  - Registers the three representative instruments (`mcp.tool.invocation` counter, `mcp.tool.invocation.duration` histogram, a constant-1 info gauge) and diffs real scrape output against the intended frozen names.
  - **Finding: the exporter's *default* transformation already produces `mcp_tool_invocation_total`, `mcp_tool_invocation_duration_seconds`, and `mcp_build_info` — no deprecated option needed, and none is used anywhere in this file.**
  - Confirms `WithoutScopeInfo()` removes `otel_scope_*` from every series.
  - Confirms `target_info` is present and carries resource attributes (validates the Story 34 approach).
  - Evaluates all three supported approaches in order per ADR-007's fallback ladder: the default transformation, `WithTranslationStrategy` (both `NoTranslation` and `UnderscoreEscapingWithSuffixes`, explicitly — this is the non-deprecated replacement the deprecation notices point at), and adjusted instrument names/units.
[...]

PR #357, opened 2026-08-31
SOL-152091: Serve /metrics on a dedicated listener so Prometheus scrapes without an MCP network-policy change

solace-wmaryam:
## Summary

Adds a Prometheus-format `/metrics` endpoint on a dedicated listener (default `:9091`), gated by `OBS_METRICS_ENABLED` and off by default. This is Story 14 (SOL-152091); it owns the OpenTelemetry meter provider and `/metrics` handler that the downstream metrics stories register instruments against.

Fixes SOL-152091

## Type of Change

- [x] Dependency update

## Motivation and Context

Customer NOC operators need to scrape MCP-server metrics with Prometheus without changing the network policy that governs MCP traffic. Serving `/metrics` on a separate listener lets operators apply an independent network policy to scraping. This story also settles the instrument root and exposition options every downstream metrics story builds on: it owns the OTel meter provider, the `client_golang` registry, and the `/metrics` handler.

## Changes Made

- **`internal/observability/metrics/provider.go`** — `Provider`: one `client_golang` registry holding both the OTel Prometheus exporter (`WithoutScopeInfo()`) and the Go/process collectors, plus the meter provider all instruments register against. Exposes `MeterProvider()`, `Meter()`, `Handler()` (OpenMetrics enabled), `Shutdown()`.
- **Three self-observation instruments** — `mcp_build_info{version}`, `mcp_schema_version{metrics_schema,audit_schema}`, and the `mcp_metrics_scrape_total` counter (incremented per scrape via a handler wrapper).
- **`internal/config/observability.go` + `internal/defaults/defaults.go`** — new `observability.metrics_bind_address` field (default `:9091`, `${VAR}` substitution), defaulted when empty.
- **`cmd/server/main.go`** — starts the second listener when enabled, registered as a required `metrics_endpoint` readiness probe so a bind failure surfaces on `/readyz` as `metrics_endpoint: <err>`. The probe reflects live listener state via an `atomic.Pointer[error]`.
- **Tests** — golden-file schema test (plain-text scrape, `-update` to regenerate), scrape-counter increment test, and a guard test that fails if the deprecated `WithoutUnits()`/`WithoutCounterSuffixes()` options appear.
- **`CHANGELOG.md`** — `[Unreleased]` Added entry.
- **`go.mod`/`go.sum`** — adds `go.opentelemetry.io/otel` (v1.46.0), `.../sdk/metric`, `.../exporters/prometheus` (v0.68.0), `github.com/prometheus/client_golang` (v1.24.1); versions aligned with SOL-152417 (PR #355).

## Testing

**Test Environment:**
- Go version: per `go.mod` (1.25.0)
- OS: darwin (local dev)
- Broker version (if applicable): N/A — no broker-facing code

**Tests Performed:**
- [ ] Integration tests added/updated (if applicable) — N/A
- [ ] E2E tests added/updated (if applicable) — N/A (no broker-facing surface)
- [x] Tested locally with `go test -race ./...` (via `make test-cover`)

**Test Coverage:**
- `make test-cover` — clean, 88.7% total coverage (85% gate).
- Golden file pins the exact published names — `mcp_build_info`, `mcp_schema_version`, `mcp_metrics_scrape_total` — over a deterministic plain-text scrape with the runtime collectors filtered out.
- Verified inert when `OBS_METRICS_ENABLED` is off (no listener, no instruments).
- `/check-logs` — no violations in the added logging.

## Breaking Changes

None. The capability is off by default; the new config field has a default.


THIRD_PARTY_BUILD_TEST.md
```diff
@@ -56,18 +56,97 @@ the conclusion, but the claim is scoped rather than asserted.
...
+component that must appear in its `in_use` set — and `in_use` is built only
+from checks 1–4 (`test/e2e-*` go.mod closures, npm lockfiles under `test/`,
+GitHub Actions, container images), which by design never includes the root
+module. Using the real table markup here was tried first and confirmed to
+fail closed: every row below was flagged `is listed but nothing in the
+repository uses it any more`, for the packages this very entry exists to
+document. A bullet list keeps this section honest and human-readable without
+tripping a check that structurally cannot see the root module.
+
+**Not a release-inventory change.** `THIRD_PARTY_LICENSES.md` (the shipped-binary
```
solace-aross: This section's own contract: "that changes only once Story 14 makes these real production dependencies, at which point this section is removed and the rows move to THIRD_PARTY_LICENSES.md instead." This PR is Story 14 (provider.go links otel/client_golang into cmd/server) — the migration this sentence promises hasn't happened yet, which is the root cause of the five failing CI checks. See the top-level comment for the fix sequence.

cmd/server/main.go
```diff
@@ -537,6 +540,56 @@ func startServer(srv *http.Server, tlsCertFile, tlsKeyFile string) <-chan error
...
+// registers a "metrics_endpoint" readiness probe. Returns the provider for
+// instrument registration, or nil if the engine failed.
+//
+// No explicit shutdown: /metrics is pull-based, so nothing is buffered to lose,
+// and process exit reaps the listener. TODO(SOL-152449): once the shared
+// shutdown-hook registry lands, register provider.Shutdown against it.
+func startMetricsEndpoint(cfg *config.ServerConfig, readiness *health.ReadinessState) *metrics.Provider {
+	provider, err := metrics.New(version.Version())
+	if err != nil {
+		slog.Error("metrics endpoint disabled: provider build failed", slog.String("error", err.Error()))
```
solace-aross: This branch runs when metrics are explicitly enabled but the provider failed to build — an actual startup failure, not a deliberate disable. "metrics endpoint disabled" reads as intentional and could send an operator debugging a missing /metrics down the wrong path. Something like "metrics endpoint unavailable: provider build failed" would say what actually happened.

cmd/server/main.go
```diff
@@ -537,6 +540,56 @@ func startServer(srv *http.Server, tlsCertFile, tlsKeyFile string) <-chan error
...
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", provider.Handler())
+	srv := newHTTPServer(cfg.Observability.MetricsBindAddress, mux)
+
+	// Open the port and wait, so a failure (e.g. port in use) shows on /readyz.
+	var lc net.ListenConfig
+	ln, err := lc.Listen(context.Background(), "tcp", srv.Addr)
+	if err != nil {
+		slog.Error("metrics endpoint failed to bind",
```
solace-aross, line 568: Verified this path by injection rather than reading: wrote a probe test that occupies the bind address first, calls startMetricsEndpoint, and checks readiness.Evaluate() — confirmed the bind failure correctly surfaces as unready with the metrics_endpoint reason, and a successful bind correctly reports ready. The implementation is right; there's just no test in this PR proving it, on a path that gates real traffic-routing decisions. Worth a test in cmd/server covering both the bind-failure and successful-bind cases, matching the coverage this file already has for the main HTTP listener.

internal/observability/metrics/provider.go
```diff
@@ -0,0 +1,158 @@
...
+func (p *Provider) Handler() http.Handler {
+	base := promhttp.HandlerFor(p.registry, promhttp.HandlerOpts{EnableOpenMetrics: true})
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Count before serving so this scrape's own render includes the increment.
+		p.scrapeCounter.Add(r.Context(), 1)
+		base.ServeHTTP(w, r)
+	})
+}
+
+// Shutdown flushes and stops the meter provider. Wired as a shutdown hook
```
solace-aross: This doc comment states Shutdown() "is wired" as a shutdown hook, but it isn't — cmd/server's own comment on startMetricsEndpoint says explicitly "No explicit shutdown... TODO(SOL-152449): once the shared shutdown-hook registry lands, register provider.Shutdown against it." Low functional impact today (the Prometheus exporter is pull-based, so there's nothing buffered to lose on exit), but the comment describes an intended future state as already true. Suggest "Flushes and stops the meter provider. Not yet wired into the shutdown path — see the TODO in cmd/server."

solace-aross (changes requested):
CI is red on this PR, and it's caused by this diff, not baseline noise — checked against the last 15 runs of `CI - Pull Request`, including the current main head and sibling PR #355; this branch is the only failure. Five checks fail for one reason: `client_golang`, `otel`, `otel/sdk/metric`, `protobuf`, and 15 more transitive modules are now compiled into `cmd/server` but aren't in `THIRD_PARTY_LICENSES.md`, and four modules ship a NOTICE Apache-2.0 §4(d) attribution this repo doesn't propagate.

The PR this one builds on (#355) already named the fix: `THIRD_PARTY_BUILD_TEST.md` §"Root Module (test-only)" says explicitly *"that changes only once Story 14 makes these real production dependencies, at which point this section is removed and the rows move to `THIRD_PARTY_LICENSES.md` instead."* This PR is Story 14, and that migration hasn't happened — `THIRD_PARTY_LICENSES.md` only picked up the unrelated `testify`/`x/sys`/`x/text` bump.

`make refresh-third-party-inventory` gets most of the way but won't finish the job:
- It will resolve 18 of 19 missing rows, then exit 1 on `github.com/prometheus/client_golang` — `go-licenses` can't resolve a single unambiguous license for it (it vendors code under its own `internal/` with a distinct license), so that row needs adding by hand from its own `LICENSE` file. Expect a partially-rewritten tree when this happens; that's the script working as designed, not a break.
- `NOTICE` is out of the script's scope entirely. Four modules need an Apache-2.0 §4(d) attribution added by hand: `client_golang`, `client_model`, `common`, `procfs`.
- The `THIRD_PARTY_BUILD_TEST.md` "Root Module (test-only)" section is hand-maintained (deliberately a bullet list so the checker can't see it) and needs deleting per its own stated contract, now that these are real production deps.

Three non-blocking notes below, verified rather than assumed.

solace-aross (dismissed):
Blocking issue from my prior review is fixed. Verified locally rather than trusting the green checkmarks:

- `licenses-check.sh` now exits 0 in a fresh worktree at this head — the `client_golang` ambiguity (it vendors a BSD-3-Clause `golang/gddo` copy under its own `internal/`, distinct from its Apache-2.0 parent) is handled correctly: added to `REQUIRED_COMPONENTS` with the license read from the vendored file itself, matching the pattern the script already uses for `go-jose/go-jose/v4/json`.
- `NOTICE` now carries all four attributions I flagged missing (`client_golang`, `client_model`, `common`, `procfs`).
- `Third-party licenses current`, `Third-party inventory refresh scripts self-test (network)`, `(offline) (ubuntu-latest)`, and `SBOM check script self-test` are all green.

**One check is still red — `Third-party inventory refresh scripts self-test (offline) (macos-latest)` — and it's not this PR's bug.** Reproduced locally: the self-test's own baseline case ("an already-matching file is a clean no-op") fails on this machine too, dropping the `procfs` row from an untouched, already-correct file. Root cause, confirmed directly rather than assumed:

```
$ go list -deps ./cmd/server | grep -c prometheus/procfs        # darwin
0
$ GOOS=linux go list -deps ./cmd/server | grep -c prometheus/procfs
3
```

`client_golang`'s process collector only imports `procfs` on Linux — it's a `/proc` reader, meaningless on macOS. The self-test (and, presumably, `licenses-check.sh` if ever run on a Mac) calls `go list -deps` natively rather than pinning `GOOS=linux` for a check about a Linux-shipped binary, so it disagrees with itself depending on which runner OS executes it. This is a pre-existing gap in the tooling that any future Linux-only transitive dependency would trigger — not something in scope for this PR to fix, and not a required check (checked the branch ruleset: only `Third-party licenses current` is required, and that one's green). Worth its own ticket; happy to file it if nobody already has.

Also checked: my earlier notes on the misleading log message, the untested-but-correct readiness path, and the `Shutdown()` doc comment weren't touched in this fix commit (it's scoped tight to the inventory, which is the right call) — they're still open as non-blocking. `e2e-monitoring` is still running as I write this; nothing in this PR's actual diff touches that surface.

Approving.

solace-wmaryam:
@solace-aross thanks for the detailed review — all points are addressed across two commits.

**Inventory (the blocking issue)** — c111766:
- Added the missing rows to `THIRD_PARTY_LICENSES.md` (each license read from the module's own LICENSE file, every URL checked).
- Hand-added `client_golang` plus its vendored BSD-3-Clause `internal/.../gddo` row that `go-licenses` can't resolve, and registered `gddo` in `licenses-check.sh`'s `REQUIRED_COMPONENTS` so the row can't be silently dropped.
- Propagated the four §4(d) NOTICE attributions (`client_golang`, `client_model`, `common`, `procfs`).
- Removed the "Root Module (test-only)" section from `THIRD_PARTY_BUILD_TEST.md` per its own stated contract, now that these are real production deps.

`Third-party licenses current` is green.

**Non-blocking notes** — f23b917:
- Log line reworded "disabled" → "unavailable: provider build failed".
- `Shutdown()` doc corrected — not yet wired (TODO SOL-152449).
- Package comment rewritten (no longer a "skeleton").
- `startMetricsEndpoint` now shuts the provider down and returns nil on a bind failure, matching its doc.
- Added a `cmd/server` test covering both the bind-failure (unready) and successful-bind (ready) paths via readiness.

**One item left for a separate Jira:** the `macos-latest` inventory self-test fails on `procfs`. It is Linux-only but required for the shipped Linux binary, so the macOS run sees the row as extra. That is a gap in the shared tooling (the first GOOS-specific production dep), not something this PR should work around — I'll file a Jira to make the refresh/check pin `GOOS=linux` (the release target).

solace-wmaryam:
@solace-aross filed the macOS/procfs tooling gap as SOL-153873 (under the MCP 7 epic), so no need to file a duplicate. It captures the root cause and your `GOOS=linux` direction — pin the release target in `licenses-check.sh`, the refresh scripts, and the offline self-test. Thanks for the thorough review and the approval.

amitmorade (changes requested):
@solace-wmaryam  — nice work on this.

Because metrics is optional and off by default, someone can ship MCP today on port `9091`. Nothing in this change reserves that port, and they have no reason to think about a scrape listener they have not turned on. A month later they set `OBS_METRICS_ENABLED=true` and leave YAML alone. Metrics still defaults to `:9091`. Config load does not compare that to the MCP bind, so startup looks fine until listen fails with “address already in use.”

What makes that more than a noisy log: this PR registers the metrics bind as a required `metrics_endpoint` readiness probe. So opting in can mark the whole process unready, or metrics can win the race and MCP itself fails to bind. Defaults (`9090` vs `9091`) do not collide; this is only “MCP has been on 9091 for a while, then we turn metrics on.”

Could we either catch overlapping binds in `validate()` here, or keep bind-time failure and file a follow-up (with a link) so it isn’t left implicit? I’m happy with either; I just don’t want to merge until we’ve said which one we mean.

solace-aross (commented):
Concur with the port-collision finding, and the severity is worse than either of us first framed it.

Reproduced both race directions at PR head (`300b800`) in a scratch worktree, running the real `main()` ordering (`startServer` then `startMetricsEndpoint`) 300 times per config:

| GOMAXPROCS | MCP binds first | MCP fails to bind |
|---|---|---|
| default (10-core) | 190 | 110 (37%) |
| 2 | 274 | 26 (9%) |
| 1 | 1 | 299 (99.7%) |

`go.mod` pins `go 1.25.0`, so container-aware GOMAXPROCS is active — anyone running this under `limits.cpu: 1` gets the MCP-never-binds outcome on essentially every boot. This repo's own shipped `deployment.yaml` doesn't set a CPU limit, but that's the operator's line to change, and 9-37% is already a coin flip on more common shapes.

Neither direction is benign under this repo's own manifests. `deployment.yaml`'s `startupProbe` hits `/readyz` every 2s with `failureThreshold: 30`, and `readiness.go` holds `/readyz` unready forever once a registered probe returns an error. A metrics bind failure registers exactly that probe. So whichever listener loses the race, the pod CrashLoopBackOffs at ~60s — this isn't "surfaces on `/readyz` instead of failing silently," it's a startup-probe kill either way, and at two replicas (SOL-152089) that's a rolling outage, not a log line.

"Who runs MCP on 9091" is also already answered inside this repo: `test/health/test-health.sh` sets `SECONDARY_MCP_PORT=9091`, and `.github/workflows/dev-build.yaml` documents the lab dev instance picking up builds on port 9091. This isn't a hypothetical operator misconfiguration, it's the team's own lab.

Given that, I don't think the follow-up-ticket option should stay on the table — a follow-up ships a nondeterministic crash-loop on a port this team already uses. Asking for a `validate()`-time guard in this PR:

- `validate()` already does cross-field collision checks (broker alias uniqueness) and validates `listen_address` form specifically "so an unbindable host fails at startup rather than deep inside `ListenAndServe`" — same principle, just not extended to `metrics_bind_address`.
- One detail that'll make a naive fix half-work: `BindAddress()` is `net.JoinHostPort(ListenAddress, Port)`, so `listen_address: ""` renders `:9091` while `listen_address: "0.0.0.0"` renders `0.0.0.0:9091` — those collide in practice but won't string-match. The guard needs to split host:port and treat `""`, `0.0.0.0`, and `::` as wildcards overlapping any host on the same port.

Non-blocking: the guard fixes the self-collision, but `startServer` returning before its goroutine's `net.Listen` runs leaves the same nondeterminism for any *external* port conflict (a sidecar, host networking). Having it confirm its own bind before returning would remove the race class entirely rather than just this one instance of it. Not worth blocking on here.

I'll approve once the guard and a test for it land.

solace-aross (dismissed):
Guard and test both landed as asked. Verified by injection rather than reading, in a scratch worktree at `f1a781b`: disabling `validateMetricsBindAddress` makes exactly the three collision cases in `TestValidate_MetricsBindAddressCollision` go red (wildcard-vs-loopback, explicit wildcard, localhost-vs-127.0.0.1); the two no-collision cases and the metrics-disabled case stay green either way. The test earns its green.

Also checked the defaulting order, since that's where a guard like this usually breaks: `ListenAddress` and `MetricsBindAddress` are set in `applyDefaults`, `MetricsEnabled` in `applyEnvOverrides`, both ahead of `validate()`. So the guard sees final values, including the oauth mode's empty (wildcard) `ListenAddress`.

The one red check, `Third-party inventory refresh scripts self-test (offline) (macos-latest)`, is the pre-existing GOOS gap from my last pass on this PR, tracked as SOL-153873. Not this diff's problem.

Approving. This closes the port-collision finding, wildcard and loopback edge cases included.

amitmorade (dismissed):
LGTM, thank you for fixing that

amitmorade (approved):
LGTM

PR #358, opened 2026-09-01
SOL-153834: make ingress session routing a required, documented step

solace-awaheed:
The manifests default to replicas: 2, and docs/authentication.md tells operators to front the Service with a TLS-terminating ingress. That ingress load-balances to pod IPs, bypassing the Service's sessionAffinity: ClientIP, and MCP sessions live in one pod's memory — so each call after initialize has roughly a 1/replicas chance of reaching the pod that owns the session. A 10-call session survives with probability (1/2)^10. Near-certain failure in the topology this project recommends for production OAuth.

The guidance existed only as an aside in observability.md, phrased as an explanation rather than a requirement. Promote it to a required step in authentication.md and ship deploy/kubernetes/ingress.yaml.example.

The working hash key is $remote_addr, not the session ID. The intuitive upstream-hash-by: "$http_mcp_session_id" is broken: the initialize POST carries no such header — the server mints the ID onto the response — so it hashes on a constant empty key while the ID it returns hashes independently, stranding half of all sessions at two replicas. The session ID is the exact routing key for every request except the one that creates the session. Also documented as not working: cookie affinity, whose behaviour depends on the client's HTTP stack, and these annotations under a non-nginx ingressClassName, which are ignored silently.

The example routes the whole /.well-known prefix, not just oauth-protected-resource: buildRootHandler's catch-all exists so the discovery paths clients probe during (re-)authentication return JSON rather than the plain-text 404 that breaks them, and routing one path would have reintroduced exactly that. Its SSE proxy timeouts are 7200s to match the server's 2h session idle timeout, so nginx cannot sever a stream the server still holds.

Reconcile the five other places carrying this guidance so they cannot drift: observability.md and user-guide.md now point at the new section rather than restating it, service.yaml, both READMEs and packaging-release.md follow, and the SOL-152089 CHANGELOG entry no longer prescribes the session-ID hash. The user-guide troubleshooting row leads with an ingress check, since the Service still reports sessionAffinity: ClientIP in precisely the broken case.

Docs and example manifests only; no code change. The example is schema-valid under kubeconform -strict, and its .example suffix keeps the directory-wide kubectl apply from picking it up.

## Summary

Makes ingress session routing a required, documented deployment step and ships `deploy/kubernetes/ingress.yaml.example`. Since the manifests default to `replicas: 2`, an operator following the TLS-terminated-upstream pattern currently gets a deployment that 404s on essentially the first call.

Tracked under [SOL-153834](https://sol-jira.atlassian.net/browse/SOL-153834). Builds on SOL-152089 (#354).

## Type of Change

- [x] Documentation update

## Motivation and Context

An Ingress load-balances to pod IPs, so it bypasses the Service's `sessionAffinity: ClientIP` (kube-proxy enforces that on the ClusterIP path only). MCP sessions live in one pod's process memory, so each call after initialize has roughly a `1/replicas` chance of reaching the pod that owns the session — a 10-call session survives with probability `(1/2)^10` ≈ 0.1%.

That is near-certain failure, and it happens in exactly the topology `docs/authentication.md` recommends for production OAuth. The guidance existed only as an aside in `docs/observability.md`, phrased as an explanation rather than a requirement.

## Changes Made

- **New** `deploy/kubernetes/ingress.yaml.example` — copy-and-edit Ingress with the required annotation. `.example` so the directory-wide `kubectl apply` skips it.
- **New** `docs/authentication.md` § "Session Routing at the Ingress (Required Above One Replica)" — the authoritative location, with the annotation, the quantified failure rate, and what does *not* work.
- Reconciled the five other places carrying this guidance so they cannot drift: `docs/observability.md`, `docs/user-guide.md`, `deploy/kubernetes/service.yaml`, `deploy/kubernetes/README.md`, `docs/internal/packaging-release.md`.
- `CHANGELOG.md` — new `[Unreleased]` entry, and corrected one clause in the SOL-152089 entry that prescribed the session-ID hash.

**⚠️ Deviation from the ticket's acceptance criteria.** The AC specified `upstream-hash-by: "$http_mcp_session_id"`. That does not work, so this PR ships `$remote_addr` instead and documents the session-ID hash as a named wrong turn.

The initialize POST carries no `Mcp-Session-Id` header — the server mints the ID onto the *response* (verified in go-sdk v1.7.0 `mcp/streamable.go:633-653`). So it hashes on a constant empty key while the ID it returns hashes independently, stranding half of all sessions at two replicas. The session ID is the exact routing key for every request *except* the one that creates the session, and that exception disqualifies it.

**The same AC's Gateway API `sessionPersistence` requirement is also not viable.** Per GEP-1619 it is gateway-assigned persistence — the gateway issues a token the client echoes back — so it fails for the same reason cookies do, and it is still experimental-channel with thin support. There is no portable Gateway API equivalent of `upstream-hash-by`, so `docs/authentication.md` § "Gateway API and service meshes" points at the controller's own source-address hash instead (Istio `consistentHash.useSourceIp`, Envoy Gateway `consistentHash.type: SourceIP`).

Also documented as not working: cookie affinity (behaviour depends on the client's HTTP stack — the Go SDK keeps no cookie jar, httpx-based clients do), and these annotations under a non-nginx `ingressClassName`, which are ignored silently.

## Testing

Documentation and example manifests only — no Go code, so no unit tests apply.

- `ingress.yaml.example` validates against the real Kubernetes schema: `kubeconform -strict` → 1 resource, valid.
- Verified the `.example` suffix excludes it from `kubectl apply -f deploy/kubernetes/`.
- Verified every new anchor link resolves to its heading slug.
- Repo-wide grep confirms no remaining "configure stickiness on `Mcp-Session-Id`" advice.
- `make check` is unchanged by this PR — the 17 pre-existing staticcheck `SA5011` findings in Go test files reproduce identically on a clean tree, and this diff contains zero `.go` files.

**Not verified:** the annotation's runtime behaviour against a live ingress-nginx. A kind cluster with 2 replicas running a 10-call session would confirm it end to end.

## Checklist

- [x] Self-review completed
- [x] No credentials in code
- [x] docs/ updated
- [x] CHANGELOG.md updated
- [x] Examples updated
- [x] Branch up to date with main, no conflicts

## Additional Notes

Items raised in review and left out of scope:

1. **A hard `mode: static` gate on the manifest.** Per review, `ingress.yaml.example` now carries a one-line cross-reference to the shared-token warning, for anyone who opens that file without reading the README first. A real gate is still out of scope: this is the first manifest the project ships that makes the server publicly reachable, and the existing guidance is phrased only as "keep the Service `ClusterIP`" — which an operator applying an Ingress satisfies while still publishing a shared-token, broker-admin-backed surface. Worth a follow-up ticket.
2. **Rate-limit / source-range annotations** (`limit-rps`, `whitelist-source-range`) on the example.

Out of scope per the ticket: `StreamableHTTPOptions.Stateless` or a shared session store, either of which would remove the need for affinity entirely.

## Related Issues/PRs

- Builds on #354 (SOL-152089)
- Raised in review of #354 by @bczoma

---

**For Reviewers:**

**Focus Areas**: The `$remote_addr` vs `$http_mcp_session_id` deviation is the main thing to check — the reasoning is in "Changes Made" above and in `docs/authentication.md` § "What does not work". Also worth a look: the example routes the whole `/.well-known` prefix rather than just `oauth-protected-resource`, because `buildRootHandler`'s catch-all (`cmd/server/main.go:1103`) exists so the paths clients probe during (re-)authentication return JSON instead of the plain-text 404 that breaks them.


bczoma:
Ran this through review (including adversarial challenges on the more severe-looking findings — most turned out to be already mitigated elsewhere in this same PR, so they're not blockers). A few small proposals worth considering, none blocking:

1. **`docs/internal/packaging-release.md`** — the `service.yaml` bullet still points to `docs/observability.md` for the ingress/mesh-bypass caveat, but `observability.md` itself now defers to `docs/authentication.md` as the authoritative location. Proposal: point this bullet directly at `../authentication.md#session-routing-at-the-ingress-required-above-one-replica`, matching the `ingress.yaml.example` bullet right below it.

2. **`deploy/kubernetes/ingress.yaml.example`** — the existing `mode: static`/OAuth warning already sits immediately upstream in `configmap.yaml`'s comment and the README's "Client authentication" section (right before "TLS and ingress"), so this isn't a real gap — but a one-line cross-reference in this file's own top comment would help anyone who jumps straight here without reading the rest of the README first.

3. **`CHANGELOG.md`** — the SOL-153834 entry names only `/.well-known/oauth-protected-resource`, while the shipped example routes the whole `/.well-known` prefix (correctly, per the manifest's own comment). Proposal: reword to "routes the whole `/.well-known` prefix" for accuracy, even though the manifest's inline comment already covers anyone who'd act on this.

4. **PR body's "⚠️ Deviation" section** — covers the `$remote_addr` vs. session-ID substitution but not that AC1's Gateway API `sessionPersistence` requirement was also found non-viable (verified correct against GEP-1619 — it's cookie/token-based, same failure mode as cookies). The gap is already well-documented in the CHANGELOG and `docs/authentication.md` itself, so this is process/disclosure only — proposal: one added sentence there for anyone skimming just that section.

One item I initially flagged (`docs/user-guide.md`'s jsonpath command) turned out to be correct as written when tested against real `kubectl` — no change needed there, despite Copilot's earlier comment.

Nice catch overall on the `$http_mcp_session_id` deviation — good that this got verified against the go-sdk source rather than shipped on the ticket's assumption.

solace-aross (approved):
Verified the core technical claim independently: checked go-sdk v1.7.0 mcp/streamable.go directly. The initialize POST carries no Mcp-Session-Id header; the server mints the ID onto the response only, and reads it from the request header on every later call. That confirms $remote_addr over $http_mcp_session_id for the upstream-hash-by annotation, and the deviation from the ticket's AC.

ingress.yaml.example reads as a structurally valid Ingress (apiVersion/kind/metadata/spec.rules/tls) — no kubeconform or live cluster available in this environment to reproduce the exact command from the PR description, but nothing in the shape is off.

bczoma's four review proposals: the CHANGELOG wording, the packaging-release.md cross-reference, and the ingress.yaml.example mode: static note are already in 05871bd. The fourth (PR body missing the Gateway API sessionPersistence deviation) is description-only, already called out as non-blocking.

Docs and example manifest only, no code path affected. Approving.

bczoma (approved):
Approving. Full review posted above (multi-lens + adversarial challenge on the more severe-looking findings) — no showstopper survived scrutiny. CI is green across the board and mergeStateStatus is CLEAN.

The follow-up proposals in my prior comment (packaging-release.md cross-reference, CHANGELOG wording, PR-body Deviation-section addendum) are all minor and non-blocking — fine to land as-is or fold into a quick follow-up.

PR #359, opened 2026-09-01
SOL-152449: own the shutdown-flush budget in one place

solace-awaheed:
Add a shutdown-hook registry (internal/observability/hooks) so Stories 14, 25, and 46 can each register a bounded OTel provider flush without editing cmd/server's shutdown path three separate times.

Hooks run concurrently and are bounded by a single new DefaultShutdownHookTimeoutSeconds budget (3s, carved from the 5s of previously-undocumented K8s grace-period slack) after the HTTP server's own graceful shutdown has settled - never while requests may still be in-flight, and never on a second-signal forced close. A panicking or failing hook is contained (via safego.Go) and logged, and never affects the process exit status. With no hooks registered the registry is a true no-op.

Updates the K8s terminationGracePeriodSeconds accounting in deployment.yaml and its README to include the new hook-budget line item.

## Summary

Adds a shutdown-hook registry (`internal/observability/hooks`) and a decided split of the shutdown timeout, so Stories 14, 25, and 46 each have a seam to register a bounded OTel provider flush without editing `cmd/server`'s shutdown path three separate times.

Fixes SOL-152449

## Type of Change

## Motivation and Context

Three upcoming stories (meter provider, tracer provider, OTLP metrics reader) each need to flush an OTel provider on SIGTERM within a bounded timeout. Without a shared registry, three inline edits to the same shutdown function produce either three sequential full-budget waits (blowing `terminationGracePeriodSeconds`) or three uncoordinated fractions nobody reconciles. This is also the last comfortable moment to add the seam — `drainAndShutdown` is the same function SOL-151437's force-close path threads through, so after three provider stories touch it, adding this becomes a merge conflict instead of a story.

## Changes Made

- **`internal/observability/hooks/hooks.go`** — `Registry` with `Register(name, fn)` and `RunAll(ctx)`. Hooks run concurrently via `safego.Go` (panic-safe); `RunAll` returns within `ctx`'s deadline regardless of hook count, and one stuck/failing/panicking hook never blocks or affects the others. Nil/empty registry is a true no-op.
- **`internal/defaults/defaults.go`** — new `DefaultShutdownHookTimeoutSeconds = 3`, documented next to the existing shutdown-timeout/drain-delay constants.
- **`cmd/server/main.go`** — threads `*hooks.Registry` through `drainAndShutdown`/`gracefulShutdown` (the `shutdowner` interface itself is untouched). Hooks run only after the HTTP server's own shutdown has settled — never while requests may still be in-flight, and never on a second-signal forced close. `main()` builds one empty registry, inert until a later story registers against it.
- **`deploy/kubernetes/deployment.yaml`, `deploy/kubernetes/README.md`** — updated the `terminationGracePeriodSeconds` accounting comment to include the new hook budget (was 30s drain + 5s buffer; now 30s drain + 3s hooks + 2s buffer, still 45s total).
- **Tests** — new `hooks_test.go` (bounded total, no starvation, error/panic doesn't propagate, no-op when empty); two new `cmd/server/shutdown_test.go` tests (hooks run after a normal shutdown; hooks skipped on forced close). The 6 existing call sites needed a mechanical `nil` appended for the new parameter — no assertions changed.

## Testing

**Test Environment:**
- Go version: 1.25.0 (per `go.mod`)
- OS: darwin (local dev)
- Broker version (if applicable): N/A — no broker-facing code

**Tests Performed:**
- [ ] Integration tests added/updated — N/A
- [ ] E2E tests added/updated — N/A

**Test Coverage:**
- `make test-cover` — 89.5% total (85% gate)
- All 8 pre-existing `shutdown_test.go` tests still pass with identical assertions
- `/check-logs` — 1 MEDIUM (generic `err.Error()` logging on a caller-supplied hook function; not blocking, no hook is registered yet so nothing flows through it today)

## Breaking Changes

None. The registry is empty and inert until a later story calls `Register`.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code
- [x] Followed secure logging rules (`/check-logs`: 1 MEDIUM, non-blocking — see Testing)
- [ ] Credential-carrying types implement `slog.LogValuer` — N/A, none added

### Testing
- [x] Edge cases covered by tests (bounded total, starvation, panic, error, empty/nil)
- [x] Error paths tested

### Documentation
- [ ] README.md updated — N/A, no user-facing change
- [x] docs/ updated — `deploy/kubernetes/README.md` grace-period accounting
- [ ] CHANGELOG.md updated — N/A, no observable behavior change (registry is inert)
- [ ] Examples updated — N/A

### Git & CI
- [x] No merge conflicts
- [ ] CI checks passing — pending first run

## Additional Notes

`make lint` locally reports 65 pre-existing staticcheck (SA5011) findings, none in files this PR touches — confirmed identical on a clean `main` via `git stash`. CI is the source of truth.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -45,6 +45,7 @@ import (
 	"github.com/SolaceProducts/solace-broker-mcp/internal/oauth/cache"
 	"github.com/SolaceProducts/solace-broker-mcp/internal/observability/correlation"
 	"github.com/SolaceProducts/solace-broker-mcp/internal/observability/health"
+	"github.com/SolaceProducts/solace-broker-mcp/internal/observability/hooks"
 	"github.com/SolaceProducts/solace-broker-mcp/internal/semp"
 	"github.com/SolaceProducts/solace-broker-mcp/internal/semp/sempv2"
 	"github.com/SolaceProducts/solace-broker-mcp/internal/semp/sempv2/specs"
@@ -568,7 +569,7 @@ func startServer(srv *http.Server, tlsCertFile, tlsKeyFile string) <-chan error 
 // skipping the graceful wait. Readiness has already flipped to 503 above, so a
 // forced exit still leaves /readyz correct. A nil forceSig (the startup-error
 // path, which never drains) simply never fires that select case.
-func drainAndShutdown(srv shutdowner, readiness *health.ReadinessState, drainDelay, shutdownTimeout time.Duration, forceSig <-chan os.Signal) error {
+func drainAndShutdown(srv shutdowner, readiness *health.ReadinessState, drainDelay, shutdownTimeout time.Duration, forceSig <-chan os.Signal, shutdownHooks *hooks.Registry) error {
 	readiness.SetShuttingDown()
 	slog.Info("draining before shutdown",
 		slog.String("reason", "signal"),
@@ -584,7 +585,7 @@ func drainAndShutdown(srv shutdowner, readiness *health.ReadinessState, drainDel
 			return nil
 		}
 	}
-	return gracefulShutdown(srv, shutdownTimeout, forceSig)
+	return gracefulShutdown(srv, shutdownTimeout, forceSig, shutdownHooks)
 }
 
 // forceClose is the shared "second signal" reaction (SOL-151437): a single WARN
@@ -646,7 +647,12 @@ type shutdowner interface {
 // returned via the force path — no goroutine leak. A nil forceSig (the
 // startup-error path) never fires that case, so that path keeps its exact
 // prior behavior, including the Shutdown-timed-out forced-close fallback.
-func gracefulShutdown(srv shutdowner, timeout time.Duration, forceSig <-chan os.Signal) error {
+//
+// shutdownHooks runs (SOL-152449) once srv.Shutdown has settled, success or
+// timed-out-then-forced-close alike — never while requests may still be
+// in-flight. A second signal returns before that point, so it skips hooks
+// entirely: "stop now" outranks a telemetry flush.
+func gracefulShutdown(srv shutdowner, timeout time.Duration, forceSig <-chan os.Signal, shutdownHooks *hooks.Registry) error {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
@@ -661,6 +667,9 @@ func gracefulShutdown(srv shutdowner, timeout time.Duration, forceSig <-chan os.
 				slog.Error("forced close failed", slog.String("error", closeErr.Error()))
 			}
 		}
+		hookCtx, hookCancel := context.WithTimeout(context.Background(), time.Duration(defaults.DefaultShutdownHookTimeoutSeconds)*time.Second)
+		defer hookCancel()
+		shutdownHooks.RunAll(hookCtx)
 		return err
 	case sig := <-forceSig:
 		forceClose(srv, sig)
@@ -1071,6 +1080,10 @@ func main() {
 	readiness := health.NewReadinessState()
 	mux := buildMux(readiness)
 
+	// shutdownHooks is empty until a later story (SOL-152091, SOL-152420,
+	// SOL-152418) registers an OTel provider flush against it.
+	shutdownHooks := hooks.NewRegistry()
+
 	// Create MCP handler
 	mcpHandler := mcp.NewStreamableHTTPHandler(func(req *http.Request) *mcp.Server {
 		return server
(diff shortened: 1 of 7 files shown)


solace-aross (dismissed):
Verified by injection in a throwaway worktree, not by reading.

**Second-signal skip is real, and correctly bounded.** Injected a defect that ran hooks on the forced-close path too — `TestGracefulShutdown_SecondSignalSkipsHooks` caught it immediately. The design is sound.

**Ran an untested scenario the shipped suite does not cover:** every `hooks_test.go` case uses a hook that itself watches `ctx.Done()`. I added a hook that ignores `ctx` entirely and sleeps 2s. `RunAll` still returned in ~53ms against a 50ms budget — the abandon-and-move-on guarantee in the doc comment is real, just unverified by the existing tests. Worth a test with a non-cooperating hook, since that is the one case a future edit to `RunAll` could break silently.

Two more notes, from an independent Opus pass I ran before approving (this touches a security-logging rule and a documented behavior claim, so I wanted a second read before either signing off or contesting):

1. **`hooks.go`'s `slog.String("error", err.Error())` is inert today** (registry is empty, the check-logs MEDIUM classification is correctly non-blocking per this repo's own rubric) **but the audit boundary is in the wrong place.** `Register(name, fn func(context.Context) error)` hands the registry an opaque error it cannot audit. The three stories about to register against this (SOL-152091/152420/152418) will each pass through an OTel SDK/collector error here — exactly the "external system" class Rule 5 exists for — and none of their diffs will show this line. Cheaper to close now than to migrate three call sites later: have the hook itself own logging its audited detail (matching `brokerstatus.logHardwareStepFailure`'s pattern of logging the unwrapped error type, not the message), and say so in `Register`'s doc comment.
2. `gracefulShutdown`'s comment ("a second signal... skips hooks entirely") overclaims: a second signal arriving *during* the ≤3s hook window is read into the reserved buffer slot but not acted on until process exit. Bounded and harmless in practice, but the comment should say so rather than "entirely."

Neither blocks. Correct k8s grace-period arithmetic (10+30+3+2=45), clean `go vet`, and the core design — hooks run after shutdown settles, never during, bounded independent of signal — holds up.

solace-lootii (dismissed):
Pulled latest `main`, checked out this branch, and went through it against Story 48's acceptance criteria (SOL-150251 discovery doc) plus a full local verification pass:

- `go build ./...`, `go vet ./...`, `golangci-lint run` — clean
- `go test ./...` (full suite, not just the touched packages) — all green, including the existing `cmd/server/shutdown_test.go` suite with unchanged assertions
- `internal/observability/hooks` — 100% statement coverage; the bounded-total, slow-doesn't-starve-fast, panic-contained, and error-swallowed cases all check out
- Confirmed no call sites of `drainAndShutdown`/`gracefulShutdown` outside `main.go` and `shutdown_test.go`, matching the "6 mechanical `nil` appends, no assertion changes" claim

Checked off against Story 48's ACs: HTTP drain before hooks ✅, budget split decided and documented in one place ✅ (and the K8s `terminationGracePeriodSeconds` total correctly stays at 45s), bounded total regardless of hook count ✅, one slow/failing hook can't starve others ✅, hook failures logged without affecting exit status ✅, second-signal force-close skips hooks ✅, no-hooks path is a true no-op ✅.

One thing worth a note rather than a blocker: the work breakdown says to "wire the force-signal channel so a second signal abandons remaining hooks." As written, `forceSig` is checked before hooks start (so a second signal arriving before that point correctly skips them entirely), but once `gracefulShutdown` is inside the hook-running branch, a second SIGTERM arriving during that ≤3s window isn't watched for — hooks just run out their bounded budget. Worst case that's a capped 3s, so I don't think it needs to hold up this PR, but a short comment on `gracefulShutdown` calling that out (or a follow-up if you want it exact) would close the gap between the code and that AC's literal wording.

Nice touch giving every hook the full budget concurrently rather than dividing it per-hook — better than the AC's literal phrasing and well-explained in the `RunAll` doc comment.

Approving — this is a clean, well-tested seam that unblocks Stories 14/25/46 without touching `shutdowner` or the existing shutdown semantics.

amitmorade:
@solace-awaheed Curious how you read the 3s budget — I can see two readings and I’m not sure which one you meant.

This PR takes the 3s from the old 5s Kubernetes slack (10+30+3+2=45), which leaves 2s before SIGKILL. The other reading is taking it out of the 30s HTTP window so that 5s buffer stays.

Once the three OTel providers register, worst case is 43s of work inside a 45s grace period. If that’s the trade-off you intended, this looks good to me — just wanted to check I had it right.

On Andrea’s err.Error() note: nothing is registered today so it’s inert, but Stories 14/25/46 will pass collector errors through that line. Happy either way — did you want to handle it here, or fold it into Story 14?

solace-awaheed:
@amitmorade the 3s comes from the old K8s slack, not the 30s HTTP window. TestDrainAndShutdown_ShutdownGetsFullBudgetAfterDrain locks this in by asserting that srv.Shutdown gets the full shutdownTimeout on its context, so I couldn't reduce that 30s without breaking an existing test.

So the breakdown is 10s (drain) + 30s (HTTP) + 3s (hooks) + 2s (buffer) = 45s. In the worst case, once all three providers register, that's 43s of actual work in a 45s grace period, leaving 2s before SIGKILL. That's the trade-off I was going for. If 2s feels too tight, I'm happy to increase the buffer by reducing the hook budget.

For the err.Error() note, I fixed it here rather than deferring it. I removed the raw error text from the logs, so now we just log the hook name.

solace-aross (approved):
Re-reviewed the follow-up commit addressing Copilot's two findings from the prior push (my earlier approval was auto-dismissed by that push, not by any finding of mine).

Both fixes read correctly and I verified each rather than trusting the description:

- Mid-flush second signal: `runShutdownHooks` now races the hook flush against `forceSig` and cancels the hook context immediately on a second signal, instead of always waiting out the 3s budget. Sabotaged it in a scratch worktree (dropped the `select`, ran `RunAll` unconditionally) and confirmed `TestGracefulShutdown_SecondSignalDuringHookFlushCutsItShort` goes red exactly as expected (`gracefulShutdown did not return after a second signal during hook flush`); restored, green again.
- Log line: `hooks.go` no longer logs `err.Error()` on a failed hook, only the hook name — correct per secure-logging-rules Rule 5, since a hook can wrap an unaudited external error (OTel collector).

`go build ./...` and `go test ./cmd/server/... ./internal/observability/hooks/... -race` both clean.

Approving.

amitmorade (approved):
LGTM

PR #363, opened 2026-09-01
SOL-153454: stop inventory refresh self-tests from failing on unrefreshed Dependabot PRs

amitmorade:
## Summary

The third-party inventory refresh self-tests (SOL-152956) whole-file-diffed the refresh script's output against the *committed* `THIRD_PARTY_LICENSES.md` / `THIRD_PARTY_BUILD_TEST.md`, treating that file as already-correct ground truth. That's false on every fresh Dependabot PR: `go.mod` or an Action SHA has moved and the inventory hasn't been refreshed yet, so the suites failed with `exit 0, but the file didn't fully converge to the committed one` — a misleading false positive on the exact situation these scripts exist to handle (confirmed live on PRs #320/#326).

Jira: SOL-153454

## Type of Change

## Motivation and Context

`assert_refresh` in both suites compared the refresh script's output to `$REPO_ROOT/$DOC`. That assumption holds on `main`, where the inventory is always converged, but not on a branch mid-drift. Not merge-blocking (neither self-test job is a required check), but it reads as "the refresh scripts are broken" on every future Dependabot bump.

## Changes Made

- `.github/scripts/refresh-licenses-inventory.test.sh` and `.github/scripts/refresh-build-test-inventory.test.sh`: `assert_refresh` now diffs against a baseline computed once per suite run (`compute_converged_baseline`), by running the real refresh script against an unmutated copy of the repo — not against the committed file.
- Added `INVENTORY_SOURCE` to let a case start from a stale copy of the inventory instead of the committed one.
- Added an in-suite ambient-drift case to each suite that manufactures the Dependabot shape (real inputs current, inventory stale) so this regression is covered on every `main` run, not only when a live drifted PR happens to exist.
- No changes to `refresh-licenses-inventory.sh` or `refresh-build-test-inventory.sh` — confirmed working correctly; this is a test-harness-only fix, as scoped by the ticket.

## Testing

**Tests Performed:**
- [x] Unit tests added/updated (the two new ambient-drift cases per suite)
- [x] Tested locally with `SKIP_NETWORK_TESTS=1` for both suites on a converged tree — all pass
- [x] Reproduced the original failure in an isolated git worktree by staling `golang.org/x/sync`'s version (licenses) and `actions/checkout`'s SHA (build-test) without running refresh, confirmed both suites still pass against that drift with the fix, confirmed the drift is left untouched afterward (harness doesn't write back)

**Test Coverage:**
- Converged tree (`main`-shaped): both suites pass, same as before
- Ambient inventory drift (Dependabot-shaped, unrefreshed): both suites now pass instead of failing
- Refusal cases (unparseable row, required differing-licence drop, unpinned action, new container image/npm package) still exit nonzero and leave the file untouched

## Breaking Changes

None — test harness only.

### Code Quality
- [x] Self-review completed
- [x] Comments explain "why" (baseline vs. committed file), not "what"

### Testing
- [x] All tests pass locally (`SKIP_NETWORK_TESTS=1` both suites)
- [x] Edge cases covered (refusal paths still refuse; ambient drift now converges)

### Git & CI
- [x] Commit signed off (DCO)

## Related Issues/PRs

- Related to SOL-152956 (#316), which introduced these self-tests
- Confirmed against the same shape of failure seen on PRs #320 and #326

---

**For Reviewers:**

**Focus Areas**: whether comparing against a per-run computed baseline (vs. the committed file) still gives the mutation/refusal cases real teeth — see the "Every exit-0 case ... should converge to exactly the computed baseline" comment in both files, which documents the collateral-damage check this preserves.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/scripts/refresh-build-test-inventory.test.sh
+++ b/.github/scripts/refresh-build-test-inventory.test.sh
@@ -200,9 +200,13 @@ setup_refresh_fixture() { # <tmp>
 # compute_converged_baseline
 #   Runs refresh against an unmutated fixture and saves the result as
 #   CONVERGED_DOC. Aborts the suite if that run does not exit 0 — every
-#   later exit-0 case is meaningless without this oracle.
+#   later exit-0 case is meaningless without this oracle. Not reachable
+#   today (this always runs before any case sets INVENTORY_SOURCE), but
+#   shadowed with a local, unset value regardless: the baseline must stay
+#   an unmutated oracle even if a future case sets INVENTORY_SOURCE before
+#   calling this.
 compute_converged_baseline() {
-    local tmp out_file got=0
+    local tmp out_file got=0 INVENTORY_SOURCE=
     tmp=$(mktemp -d)
     out_file=$(mktemp "$tmp/refresh-out.XXXXXX")
     setup_refresh_fixture "$tmp"
--- a/.github/scripts/refresh-build-test-inventory.test.sh
+++ b/.github/scripts/refresh-build-test-inventory.test.sh
@@ -6,8 +6,12 @@
 # copy, the tree symlinks the real submodules and workflows so expected values
 # come from real `go list`/`gh api` calls rather than a fixture that goes
 # stale, and every fix/add case asserts the produced row is byte-identical to
-# the committed one — not just that the refresh script's exit code was 0,
-# which a coincidentally-also-valid row would satisfy too.
+# the *computed* baseline (the real refresh script run once against an
+# unmutated copy of the repo), not the committed file. The committed file is
+# stale on every fresh Dependabot PR (SOL-153454); treating it as ground
+# truth made this suite fail for the exact situation the refresh scripts
+# exist to fix. Exit code 0 alone would also pass a coincidentally-also-valid
+# row, which is why the per-row check stays.
 #
 # This suite needs network access (`gh api` for action licences, `go run
 # github.com/google/go-licenses` for new Go-module rows) and a warm module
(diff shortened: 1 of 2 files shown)


solace-aross (dismissed):
Verified by injection in a worktree, not by reading. Staled the committed `THIRD_PARTY_LICENSES.md` version for `golang.org/x/sync` on main (pre-fix): the self-test false-failed exactly as described (3 cases wrongly red against the stale committed file). Same staling on this PR's head: 11 passed, 0 failed.

Sabotage check: disabled the version-fix line in `refresh-licenses-inventory.sh` and reran on this PR's head. 3 cases correctly went red. The new baseline still has teeth, it's not just papering over the ambient-drift shape.

On Copilot's two comments, checked the actual call order and traps before agreeing with either:
- `INVENTORY_SOURCE` leaking into `compute_converged_baseline`: not live today (the function runs before `INVENTORY_SOURCE` is ever set in either file, confirmed by line order in both scripts). And even if it leaked, a non-convergent source aborts the whole suite with `FATAL` rather than going silent, so "weaken the suite" overstates it. A defensive `local INVENTORY_SOURCE=` inside the function is a cheap one-liner worth taking, not a blocker.
- Per-case tmpdir outside `SUITE_TMP`: pre-existing since before this PR, neither file had a trap on main. This PR adds the first one. Not a regression in this diff.

Approving. The hardening above is optional.

solace-aross (approved):
Follow-up commit does exactly the optional hardening I flagged: `local ... INVENTORY_SOURCE=` shadows the caller's value for the duration of `compute_converged_baseline`.

Verified by injection again on this head, not by reading. Ran the clean suite (11 passed, 0 failed) with an ambient `INVENTORY_SOURCE` exported right before the real `compute_converged_baseline` call, both with the shadow in place and with it stripped back out. Both passed identically. That's not a hole, it's the convergence property doing its job: the class of drift a stray `INVENTORY_SOURCE` could introduce here (a stale version) is exactly what `refresh` always corrects regardless of starting state, so the baseline lands the same either way. The shadow's real value is for a narrower case refresh can't self-correct (e.g. a licence field that's wrong but present and version-matched, which nothing currently exercises) or a fixture that would make refresh refuse, which already aborts loudly via the existing `FATAL` path rather than silently. Matches the commit message's own "not reachable today" framing. Cheap, correctly scoped, no regressions.

Approving.

PR #367, opened 2026-09-02
SOL-153873: Pin GOOS=linux in third-party inventory checks so Linux-only deps (procfs) don't fail on macOS runners

solace-wmaryam:
## Summary
- `Third-party inventory refresh scripts self-test (offline) (macos-latest)` failed on PR #357 because `go list -deps ./cmd/server` resolves a different dependency closure depending on the runner OS: `github.com/prometheus/procfs` (pulled in transitively by `client_golang`'s process collector) is Linux-only, so it drops out of the closure on macOS even though the committed inventory row is correct for the shipped (Linux distroless) binary.
- Pins `GOOS=linux` in `licenses-check.sh` and in `refresh-build-test-inventory.sh`'s own `go list -deps` call, so the check agrees with itself regardless of which OS runs it.
- `refresh-licenses-inventory.sh` needs no change — it delegates to `licenses-check.sh` rather than calling `go list` itself, so it inherits the fix.
- `build-test-licenses-check.sh` is intentionally left alone (documented in-code) — it checks the test submodules' own closures, not `./cmd/server`, and this ticket is about the root-module closure specifically.

## Test plan
- [x] `SKIP_NETWORK_TESTS=1 .github/scripts/licenses-check.test.sh` — 10/10 passed
- [x] `SKIP_NETWORK_TESTS=1 .github/scripts/refresh-build-test-inventory.test.sh` — 12 passed, 1 failed (pre-existing, reproduced identically on unmodified `main`: a sandbox network limitation resolving `golang.org/x/oauth2`'s licence via go-licenses, unrelated to this change), 3 skipped (network tests)
- [x] Confirmed via `go list -deps ./cmd/server` vs `GOOS=linux go list -deps ./cmd/server` that this is the exact root cause (0 vs 3 `procfs` hits)

Fixes SOL-153873.

### DIFF
--- a/.github/scripts/build-test-licenses-check.sh
+++ b/.github/scripts/build-test-licenses-check.sh
@@ -215,9 +215,13 @@ while read -r gomod; do
     # so the closure silently empties and the reverse-direction check then
     # reports every real component as unused — telling a maintainer to delete
     # legitimate rows from a compliance artifact.
+    #
+    # GOOS pinned to linux for the same reason as licenses-check.sh: the
+    # closure is platform-dependent, and linux (what CI runs on) is the ground
+    # truth this inventory documents — not the maintainer's laptop.
     list_out=""
     list_rc=0
-    list_out=$(cd "$sub" && go list -deps -test -f '{{with .Module}}{{.Path}} {{.Version}}{{end}}' ./... 2>&1) || list_rc=$?
+    list_out=$(cd "$sub" && GOOS=linux go list -deps -test -f '{{with .Module}}{{.Path}} {{.Version}}{{end}}' ./... 2>&1) || list_rc=$?
     if [ "$list_rc" -ne 0 ]; then
         err "$DOC" "'go list' failed in $sub, so its dependency closure is unknown. Fix that before trusting this check. Output: $(head -3 <<<"$list_out" | tr '\n' ' ')"
         continue
--- a/.github/scripts/licenses-check.sh
+++ b/.github/scripts/licenses-check.sh
@@ -69,9 +69,16 @@ done
 ROOT_MODULE=$(go list -m)
 
 # Expected: "<module path> <version> <dir>" for every module in the binary's
-# closure, minus this repository itself.
+# closure, minus this repository itself. GOOS is pinned to linux because the
+# closure is platform-dependent — github.com/prometheus/procfs is only pulled
+# in on linux (client_golang's process collector) — and linux is the ground
+# truth: it is what this check's CI job runs on and what the shipped container
+# image is. Unpinned, a macOS run computes a smaller closure and reports the
+# procfs row as "listed but not in the dependency closure; drop the row" —
+# advice that, followed, deletes a correct row from a compliance artifact
+# (PR #365 did exactly that via the refresh script).
 closure=$(
-    go list -deps -f '{{with .Module}}{{.Path}} {{.Version}} {{.Dir}}{{end}}' "$TARGET" |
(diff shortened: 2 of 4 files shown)


solace-aross (dismissed):
Verified by injection on macOS (arm64), not by reading. Reverted the `GOOS=linux` pin in `licenses-check.sh` and reran the offline self-test: reproduced the exact SOL-153873 failure (`the committed artifacts pass` and the row-resolution case both false-fail, 8 passed/2 failed). Reapplied the fix: clean, 10/0. Confirms the root cause and the fix.

One correction on Copilot's comment. It's right on the literal fact, wrong on why it matters: `release.yml`'s build-binaries matrix (lines 280-289) does ship `darwin/amd64` and `darwin/arm64` archives alongside linux, so "the shipped binary is Linux-only" is inaccurate as written. But the darwin archives were never what `THIRD_PARTY_LICENSES.md` tracked either way: `procfs` is already a committed row (added when `client_golang`'s process collector landed it), which only makes sense if the doc was always implicitly tracking the Linux/Docker closure. Before this PR, a macOS *runner* validated that Linux-shaped doc against a darwin-native `go list` and got a false mismatch depending on which CI machine happened to run it, not because the check was validating the darwin binary on purpose. Pinning `GOOS=linux` doesn't create a new gap here, it makes the existing implicit assumption explicit and deterministic. Worth fixing the comment's wording (say "the committed inventory tracks the Linux/Docker build" rather than "the shipped binary is Linux-only"), and worth a separate ticket if anyone wants the darwin archives' own closure audited, since nothing does that today, before or after this PR. Not a blocker either way.

Approving.

solace-aross:
Heads up: #370 (SOL-153937) fixes the same GOOS-dependent closure bug, filed after it bit again on #365 in production (a macOS refresh deleted the `procfs` row from `THIRD_PARTY_LICENSES.md`, because two more scripts have the same unpinned closure this PR doesn't touch). #370 is a strict superset: it covers `licenses-check.sh` and `refresh-build-test-inventory.sh` the same way this PR does, plus `build-test-licenses-check.sh` and `sbom-check.test.sh`, which this PR leaves unpinned.

Both branches came off the same pre-#367 main and touch the same lines in `licenses-check.sh`/`refresh-build-test-inventory.sh` with different comment wording, so whichever merges second will conflict, not land silently.

My approval on this PR stands for what it covers, verified by injection. But recommend closing this one in favor of #370, or rebasing it to add the two missing files, rather than merging both independently.

solace-wmaryam:
@solace-aross, I've rebased this PR against #370 and it now carries the full fix. Please review when you have a moment!

amitmorade (approved):
LGTM

solace-aross (approved):
This branch is now byte-identical to #370 (same single commit, `4f57bccc`, confirmed via `gh pr view --json commits` on both). I already did full Gate 2 verification of this exact commit there: reverted and reproduced the macOS false-failure in `licenses-check.sh`/`refresh-build-test-inventory.sh`, confirmed `build-test-licenses-check.sh`'s pin is forward-looking hardening rather than a live fix (procfs isn't in a test-submodule closure today), and confirmed `sbom-check.test.sh`'s throwaway-GOBIN approach actually changes the SBOM's dependency set (0 vs 8 `procfs` hits with/without the `GOOS=linux` pin), not just its exit code. See #370's review for the full evidence. Nothing to add by re-doing that work here on the same content.

CI is green here too (only the known non-required `Guardian gate: skipping`).

One housekeeping note: with both PRs now carrying the identical commit, merging one will make the other a no-op merge of the same tree. Worth closing whichever doesn't merge first rather than merging both.

Approving.

PR #369, opened 2026-09-02
SOL-152087: Carry the verified human principal through every request so audit events can name who acted (Story 20)

solace-awaheed:
## Summary

Makes `auth.Principal` the single canonical caller identity, built once per request
and carried on the context, and re-sources the existing audit-log `Identity` as a
projection of it. Emitted log output is unchanged.

Jira: SOL-152087 (Story 20)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

Four separate log sites each reached into the MCP SDK's token structure and rebuilt
caller identity independently. Nothing structural kept them in agreement, and the
audit events, metrics and denial records in Stories 21-24 all build on that identity.
This collapses them into one extraction with one sanitization layer.

While implementing it I hit a defect worth flagging for review. The first version
attached the principal in an HTTP middleware, which passes every unit test and is
wrong in this transport: in stateful streamable HTTP the go-sdk builds its connection
once, from the POST that establishes the session, so per-request context values never
reach a tool handler. That version named the token which opened the session for the
session's whole life, through every client token refresh, while keeping the log line's
exact shape. Population is therefore an MCP receiving middleware over the SDK's
per-request extras, which is the only carrier that tracks the caller.

## Changes Made

- `auth.Principal` now carries the committed claim set (`sub`, `scope`, `client_id`,
  `iss`, `jti`) behind accessors over unexported fields, so `NewPrincipal` is the only
  way to build one and therefore the only place claims are read and sanitized.
  `preferred_username` and `email` are excluded per Q-013, with a reflection test
  freezing the field set.
- New `auth.PrincipalMiddleware`, an MCP receiving middleware that populates the
  principal per request. Registered last in `cmd/server` on purpose:
  `AddReceivingMiddleware` wraps the current handler, so the last registration runs
  first, and the tools/list filter reads the principal.
- `tools.Identity` is now a projection at all five audit emit sites.
  `NewIdentityFromTokenInfo` is deleted. Type, fields and log output are unchanged.
- `AbsentSentinel` and `NormalizeAbsent` moved into the shared `sanitize` package, so
  both reserved audit values live in one place.
- Corrected a comment in `register.go` asserting that the HTTP request context reaches
  a tool handler. It does not, and that claim is what produced the defect above.
- Corrected the `sanitize` package's claim that its CPU cost is bounded by the output
  cap. Behaviour unchanged; the comment now states the real bound.
- Docs: both sentinels documented for operators; architecture page no longer describes
  the principal as an empty skeleton.

## Testing

**Test Environment:**
- Go version: 1.26.5 (module targets 1.25.0)
- OS: macOS 26.6.2, arm64
- Broker version: n/a, no broker involved in these paths

**Tests Performed:**
- [ ] E2E tests added/updated

**Test Coverage:**
- Golden test pinning the `tool invoked` line byte for byte, in authenticated and
  disabled modes. Landed as its own first commit against the unrefactored code, so
  the byte-identical requirement is enforced by CI rather than by memory.
- Integration test driving a real HTTP session, calling a tool twice with different
  tokens, asserting each audit record names its own token. This is the case no unit
  test could see.
- All five emit sites, asserting field values rather than key presence. Four had no
  coverage at all before this: blanking identity at the primary audit chokepoint left
  the whole suite green.
- Middleware registration order, driven through the real server wiring.
- Sample JWTs through the real verifier for claim presence and absence; sanitization
  of newline, null byte, bidi override and over-cap values; disabled mode emitting no
  identity keys.
- Every new test was verified by deliberately breaking the code it covers.

## Breaking Changes

None. Log output, MCP tool schemas and operator configuration are unchanged.

### Code Quality
- [x] Self-review completed

### Security
- [x] No credentials in code
- [x] Credential-carrying types implement `slog.LogValuer` (n/a: no new
      credential-carrying types; `Principal` holds identity claims, not secrets)

### Testing
- [x] Error paths tested

### Documentation
- [ ] README.md updated (not user-facing)
- [x] docs/ updated
- [x] CHANGELOG.md updated
- [ ] Examples updated (n/a)

### Git & CI
- [ ] Branch is up to date with main (3 commits behind; merges cleanly, no conflicts)
- [x] No merge conflicts
- [ ] CI checks passing (not yet run)

## Additional Notes

The Docker-backed E2E suites have not been run locally, so CI is the first place
`e2e-oauth` exercises this path.

Two things I would like an opinion on. `scope` is stored as `[]string` to match the
SDK, which costs `Principal` its comparability and use as a map key; a space-joined
string would keep both, and there is no production consumer either way.
`auth.WithPrincipal` is exported, so any package in the module could attach a
principal no token backs; it is reachable from tests only today, and unexporting it
behind a test-only seam is the stricter option.

## Related Issues/PRs

- Related to SOL-153935: the raw subject token used for RFC 8693 broker token
  exchange rides the identical frozen carrier, so exchange replays the session's
  first credential. Pre-existing since 0.6.0, reproduced on `main`, untouched here.
- Unblocks SOL-152090, SOL-152096, SOL-152097, SOL-152099 (Stories 21-24).


solace-aross (dismissed):
Solid refactor, thoroughly self-documented, and the author's own PR description already flags the two subtlest risks correctly. Verified the core design, not just read it.

**Extra vs. ctx, verified against the SDK source.** Traced `go-sdk@v1.7.0/mcp/streamable.go`: `servePOST` attaches the current POST's `TokenInfo` onto `jreq.Extra` fresh per message (`streamable.go:1553-1556`) before the message crosses onto the `c.incoming` channel, while the ctx that eventually reaches a method handler descends from the session-establishing POST (`connectStreamable(req.Context(), ...)` at connect time, `streamable.go:700-703`). So `PrincipalMiddleware` reading `req.GetExtra().TokenInfo` rather than ctx is the only correct choice in this transport, and `TestPrincipalMiddleware_isPerRequest` (calling the middleware's returned handler twice with different `jti`s) correctly proves that. `TestPrincipal_fieldSetIsCommittedClaims`'s reflection check does freeze the field set and does reject `username`/`email` substrings, confirmed by reading it.

**Sabotage finding: `TestPrincipalReachesListFiltering` doesn't pin what its own comment claims.** Its docstring says "It fails if the two middlewares are registered in the wrong order" and "Exactly the order main() uses." It isn't: the test calls `installToolListFiltering` and `server.AddReceivingMiddleware(auth.PrincipalMiddleware())` itself, hardcoding its own copy of the correct order, rather than exercising `main()`'s actual registration code. Proved it by swapping the two lines in `cmd/server/main.go` and rerunning: `TestPrincipalReachesListFiltering` still passes. Checked whether e2e backstops it — `test/e2e-oauth/test-rbac-scenarios.sh` covers `tools/call` authorization audit records but never calls `tools/list`, so nothing in CI would catch a future reordering of those two lines in `main()`, exactly the silent-regression scenario this PR's own comment warns about. Same shape of gap as the `cmd/server` wiring-coverage exemption in `docs/internal/unit-test-coverage.md` (I raised the identical pattern on #368 today), so not a blocker, but worth a follow-up: either have the test call a small extracted `registerAuditMiddleware(server, ...)` that `main()` also calls, or drive it through `TestMain`-style server construction.

**Investigated the correlation_id claim in the CHANGELOG, found the author already caught the real bug.** The entry says the non-string-claim ERROR "now carries the request's correlation_id, so it can be joined to the audit line." I wrote a test to check this end to end and my first attempt was methodologically broken (mis-tagged which physical POST got which header, given the SDK sends `server/discover` then `initialize` before any tool call). Redid it tagging by JSON-RPC method instead, and ran an independent adversarial pass on my own finding before trusting it. Confirmed: `correlation_id` is session-scoped, not request-scoped, wherever it's read off a context in the tool path — `Meta`, `tool invoked`, `tool authorization`, `tool list filter`, outbound SEMP correlation headers, token-exchange logs — while the `X-Correlation-Id` *response header* is genuinely per-request (set in `correlation.Middleware` before `next` runs, on that POST's own context). So a client following the documented response-header contract cannot find its own request in the server's logs; it gets the session's first (`initialize`) request's ID instead. This is pre-existing on `main`, unrelated to this diff, reproduces identically at the merge-base commit, and this PR is the one that finally diagnosed the mechanism correctly, in `register.go`'s corrected comment ("This comment claimed the opposite and was measured wrong... Not addressed here"). Two asks, neither blocking:
1. The CHANGELOG line's "the request's correlation_id" should say "the session's" — the ERROR and the `tool invoked` line do carry the same value as each other (the join works), but neither identifies the specific request, which is what the current wording implies.
2. File a ticket for the underlying bug. Its blast radius is wider than this PR's comment describes (the response-header-vs-log-line mismatch above is a concrete, client-visible defect in already-shipped SOL-151282 behavior), so it deserves tracking on its own rather than living only in a code comment.

**Copilot's `auth.InjectPrincipal` comments are right.** Grepped the diff: no such symbol exists anywhere in `principal.go` (the real names are `WithPrincipal`, `PrincipalFrom`, `PrincipalMiddleware`). Both flagged comments (`register.go:176`, `authorization_audit_schema_test.go:156`) should say `auth.PrincipalMiddleware`/`auth.WithPrincipal`. Small, easy fix.

**The two questions you asked for an opinion on:**
- `scope` as `[]string`: keep it. OAuth scope values are themselves space-delimited by spec, so a join/split round-trip is lossless either way, and there's no comparability consumer today. Converting later if one shows up costs less than guessing now.
- `auth.WithPrincipal` exported: take the stricter option. This backs audit-trail identity; any package being able to attach an unbacked principal is exactly the kind of footgun that's cheap to close now and expensive to notice once something depends on the looser version. A test-only seam (build-tag-gated `_test.go` export, or an internal test helper package) costs little today since it's only reachable from tests anyway.

Approving. Nothing above blocks merge; the wiring-test gap and the correlation ticket are both pre-existing-shaped, not regressions this PR introduces.

internal/auth/principal.go
```diff
@@ -31,23 +46,156 @@ import "context"
...
+// With no TokenInfo on the request it attaches nothing: auth mode "disabled"
+// verifies no token, so PrincipalFrom keeps returning the zero value and
+// absence stays a single code path. It never rejects — RequireBearerToken
+// upstream is the authority on that.
+//
+// Registration order matters. AddReceivingMiddleware wraps the current
+// handler, so the LAST middleware registered runs FIRST; this one must be
+// registered after any middleware that reads the principal (today
+// tools.WithListFiltering). cmd/server does that, and
+// TestPrincipalReachesListFiltering pins it.
```
solace-wmaryam: This godoc says "TestPrincipalReachesListFiltering pins it." `solace-aross` showed that swapping the two registration lines in `cmd/server/main.go` does not fail that test — it builds its own server with its own order and does not touch `main()`'s wiring. The same correction noted for the `main.go` comment applies here: this claim should be removed or softened so future readers do not rely on coverage that does not exist.

solace-wmaryam (dismissed):
Solid refactor. The freshness invariant design and the -style sabotage evidence are both strong. One comment below not raised elsewhere.

amitmorade (dismissed):
LGTM

solace-wmaryam (approved):
LGTM!

amitmorade (approved):
LGTM

PR #370, opened 2026-09-02
SOL-153937: pin inventory closure computation to GOOS=linux

amitmorade:
## Summary

Pins every dependency-closure computation in the third-party inventory scripts to `GOOS=linux`, so a refresh or check run on macOS computes the same closure CI validates on ubuntu — instead of a smaller one that reports Linux-only rows as stale and tells the maintainer to delete them.

Jira: [SOL-153937](https://sol-jira.atlassian.net/browse/SOL-153937)

## Type of Change

## Motivation and Context

The inventory scripts compute the binary's dependency closure with the host's `GOOS`, but linux is the ground truth: it's what the checks' CI jobs run on and what the shipped container image is. `github.com/prometheus/procfs` is only in the closure on linux (client_golang's process collector), so a macOS run of `licenses-check.sh` reported the correct procfs row as "listed but not in the dependency closure; drop the row." The refresh script followed that advice on #365 — deleting a valid row from a compliance artifact and turning `Third-party licenses current` and the SBOM self-test red on a commit meant to fix them.

## Changes Made

- **`.github/scripts/licenses-check.sh`** — pins the `go list -deps ./cmd/server` binary closure to `GOOS=linux`, with a comment recording the procfs failure mode.
- **`.github/scripts/build-test-licenses-check.sh`** — pins the per-submodule `go list -deps -test` closures the same way.
- **`.github/scripts/refresh-build-test-inventory.sh`** — pins submodule resolution to match the check's closure exactly.
- **`.github/scripts/sbom-check.test.sh`** — builds cyclonedx-gomod natively into a throwaway GOBIN, then runs it with `GOOS=linux`. `GOOS=linux go run` can't do this: it cross-compiles the tool itself into a binary the host can't exec. The throwaway GOBIN preserves the property the plain-`go run` shape had — a stale pre-installed tool can't mask what real CI would do.

Deliberately unchanged: go-licenses licence *resolution* for packages that don't build on the host. The refresh already refuses loudly and directs a hand-verified row, which is the right behavior for a compliance file.

`refresh-licenses-inventory.sh` needs no edit — it derives its work entirely from `licenses-check.sh`'s diagnostics, so pinning the check pins the refresh.

## Testing

**Test Environment:**
- Go version: go1.26.0
- OS: macOS (darwin/arm64) — the platform where the bug lived
- Broker version: n/a (CI scripts only)

**Tests Performed:**
- [x] All five self-test suites pass locally: licenses-check 10/10, sbom-check 10/10, build-test-licenses-check 26/26, refresh-build-test 15/15 (3 network cases skipped offline), refresh-licenses 11/11 (1 skipped)
- [x] Both real checks pass against the tree on macOS (47 modules, procfs included) — `licenses-check.sh` failed on macOS before this change
- [x] `make refresh-third-party-inventory` on macOS makes no changes — a Mac refresh can no longer delete Linux-only rows

**Test Coverage:**
- CI's ubuntu jobs are behaviorally unaffected (`GOOS=linux` was already the effective value there); the macos-latest self-test job now exercises the pinned path and should produce output identical to ubuntu.

## Breaking Changes

None.

## Additional Notes

Known residual gap, out of scope by design: adding a *new* Linux-only dependency still requires one hand-verified row when refreshing from macOS, because go-licenses can't inspect a package that doesn't build on the host. The script's refusal message directs exactly that.

## Related Issues/PRs

- Related to #365 (the Dependabot PR where the darwin refresh deleted the procfs row)

### DIFF
--- a/.github/scripts/build-test-licenses-check.sh
+++ b/.github/scripts/build-test-licenses-check.sh
@@ -215,9 +215,13 @@ while read -r gomod; do
     # so the closure silently empties and the reverse-direction check then
     # reports every real component as unused — telling a maintainer to delete
     # legitimate rows from a compliance artifact.
+    #
+    # GOOS pinned to linux for the same reason as licenses-check.sh: the
+    # closure is platform-dependent, and linux (what CI runs on) is the ground
+    # truth this inventory documents — not the maintainer's laptop.
     list_out=""
     list_rc=0
-    list_out=$(cd "$sub" && go list -deps -test -f '{{with .Module}}{{.Path}} {{.Version}}{{end}}' ./... 2>&1) || list_rc=$?
+    list_out=$(cd "$sub" && GOOS=linux go list -deps -test -f '{{with .Module}}{{.Path}} {{.Version}}{{end}}' ./... 2>&1) || list_rc=$?
     if [ "$list_rc" -ne 0 ]; then
         err "$DOC" "'go list' failed in $sub, so its dependency closure is unknown. Fix that before trusting this check. Output: $(head -3 <<<"$list_out" | tr '\n' ' ')"
         continue
--- a/.github/scripts/licenses-check.sh
+++ b/.github/scripts/licenses-check.sh
@@ -69,9 +69,16 @@ done
 ROOT_MODULE=$(go list -m)
 
 # Expected: "<module path> <version> <dir>" for every module in the binary's
-# closure, minus this repository itself.
+# closure, minus this repository itself. GOOS is pinned to linux because the
+# closure is platform-dependent — github.com/prometheus/procfs is only pulled
+# in on linux (client_golang's process collector) — and linux is the ground
+# truth: it is what this check's CI job runs on and what the shipped container
+# image is. Unpinned, a macOS run computes a smaller closure and reports the
+# procfs row as "listed but not in the dependency closure; drop the row" —
+# advice that, followed, deletes a correct row from a compliance artifact
+# (PR #365 did exactly that via the refresh script).
 closure=$(
-    go list -deps -f '{{with .Module}}{{.Path}} {{.Version}} {{.Dir}}{{end}}' "$TARGET" |
+    GOOS=linux go list -deps -f '{{with .Module}}{{.Path}} {{.Version}} {{.Dir}}{{end}}' "$TARGET" |
         grep -v '^$' |
         grep -v "^${ROOT_MODULE} " |
         sort -u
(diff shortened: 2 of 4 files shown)


solace-aross (approved):
Verified by injection on macOS (arm64), not by reading, across all four files.

`licenses-check.sh` / `refresh-build-test-inventory.sh`: same fix I already verified on #367 (see below for the overlap). Reproduces and clears the same way.

`build-test-licenses-check.sh`: reverted the `GOOS=linux` pin and ran the real check against this repo. It still passed (46/46) even unpinned, since `procfs` isn't in any test submodule's closure today, only the root module's. So there's no live bug here yet, this is consistent, forward-looking hardening against the exact failure mode that already hit `licenses-check.sh`, not a fix for a reproduced defect. Said so plainly rather than claiming a catch that isn't there.

`sbom-check.test.sh`: this one I dug into hardest, since it's the most structurally different change (native build into a throwaway `GOBIN`, then `GOOS=linux` on the run, instead of `GOOS=linux go run`). Confirmed the stated reason directly: `GOOS=linux go run github.com/CycloneDX/cyclonedx-gomod/...@v1.10.0 --version` on this machine fails with `exec format error`, because `go run` would cross-compile the tool binary itself, not just point its analysis at linux. Then confirmed the fix actually changes the SBOM's content, not just its exit code: built the pinned tool into a throwaway `GOBIN` and ran it against `cmd/server` twice, once natively and once with `GOOS=linux`. `prometheus/procfs` appears 0 times in the native SBOM and 8 times in the `GOOS=linux` one. The mechanism is real, not just plausible.

The self-test's own "could not generate a real SBOM to test against" failure I hit locally is a pre-existing environment gap, not a regression: reproduced the identical failure (`git: reference not found`) on unmodified `main` in the same detached-HEAD worktree setup, because `cyclonedx-gomod` needs `go version -m`-style git describe info a worktree checkout doesn't have. CI's own real checkout doesn't hit this, and CI is green (SBOM check self-test, both third-party inventory self-test jobs including the previously-flaky macos-latest one, and Third-party licenses current all pass on this PR).

**Overlap with #367, flagging for whoever manages merge order:** this PR and #367 fix the same underlying bug independently and don't know about each other, both branched from the same pre-#367 `main` and touch `licenses-check.sh`/`refresh-build-test-inventory.sh` with the same semantic change (different comment wording, so they will conflict, not silently coexist). This PR is a strict superset: it also covers `build-test-licenses-check.sh` and `sbom-check.test.sh`, which #367 leaves unpinned. That gap is exactly what let the same bug bite again on #365, in real production history, after #367 had already been opened: a maintainer ran `make refresh-third-party-inventory` on macOS, and because two of these four scripts still resolved a GOOS-dependent closure, it deleted the `procfs` row from `THIRD_PARTY_LICENSES.md`. Recommend closing #367 in favor of this one, or rebasing #367 to add the two missing files. Left a note on #367 pointing here so it doesn't get merged independently.

Approving on the merits.

amitmorade:
Closing this one, https://github.com/SolaceProducts/solace-broker-mcp/pull/367 rebassed with this branch so it will carry the additional changes introduced here. Thank you @solace-aross for helping coordinate between the prs

PR #379, opened 2026-09-04
SOL-152086: Expose per-tool call-rate and latency metrics so teams can dashboard tool SLOs

solace-wmaryam:
## Summary

Exposes per-tool RED (Rate/Errors/Duration) metrics and an in-flight request gauge on the Prometheus `/metrics` endpoint, so NOC/SRE teams can build tool-level SLO dashboards and alerts with no custom instrumentation. Gated by `OBS_METRICS_ENABLED` (off by default).

Tracks [SOL-152086](https://sol-jira.atlassian.net/browse/SOL-152086) (Story 15).

## Type of Change

- [x] Documentation update

## Motivation and Context

This is the core Rate/Errors/Duration signal for every tool the server exposes. It also establishes the shared `outcome` label (and companion `error_type`) that the audit stream and traces reuse, so a single key joins metrics, logs, audit, and spans and they cannot disagree about why a call failed. Builds on the meter provider and `/metrics` endpoint from SOL-152091 (Story 14).

## Changes Made

- **New instruments** (`internal/observability/metrics/instruments.go`): `mcp_tool_invocation_total` (counter), `mcp_tool_invocation_duration_seconds` (histogram, buckets `1ms…10s`), and the unlabelled `mcp_http_active_requests` gauge. All methods are nil-safe, so a disabled server records nothing.
- **Wiring**: `Provider.ToolMetrics()` builder; a nil-safe `atomic.Pointer` recorder in the `tools` package (`SetToolMetrics`), installed from `main.go` only when `OBS_METRICS_ENABLED` is on.
- **Observation**: counter + histogram recorded at the existing deferred `logToolResult` site — one site covers every tool. `error_type` is promoted from a log-only attribute to a metric label (empty on any non-error outcome).
- **Log field rename** `status` → `outcome` (Decision #2), so the log line, the metric label, and the future audit/trace fields share one join key. Same `success`/`error` values.
- **In-flight gauge**: `ActiveRequestsMiddleware` on `/mcp`, incrementing on entry and decrementing from a `defer` so a panicking request cannot leak a count.
- **Docs/changelog**: `docs/observability.md` updated for the eleventh `error_type` value (`broker_permission_denied`, ISSUE-031) and the corrected cardinality (13 series per tool/broker); `CHANGELOG.md` `[Unreleased]` entries.

## Testing

**Test Environment:**
- Go version: go1.26.3
- OS: darwin/arm64
- Broker version (if applicable): n/a (no broker interaction)

**Tests Performed:**

**Test Coverage:**
- Label correctness on success; empty `error_type` on success (asserts no non-empty-`error_type` series is created — the cardinality guarantee).
- Histogram bucket boundaries are exactly the nine required.
- In-flight gauge inc/dec, verified with a synchronization barrier.
- Recovered-panic path records `outcome=error`, `error_type=panic` (not its own outcome).
- Middleware decrements even when the wrapped handler panics.
- Golden `/metrics` fixture extended with the tool-RED series and the gauge.
- Full `make check` passes: build, vet, golangci-lint, `-race` tests, coverage 89.4% (85% floor).

## Breaking Changes

**Impact:** The per-invocation tool log line (`msg="tool invoked"`) now carries `outcome` instead of `status`. Anything grepping or dashboarding on `status` for that line must update. This is not a schema/config or tool-name break; the values are unchanged (`success`/`error`).

**Migration Guide:** Replace `status` with `outcome` in any log query or dashboard that targets the `tool invoked` line.

### Code Quality
- [x] Self-review completed

### Security
- [x] Followed secure logging rules (`/check-logs` clean; only the log field name changed)

### Testing
- [x] Edge cases covered by tests (panic path, empty error_type, bucket edges)
- [x] Error paths tested

### Documentation
- [x] docs/ updated (`docs/observability.md`)
- [x] CHANGELOG.md updated

### Git & CI
- [x] No merge conflicts

## Additional Notes

Split into two commits for review: (1) the instrument/production changes plus the log-rename fallout to existing tests, (2) the new unit tests, golden fixture, and the test-only `go.mod` dependency (`kylelemons/godebug`, pulled in by `prometheus/testutil`). The `cancelled` outcome value is defined here but wired by a later story (Story 42).

## Related Issues/PRs

- Builds on SOL-152091 (`/metrics` endpoint and meter provider)


internal/tools/manager.go
```diff
@@ -446,6 +451,9 @@ func logToolResult(ctx context.Context, tool string, broker *string, start time.
 	}
 
 	slog.LogAttrs(ctx, slog.LevelError, "tool invoked", attrs...)
+
+	// Record tool invocation and duration metrics on the error path
+	toolMetrics.Load().Record(ctx, tool, *broker, "error", *errorType, dur)
```
solace-aross: On the `unknown_broker`/`broker_init_error` path, `brokerAlias` is never canonicalized — that only happens on the success branch a few lines up (manager.go:234-236) — so `*broker` here is whatever raw string the caller passed. Once `OBS_METRICS_ENABLED` is on, that string becomes a Prometheus label value with no allow-list (`register.go`'s `broker` param schema is `{"type":"string"}`, no enum).

Checked this against the actual OTel/Prometheus SDK versions this repo pins: the meter provider's default cardinality limit (2000 series/instrument) caps memory growth, so it's not unbounded, but once tripped every *new* (tool, broker, error_type) combination silently falls into `otel_metric_overflow`, and those 2000 slots persist for the process lifetime — the RED dashboards this PR exists to feed go blind, and that's exactly the kind of moment (callers retrying with typos or a stale alias) that would trip it. Case matters too: `broker_init_error` on a *configured* broker mints a separate series per casing variant (`dev`/`DEV`/`Dev`) since pool lookup lowercases but the label doesn't, so this isn't purely an abuse case, it happens on ordinary traffic.

Suggest canonicalizing before `Record`: resolve to `bc.DisplayName()` when the alias matches a configured broker case-insensitively, and use a fixed sentinel (e.g. `"unknown"`) otherwise. Also worth fixing the doc claim at docs/observability.md:161 ("Bounded by the number of configured brokers") since that's only true on the success path today.

docs/observability.md
```diff
@@ -763,6 +624,7 @@ Present only on `outcome: error`, drawn from a closed set of ten values:
 | `nil_result` | The tool returned no result. |
 | `output_validation_error` | The tool's output failed schema validation. |
 | `marshal_error` | The result could not be serialized. |
+| `broker_permission_denied` | The broker rejected the operation as not permitted for the authenticated identity. |
```
solace-aross: `broker_permission_denied` isn't implemented anywhere — grepping this head for it hits only this line. This table reads as the current shipped set, not something reserved for later (unlike `cancelled` in the outcome table above, which is explicitly marked reserved).

Meanwhile two real, easily-reachable values are missing from this table: `bad_request` (register.go:204, fires on any tool call whose arguments fail to unmarshal — probably the single most reachable error_type in the whole system) and `not_found` (describe_semp_schema.go:372). The actual closed set at this head is 12 values, not 11.

Given the doc says "CI enforcement planned for GA," this table is meant to be authoritative already. Worth generating it from the actual `errorType = "..."` literals in internal/tools rather than hand-maintaining it, so it can't drift again.

docs/observability.md
```diff
@@ -164,15 +162,15 @@ refused by tool authorization never reaches one, so it is absent here and counte
...
 - `outcome`: see [The Outcome Vocabulary](#the-outcome-vocabulary).
-- `error_type`: the failure cause, from the ten values in
+- `error_type`: the failure cause, from the eleven values in
   [`error_type`](#error_type). Empty on any non-error outcome.
 - Histogram buckets (seconds): `0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1, 5, 10`.
 
 **Cardinality:** `error_type` is non-empty only on the error path, so the series per `tool` and
-`broker` is bounded at `success (1) + cancelled (1) + error x 10 = 12`, not the 33 a naive
-product of the two label domains would suggest. That 12 is the worst case once every outcome
+`broker` is bounded at `success (1) + cancelled (1) + error x 11 = 13`, not the 36 a naive
```
solace-aross: Even correcting the count to 12, the arithmetic here doesn't hold: `bad_request` and `not_found` only ever carry `broker=""` (register.go never sets `brokerAlias` before the bad_request return; describe-semp-schema has no `broker` parameter at all), and `unknown_broker` by definition never carries a configured alias. "Per `tool` and `broker`" doesn't cleanly multiply through in either direction. Worth restating this section once the table above is fixed, rather than swapping the number to 14.

solace-aross (changes requested):
Solid feature overall: nil-safe instruments, panic path correctly classified, gauge decrements survive a panicking handler, and the `status`→`outcome` rename is scoped to exactly the one log line it claims and disclosed clearly in the CHANGELOG. Two issues need a fix before merge; the rest are worth a look but don't need to block.

**Blocking**

1. The `broker` metric label isn't canonicalized on the error path, so it can carry an unbounded set of caller-supplied strings once metrics are on — see inline comment on manager.go.
2. `error_type`'s documented closed set doesn't match the code — see inline comments on docs/observability.md.

**Notes (non-blocking)**

- `internal/tools/metrics.go`'s package-level `atomic.Pointer[metrics.ToolMetrics]` + `SetToolMetrics` setter is the first cross-cutting dependency in this codebase wired that way rather than through constructor injection — `pool`, `shutdownHooks`, and `readiness` are all threaded through explicit parameters instead. More metrics/audit work is already staged per the CHANGELOG (SOL-153884, SOL-153443, SOL-150254), so whatever pattern lands here is likely to get copied. Worth considering a `ToolManager` field built before tool registration in main.go, matching how `pool` is handled.
- `ActiveRequestsMiddleware` (cmd/server/main.go:1208) is wired outside `buildMCPEndpoint`, this codebase's one documented spot for `/mcp`'s middleware chain, and unconditionally regardless of `OBS_METRICS_ENABLED`. Correct today (nil-safe), but a reader auditing `buildMCPEndpoint` for the full layer order will miss it.
- That middleware's nil-safe path — the default for every deployment, since metrics are off by default — has no test exercising it directly. Every current test that touches it calls `SetToolMetrics` with a real provider first. One test with `toolMetrics` left at its zero value would close the gap.
- There's a narrow startup window where `/mcp` is already accepting connections before `SetToolMetrics` runs. Any request landing there records nothing, silently. Not a crash risk, but worth a reorder so metrics are wired before the listener starts — that's exactly the moment right after a restart when someone's watching the dashboard for a regression.

Happy to re-review once the two blocking items are addressed.

bczoma (commented):
## Summary

Solid, well-tested instrumentation with a genuinely good core design: one recording site per invocation reusing the deferred hook every tool already passes through, a real nil-object pattern (`atomic.Pointer` defaulting to nil plus nil-receiver guards) so the disabled path costs one atomic load and no allocations, and tests that assert exact exposition text via `GatherAndCompare` and a golden fixture rather than substrings. `dur` is hoisted once so the log line and the histogram report the same measurement — an easy detail to get wrong.

I checked the PR body's "all methods are nil-safe" claim rather than taking it, and **it holds**: `Record`, `IncActive` and `DecActive` each guard `if t == nil` (`internal/observability/metrics/instruments.go:60,75,83`), and since `ToolMetrics` is a concrete pointer that is legal static dispatch, so the default `OBS_METRICS_ENABLED=false` path returns immediately. The `*broker` deref at both record sites also cannot panic (always `&brokerAlias` on a value variable), and the middleware installing at `main.go:1305` before `SetToolMetrics` at `:1353` is benign because the pointer is loaded per request. The `status` → `outcome` rename is complete across all product producers and consumers.

Three security findings I filed initially were put through adversarial challenge and **all three were reduced** — an unauthenticated-`/metrics` reconnaissance claim, an in-flight-gauge manipulation claim, and a SIEM-detection-break claim. What is left of them is documentation, not risk. Details in Challenged and Dropped; I have kept the measured evidence in the findings so you can see the risk was computed rather than asserted.

Nothing I found is blocking on its own. **The PR does remain blocked by @solace-aross's independent `CHANGES_REQUESTED`**, which I verified is still unaddressed at `4bb045c` — see the note below.

---

### On the prior reviews

@solace-aross's and Copilot's comments sit at `267ef831`. Head is now `4bb045c`, and the two are diverged (ahead 12 / behind 2) because the branch was rebased onto main — it picked up 10 unrelated commits, but the PR still carries only its own two. **The rebase moved the work; it did not address the findings.**

I re-checked the crux one in code: at `internal/tools/manager.go:210` `brokerAlias` is the raw `params["broker"]` string, canonicalization to `bc.DisplayName()` happens only at `:234-235` on the success path, and the error returns at `:220`/`:226` reach `logToolResult` with the raw value. @solace-aross's cardinality finding is **still live**. I have not re-filed it, and the four findings below deliberately avoid re-treading it or Copilot's two.

---

### Issues

#### 🟡 The `broker` label is keyed from two different identity sources
`internal/tools/manager.go:361` vs `:455` (flagged by `good:architecture`)

On the success path `brokerAlias` has been replaced by `bc.DisplayName()` (configured casing, `:234-235`) before `Record`. On the pre-resolution error paths it is still whatever the caller typed. So one logical broker emits `broker="Dev"` on success and `broker="dev"` on error.

This is **distinct from @solace-aross's cardinality comment and survives fixing it**: even with the label bounded, the numerator and denominator of a per-broker error rate land on different series, so `errors/total` grouped by `broker` is silently wrong for any caller who mistypes the casing. That is precisely the dashboard this PR exists to enable, which is why I am calling it out separately even though it lives in the same code region.

Filed at 🟢 by the lens; I am carrying it at 🟡 on the strength of that consequence, not re-rating it after challenge. Fixing it alongside @solace-aross's canonicalization is one change, not two: resolve the label through a single canonicalization step before recording on every branch, or record the resolved display name plus a sentinel when resolution failed.

#### 🟡 A test that cannot fail
`internal/tools/metrics_test.go` — `TestActiveRequestsMiddleware_PanicStillDecrements` (flagged by `good:code`)

The only assertion is `activeRequests(t, p) == 0`, and the `activeRequests` helper **returns 0 when the series is absent**. A middleware that incremented nothing, or a `SetToolMetrics` that stored nothing, passes this test identically — mutate `IncActive` to a no-op and it stays green. So the PR body's "a panicking request cannot leak a count" guarantee is not actually guarded.

Assert the gauge *exists* and reads 1 inside the panicking handler before it panics, as `TestActiveRequestsMiddleware_IncDecAroundRequest` does via its `during != 1` check; or have the helper `t.Fatal` on a missing series and return an explicit `found bool`. The `after` check in the IncDec test has the same latent weakness, mitigated there only by that `during` assertion.

#### 🟡 The reserved `cancelled` outcome has no structural forcing function
`internal/tools/manager.go:361, 455` (flagged by `good:architecture`)

Recording lives inside `logToolResult` — a function whose name, contract and doc comment are about emitting the audit line — and is duplicated once per branch. The `outcome` vocabulary documents a third value, `cancelled`, as reserved and wired by a later story. Nothing structural forces that story to add a third `Record` call, so the likely outcome is a cancelled invocation that logs but never counts, under-reporting the denominator of every SLO built on this metric. The same coupling means any future change to logging (a level gate, an early return on a non-audited path) silently drops metrics with it.

The single-site idea is right; the hosting function is wrong. Extract a `recordToolOutcome(...)` that computes `outcome`, `errorType` and `dur` once and fans out to both sinks with exactly one `Record` call — then a new outcome cannot reach one sink and not the other. This also subsumes the DRY duplication `good:code` flagged at the same lines.

#### 🟡 Cardinality is bounded by discipline, not by construction
`internal/observability/metrics/instruments.go:57-70` (flagged by `good:code` + `good:architecture`, independently)

`Record(ctx, tool, broker, outcome, errorType string, dur)` takes four adjacent bare `string`s, two of which are documented closed sets (3 and 11 values) from which the docs derive a hard bound of 13 series per tool/broker. Nothing at the type level or the recording boundary enforces that closure — any caller can mint a new series, and transposing `tool`/`broker` or typoing `"succes"` compiles cleanly. The docs themselves note CI enforcement is only planned.

`type Outcome string` / `type ErrorType string` with exported consts, taken by `Record`, which maps an unrecognised value to an `other` sentinel rather than passing it through. That makes the documented bound structural and gives the planned CI check something to assert against. It is also the cheapest available mitigation for @solace-aross's finding.

#### 🟡 The house logging template still teaches the old field name
`.claude/skills/add-logs/SKILL.md:76` (flagged by `docs` + `good:code`, independently)

The "Good log calls" example for the tool-invocation line still shows `slog.String("status", "success")`. This file is the template engineers and future `/add-logs` runs copy from, so anyone adding a new tool-invocation call site re-emits `status`, silently reintroducing the field this PR removes and reopening exactly the breakage the Breaking Changes note warns about. Not in the diff. One-word fix.

#### 🟢 The secure-logging spec no longer describes what the code emits
`docs/internal/secure-logging-rules.md:31`, mirrored in `.claude/skills/check-logs/SKILL.md:64` and `add-logs/SKILL.md:39`

The Always-Log table still lists `status` as required for "Tool invocation". `CLAUDE.md` and the PR template both make this file the authority for the mandatory `/check-logs` pre-commit gate, so left stale it makes that gate emit a **false L-01 violation against correct code** — and a security gate that cries wolf stops being read. No CI job enforces the table, so nothing breaks a build; this is spec hygiene.

I originally filed this as a broken SIEM detection. A challenger established that framing does not hold: `docs/observability.md:44-76` publishes this schema deliberately ahead of the code, tags Metrics and Audit as **[Planned]**, and openly solicits pilot feedback "up to the GA freeze" at version 0.8.0 with no GA release cut. The consumer is prospective. Worth widening the CHANGELOG line from "grepping or dashboarding" to include alerting/SIEM queries — one word, not a blocker.

#### 🟢 `/metrics` is unauthenticated and defaults to a wildcard bind, documented nowhere
`cmd/server/main.go:578-587`, `internal/defaults/defaults.go:32`

`DefaultMetricsBindAddress = ":9091"` — empty host, all interfaces — on a bare `ServeMux` with no auth and no TLS. Worth noting the asymmetry: the *MCP* listener deliberately defaults to loopback for its unauthenticated dev modes (`internal/config/config.go:683-690`, with a comment that dev modes must opt in to a routable bind). The metrics listener does not follow that convention, and nothing in `observability.md`, `configuration.md` or the threat model records the default or the missing auth. `deploy/kubernetes/` ships no NetworkPolicy.

This started as a 🟡 reconnaissance-surface claim and a challenger reduced it. The measured reasons: the new instruments have **no zero-value preregistration**, so a series exists only for tools actually invoked — it is not an inventory of everything the server exposes; and in this public repo all tool names are already published in `docs/tools-reference.md`, so the `tool` label discloses nothing an attacker cannot read on GitHub. Marginal disclosure reduces to broker aliases plus usage timing, on a listener that is absent unless `OBS_METRICS_ENABLED` is set, on a port no k8s manifest exposes. The pre-existing `mcp_build_info{version}` series is arguably more attack-useful.

A short Exposure note in `docs/observability.md` — unauthenticated, unencrypted, default `:9091` on all interfaces, restrict by NetworkPolicy or bind to loopback for a sidecar scraper — is worth taking regardless of magnitude, because the default is safe today by coincidence of gating rather than by design.

#### 🟢 `mcp_http_active_requests` does not mean what its doc says
`docs/observability.md:140,146` (mechanism at `internal/tools/metrics.go:30-37`, wiring at `cmd/server/main.go:1305`)

The gauge is offered as the signal separating a capacity problem from a tail-latency problem, and documented as `Labels: none`. But the middleware sits outside auth, the body limit and cross-origin protection, so it counts requests rejected 401/403/413 as in-flight; and despite the generic name it is wired on `/mcp` alone, so `/readyz`, `/metrics` and the OAuth discovery routes contribute nothing. Neither fact is written down.

This too began as a 🟡 — an unauthenticated client driving the gauge at will. A challenger refuted the availability half on a detail worth recording: the increment happens *inside* `ServeHTTP`, after headers are already read, so **held connections never enter the gauge**, and every pre-auth layer is header-only and non-blocking with both auth backends local. By Little's law a ~50 µs 401 path needs ~200k req/s to hold the gauge at 10 — a rate that is itself the genuine load event the gauge exists to report. Nothing automated consumes the series either: no HPA, no PrometheusRule, no dashboard in `deploy/`, and readiness is backed by `health.NewReadinessState()`.

One sentence on the bullet: covers `/mcp` only, and includes requests rejected before authentication, so a spike can mean rejected traffic rather than accepted work. An `authenticated`-class label would also work but trades against the "Cardinality: trivial" claim at `:150`, so the doc note is better value.

#### 🟢 `ToolMetrics()` is a factory named like a getter, with an operator-invisible degraded mode
`internal/observability/metrics/provider.go:185` (flagged by `good:code` + `good:architecture`)

Every call runs `NewToolMetrics` and re-registers three instruments, returning a fresh pointer; nothing memoizes it. It also deviates from the package's own convention — the other first-party instruments register once in `registerInstruments` at `New` time (`:113-153`), where a failure fails construction. Here failure surfaces later at `main.go:1353`, is logged and execution continues, so `/metrics` is up and healthy with the tool series simply absent, indistinguishable from "no traffic yet".

Register the tool instruments in `registerInstruments` and make `ToolMetrics()` a plain accessor; or memoize with `sync.Once`. If log-and-continue is deliberate, reflect it in the readiness surface so the gap is detectable.

#### 🟢 Log and metric disagree about "no broker" from one shared site
`internal/tools/manager.go:186`, `register.go:317`, `describe_semp_schema.go:339`

`list-brokers` and `describe-semp-schema` resolve no broker. The log line deliberately **omits** the field in that case (asserted by `register_test.go:307`, `describe_semp_schema_test.go:196`) while `Record` passes `*broker` unconditionally, so the metric emits `broker=""`. The empty-label documentation gap is Copilot's comment; this is the other half — the two sinks representing the same state differently from one call site. A `none` sentinel applied to both is dashboard-friendlier than `""`.

#### 🟢 Smaller items

- `internal/tools/metrics.go:24-27` — the recorder is a package-level `atomic.Pointer` mutated by an exported `SetToolMetrics`, with no owner, while `ToolManager` and the `register.go` closures are both constructed once at startup and would serve. `metrics_test.go` installs a real provider globally and asserts an exact count of 1, and `describe_semp_schema_test.go` already uses `t.Parallel()` — so a future parallel test that invokes a tool during that window inflates the count and the assertion flakes. Threading `*metrics.ToolMetrics` through `NewToolManager` would remove the global; at minimum make it unexported. (`good:architecture` 🟡 — carried here for length, not because it is minor.)
- `internal/tools/metrics.go:30-37` — an HTTP middleware in a domain package, which is what *forces* the global; in `internal/observability/metrics` it would simply take `*ToolMetrics` as a parameter. It also pulls `net/http` into `internal/tools`, whose only other middleware is an `mcp.Middleware`. (`good:architecture` + `good:code`)
- No explicit test asserts the metrics-disabled path, though every pre-existing `internal/tools` test exercises it incidentally with `toolMetrics` unset. Worth one intentional test naming the invariant.
- No exported symbol in `metrics.go` or `instruments.go` carries a doc comment — notable because the reason `SetToolMetrics(nil)` is legal is exactly the kind of invariant that belongs where the setter lives.
- `docs/superpowers/plans/2026-05-27-sol-149606-...md:372-387` — four worked `tool invoked` examples still show `"status":"success"`. Historical plan doc; a dated note is fine.
- `go.mod:33` — the only change is `godebug v1.1.0 // indirect`, a test-only transitive of `client_golang` and the OTel Prometheus exporter promoted by `go mod tidy`. MIT, FOSSA clean. Worth a line in the PR body saying why it appeared, and confirming whether `make refresh-third-party-inventory` needs to run.
- `mcp_http_active_requests` is unlabelled; adding a bounded `route` label now makes a later per-route gauge additive rather than breaking.
- The `status`→`outcome` rename ships **outside** the `OBS_METRICS_ENABLED` gate, so the flag does not bound the PR's blast radius — backing out the breaking half means reverting the whole change. Correctly disclosed; worth calling out at release time independently of whether operators enable metrics.

#### 🟢 PR body asserts what reviewers have contradicted

The Changes Made section claims docs were corrected for "the eleventh `error_type` value (`broker_permission_denied`)" and "the corrected cardinality (13 series per tool/broker)" — the two specific claims @solace-aross's comments dispute — and the checklist ticks Self-review completed. Worth either landing `broker_permission_denied` in code or marking those doc claims provisional until the open comments resolve, so the body and the review thread stop disagreeing.

---

### Challenged and Dropped

Every finding meeting the challenge threshold was challenged; all three were reduced.

| Original | Filed by | Verdict | Why it didn't stand |
|---|---|---|---|
| 🟡 `main.go:578-587` — unauthenticated `/metrics` is a "ready-made reconnaissance map" | good:security | → 🟢 | No zero-value preregistration, so series exist only for invoked tools; and every tool name is already published in this public repo's `docs/tools-reference.md`. Marginal disclosure is broker aliases plus timing, on a listener absent by default. |
| 🟡 `internal/tools/metrics.go:31-36` — unauthenticated clients can drive the in-flight gauge to arbitrary values | good:security | RECHARACTERIZE → 🟢 | The increment is inside `ServeHTTP`, after headers are read — held connections never enter the gauge. Every pre-auth layer is header-only and non-blocking, so ~200k req/s is needed to hold it at 10. No HPA, alert rule or dashboard consumes it. |
| 🟡 `secure-logging-rules.md` — the rename silently breaks SIEM detections | good:security | RECHARACTERIZE → 🟢 | Premise confirmed verbatim, consequence not: `observability.md:44-76` publishes the schema pre-freeze with Metrics/Audit **[Planned]** and solicits pilot feedback up to GA, at 0.8.0 with no GA cut. The consumer is prospective; the stale spec survives as gate hygiene. |

Reported **unchallenged** because the cap of three bound, listed so you can ask for a challenge on any: `internal/tools/manager.go:361,455` (reserved `cancelled` outcome), `internal/observability/metrics/instruments.go:57-70` (cardinality by discipline), `internal/observability/metrics/provider.go:185` (factory named as getter), `internal/tools/metrics.go:24-27` (package-level global).

---

### Strengths

- The nil-object design is the right call for an opt-in subsystem: `atomic.Pointer` plus three receiver guards, genuinely allocation-free when disabled, instead of a null-object interface or an `if enabled` at every call site.
- Hoisting `dur := time.Since(start)` once means the log line and the histogram report the *same* duration rather than two slightly different measurements.
- `error_type` is a closed vocabulary of in-code literals — no `err.Error()` text, no SEMP response body, no HTTP status message reaches a label. Internal error detail stays in the log `detail` field and never crosses onto the scrape surface. That is the correct call and it is what made the security review boring.
- Tests assert exact exposition text via `GatherAndCompare` and the golden fixture rather than substrings, so label-set drift fails loudly — the cardinality guarantee actually enforced, not just asserted in prose. `TestToolMetrics_ActiveRequestsIncDec` uses a real two-channel barrier across five goroutines with a comment explaining why the naive version would race.
- The `status`→`outcome` rename was carried through exhaustively across six test files, the golden log fixtures and an integration test in another module path, with a CHANGELOG migration note naming the affected consumers.
- Dependency direction is right and acyclic: `internal/tools` → `internal/observability/metrics`, which imports nothing from the domain layer.

---

### Summary Verdict

|---|---|---|
| Security | ✅ | good:security (re-examined: all three 🟡 reduced under challenge; only documentation gaps remain) |
| Documentation | ⚠️ | docs |

Not independently verified: the claimed 89.4% coverage figure — I did not run `make check`.

Nothing I found is blocking: no finding survived challenge at 🔴, and the two most alarming security claims were reduced to documentation gaps by measurement rather than by opinion.

To be unambiguous about what this verdict does and does not do: **it is not an unblock.** @solace-aross's `CHANGES_REQUESTED` stands, and I verified their canonicalization finding is still unaddressed at `4bb045c`. The highest-value next step is fixing that one, since the broker-label casing divergence above is the same code region and the type-safety change at `instruments.go:57-70` is the cheapest structural guard against the class.

solace-wmaryam:
All review feedback (from Copilot, @solace-aross, @bczoma, and the Jira notes on SOL-152086) has been addressed. There were a lot of items, so they're split into separate commits for easier review:

- `50a25bf` — **blocking fixes**: canonicalize the `broker` metric label on the error path; correct the `error_type` docs to the real closed set of **12** values (dropped `broker_permission_denied`, which isn't implemented; added the real `bad_request` and `not_found`).
- `9faaee5` — inject `ToolMetrics` (removes the package-level global; wired before the `/mcp` listener starts); typed `Outcome`/`ErrorType` consts with unknown→`other` coercion; align the log `broker` field to the `none` sentinel; move the gauge middleware into `buildMCPEndpoint`; the `_total` registration-string fix; strengthened panic test + added a metrics-disabled test.
- `3d3d2d8` — docs: house logging templates `status`→`outcome`; `/metrics` exposure note; `mcp_http_active_requests` semantics note; span-row `error_type` softened.
- `d63990a` — memoize `ToolMetrics()` so it registers once; tighten comments.

Two items intentionally deferred (non-blocking): the `cancelled`-outcome forcing function (that value is wired by a later story) and the `manager.go` defer-ordering coordination with Story 26 (not merged yet — whoever lands second rebases).

`make check` passes (coverage 89.6%).

solace-aross (approved):
Both blocking findings are fixed, and verified by mutation rather than by reading the diff:

- `canonicalBrokerLabel` (metrics.go) now resolves the metric's `broker` label to the configured display name, or a fixed `unknown`/`none` sentinel — never the raw caller string. Confirmed by reverting it to pass through the raw alias in a throwaway worktree: `TestCallTool_UnknownBrokerMetricLabelCanonicalized` went red immediately, on both assertions (missing sentinel, raw alias leaked). Restored cleanly.
- `docs/observability.md`'s `error_type` table is now the real 12-value set (`bad_request`, `not_found` added; the unimplemented `broker_permission_denied` dropped), and the cardinality section states the actual constraint — several error types only ever pair with `broker=none`/`unknown`, so it's correctly described as bounded rather than given a wrong precise number this time.

Also worth calling out, unprompted: this revision goes well past the two items I blocked on. It picks up bczoma's parallel review too — typed `Outcome`/`ErrorType` consts with an `other` coercion (cardinality bounded by construction, not discipline), the metrics recorder now injected through the constructor instead of a package-level global, the gauge middleware moved inside `buildMCPEndpoint` so the `OBS_METRICS_ENABLED` gate is visible from `main.go`, the provider built before the tool manager and the `/mcp` listener so there's no startup window recording into a nil-but-not-yet-wired recorder, `ToolMetrics()` memoized via `sync.Once`, and the panic-gauge test now asserts the series actually exists rather than defaulting to a vacuous pass. The one bczoma flagged and this PR explicitly defers — extracting a single `recordToolOutcome` site ahead of the `cancelled` outcome shipping — is disclosed and reasonable, since that value isn't wired yet.

CI green throughout. Approved.

bczoma (approved):
## PR Review: #379 — SOL-152086: Expose per-tool call-rate and latency metrics so teams can dashboard tool SLOs

**Author:** solace-wmaryam | **Changes since my last review:** +1497-line incremental diff across 27 files (4 commits)
**Follow-up review:** incremental diff since `4bb045c5` — parent-only pass, no lenses dispatched

---

### Verdict

**Recommendation:** ✅ Approve

*19 prior findings settled — 15 addressed, 4 accepted as-is (each one my own review offered as optional). No new findings.*

---

### Summary

Every finding from the last pass is addressed, and three of them are fixed better than I proposed. I verified each in code at `d63990a2` rather than from the commit messages, and ran the three touched packages — all green.

The two that mattered most are now structural rather than conventional. `canonicalBrokerLabel` (`internal/tools/metrics.go:41`) maps `""` → `none`, a resolvable alias → `DisplayName()`, and anything else → `unknown`, so the label domain is the configured aliases plus two sentinels — finite by construction. That closes my casing-divergence finding *and* @solace-aross's unbounded-cardinality finding with one function, and `recordToolInvocation` is now a single call site in the defer rather than one per branch, so a future `cancelled` outcome needs one change instead of two. Separately, `Record` coerces an out-of-set `errorType` to `ErrorTypeOther`, which makes the documented bound enforced rather than reviewed.

The `SetToolMetrics` global is gone outright — threaded through `NewToolManagerFromComposite`, `RegisterListBrokers`, `RegisterDescribeSempSchema` and `buildMCPEndpoint`, with `startMetricsEndpoint` split into build-then-serve so the recorder exists before anything is served. That also removed the reason the HTTP middleware needed to live in a domain package, and the build-failure path now registers a readiness listener, which closes the operator-invisible degraded mode I flagged alongside the memoization.

The panic-gauge test is fixed more thoroughly than I asked: `activeRequests` returns `(float64, bool)` so absence is representable, and the test reads it *inside the handler before the panic* and asserts `!seen || during != 1`. I checked whether the `after, _ :=` discard reintroduced the original weakness — it does not, because presence is already proven by that first assertion.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ parent (follow-up) — verified in code at `d63990a2`; `internal/tools`, `internal/observability/metrics` and `cmd/server` all pass |
| Code Quality | ✅ parent (follow-up) — the global removed by injection, typed `Outcome`/`ErrorType` with sentinel coercion |
| Test Coverage | ✅ parent (follow-up) — the vacuous panic assertion replaced with a presence-and-value check; a disabled-path test added |
| Documentation | ✅ parent (follow-up) — `/metrics` exposure note, gauge scope and pre-auth caveat, secure-logging Always-Log row, and both skill templates all corrected |
| Architecture | ✅ parent (follow-up) — dependency injection replaces process-global state; recording collapsed to one site |
| Security | — not re-reviewed (no security surface in the push; the exposure finding was a doc gap and is now documented) |
| PR Scope & Description | ✅ parent (follow-up) — the `error_type` set corrected to the real twelve and `broker_permission_denied` dropped |

Not independently verified: the coverage-gate percentage.

PR #380, opened 2026-09-04
SOL-154037: count recovered panics and tag the tool-path panic log

solace-aross:
## What this does

A recovered panic is invisible outside the process today. This adds the counter that makes it alertable, and the log attribute that makes the existing log-based alert cover both request paths.

`mcp_panic_recovered_total{boundary}` — one counter, registered against the existing meter provider (SOL-152091), incremented at both recovery nets that guard a request's own goroutine:

| boundary | Site | What the caller gets (unchanged) |
|---|---|---|
| `http` | `recovery.HTTPMiddleware` | clean HTTP 500 |
| `tool` | `withRecovery` in `internal/tools` | `isError: true`, generic message, `retryable: false` |

The tool-path panic log now carries `event="panic_recovered"`, matching `recovery/middleware.go`, `safego`, `tokenexchange`, and `tracing/selfstats`. Before this, an alert keyed on that attribute silently missed every tool panic.

## Why it exists

SOL-151286 (Story 12) and SOL-151287 (Story 13) shipped the two recovery paths and were both closed with a panic counter in their acceptance criteria. Neither counter was built. Found during the SOL-150251 alignment audit, not from an incident — but it is exactly the gap that would slow one down.

## Both series are published at zero from startup

This is the part worth reviewing closely. An OTel counter renders no series until it has a data point, and PromQL's `increase()` needs two samples in the window, so the sample that *creates* a series is only a baseline. Without seeding, `increase(mcp_panic_recovered_total[5m]) > 0` — the alert the schema prescribes — would not have fired on a process's first panic. For a bug that reached production that is the modal case, not an edge case.

`Register` now records zero on both boundaries before returning. Three consequences:

- The first panic fires the alert.
- A flat zero reads as "nothing panicked" instead of "No data", which a dashboard renders the same as "metrics misconfigured".
- `absent(mcp_panic_recovered_total)` becomes a usable "metrics are on but the counter was never wired" alert.

## What the counter does not cover

Only panics on a request's own goroutine. A panic on a goroutine a handler *spawns* is recovered elsewhere and is not counted, even though it happened during a request:

| Site | Where it fires |
|---|---|
| `internal/safego` | SEMP fan-out workers in the broker-status, queue-metrics and discard-stats handlers, and in the composite executor |
| `internal/tokenexchange` | The singleflight goroutine performing the IdP round trip |
| `internal/observability/tracing` | The periodic OTLP self-stats loop (genuinely off the request path) |

All three convert the panic into an error that returns through the handler normally, so `withRecovery` never sees it. They all still log `event="panic_recovered"`, so the log attribute has wider reach than the metric. `docs/observability.md` names the sites and says so, rather than leaving an operator to work it out from a flat graph. Widening the counter to cover them is a deliberate follow-up, not something to do by accident.

## Telemetry only

No change to what either caller receives. An architecture review (2026-09-04) confirmed both shipped shapes are correct: tool-path panics return `CallToolResult{IsError:true}` over a successful MCP call per the spec's application-error convention, and HTTP-boundary panics return a clean 500. Both are now **pinned by tests that assert the response** rather than discarding it — `TestPanicAuditedAsError` previously threw the result away and checked only the audit line.

Both increments run **after** the caller's response is produced. `recover()` has already consumed the original panic, and the SDK runs tool handlers on a goroutine with no `recover()` of its own, so a second panic raised in that deferred function would kill the process — the exact failure `withRecovery` exists to prevent. Nothing in `panics.Recovered*` or OTel's `Add` can panic (traced through `sdk/metric` to the atomic add), but ordering the record last means a telemetry fault could only ever cost a metric.

`http.ErrAbortHandler` stays exempt on the `http` boundary, from the log and the counter alike. `net/http` uses it as a client-disconnect sentinel and the MCP SSE path raises it on ordinary teardown, so counting it would fire a panic alert on routine traffic.

## Design notes

**One metric with a label, not two names.** A single alert covers both nets.

**Cardinality is closed by the API.** `internal/observability/panics` exposes `RecoveredHTTP(ctx)` and `RecoveredTool(ctx)` and no way to pass an arbitrary boundary, so there is no exported var to reassign and no zero value to guard against. Adding a boundary means adding a function there and a row in `docs/observability.md`.

**`Register` takes the `metric.MeterProvider` API interface**, not the concrete SDK type, so the outermost panic net does not pull the OTel SDK into its import graph.

**The counter is process state, not a parameter.** Injecting it would add a telemetry argument to `tools.RegisterWithServer` — a security-relevant signature already carrying the authorization policy and the write gate, with 20+ call sites — and a second parameter to the one-argument `HTTPMiddleware` decorator. Both sites would still need a nil branch for the metrics-disabled case, which is exactly the no-op the package already provides. Both already reach for process-global telemetry (`slog.Default()`) for the same reason. The reasoning is written down in the package doc.

**Off is off.** With `OBS_METRICS_ENABLED=false` nothing is registered and both increments are no-ops. Recovery itself stays unconditional and never depends on this package.

## Testing

- `make check` green: build, vet, lint, race-enabled tests, total coverage 89.7% against the 85% floor.
- Golden fixture regenerated. It now pins both seeded series at zero, which is exactly what a healthy process exposes.
- **Verified by defect injection**, not by reading. Each mutation applied with `go test -overlay` so the tree was never modified:

  | Injected defect | Result |
  |---|---|
  | Remove the zero-seeding from `Register` | `TestRegister_SeedsBothSeriesAtZero` and `TestGoldenSchema` fail |
  | Remove the `boundary="tool"` increment | `TestPanicIncrementsPanicCounter` fails |
  | Remove the `boundary="http"` increment | `TestHTTPMiddleware_IncrementsPanicCounter` fails |
  | Flip `IsError` to false | `TestPanicAuditedAsError` fails |
  | Move the increment above the `ErrAbortHandler` re-raise | `TestHTTPMiddleware_ErrAbortHandlerNotCounted` fails |
  | Drop the `event` attribute | `TestPanicAuditedAsError` fails |

- Two independent Opus review passes before and after opening this PR, the second briefed to argue the strongest case against merging. The zero-seeding gap and the request-path coverage claim both came out of that second pass and are fixed in the follow-up commit.

Also corrects a stale doc comment in `register.go`: go-sdk v1.5.0 → v1.7.0, matching `go.mod`.

Closes SOL-154037. Closes the outstanding acceptance criteria on SOL-151286 and SOL-151287.


internal/middleware/recovery/middleware_test.go
```diff
@@ -305,3 +310,113 @@ func TestHTTPMiddleware_CleanPathWrites500(t *testing.T) {
...
+// panicked.
+func TestHTTPMiddleware_NoPanicNoCount(t *testing.T) {
+	reader := newPanicCounterReader(t)
+
+	_ = drive(HTTPMiddleware(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})))
+
+	got := panicCountsByBoundary(t, reader)
+	if got["http"] != 0 || got["tool"] != 0 {
```
bczoma, line 352: 🟡 **Two counter tests pass when the instrument does not exist**

Flagged by `good:code`.

`TestHTTPMiddleware_NoPanicNoCount` here and `TestHTTPMiddleware_ErrAbortHandlerNotCounted` at :419 both assert only zeros — `got["http"] != 0 || got["tool"] != 0`. But `panicCountsByBoundary`'s own doc comment says an absent metric yields an empty map, and an empty map returns 0 for every key. So both tests stay green if the instrument disappears entirely: if the metric is renamed, if `Register` stops seeding, or if the increment site is deleted.

These are the two tests guarding the *does not over-count* half of the behaviour — the half that matters for the `ErrAbortHandler` exemption, since counting a client disconnect would page someone on routine SSE teardown. As written they cannot distinguish "recorded zero" from "nothing exists", so exactly the regression they exist to catch passes.

The fix is already written in this PR. `panics_test.go:82` gets it right with a presence check before the zero assertions:

```go
if len(got) != 2 {
    t.Fatalf("counts = %v, want both boundaries present", got)
}
```

Add that to both tests here, and to the pre-call assertion in `internal/tools/register_test.go:420`, which has the same shape. Alternatively make the helper return `(map[string]int64, bool)` so absence becomes representable rather than indistinguishable from zero.

The positive-increment tests are fine — they assert `== 1` against a seeded-zero baseline, so those cannot pass vacuously.

internal/observability/panics/panics_test.go
```diff
@@ -0,0 +1,156 @@
...
+	t.Cleanup(func() { counter.Store(nil) })
+	if err := Register(mp); err != nil {
+		t.Fatalf("Register() error = %v", err)
+	}
+	return reader
+}
+
+// countsByBoundary collects mcp.panic.recovered and returns its value per
+// boundary label. An absent metric yields an empty map.
+func countsByBoundary(t *testing.T, reader *sdkmetric.ManualReader) map[string]int64 {
```
bczoma, line 50: 🟡 **The counts helper is copied three times and has already diverged**

Flagged by `good:code`.

This helper exists in three places: here, `internal/middleware/recovery/middleware_test.go:337-363`, and `internal/tools/register_test.go`. The middleware and register copies are byte-identical at ~28 lines each; this one differs only in fataling on a missing `boundary` attribute. The reader-install helper is duplicated twice more (`newTestReader` / `newPanicCounterReader`) plus a third inline copy in `TestPanicIncrementsPanicCounter` — roughly 90 lines of duplicated scaffolding across three packages.

The divergence is not hypothetical, it has already happened, and it is the reason the sibling finding on `middleware_test.go:400` exists. Two copies do `v, _ := dp.Attributes.Value("boundary")` and silently bucket an attribute-less point under `""`; this copy fails loudly instead. That is the same class of bug in the helper that the tests using it are supposed to catch in the code.

Extract a test-only helper package — e.g. `internal/observability/panics/panicstest` exposing `InstallReader(t) *sdkmetric.ManualReader` and `Counts(t, reader) map[string]int64` — and have all three call sites use it. That also gives one place to put the presence check from the other thread, rather than three.

Worth doing now rather than later: any future change to the metric name or the `boundary` label currently has to land in three files, and the copy that gets missed fails silently by construction.

docs/observability.md
```diff
@@ -45,7 +45,7 @@ capability headings carry the same tag:
 | Capability | Status | Notes |
 |---|---|---|
 | Correlation ID | **[Implemented]** | Wired and on by default (`OBS_CORRELATION_ID_ENABLED`). |
-| Metrics | **[Planned]** | The `/metrics` endpoint and instruments are not yet wired; the names and labels here are the proposal under review. |
+| Metrics | **[Planned, with exceptions]** | Most instrument names and labels here are still the proposal under review. Wired and emitted today: the `/metrics` endpoint itself, `mcp_build_info`, `mcp_schema_version`, `mcp_metrics_scrape_total`, the OTLP export-health counters, and `mcp_panic_recovered_total` (see [Panic Recovery](#panic-recovery--implemented)). Assume any other metric below is not yet emitted. |
```
bczoma: 🟡 **This row contradicts the `## Metrics` section 60 lines below**

Flagged by `docs`.

This row now reads `[Planned, with exceptions]` and lists `mcp_panic_recovered_total` among the instruments "Wired and emitted today". The `## Metrics — [Planned]` heading at :110 and its blockquote at :112-113 — untouched by this PR — still state flatly:

> _Status: **[Planned]**. The `/metrics` endpoint and instruments are not yet wired in the build._

Same capability, same file, opposite claims.

An operator who reads past this summary table into the `## Metrics` section itself — the more detailed and presumably authoritative part — will conclude nothing on `/metrics` is real yet, including this counter, and skip wiring the panic alert this PR exists to enable. That is a worse outcome than the doc simply being terse, because the contradiction resolves the wrong way for anyone who reads further.

Update the `## Metrics` heading and blockquote to `[Planned, with exceptions]` and carry the same "Wired and emitted today: …" sentence, or point the blockquote back at this row as the authority. Everything else in the Panic Recovery section is accurate and complete — the three uncounted sites are named in a table with their practical consequence stated, the zero-seeding is explained with the PromQL reason and both derived alert idioms, the `ErrAbortHandler` exemption is documented, and the cardinality of 2 is correct.

bczoma (dismissed):
## PR Review: #380 — SOL-154037: count recovered panics and tag the tool-path panic log

**Author:** solace-aross | **Branch:** `aross/sol-154037-panic-counter` → `main` | **Changes:** +733 / -11 across 11 files
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-380`
**Jira:** `SOL-154037` — all nine acceptance criteria trace to a named change or test in the diff

---

### Verdict

1. 🟡 `internal/middleware/recovery/middleware_test.go:400` — two counter tests pass when the instrument does not exist; add the presence check the sibling test already has
2. 🟡 `internal/observability/panics/panics_test.go:42` — the counts helper is copied three times and has already diverged; extract it
3. 🟡 `docs/observability.md:48` — the new status row contradicts the `## Metrics — [Planned]` section 60 lines below

---

### Summary

Telemetry-only, and unusually careful about it: the load-bearing safety claims in the description are all true in the code, which is not the norm. Both increments really are the last statement in their deferred recovery function, `record` really is a no-op before `Register`, the zero-value `boundary` really is unrepresentable because `boundaryHTTP`/`boundaryTool` and `record` are all unexported, and `internal/middleware/recovery`'s import graph really does stay clear of the OTel SDK — `go list -deps` shows only `otel/metric`, `otel/metric/embedded` and `otel/attribute`. The zero-seeding is pinned in three independent places, and the startup reorder is genuinely inert: `startServer` only launches a goroutine and returns a channel, and `SetInitialized` still runs after both registrations.

One architectural finding was filed and reduced by measurement rather than opinion: the claim that the counter misses the likeliest panic class turned out to invert on inspection — the deref-heavy curation code (`Curated()`, `assembleEnvelope`) runs on the *counted* request goroutine after `g.Wait()`, while the `safego` workers only do error-returning typed decode, and the composite executor's one named panic class is already recovered locally in `safeTemplateExecute`. What survived is the 🟢 on the HELP wording below.

---

### Minor notes

- 🟢 `internal/observability/panics/panics.go:111` — the description renders as the metric's HELP line and reads "Panics trapped by a request-path recovery net", but `internal/safego` and `internal/tokenexchange` *are* request-path nets by this package's own doc and are not counted. The HELP string is the only description an operator sees when writing the `increase() > 0` rule the docs prescribe; the caveat lives in the package doc and `docs/observability.md:313-321`, neither of which is in that path. Narrow it to "trapped on a request's own goroutine … spawned-goroutine panics are logged but not counted".
- 🟢 `internal/observability/metrics/provider_test.go:88-90` — `TestGoldenSchema` calls `panics.Register(...)`, mutating another package's process global with no `t.Cleanup` and no way to unregister. Under `-shuffle` a later metrics test sees a counter bound to this test's provider, and the provider's own contract test now imports a consumer of the provider.
- 🟢 `internal/observability/panics/panics.go:87` — the instrument is a package-level `atomic.Pointer` with no `sync.Once`, no unregister and last-writer-wins. Dependent packages cannot reset it, so `middleware_test.go:315-325` deliberately leaks a live `ManualReader` provider to stay correct, and the no-parallel constraint is a comment. A `sync.Once` or a test-only reset seam would remove the need to leak. (`good:architecture` + `good:code`)
- 🟢 `internal/tools/register.go:116-123` — `panics.RecoveredTool(ctx)` sits before `stampCorrelationID`, while the comment says a telemetry fault "could only ever cost a metric". It would also skip the correlation stamp. Moving the increment to the last statement makes the comment literally true; the HTTP site already does it that way.
- 🟢 `docs/observability.md` (Panic Recovery section) — worth one line saying this series is externally influenceable. Unlike every other metric emitted today, which is static or self-referential, a panic counter gives a scraper a one-bit-per-boundary oracle for "did my input crash a handler". It needs scrape reachability to matter, so this is a doc note rather than a code change.
- 🟢 `internal/middleware/recovery/middleware.go:95` — the `http.ErrAbortHandler` re-raise is an identity compare placed before both the log and the counter, which is right for client disconnects. But a genuine bug that panics with or re-panics that exact sentinel is then invisible in *both* detection channels. A DEBUG line before the re-raise would make it greppable without touching the paging signal.
- 🟢 `internal/observability/panics/panics.go:100-107` — the `meterProvider == nil` guard cannot catch a typed nil, which is the shape the only production caller can produce. Not reachable today, since `metrics.New` cannot return a `Provider` with a nil `meterProvider`, but the guard advertises protection it does not give at exactly the seam where the interface meets a concrete pointer. (`good:code` + `good:architecture`)
- 🟢 `internal/observability/panics/panics.go:83` — `atomic.Pointer[metric.Int64Counter]` stores a pointer to an interface, forcing `counter.Store(&c)` and `(*c).Add(...)`. A one-field struct would read better.

…and 6 further minor items — comment-to-code ratio in the new package (~85 lines of prose defending design choices against alternatives, which will drift as `safego`/`tokenexchange` change), the cancelled-context test pinning SDK rather than local behaviour, `Register` failure being visible only in a log while `startMetricsEndpoint` degrades readiness, the tool path having no `ErrAbortHandler` exemption, no "Testing" heading in the PR body, and Copilot's approval covering only `737a681` and not the zero-seeding commit. Ask if you want them written up.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

Not independently verified: I did not run `make check`, so the coverage gate is unconfirmed.

---

### Housekeeping

Cleanup skipped this pass — it ran earlier in this session and nothing has merged or closed since, so it would rediscover nothing.

internal/middleware/recovery/middleware_test.go
solace-aross: Added the presence check: both zero-count tests now `t.Fatalf` if `len(got) != 2` before asserting the zero values. Addressed in bae927f.

internal/observability/panics/panics_test.go
solace-aross: Extracted internal/observability/panics/panicstest (InstallReader + Counts, with the loud presence check) and moved the middleware and tools copies onto it. This package's own panics_test.go keeps its local copy rather than importing panicstest: that file is package panics (white-box, needs the unexported counter), and panicstest imports panics, so importing it back is a real cycle, not a style call — noted in both files' comments. Addressed in bae927f.

docs/observability.md
solace-aross: Updated the ## Metrics heading and blockquote to [Planned, with exceptions], carrying the same "Wired and emitted today" sentence as the summary row. Also updated both #metrics--planned anchor references elsewhere in the doc, since the heading text change shifts the auto-generated anchor. Addressed in bae927f.

sweep-notes.md
```diff
@@ -0,0 +1,116 @@
+# andrea-sweep session notes — solace-broker-mcp
```
bczoma: 🟡 **An agent-session scratch file is riding along in a PR about panic counters**

Flagged by the parent review.

`sweep-notes.md` is 116 lines of `/andrea-sweep` session notes — the phase-by-phase log of a code-sweep run: Jira project and custom-field configuration, the worktrees created per ticket, which findings the scan subagents were told to skip, the `.ua/` knowledge-graph staleness check. It has nothing to do with SOL-154037, and it landed in `fd75a5f`, the commit that adds the panic counter. It is not on `main`, and it was not present at the head I reviewed last pass (`5189095`), so it arrived with the force-push rather than being something the earlier review missed.

**Why it is worth a thread and not a 🟢.** It is a new file at the repository root of a public repo, so it lands in the tree permanently and ships in the release tarball, and the root is where a reader looks for files that describe the project. Nothing references it, nothing ignores it, and no reader six months from now will be able to tell whether it is a deliberate artifact or debris. Removing it is `git rm sweep-notes.md`; leaving it is a decision nobody made.

**What I checked before writing this, because my first read overstated it.** My initial concern was that it publishes internal security work — it lists ticket keys with one-line topics like "broker URL leaking into logs", "token-exchange audience trust", "SEMPError raw-body logging". That framing does not survive checking: every ticket key in the file except the parent epic `SOL-153075` is already cited by key in this repo's own public `CHANGELOG.md`, and `andrea.ross@solace.com` is already in the public git history as a commit-author email on `main`. So this is not an exposure, and I am not asking for a history rewrite or anything beyond deleting the file from the branch.

What genuinely is new to the public tree is small and still pointless to publish: the epic key, the Jira custom-field ids and option ids, the internal Slack channel, and the local worktree/branch layout of someone's machine.

**Recommended fix.** `git rm sweep-notes.md` and force-push. If the notes are worth keeping, they belong outside the repo, or under a path the repo already ignores. Worth a quick look at whether the sweep tooling writes this file into the working tree by default — if it does, adding it to `.gitignore` stops the next one from being committed by accident.

docs/observability.md
```diff
@@ -107,10 +107,14 @@ avoid. Pin dashboards to `mcp_schema_version` and SIEM queries to `audit_schema_
 
 ---
 
-## Metrics — [Planned]
+## Metrics — [Planned, with exceptions]
 
-> _Status: **[Planned]**. The `/metrics` endpoint and instruments are not yet wired in the
-> build; the following names, types, and labels are the proposal under review._
+> _Status: **[Planned, with exceptions]**. Most instrument names, types, and labels below
+> are the proposal under review, not yet wired in the build. Wired and emitted today: the
```
bczoma, line 113: 🟡 **The new "wired today" whitelist omits three instruments the build actually emits**

Flagged by `docs`.

Fixing last pass's contradiction replaced a blanket "nothing is wired" with an explicit list, here and in the status row at `:48`. That is the right move, and it makes the list load-bearing: both places now close with *"Assume any other metric below is not yet emitted."* An operator is being told to trust the list as the authoritative set of live instruments.

Three emitted instruments are missing from it:

- `mcp_tool_invocation_total`
- `mcp_tool_invocation_duration_seconds`
- `mcp_http_active_requests`

All three are constructed in `internal/observability/metrics/instruments.go` and wired in production in `cmd/server/main.go`; the first two arrived with #379, which merged into `main` after this branch was cut. The golden fixture in this very PR contradicts the doc directly — `internal/observability/metrics/testdata/metrics_golden.txt:4-6` pins `mcp_http_active_requests` as emitted, and the tool-invocation pair is there too. `mcp_http_active_requests` also sits in the same table at `:144` as three of the health metrics the whitelist *does* name, which makes the omission read as a deliberate exclusion rather than an oversight.

**Why it matters, in this PR specifically.** The argument for seeding both panic series at zero — the part of this PR I would defend hardest — is that an alert which does not exist is the harm, and that "No data" and "nothing happened" must not look the same to an operator. The same harm applies here one level up: an operator who reads this list and believes it skips the RED dashboard and the error-rate/latency alerts on `mcp_tool_invocation_*`, and skips the saturation panel on `mcp_http_active_requests`, for signals that exist in the build today.

**Recommended fix.** Add the three names to both lists (`:48` and `:113-116`). Deriving the list from `metrics_golden.txt` is what keeps it honest as instruments land — that fixture is already the pinned record of what a healthy process exposes, so a whitelist that disagrees with it is a test-visible defect rather than a judgement call: `mcp_build_info`, `mcp_http_active_requests`, `mcp_metrics_scrape_total`, `mcp_panic_recovered_total`, `mcp_schema_version`, `mcp_tool_invocation_total`, `mcp_tool_invocation_duration_seconds`, plus the tracing-gated OTLP export-health pair.

This is a fair thing to hand to whoever rebases onto #379 rather than fixing blind, since the rebase is where the two sets meet.

bczoma (dismissed):
## PR Review: #380 — SOL-154037: count recovered panics and tag the tool-path panic log

**Author:** solace-aross | **Branch:** `aross/sol-154037-panic-counter` → `main` | **Changes:** +841 / -9 across 13 files
**Jira:** `SOL-154037` — closes the outstanding panic-counter ACs on SOL-151286 and SOL-151287

---

### Verdict

All three 🟡 from my previous pass are addressed, and addressed at the seam each one asked for rather than worked around. Two new 🟡, neither a showstopper.

1. 🟡 `sweep-notes.md:1` — a 116-line agent-session scratch file at the repo root, unrelated to this ticket, added by `fd75a5f`; delete it from the branch
2. 🟡 `docs/observability.md:113` — the new "Wired and emitted today" whitelist omits three instruments the build does emit, under a sentence telling operators to assume everything unlisted is not emitted

---

### Summary

The response to the previous review is the good kind: the presence check went in as a `t.Fatalf` in all three places it was asked for including the easily-missed pre-call assertion in `register_test.go`, and the helper extraction deleted the old copies rather than adding a fourth — the version that survived is the strictest of the three, so the fix also permanently closes the divergence the finding described. `panics.Register` was moved so that its comment's "no window" claim became true, rather than the comment being softened to match the old position; I verified the ordering in `main()` and the claim is accurate as written. The one item not changed came back with a checkable engineering argument, and it holds. Both new findings are hygiene rather than behaviour, and one of my own framings was reduced by measurement before it reached this review.

*11 prior findings settled — 3 addressed, 8 accepted as-is.*

---

### Minor notes

- 🟢 `internal/observability/panics/panicstest/panicstest.go:73` — `Counts` still returns an empty map for an absent metric and delegates "is anything there at all" to each caller via a doc comment. That is the exact trap that produced last pass's 🟡; the new shared API preserves it in prose instead of closing it. Fatal inside `Counts` when the metric is absent (it is seeded at registration, so absence is always a defect), or add a `CountsBothBoundaries` helper that fatals unless `len == 2`.
- 🟢 `internal/observability/panics/panicstest/panicstest.go:55` — `InstallReader` names only its return value; its primary effect is registering the process-level counter, which is why the caller cannot be `t.Parallel()`. `InstallCounter` (returning the reader) would put that in the name instead of six lines of comment.
- 🟢 `docs/observability.md:1008` — "Until a session gauge exists (metrics are `[Planned]` above…)" is the one surviving use of the bare `[Planned]` tag this PR retired for the Metrics capability, so grepping the tag now gives an inconsistent answer.
- 🟢 `docs/observability.md:55` — the legend still says "Only capabilities tagged **[Implemented]** … are live today", but Metrics is now `[Planned, with exceptions]` and contains a subsection tagged `### Panic Recovery — [Implemented]`. The legend no longer describes its own table; naming the partly-live state would fix it.
- 🟢 `docs/observability.md:292-346` — the Panic Recovery section still has no note that this series is externally influenceable (a scrape reader gets a one-bit-per-boundary oracle for "did my input crash a handler"). Offered as optional last pass and reasonably left; repeating it once here only because a 🟢 gets no thread and this is the only record it will have.

---

### Summary Verdict

|---|---|---|
| Architecture | — not re-reviewed (unchanged since `5189095d`; the incremental diff is tests and docs only) | — |
| Security | — not re-reviewed (unchanged since `5189095d`; the incremental diff is tests and docs only) | — |
| Documentation | ⚠️ | docs |
| PR Scope & Description | ⚠️ | parent (follow-up) |

Scope note: this pass reviewed `60b88f9..bae927f` plus the two other changes since my prior review — the `cmd/server/main.go` registration move and the added `sweep-notes.md`. Everything else in the PR is byte-identical to what the previous review examined, verified by comparing the PR's cumulative diff then against now; it was not re-reviewed and the marks above do not re-assert it.

sweep-notes.md
solace-aross: Removed (git rm) and added sweep-notes.md to .gitignore alongside watch-my-prs-notes.md/watch-for-prs-notes.md — same class of session scratch, same fix. Thanks for checking whether it was an exposure before flagging it as one; agreed it wasn't, but it didn't belong here either way. Addressed in c2272dd.

docs/observability.md
solace-aross: Added all three to both lists (:48 and the Metrics blockquote). Good catch that the golden fixture already pinned them — should have diffed the whitelist against metrics_golden.txt instead of writing it from memory. Addressed in c2272dd.

bczoma (approved):
## PR Review: #380 — SOL-154037: count recovered panics and tag the tool-path panic log

**Author:** solace-aross | **Branch:** `aross/sol-154037-panic-counter` → `main` | **Changes:** +734 / -9 across 13 files
**Jira:** `SOL-154037` — closes the outstanding panic-counter ACs on SOL-151286 and SOL-151287

---

### Verdict

**Recommendation:** ✅ Approve

Both 🟡 from the previous pass are addressed, and nothing new. Follow-up pass over `bae927f..c2272dd` (3 files, +9/−118); no lenses dispatched.

*7 prior findings settled — 2 addressed, 4 accepted as-is, 1 declined.*

---

### Summary

The stray `sweep-notes.md` is gone from the branch, and the fix went one better than the finding asked: rather than only deleting it, the file is now in `.gitignore` with a comment recording what it is and how it got committed, so the next `/andrea-sweep` run cannot repeat the accident. The metrics whitelist is complete — I checked all seven instruments in `metrics_golden.txt` against both lists (`docs/observability.md:48` and `:112-117`) and every one is now named, which was the point of deriving the list from the fixture rather than by hand. `go build ./...` is clean at this head.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ✅ (parent, follow-up) |
| PR Scope & Description | ✅ (parent, follow-up) |
| Correctness | — not re-reviewed (unchanged since `bae927f1`) |
| Code Quality | — not re-reviewed (unchanged since `bae927f1`) |
| Test Coverage | — not re-reviewed (unchanged since `bae927f1`) |
| Architecture | — not re-reviewed (unchanged since `bae927f1`) |
| Security | — not re-reviewed (unchanged since `bae927f1`) |

The five categories above marked not re-reviewed were examined at `5189095d` and `bae927f1`; this push touched only `.gitignore` and `docs/observability.md`, so those marks are not re-asserted here.

PR #381, opened 2026-09-04
SOL-152090 / SOL-152096: enforce the audit-record schema, and emit one for every destructive tool call (Stories 21 and 22)

solace-aross:
Two stories, one logical change: the audit-record schema, and the first thing that emits one.

Kept as a single PR because Story 21 emits nothing on its own. Reviewing the constructor without its first caller means reviewing a table of rules against no evidence they are the right rules. Two commits, in order, so each story is still reviewable on its own.

- **SOL-152090 (Story 21)** — https://sol-jira.atlassian.net/browse/SOL-152090
- **SOL-152096 (Story 22)** — https://sol-jira.atlassian.net/browse/SOL-152096

## What changes

`docs/observability.md` has described an audit-record schema whose field applicability varies by record type. Nothing enforced it, and three emission sites are due to land against that description. Each would have had to reproduce the rules correctly from prose.

**Story 21** makes the schema executable. `audit.NewEvent` is the single constructor. It validates the field-applicability table and returns an error rather than a partial record:

| Rule | Why |
|---|---|
| `error_type` only on `operation`, only when `outcome` is `error` | a success cannot carry a cause; a failure cannot omit one |
| `reason` only on the three failure/denial types, each against its own closed vocabulary | an authentication reason on an authorization record corrupts the query a reviewer runs |
| `arguments_hash` never on a denial | the call was refused before its arguments were dispatched |
| `audit_drop` carries the common fields and nothing else | it reports that a record is missing; there is none to describe |

Identity comes from the request principal (SOL-152087), never from the caller, so a record cannot name someone other than the authenticated caller. Only `principal.sub` and `agent_client_id` are written (Q-013).

`audit.HashArgs` computes `arguments_hash` as a hex SHA-256 over the RFC 8785 (JCS) form of the arguments, via `github.com/gowebpki/jcs` (Apache-2.0). Using a published canonicalisation spec rather than sorting keys ourselves is the whole point: number serialisation, Unicode handling and recursive key ordering are where a home-grown canonicaliser diverges silently, producing a digest the customer's own tooling cannot reproduce and our tests would not catch.

**Story 22** replaces the bare destructive-operation WARN. With `OBS_AUDIT_LOG_ENABLED` on, a destructive call emits one `operation` record at completion carrying the identity chain, the arguments hash, the outcome and the duration. With the flag off, the WARN is emitted exactly as before and no record is written; the golden test on that line is unchanged and still passes.

Drop detection ships with it. An audit write that does not land produces an `audit_drop` record that keeps `event="audit"`, so the one record saying audit records are missing stays inside the filter customers route the audit stream on.

## Three schema corrections

All of them documentation catching up to code. No emitted record changes.

1. **`audit_event_type` is seven, not six** — `broker_authz_denied` for a hop-2 broker denial. Its emission site ships under SOL-153332; the constructor accepts it now so a SIEM rule can be written against it.
2. **`error_type` is twelve, not the ten `docs/observability.md` listed.** Four call sites reach `logToolResult`, not one. `bad_request` and `not_found` come from `list-brokers`, `describe-semp-schema` and the argument-parsing guard ahead of every tool. Counts taken from the main dispatch path alone missed them, twice.
3. **The field-applicability table gains its `broker_authz_denied` row.**

`internal/tools/audit_error_type_drift_test.go` now derives the `error_type` set from the emit sites' AST and fails if it ever diverges from the vocabulary the constructor accepts. The next addition breaks the build instead of the audit trail.

## What "one record per call" does and does not mean

It bounds the `operation` record type, not the total. A call the broker denies at hop 2 will produce its `operation` record **and** a `broker_authz_denied` record once SOL-153332 lands, and both are correct. The tests count records with `audit_event_type="operation"` rather than asserting a total, so that landing does not break them, and `docs/observability.md` now tells SIEM authors the same thing.

## Review already applied

An independent principal-level review ran over both commits before this PR opened. It found four things worth fixing, all in the third commit:

1. **The audit trail vanished silently at `log_level: error`.** An `operation` record is `INFO` and the drop notice reporting its suppression was `WARN`, so at the highest configurable level both were filtered out: no records, no notice, nothing. The drop notice is now `ERROR`, so it survives every level an operator can set.
2. **The `error_type` drift guard had three holes** — it read only inline literals (a named constant slipped a thirteenth value past it), silently skipped shapes it could not parse, and could lose sight of `register.go` entirely without noticing, since all three of that file's values are produced elsewhere too. All three escapes were reproduced and are now caught.
3. **Nothing pinned `event="audit"` or `msg="audit"` as literals.** Every assertion compared against the constant, so a rename would have shipped green while every SIEM route in the field stopped matching.
4. **The field-presence matrix in `docs/observability.md` contradicted the table it calls enforced** — identity shown as always present when it is absent in `disabled` auth mode, `correlation_id` shown as universal when it is omitted rather than written empty.

Two findings were investigated and dismissed with evidence rather than changed: the emission placement, and going through `slog.Default().Handler()` (`cmd/server` binds resource attributes onto the handler, not the logger, so nothing is lost).

A second, harder pass then ran 30 defect injections. The constructor and the hash killed everything thrown at them; every survivor was on the integration seam or in the guard test itself. Fixed in the fourth commit:

1. **`docs/observability.md` shipped a false claim.** The audit `error_type` field pointed at all twelve values, but only five can reach an `operation` record — the destructive gate sits after tool lookup, broker resolution and validation, so `unknown_tool`, `missing_broker`, `unknown_broker`, `broker_init_error`, `validation_error`, `bad_request` and `not_found` are structurally unreachable there. A SIEM rule for `error_type: validation_error` on `event="audit"` would never have fired. The twelve-value set is still correct for the metric and the log line, so the presentation is split rather than the vocabulary shrunk. Four other statements still saying "ten" corrected in the same pass.
2. **The package's "audit emission never fails the operation" promise was not enforced.** A handler panicking in `Enabled` or `Handle` unwound to `withRecovery`, which would report a destructive call that had already mutated the broker as failed. `write` now recovers. The test for this was itself vacuous on first write and was caught by mutating the fix.
3. **Both drop-notice call sites were untested** — deleting either `EmitDrop` left the suite green. They are the safety net under the drift guard. The hash-failure branch is unreachable from the wire (verified: validation rejects it first), so it is extracted into `hashArgsForAudit` where a test can reach it.
4. **`duration_ms` was never asserted at the integration level.** Hardcoding it to zero passed; so did a first-attempt "non-negative" assertion. The destructive stub now sleeps so the assertion has a real floor.
5. **Three more drift-guard escapes closed** — an unresolvable call expression, a `var` declaration, and an emit site whose local is not named `errorType`. The last is caught by checking every `logToolResult` call site passes `&errorType`. The guard's comment now states plainly what it still cannot do.

Five categories came back clean: the config flag is fully plumbed and off by default; nothing outside Go unit tests depends on the destructive WARN; `withAuthorization` is correctly scoped to SOL-152097; the secure-logging treatment of both error paths holds; and the audit package sits at 96.0% coverage with the repo at 90.0%.

## Testing

`make check` green, coverage 89.9%.

Every new assertion was mutation-checked rather than assumed: dropping a vocabulary entry, neutering the applicability check, skipping RFC 8785 canonicalisation, hashing the empty map instead of the real arguments, and dropping `panic_recovered` each fail the tests that claim to guard them.

Hash-stability goldens carry expected digests computed **outside Go**, from hand-derived canonical forms, so they fail if canonicalisation output ever changes rather than tracking whatever the code does. Set sizes assert against `len()` of the table, never a literal.

## Deliberately not in scope

- **`mcp_audit_events_dropped_total`.** The `audit_drop` record ships with no dependency; the counter waits on audit instruments being registered against the meter provider. Documented as not registered rather than claimed.
- **The broker-side SEMP-log assertion** in Story 22's acceptance criteria. The identity chain is asserted server-side (the audit record names the same principal as the `tool invoked` line, from one source), and `test/e2e-oauth` already proves the broker enforces against the *exchanged* token rather than the agent's, through the audience-isolation and insufficient-permission scenarios. Asserting on the broker's own event log needs a CI capability that does not exist yet, and it belongs with SOL-153332, which is the story that actually inspects hop-2 outcomes.
- **Story 49 itself.** Nothing here emits `broker_authz_denied` or sets `broker_permission_denied`; the schema simply stops rejecting them.


internal/tools/manager.go
```diff
@@ -246,12 +280,23 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
...
+	// not which calls produce one. Arguments are hashed here, while they are
+	// still the validated map the handler is about to receive; the record
+	// itself is emitted from the defer, once the outcome is known.
 	if rt.annotations.Destructive != nil && *rt.annotations.Destructive {
-		slog.Warn("executing destructive operation",
-			slog.String("tool", name),
-			slog.String("broker", brokerAlias),
-			slog.Any("", id))
+		if m.auditLog {
+			auditArgsHash = hashArgsForAudit(ctx, name, brokerAlias, params)
```
bczoma: 🟡 **Hash a redacted projection of `handlerParams`, not raw `params`**

Flagged by `good:security` + `good:code` + Copilot, independently. Two distinct defects share this one argument, and one change fixes both.

**1. The digest can cover a value on the repo's own Never-Log list.** `msgVpnConfig` is declared `type: object` with a description and nothing else (`internal/composite/definitions/tools.yaml:848`), and `buildCompositeInputSchema` (`internal/tools/composite_handler.go:136-167`) sets neither `properties` nor `additionalProperties` for a param — so any key validates. The bundled 10.26.5 spec's `MsgVpn` PATCH body exposes `replicationBridgeAuthenticationBasicPassword` and `replicationBridgeAuthenticationClientCertPassword`, both writable on the exact `updateMsgVpn` operation `update-message-vpn` calls. Nothing between the wire and the digest strips secret-shaped keys.

`docs/internal/secure-logging-rules.md:13` names "SEMP passwords" on the Never-Log list, and Rule 3's `ReplaceAttr` net **structurally cannot reach this**: the secret is in the digest *pre-image*, while the emitted attribute key is `arguments_hash`, which matches no redaction pattern. That is the finding — a policy nonconformance on a stream the operator ships off-host with long retention.

Measured, so the severity is honest. I filed this initially as a brute-force oracle and a challenger reduced that framing: full exploitation needs the flag on (off by default), a rare replication-password change actually routed through the agent, a guessable password, and a hostile reader of the audit stream. Too long a chain to lead with. But no probe defended the mechanism — no schema constraint, no stripping, no salt, no HMAC (`hash.go:62` is `json.Marshal` → `jcs.Transform` → `sha256.Sum256`, and `jcs.Transform` takes only `jsonData`). Two details make it worse than filed: the hash is computed **before** `handler.Handle`, so the value is digested even if the broker rejects the request; and both fields are *absent from a SEMP GET*, so the digest would be the only recovery channel for that specific secret.

**2. The digest is salted by the caller's broker casing.** `brokerAlias` is rewritten to `bc.DisplayName()` at `:264-269`, but `params["broker"]` still holds whatever casing the caller typed. So the record's `broker` field carries the canonical name while the digest covers the raw one: two byte-identical operations differing only in capitalisation produce two different `arguments_hash` values, both records reading `broker: dev`. An auditor grouping by digest, or recomputing it from the canonical name, gets no match. `TestDestructiveCall_emitsExactlyOneOperationRecord` hashes the same map with a lowercase alias, so nothing in the suite notices.

**The fix for both.** Hash a projection: walk the map recursively, replace the value of any key matching the existing Rule 3 pattern (`password`, `secret`, `token`, `credential`, `api_key`, `private_key`) with a fixed placeholder, and use `handlerParams` — which is what the comment at `:286-288` already claims. If `broker` must stay in the digest, insert the resolved `brokerAlias` rather than the raw value. Then document the exclusion in `docs/observability.md`'s "Reproducing the hash" section, next to the broker-casing note already there, so the digest stays customer-reproducible. Prefer this to HMAC-with-a-server-key, which would break the reproducibility promise the feature is built on.

Worth a test that calls with `"broker": "DEV"` and asserts the digest matches the `"dev"` call.

Latent, not in this PR: `create-message-vpn` takes the same free-form `msgVpnConfig` and is safe only because it is `destructive: false`. It would inherit this the day that annotation flips.

internal/tools/manager.go
```diff
@@ -306,6 +351,83 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
...
+	}
+	// The Go type only, never the message: the wrapped encoding/json error
+	// renders the offending value, and that value is a tool argument — the one
+	// thing an audit record must never carry
+	// (docs/internal/secure-logging-rules.md).
+	slog.ErrorContext(ctx, "audit: could not hash tool arguments; recording a drop instead of an operation record",
+		slog.String("tool", tool),
+		slog.String("broker", broker),
+		slog.String("detail", fmt.Sprintf("%T", err)))
+	audit.EmitDrop(ctx)
```
bczoma: 🟡 **A hash failure emits an unattributable drop, and the destructive call still executes**

Flagged by `good:architecture`, with a related note from `good:security`.

On a hash failure this logs an ERROR, calls `audit.EmitDrop(ctx)` and returns `""` — and `CallTool` then proceeds to execute the destructive operation. So the broker gets mutated and the audit stream's only trace is a record naming nothing: `applicabilityByType[EventAuditDrop]` forbids `tool`, `broker`, `outcome` and identity.

The compensating context is in the `slog.ErrorContext` line right here, which carries `tool` and `broker` — but that line is **outside** `event="audit"`, and that predicate is the one routing rule `docs/observability.md` tells customers is sufficient to capture the audit stream. So an auditor working inside the stream as documented sees a mutation they cannot attribute. The same shape applies on the constructor-rejection path at `:414-426`.

A challenger established that the drop is not *entirely* content-free — the common block in `NewEvent` runs before the applicability gate, so `timestamp_utc` and `correlation_id` are stamped, and correlation is on by default. That makes the drop joinable to the operational `tool invoked` line, which recovers the context for a **failed** call (ERROR, survives any level). A **successful** destructive call's line is INFO, so at `log_level: warn` or above the join target is filtered and nothing recovers it.

Two options, either is fine:

- Give the drop enough to be actionable inside its own stream — a `dropped_audit_event_type`, and `tool`/`broker` where the drop is not itself the leak risk. Explicitly not raw arguments.
- Or reconsider whether `arguments_hash` needs to be `fieldRequired` on `operation`. It is required by table choice, not by necessity; a hash-less operation record with an explicit `arguments_hash_unavailable` marker carries strictly more forensic value than an anonymous gap.

The doc comment above this function makes a fair point that the branch is unreachable from the wire. That is a reason the code is hard to test, not a reason the record shape is right — and if it really is unreachable, the cheap version of this fix costs nothing.

internal/observability/audit/event.go
```diff
@@ -0,0 +1,577 @@
...
+package audit
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"sort"
+	"time"
+
+	"github.com/SolaceProducts/solace-broker-mcp/internal/auth"
```
bczoma, line 24: 🟡 **This import direction blocks the auth emitters the schema exists for**

Flagged by `good:architecture`.

Reading identity via `auth.PrincipalFrom(ctx)` makes the dependency `audit → auth → {authz, idpclient}`. That is the right call for *this* PR — it is what makes identity unforgeable, and I verified `NewEvent` ignores `Fields` entirely for identity so a call site structurally cannot name another caller.

The problem is the next two record types. `auth_success` and `auth_failure` are produced in `internal/auth/middleware.go`, and hop-1 `authz_denied` in the authz path — `docs/observability.md:373` says both "land with SOL-152097". Those packages **cannot import `audit`**: that is an import cycle. So the schema package that exists to be the single constructor for seven record types has, by construction, locked out the emitters of three of them.

The workarounds available under deadline are the bad ones: emit auth records from a wrapper in `cmd/server`, where the failure *reason* is no longer known — and `reason` is validated against a per-record-type closed vocabulary, so a wrapper that has lost it cannot build a valid record; or hoist auth middleware into `internal/observability`, which is worse.

Cheap now, structural later. Either:

- move the principal context key and accessor into a leaf package both `auth` and `audit` import — `internal/auth/principal` or `internal/identity`; or
- have `NewEvent` read identity through a small interface registered once at startup.

The first is a mechanical move and keeps the unforgeability property exactly as it is. Worth doing in this PR while the package has one consumer rather than four.

internal/observability/audit/event.go
```diff
@@ -0,0 +1,577 @@
...
+// errorTypeVocabulary is the closed error_type set, and the single source of
+// truth for it.
+//
+// These are exactly the values internal/tools computes and passes to
+// logToolResult today. They are NOT re-derived here from a ticket: the count
+// has drifted twice already because each revision counted only the call site
+// it happened to be looking at. internal/tools/audit_error_type_drift_test.go
+// walks the AST of every emit site in that package and fails if the two sets
+// diverge, so adding a value there without adding it here breaks the build
+// rather than silently shipping a record the constructor rejects.
```
bczoma, line 104: 🟡 **The `error_type` set is duplicated across packages and policed by name, not by type**

Flagged by `good:architecture`, with `good:code` finding two remaining gaps in the guard.

The vocabulary lives twice: as string literals in `internal/tools`, and as this map in `audit`. The guard is a ~430-line test that parses the production package's AST, tracks a variable **by name** (`errorType`), resolves identifiers through `ast.Ident.Obj`, and asserts set equality in both directions.

Credit where it is due — I mutation-tested it and it fails closed on both escapes I tried: changing a literal in `register.go:324` was caught with the file:line and both directions of divergence reported, and moving a value behind a cross-file package-level named constant was caught loudly as "an expression the drift guard cannot read". Refusing to guess is the right failure mode.

But it is a convention policed by a reviewer-substitute where a type would do, and it has three structural limits the mutations do not reach:

1. **It is per-package.** It walks `internal/tools` only, so the first emitter outside that package gets no drift protection at all — and the reverse assertion ("audit accepts X but nothing in `internal/tools` computes it") quietly encodes "`internal/tools` is the sole producer of `error_type`", which the SOL-152097 auth stories break.
2. **It sees assignments, not comparisons.** `manager.go` sets `PanicRecovered = errorType == "panic"` and `audit.validate` enforces `panic_recovered ⇒ error_type == "panic"` — a cross-package coupling on a bare literal that appears in five places. The guard reads the assignments and not either comparison, so half the coupling is unguarded.
3. **It leans on `ast.Ident.Obj`**, deprecated in `go/ast` and slated for removal. When it goes, `stringConstant` returns false and the guard errors — the right direction, but `internal/tools`' CI then breaks on a toolchain bump for reasons unrelated to the audit schema.

The fix collapses all three: export typed constants from `audit` — `type ErrorType string` with `ErrorTypePanic`, … — make `Fields.ErrorType` that type, and key `errorTypeVocabulary` on them. `internal/tools` then references the constants, the set becomes compiler-closed across every package present and future, and this test plus its by-name fragility can be deleted. If the literals must stay strings for the operational log line, keep the constants and use them at both sites.

bczoma (dismissed):
## PR Review: #381 — SOL-152090 / SOL-152096: enforce the audit-record schema, and emit one for every destructive tool call

**Author:** solace-aross | **Branch:** `aross/sol-152090-audit-event-constructor` → `main` | **Changes:** +3611 / -56 across 18 files
**Jira:** `SOL-152090` (Story 21) and `SOL-152096` (Story 22) — all ACs traced to the diff; SOL-152090 lists SOL-152096 as a Blocks consumer, so the bundling holds

---

### Verdict

1. 🟡 `internal/tools/manager.go:291` — hash a redacted projection of `handlerParams`, not raw `params`: the digest can cover a Never-Log value, and is salted by the caller's broker casing
2. 🟡 `internal/tools/manager.go:377` — a hash failure emits an unattributable drop and the destructive call still executes
3. 🟡 `internal/observability/audit/event.go:24` — the `audit → auth` edge blocks the auth emitters this schema exists for
4. 🟡 `internal/observability/audit/event.go:104` — the `error_type` set is duplicated across two packages and policed by name, not by type

- 🟡 PR-level — commit `dec4e9f` is titled "close what a second review pass found", but all three Copilot threads are still open at this head: the `docs/user-guide.md:174` fragment, the `hash_test.go` escaping comment, and the `manager.go` `params`/`handlerParams` comment. A reviewer reading that title plus the body's itemised fix list will take prior feedback as cleared. Either close them or say in the body that they remain open.

---

### Summary

This is a carefully built change and the hard claims in it hold: the panic recover in `write` genuinely encloses both `h.Enabled` and `h.Handle`, `Event`'s fields are unexported with `Attrs()` returning a copy, identity comes only from `auth.PrincipalFrom(ctx)` so a call site structurally cannot name another caller, all eight hash goldens recompute correctly from their stated canonical forms outside Go, and the AST drift guard fails closed on both escapes it was mutation-tested against. The `gowebpki/jcs` dependency checks out — Apache-2.0, stdlib-only non-test imports, no `init()`, no `unsafe`, correctly inventoried — and taking it beats hand-rolling canonicalisation.

Three findings were filed at 🟡 and put through adversarial challenge; two were reduced to 🟢 by measurement. The `log_level` interaction that three lenses independently flagged turns out to produce a loud ERROR-per-call stream at a non-default level the docs explicitly tell operators not to use, with the diagnosis already written down — not the silent compliance gap it looked like. The delivery-contract concern was misattributed to this PR: `logToolResult` already writes synchronously to the same stderr from the same defer on every call, so the blocking property is pre-existing. What survived of both is in the notes below.

CI is green as of this review, including FOSSA and Prisma on the new dependency — worth noting because the body's "make check green" was a local claim while those were still pending.

**On what was reviewed.** The five lenses and three challengers ran against `dec4e9f4`. `ffd7f961` landed while they were working, and GitHub correctly refused to re-anchor this review onto it. That commit is test-only — +137 lines to `internal/tools/destructive_audit_test.go` — and I checked it rather than assume: `TestOperationRecord_reachableErrorTypes` drives the real `CallTool` path per case and asserts whether an operation record appears, which pins the five-of-twelve reachability claim `docs/observability.md` makes to customers instead of leaving it in prose. It passes, and it is the right instinct: that claim is about where the destructive gate sits relative to each failure, and prose cannot hold it. None of the four findings below is in a file that commit touched, so all four stand at this head.

---

### Minor notes

- 🟢 `internal/config/config.go:1016` / `cmd/server/main.go:1091` — `AuditLogEnabled` is never cross-validated against `LogLevel`. With the flag on and `log_level: warn`/`error`, every destructive call yields an ERROR `audit_drop` and no `operation` record, while startup reports `audit_log: true` with no caveat. Loud rather than silent, and `docs/observability.md:405` gives the diagnosis — but this repo validates other cross-field config at the boundary (`validateMetricsBindAddress`, the TLS cert/key pair), so this pair is an unexplained exception. Emit a startup WARN when `AuditLogEnabled && effectiveLevel > slog.LevelInfo`.
- 🟢 `docs/observability.md` ("Audit Delivery") and `internal/observability/audit/emit.go:24-30` — the contract claims delivery is "non-blocking by design" and that a backpressured sink means the event is "dropped rather than buffered". Neither exists: `write` is a synchronous `Handle` to `os.Stderr` with no queue, deadline, or non-blocking path, and no drop-on-backpressure mechanism at all. The same paragraph's "three things produce a drop today" list is correct and omits backpressure, so the section contradicts itself two sentences later. Narrow the prose to best-effort synchronous write, or add the bounded channel the sentence promises.
- 🟢 `internal/observability/audit/event_test.go:141`, `:181` — the `if len(got) != len(want)` blocks sit *after* `reflect.DeepEqual` + `t.Fatalf`. `DeepEqual` already compares length and the Fatalf aborts, so neither block can ever fail — in the two tests whose comments advertise the `len()`-not-a-literal discipline. Delete both.
- 🟢 `internal/tools/manager.go:212`, `:393`; `internal/observability/audit/event.go`; `register.go:314`; `describe_semp_schema.go:336` — the `error_type` value `"panic"` is a bare literal in five places across two packages, and it is load-bearing: `manager.go` sets `PanicRecovered = errorType == "panic"` while `audit.validate` enforces `panic_recovered ⇒ error_type == "panic"`. The drift guard sees the assignments but not the two comparisons, so half the coupling is unguarded. Export typed constants.
- 🟢 `internal/tools/audit_error_type_drift_test.go` — two branches in an otherwise fail-closed guard stay silent: an assignment whose LHS is not a bare `*ast.Ident` (`*errorTypeOut = "x"`) is `continue`d with no diagnostic, and `collectClassifierErrorTypes` reads only `ret.Results[0]`, hardcoding that `errorType` is the first return value. Every other unreadable shape errors loudly; these two should too.
- 🟢 `internal/observability/audit/event.go` (`checkPresence`) — the `switch a` has cases only for `fieldRequired` and `fieldForbidden`, no `default`, so a mis-typed table row validates nothing. Separately `rules.identity` is never passed through `checkPresence` at all, so `identity: fieldRequired` would be silently ignored. Add a `default` that errors, and narrow identity's type so the unreachable value is unrepresentable.
- 🟢 `internal/observability/audit/emit.go:29` — `Event`'s zero value is emittable: `audit.Emit(ctx, audit.Event{})` compiles and writes a fieldless `msg="audit"` line at INFO, so "NewEvent is the single constructor" is a convention rather than a type guarantee. Have `Emit` refuse an `Event` with an empty `typ`. Related: the flag is enforced only by callers, so each of the three queued emitters must remember to check it.
- 🟢 `internal/tools/manager.go:394` — `emitOperationAudit` computes `time.Since(start)` while `logToolResult` independently computes its own elapsed from the same `start`, so the two records for one call report slightly different durations — which the comment two lines up says cannot happen. Compute once in the defer and pass it to both.

…and 8 further minor items — `OutcomeCancelled` is admitted by the table but never emitted; `Enabled(cfg)` mirrors one bool three times and is the sole reason `internal/config` is in this package's dependency closure; the `slog.Default().Handler()` route makes "no attributes on the logger" a load-bearing unguarded invariant; the drift guard leans on the deprecated `ast.Ident.Obj`; `audit_drop` could carry a `dropped_audit_event_type`; `TestEventTypes_closedSetIsSeven` hardcodes the count in its own name; `OBS_AUDIT_LOG_ENABLED` is absent from `docs/configuration.md` (as are its `OBS_*` siblings, so not this PR's fault); and roughly half of `event.go` is commentary in a PR-argument register that will age against the code. Ask if you want them written up.

---

### Summary Verdict

|---|---|---|
| Documentation | ✅ | docs |

Not independently verified: the claimed 96.0% package / 89.9% repo coverage figures.

internal/tools/manager.go
solace-aross: Fixed: hashArgsForAudit now hashes handlerParams (broker already stripped, post-validation) instead of raw params, through a new audit.RedactSensitive that walks the map recursively and replaces any key matching the same pattern cmd/server's redactedKeys uses (password/token/secret/authorization/credential/api_key/private_key) with a fixed placeholder before hashing. Excluding broker entirely also removes the casing-salt bug as a side effect — no need to reinsert the resolved alias. Added TestDestructiveCall_hashIsInvariantToBrokerCasing and TestDestructiveCall_hashRedactsSensitiveArguments (plus unit coverage on RedactSensitive itself in hash_test.go). docs/observability.md's hash-reproduction recipe now documents both transforms. Addressed in 0df5f6a.

internal/tools/manager.go
solace-aross: Took the first option: audit_drop gained optional dropped_audit_event_type, tool, and broker fields (EventAuditDrop's row in applicabilityByType), set on both drop sites here (hash failure, constructor rejection). A reviewer inside event="audit" can now attribute a gap without joining to the operational line, which matters most exactly where you said — at log_level above INFO a successful call's own tool-invoked line is filtered out too, not just a failed one. Left arguments_hash fieldRequired on operation rather than the second option; narrowing the drop's own shape felt like the smaller change for the same outcome. Additive: audit_schema_version 1.0 -> 1.1. Addressed in 0df5f6a.

internal/observability/audit/event.go
solace-aross: Agreed on both the diagnosis and the fix direction — moving the principal accessor to a leaf package (internal/auth/principal or internal/identity) is the right call and keeps unforgeability exactly as-is. Not doing it in this pass: it touches internal/auth's own import graph and I'd rather give that its own focused PR and review than fold it into an already-large rebase. Filing a follow-up ticket for it rather than losing it to this thread. Leaving this open.

internal/observability/audit/event.go
solace-aross: Agreed this is the real fix, not a convention — and worth flagging what changed underneath this thread since it was filed: rebasing onto main's #379 (per-tool metrics) meant every error_type value in internal/tools became a metrics.ErrorType selector expression instead of a bare literal, so the drift guard in this PR needed extending just to keep passing (stringConstant now resolves a cross-package selector by parsing the referenced package's own const decls). That's the same duplication you're describing, now with an extra AST-resolution layer bolted onto the guard rather than fewer moving parts. Exporting typed audit.ErrorType constants and having internal/tools reference them directly — deleting the guard entirely — is the actual fix; I didn't do that here since it touches call sites in three files plus the metrics.ErrorType /audit.ErrorType relationship and deserves its own review rather than riding along on this rebase. Filing a follow-up. Leaving this open.

internal/tools/manager.go
```diff
@@ -246,12 +280,36 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
 		return buildLocalErrorResult(toolErr), nil
 	}
 
-	// Log warning for destructive tools.
+	// Destructive tools are the audit surface (SOL-152096). This is the same
+	// gate the pre-flip WARN stood at — after broker resolution and input
+	// validation — so enabling the audit log changes the FORM of the signal,
+	// not which calls produce one. Arguments are hashed here, while they are
+	// still the validated map the handler is about to receive; the record
+	// itself is emitted from the defer, once the outcome is known.
```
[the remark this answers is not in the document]
solace-aross: Overtaken by 0df5f6a: the code now actually hashes handlerParams (broker stripped, Never-Log keys redacted), not raw params, and the comment above this block matches. Same commit also fixed the real bug this pointed at — broker's caller-supplied casing was salting the digest.

docs/user-guide.md
```diff
@@ -171,7 +171,7 @@ The server exposes 24 read-only tools plus 16 write tools (40 total when write t
...
+These tools modify event broker state via the SEMPv2 action API. There is **one tool per action** so each tool's behavior is unambiguous: the destructive tools carry the MCP `destructiveHint` annotation and a description that instructs the calling LLM to obtain explicit user confirmation — restating the target (event broker, VPN, queue or client) and the effect — before invocation; the non-destructive stats-reset tools do not. The tool manager records every destructive invocation: with `OBS_AUDIT_LOG_ENABLED` on it emits one audit record per call carrying the caller, the broker, a hash of the arguments, and the outcome (see [Audit Trail](observability.md#audit-trail--interim--operation-records-only)); with it off, the WARNING line it has always logged.
```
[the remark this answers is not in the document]
solace-aross: Fixed: "with it off, it emits the WARNING line it has always logged." Addressed in e673569.

internal/observability/audit/hash_test.go
```diff
@@ -0,0 +1,275 @@
...
+			"z": []any{3, 1, 2},
+			"n": nil,
+			"m": map[string]any{"y": 1, "x": 2},
+		},
+		canonical: `{"m":{"x":2,"y":1},"n":null,"z":[3,1,2]}`,
+		want:      "22a64e4706f978b672fb5b6b629a1abfa6bf06e21f707f7ae7ae92cfbf572c59",
+	},
+	{
+		// Go's encoding/json escapes <, > and & to <, > and & by
+		// default. RFC 8785 does not. If that escaping ever leaked into the
```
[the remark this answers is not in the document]
solace-aross: Fair — the comment read as claiming the same characters before and after. Named the actual escape forms (\u003c, \u003e, \u0026) and explained why the golden's literal characters are the proof, not an oversight. Addressed in e673569.

internal/observability/audit/hash.go
```diff
@@ -0,0 +1,156 @@
...
+// sensitiveKeySubstrings mirrors cmd/server's redactedKeys — the same
+// case-insensitive substring match docs/internal/secure-logging-rules.md
+// Rule 3 applies as a ReplaceAttr safety net on log output — applied here to
+// a hash's pre-image instead. Nothing about arguments_hash's own emitted
+// attribute matches Rule 3's key patterns (the key is "arguments_hash"), so
+// that net cannot reach a secret-shaped value baked into the digest; this is
+// the control for that gap. Kept as its own copy rather than importing
+// cmd/server's var: this package must not depend on cmd/server, and the two
+// lists mean the same policy, not the same Go value.
+var sensitiveKeySubstrings = []string{"password", "token", "secret", "authorization", "credential", "api_key", "private_key"}
```
bczoma, line 36: 🟡 **No redaction pattern covers certificate/key *content* fields**

Flagged by `good:security`. Filed as a four-field oracle; measured down to one reachable field, and it survives at 🟡.

**The gap.** `sensitiveKeySubstrings` is a *log-attribute-key* list, copied byte-for-byte from `cmd/server/main.go:73`, applied here to *caller-supplied SEMPv2 argument names*. Two of the seven patterns carry underscores (`api_key`, `private_key`) and SEMPv2 arguments are camelCase, so they match nothing — verified with a probe over the list: `apiKey`→false, `api_key`→true, `privateKey`→false. The five single-word patterns survive camelCase fine (`authenticationBasicPassword`, `bearerToken`, `authenticationOauthClientSecret` all match).

That separator mismatch turns out to be unreachable — no shipped tool exposes a SEMPv2 field named `apiKey` or `privateKey`. **What is reachable is a missing pattern, not a separator problem**, and it lands on the object this control was written for:

`MsgVpn.replicationBridgeAuthenticationClientCertContent` is a PEM containing a private key. Its sibling `...ClientCertPassword` is caught by `password`. The content field is caught by nothing, and it reaches the digest through a shipped destructive tool: `update-message-vpn` declares `msgVpnConfig` as bare `type: object` (`tools.yaml:839-866`), `buildCompositeInputSchema` emits it with no `properties` and no `additionalProperties:false`, so `validateAgainstCompiledSchema` accepts any key, and `constructRequestBody` merges it into the SEMP body unchecked. Setting a replication-bridge client cert this way is a normal admin action.

**Measured risk, so the severity is not guesswork.** Reachability: routine, one field. Recovery: none — the digest goes out through `slog.Default().Handler()`, and `redactSecretAttr` sees only the attribute key `arguments_hash`, which matches nothing, exactly as `hash.go:29-35` predicts. Blast radius: one field on one tool. The other three fields named in the original finding (`authenticationKerberosKeytabContent`, `authenticationClientCertContent`, `authenticationHttpHeaderValue`) live on `MsgVpnKafkaReceiver`/`Sender`, `RestDeliveryPointRestConsumer`, `DmrCluster` and `OauthProfile` — **none writable by any shipped tool**, and the four `create-*` tools are `destructive: false` so are never hashed at all.

**Why it still matters at 🟡.** Not disclosure — only the SHA-256 pre-image includes the value. It reinstates, for this one field, the change-detection oracle `hash.go:38-43` exists to remove: two otherwise-identical calls produce different digests iff the cert changed. The "known-plaintext enumeration" reading does **not** hold here — the spec requires a private key plus one to three certificates, so the value is high-entropy. What does hold is that the same spec line says the attribute is absent from a GET: the value is write-only at the broker, so this digest is the only derivative of it that ever leaves the process, and it persists for the life of the audit stream.

The comments at `hash.go:45-52` and `manager.go:396-400` both cite `update-message-vpn`'s `msgVpnConfig` and its replication-bridge credentials as the motivating case — so the gap sits inside the control's own stated scope.

**Recommended fix.** Add `certcontent` to the list, and cheaply `keytab` and `passphrase` while you are there. Separator-insensitive normalisation (strip `_`/`-` before `Contains`) is worth taking as hygiene but is not the fix — the reachable miss is a missing term. Extend `hash_test.go:291` with `replicationBridgeAuthenticationClientCertContent`. And correct the comment at `hash_test.go:288-290`: it currently documents the camelCase `apiKey`/`privateKey` miss as a known gap, which is defensible for those two but pins the gap to the wrong axis and so reads as clearance for the cert-content case too. If byte-identity with `cmd/server`'s list must hold, put the additions in a second audit-only slice with a one-line reason.

internal/observability/audit/emit.go
```diff
@@ -0,0 +1,130 @@
...
+// The drop record carries only common fields plus DroppedEventType/Tool/Broker
+// from dctx: it reports that a record is missing, and admits just enough
+// context to be actionable inside the audit stream itself
+// (docs/observability.md, "event=audit" is the routing predicate customers are
+// told captures everything) without becoming a second place raw arguments
+// could leak. It keeps event="audit" deliberately, since a notice that fell
+// outside the customer's audit filter would go unseen in precisely the
+// situation it exists to report.
+func EmitDrop(ctx context.Context, dctx DropContext) {
+	drop, err := NewEvent(ctx, Fields{
```
bczoma, line 88: 🟡 **The drop-attribution fix ships with no test that can fail**

Flagged by `good:code`. Established by mutation, not by reading.

I deleted all three lines — `DroppedEventType`, `Tool`, `Broker` — from `EmitDrop`'s `Fields` literal and ran `go test ./internal/observability/audit/... ./internal/tools/ -count=1`. **Both packages stayed green.**

The three tests that look like they cover this do not:

- `TestEmitDrop_writesTheNotice` (`emit_test.go:243`) only ever passes `DropContext{}` and asserts `tool`/`broker` are *absent*.
- `TestHashArgsForAudit_unhashableArgumentsEmitADrop` (`destructive_audit_test.go:773`) counts the drop records but reads none of their fields.
- `event_test.go:371` covers `NewEvent` only, never the `EmitDrop` wiring.

**Why it matters here specifically.** This is the remediation of a 🟡 from the previous review — the finding was precisely that a drop was *unattributable*. The fix is good and I verified it in code: `Tool` comes from the registry and `Broker` from the resolved display name, so neither is caller-forgeable, and `event.go:281` admits the fields on `EventAuditDrop` only. But a refactor that drops the attribution — the exact regression the finding was about — is invisible to CI, on the one record type whose whole job is to explain a gap in the audit trail.

**Recommended fix.** One assertion in `emit_test.go` that `EmitDrop(ctx, DropContext{DroppedEventType: EventOperation, Tool: "delete-queue", Broker: "dev"})` produces a record carrying all three, and extend `TestHashArgsForAudit_unhashableArgumentsEmitADrop` to assert the drop it counts actually names `tool`, `broker`, and `dropped_audit_event_type="operation"`.

internal/observability/schema/version.go
```diff
@@ -24,5 +24,8 @@ package schema
 const MetricsSchemaVersion = "1.0"
 
 // AuditSchemaVersion is stamped onto every audit-log record the server emits.
-// V1 baseline. Bump on any backwards-incompatible change to the audit shape.
-const AuditSchemaVersion = "1.0"
+// Bump the minor component on an additive change (a new field), the major
```
bczoma, line 27: 🟡 **The additive-only versioning contract is enforced by nothing**

Flagged by `good:architecture`.

`AuditSchemaVersion` is consistent everywhere today — I checked: `version.go:31`, `docs/observability.md:100` and `:441`, and the `mcp_schema_version{audit_schema="1.1"}` label in the golden fixture all agree, and `event_test.go:233` asserts against the constant rather than a literal, so no stale copy exists. The 1.0 → 1.1 bump itself is correct: three optional fields added to one record type, nothing renamed or removed.

The problem is that none of that is *held*. Every assertion compares the constant against itself, so `AuditSchemaVersion` can change, or fail to change, without a single test noticing. `applicabilityByType` can gain or lose a field with no bump. `event_test.go:139` and `:711` only *tell a human* to "check whether `AuditSchemaVersion` must be bumped", and "1.1" exists as a third hand-maintained copy in the docs.

**Why it matters.** This PR's entire thesis is that a schema documented in prose drifts, so the rules belong in one constructor that refuses to build a record breaking them — and you built exactly that, well. The schema's own version, which is the field downstream SIEM queries pin against, is the one part still policed by a comment. With three emitter stories queued (SOL-152097, SOL-153332, SOL-152426), each adding fields, this is precisely where a silent breaking change ships as a minor bump.

**Recommended fix.** Pin a golden fingerprint of `applicabilityByType` — types × field applicability, rendered deterministically — next to the version constant. A change to the table that does not change the version fails, with the message distinguishing "additive → bump the minor" from "field removed or narrowed → this is a major". The same test can assert the docs' rendered table matches, which would also have caught the `panic_recovered` omission on the other thread.

docs/observability.md
```diff
@@ -405,35 +433,71 @@ audit sub-stream to a dedicated SIEM index.
...
 | `outcome` | The result; see [The Outcome Vocabulary](#the-outcome-vocabulary) | string |
-| `error_type` | Why an operation failed; present on `outcome: error` only | string (closed set) |
+| `error_type` | Why an operation failed; present on `outcome: error` only. Five of the twelve values reach an audit record, see [`error_type`](#error_type) | string (closed set) |
 | `arguments_hash` | SHA-256 over an RFC 8785 (JCS) canonicalization of the call arguments | hex string |
 | `correlation_id` | Join key to logs, traces, and the broker-side entry | string |
-| `reason` | Why authentication or authorization failed; present on `auth_failure` and `authz_denied` | string (closed set) |
-| `audit_schema_version` | The schema version, for query pinning | string (`1.0`) |
+| `reason` | Why authentication or authorization failed; present on `auth_failure`, `authz_denied`, and `broker_authz_denied` | string (closed set, one per record type) |
+| `dropped_audit_event_type` | On `audit_drop` only: which `audit_event_type` could not be built or written | string (same closed set as `audit_event_type`) |
+| `audit_schema_version` | The schema version, for query pinning | string (`1.1`) |
```
bczoma, line 505: 🟡 **`panic_recovered` is emitted but absent from both tables the doc calls exhaustive**

Flagged by `docs`.

`panic_recovered` is a real emitted field — `event.go:459`, set from `manager.go:456` — but it appears in neither the "Event Fields" table nor the field-applicability table's column list. Its only mention in the whole doc tree is one prose sentence at `:403`.

That would be a small omission anywhere else. Here the doc explicitly tells SIEM authors these tables are complete and machine-enforced: *"Not every field is on every record, so a SIEM author can tell record kinds apart from field presence alone"*, and *"This table is enforced, not merely documented … rejects any combination outside the table"*. Both statements are false for this field.

**Why it matters.** An operator building a SIEM ingest schema or a strict parser from the documented field list gets an undeclared boolean on exactly the record that says a destructive handler crashed mid-mutation — so a strict pipeline rejects or silently drops it, and the "show me every panic-affected destructive call" query has no documented predicate to key on. This is the one record whose absence the doc itself says *"must never read as 'nothing was attempted'"*.

**Recommended fix.** Add a `panic_recovered` row to the Event Fields table (type `boolean`, "present and `true` only when `outcome: error` and `error_type: panic`") and an eleventh column to the applicability table: `opt` on `operation`, `—` on the other six. Note it in the 1.1 delta, since it is a field this cycle adds — which is the same edit the versioning thread would make test-enforced.

bczoma (dismissed):
## PR Review: #381 — SOL-152090 / SOL-152096: enforce the audit-record schema, and emit one for every destructive tool call

**Author:** solace-aross | **Branch:** `aross/sol-152090-audit-event-constructor` → `main` | **Changes:** +4286 / -62 across 22 files
**Jira:** `SOL-152090` (Story 21) and `SOL-152096` (Story 22)

---

### Verdict

Both blocking-ish items from my last pass are genuinely fixed. Nothing here is a showstopper; the two prior 🟡s you deferred stay open and are counted as such below.

1. 🟡 `internal/observability/audit/hash.go:36` — no redaction pattern covers certificate/key *content* fields, so a write-only PEM private key is hashed in the clear while its password sibling is redacted
2. 🟡 `internal/observability/audit/emit.go:88` — the drop-attribution fix from last pass ships with no test that can fail; I deleted all three fields and both packages stayed green
3. 🟡 `internal/observability/schema/version.go:27` — the additive-only versioning contract is enforced by nothing, in a PR whose thesis is that schema rules belong in code
4. 🟡 `docs/observability.md:441` — `panic_recovered` is emitted but missing from both tables the doc tells SIEM authors are exhaustive and enforced

- 🟡 PR body and `SOL-152096` — the deferral record is incomplete in both places. The body's "Deliberately not in scope" list names three items, none of them the two prior 🟡s you explicitly declined, and no follow-up ticket key is given for either despite the commit message promising one would be filed; a reviewer arriving at head reads "review applied" and approves without opening the two threads you left open on purpose. Separately, the body defers SOL-152096's broker-side SEMP-log assertion to SOL-153332 on documented-gap grounds, which is the rationale that ticket's AC text pre-emptively rejects ("a required, testable assertion, not a documented gap"), and the ticket is unamended. Add a "Left open from review" section with both ticket keys, and amend SOL-152096 before merge rather than leaving the deviation only in a PR description.

---

### Summary

The two fixes from my previous pass are closed properly: `broker` is now excluded *structurally* by hashing the already-stripped `handlerParams` rather than by a special case, and the redaction list is a verbatim copy of `cmd/server`'s with the reason for not sharing the Go value written down. The drop record gained real attribution. What is left is mostly bookkeeping — a schema contract policed by prose, two tables that disagree with the code they document, and a deferral record that does not name what was deferred. Two findings were filed at higher severity and reduced by measurement: a claimed four-field credential oracle turned out to be one reachable field with a non-enumerable value, and a claimed unsatisfiable identity policy turned out to rest on the wrong function — `WithPrincipal` is exported and runs before the downstream handler, so the ctx-only rule is satisfiable for both auth types.

*13 prior findings — 3 addressed, 1 accepted as-is, 1 partially addressed, 7 still open (6 of them 🟢, listed under Minor notes rather than repeated here), 1 not re-checked.*

|---|---|---|---|
| 🟡 | The `audit → auth` import edge blocks the auth emitters this schema exists for | Not addressed — deferral **accepted**, with one condition | `event.go:24,429` — `go list -deps` confirms `audit → auth → {config, authz, idpclient, correlation, sanitize}`. `Principal`/`WithPrincipal`/`PrincipalFrom` live entirely in `internal/auth/principal.go` and `event.go:429` is their only reader outside `internal/auth`, so the fix is a file move plus one import line. Nothing here makes it harder. The condition: it must land **before** SOL-152097 starts, or the auth emitters get written against the edge and the workaround becomes the expensive thing. |
| 🟡 | The `error_type` set is duplicated across two packages and policed by name, not by type | Not addressed — deferral noted, but the stated reason no longer holds | The reply cites the cost as "touches call sites in three files". After the #379 rebase `internal/tools` already produces `metrics.ErrorType` at every site (`manager.go:270,279,311,346,358,368,379`), and `metrics` has no `audit` dependency, so `audit → metrics` is acyclic. Only a widening cast at `manager.go:252` and `Fields.ErrorType string` stand in the way. Meanwhile the guard grew ~120 lines (`audit_error_type_drift_test.go:449-563`) to re-derive at test time what the type checker already knows. Not a blocker — the guard fails closed — but the cheap fix got cheaper while the workaround got more expensive, so this should be sequenced **above** the import-edge move, not below it. |
| 🟢 | `AuditLogEnabled` is never cross-validated against `LogLevel`, so the flag can be on while every record is filtered out | Not re-checked | No lens this pass claimed it and I did not reach it. Reported as unexamined rather than settled. |

---

### Minor notes

- 🟢 `docs/observability.md:908-918` — this PR added a second `bad_request` row and a second `not_found` row to a table that already carried both, so a closed set the prose calls "twelve" now renders 14 rows, with the two `bad_request` rows giving conflicting meanings. `event.go:114-127` is the authority; delete the duplicates.
- 🟢 `docs/observability.md:641` — the doc lists three drop causes; the code has four. Constructor rejection (`manager.go:471`) is missing from both this list and the CHANGELOG's, which names both drop sites two clauses after saying "three".
- 🟢 `internal/observability/audit/hash.go:27-35` — the redaction list is a deliberate hand-copy of `cmd/server/main.go:73` with nothing asserting the two stay equal. This PR's own thesis is that security-relevant coupling gets a build-breaking test; this coupling has higher blast radius than `error_type` and no guard.
- 🟢 `docs/observability.md:552-560` — the published Python recipe implements transform 1 only; transform 2 (the recursive redaction) is left as a `# comment`. Anyone running it as written gets a mismatching digest on exactly the calls that carried a credential. Ship the six-line `redact()`.
- 🟢 `docs/observability.md:634-644` — the fail-open policy is never stated: a destructive call whose audit record cannot be built still executes. That is the right default, and for a compliance feature the first question asked will be whether it can be configured otherwise. One sentence: the operation is never blocked, a drop record is the only signal, there is no fail-closed mode, by design.
- 🟢 `internal/observability/audit/hash.go:53-88` — `RedactSensitive` is exported with a doc inviting use on "any argument map that reached the wire", but recurses only into `map[string]any` and `[]any`. Safe today because the only caller feeds it `json.Unmarshal` output; the queued auth emitters may build Go-native maps. Narrow the doc, or make it reflection-based with a depth cap.
- 💡 `docs/observability.md:482-484` and `internal/observability/audit/event.go:59-61` — both promise `principal.sub` appears on `auth_failure` "when the token parsed far enough", but nothing can produce that today. Either drop the clause, or record on the follow-up ticket that an unverified subject must never land in the `principal` group.
- 🟢 Still open from last pass, all optional: the two dead `len()` blocks after a `DeepEqual`+`Fatalf` (`event_test.go:141`, `:181`); the two `"panic"` comparisons still bare literals across packages (`manager.go:457` vs `event.go:518`) with the drift guard seeing only assignments; the guard's two silent branches (`audit_error_type_drift_test.go:350`, `:393`); `checkPresence`'s missing `default` and `rules.identity` never routed through it (`event.go:571`, `:428`); and `emitOperationAudit` recomputing `time.Since(start)` so the two records for one call disagree on duration (`manager.go:449` vs `:511`).

…and 4 further minor items (the `EventAuditDrop` doc comment now falsified by this PR's own table; an unused parameter and a dead conditional in the drift guard; no 1.0→1.1 version history in the customer doc; a CHANGELOG clause claiming the docs listed ten `error_type` values when `main` already listed twelve) — ask if you want them.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

internal/observability/audit/hash.go
solace-aross: Right call narrowing it to reachability rather than the separator. Added certcontent/keytab/passphrase to a second, audit-only list (kept sensitiveKeySubstrings a byte-identical mirror of cmd/server's redactedKeys, per your suggestion). Extended hash_test.go with replicationBridgeAuthenticationClientCertContent and corrected the comment pinning the known gap to the wrong axis. Addressed in bcb1a7e.

internal/observability/audit/emit.go
solace-aross: Confirmed by the same mutation before writing the fix. Added TestEmitDrop_carriesAttribution and extended TestHashArgsForAudit_unhashableArgumentsEmitADrop to assert the actual field values; re-ran your deletion against both and they now fail. Addressed in bcb1a7e.

internal/observability/schema/version.go
solace-aross: Took the golden-fingerprint approach, hand-maintained rather than -update-regenerated (like the metrics golden) so a mismatch forces a human to judge additive vs breaking rather than rerun a flag. Mutation-verified it catches a table change. Left the docs-table cross-check as a follow-up rather than building it into this test — a fair scope cut, not a disagreement. Addressed in bcb1a7e.

docs/observability.md
solace-aross: Added to both tables. Addressed in bcb1a7e.

docs/observability.md
```diff
@@ -518,11 +583,59 @@ only `sub` is written to the audit event. A human-readable username (`preferred_
...
+   (`internal/tools/register.go`) rather than being part of the tool's own arguments, and
+   broker aliases resolve case-insensitively — so if `broker` stayed in the digest, a call
+   sent as `"broker": "PROD"` and one sent as `"broker": "prod"` against the same broker would
+   hash differently despite being the same operation. Removing it before hashing means the
+   digest cannot be salted by the caller's casing; the record's own `broker` field (always the
+   *configured* casing) is the place to look for which broker a call targeted.
+2. **A value on the [Never-Log list](internal/secure-logging-rules.md) is replaced with the
+   fixed placeholder `[REDACTED]`, recursively.** A key matches by the same case-insensitive
+   substring test as the `ReplaceAttr` net on log output — `password`, `token`, `secret`,
+   `authorization`, `credential`, `api_key`, `private_key` — because at least one composite
```
bczoma, line 607: 🟡 **The published recipe's transform 2 no longer matches what the code redacts**

Flagged on the follow-up pass.

`hash.go:52` adds `auditOnlySensitiveKeySubstrings` — `certcontent`, `keytab`, `passphrase` — and `isSensitiveKey` (`hash.go:106-121`) checks it alongside the seven-pattern mirror. This block still enumerates only the seven, and `grep -rn "certcontent\|keytab\|passphrase" docs/` returns nothing anywhere in the doc tree.

**Why it matters.** The digest's whole value is that an auditor holding the original arguments can recompute it and prove the record corresponds to that call. Follow this recipe on an `update-message-vpn` call that set `replicationBridgeAuthenticationClientCertContent` — the exact field the new list was added for — and you redact nothing, hash the PEM in the clear, and get a digest that does not match the record. The natural reading of a mismatch is "this record does not correspond to this call", so the failure lands as a false tamper signal on precisely the calls that carried a credential. The CHANGELOG's claim that the recipe "documents both transforms" is true of the two transforms and not of the list inside transform 2.

**Recommended fix.** Add the three terms to this block, marked as audit-only so it stays clear that `sensitiveKeySubstrings` is deliberately a byte-identical mirror of `cmd/server`'s `redactedKeys` and these are not. While you are in here, the prior 🟢 about transform 2 being a `# comment` in the Python snippet (`:621`) is the same edit's other half — six lines of `redact()` makes the recipe runnable, and a recipe nobody can run is where this drift came from.

bczoma (dismissed):
## PR Review: #381 — SOL-152090 / SOL-152096: enforce the audit-record schema, and emit one for every destructive tool call (Stories 21 and 22)

**Author:** solace-aross | **Branch:** `aross/sol-152090-audit-event-constructor` → `main` | **Changes:** +4486 / -62 across 22 files
**Follow-up review** — prior review at `e6735690` was dismissed by the rebase; its findings still apply, so the diff stayed full. Parent-only pass at `9d62255`, no lenses dispatched.

---

### Verdict

All four conditions of the last verdict are addressed, and I verified two of them by mutation rather than by reading: deleting the attribution fields from `EmitDrop`'s `Fields` literal now fails `TestEmitDrop_carriesAttribution`, and the applicability fingerprint fails in **both** directions — a table change without a version bump, and a version bump without a table change. That is the fix the versioning finding asked for.

1. 🟡 `docs/observability.md:607` — transform 2 of the published hash recipe still lists only the seven log-attribute patterns; the code now redacts three more, so an auditor reproducing the digest of a cert-carrying call gets a mismatch

The 2 🟢 notes below are **not** conditions of this verdict.

---

### Summary

The audit-record constructor and the destructive-call emitter both land in good shape, and this pass is almost entirely remediation of the last one: cert-content redaction now has its own audit-only pattern list with the log-line mirror left byte-identical, the drop record's attribution has a test that actually fails without it, the schema version is pinned to the applicability table by a hand-maintained fingerprint that refuses to be regenerated by a flag, and `panic_recovered` is now in both documented tables. Hashing `handlerParams` rather than `params` is the right call and I checked that `stripBrokerParam` is a pure copy-minus-`broker` with no defaults or coercion injected, so the digest stays reproducible from the raw request by the two documented transforms. The one remaining condition is that those documented transforms no longer match the code.

*11 prior findings settled — 4 addressed (the four conditions of the last verdict), 2 declined deferrals, 1 dropped on re-check.*

|---|---|---|---|
| 🟡 | No redaction pattern covers certificate/key *content* fields | Addressed | `hash.go:52` — `auditOnlySensitiveKeySubstrings` = `certcontent`/`keytab`/`passphrase`, checked by `isSensitiveKey` at `:116`; `sensitiveKeySubstrings` left a byte-identical mirror of `cmd/server`'s list as suggested; `hash_test.go:316-327` pins `replicationBridgeAuthenticationClientCertContent`; the comment at `hash_test.go:289-295` now scopes the known gap to the separator axis only |
| 🟡 | The drop-attribution fix ships with no test that can fail | Addressed | `emit_test.go:266` — re-ran the original mutation: deleting the fields from `EmitDrop`'s literal fails with `dropped_audit_event_type = <nil>, want "operation"` |
| 🟡 | The additive-only versioning contract is enforced by nothing | Addressed | `event_test.go:923-948` — mutation-verified both directions: flipping `identity` on `operation` fails, and bumping `AuditSchemaVersion` alone fails. Docs cross-check deferred; accepted as a fair scope cut |
| 🟡 | `panic_recovered` absent from both tables the doc calls exhaustive | Addressed | `docs/observability.md:500` (row) and `:522` (eleventh column) |
| 🟢 | The hash recipe implements transform 1 only; transform 2 is left as a `# comment` | Not addressed | `docs/observability.md:621` — still a comment. Folded into the 🟡 thread, since the list it points at is now also incomplete |
| 🟢 | The "twelve values" `error_type` table renders 14 rows | Not addressed | `:972` and `:977` are two `bad_request` rows with different definitions; `:979` and `:982` two `not_found` rows. `event.go:114-127` is the authority |
| 🟢 | The doc lists three drop causes; the code has four | Not addressed | `:706` still says "Three things produce a drop today"; `EmitDrop` has four call sites — `emit.go:54` (handler refusal and level filtering), `manager.go:423` (hash failure), `manager.go:471` (constructor rejection) |
| 🟢 | The fail-open policy is never stated | Not addressed | No statement that a destructive call whose audit record cannot be built still executes; that is the right default and the first question a compliance reviewer will ask |
| 🟢 | The redaction list is a hand-copy of `cmd/server/main.go:73` with nothing asserting the two stay equal | Not addressed | Still uncovered, and the second list makes the relationship between the three copies harder to check by eye, not easier |
| 🟡 | The `audit → auth` import edge blocks the auth emitters this schema exists for | Declined | Deferral stands; the condition from last pass (land it before SOL-152097 starts) is the whole of my position and I am not re-arguing it here |
| 🟡 | The `error_type` set is duplicated across two packages and policed by name | Declined | Deferral stands, sequencing noted last pass |

Dropped on re-check: `AuditLogEnabled` is still not cross-validated against `LogLevel` (`config/observability.go:114`), but the design answer shipped — the drop record is `ERROR`, so enabling the audit log at `warn` produces a visible stream of drops rather than silence. That is a better answer than the validation I asked for, so the finding goes rather than being carried.

---

### Minor notes

- 🟢 `internal/observability/schema/version.go:29-30` — the 1.1 delta note lists only `audit_drop`'s three new fields, but `panic_recovered` (`event.go:317`) is also new this cycle. It is also the one schema field outside `applicabilityByType`, so the new fingerprint's nine columns do not cover it and its rule at `event.go:518` can change without `TestApplicabilityByType_isVersioned` noticing — the single hole left in the fix above. Add it to the delta note, and consider rendering the `PanicRecovered` rule into the fingerprint.
- 🟢 `internal/observability/audit/hash.go:66-70` — `RedactSensitive`'s doc comment says keys matching `sensitiveKeySubstrings`; `isSensitiveKey` (`:106`) now also checks the audit-only list. One clause.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ |
| Code Quality | ✅ |
| Test Coverage | ✅ |
| Architecture | — not re-reviewed (unchanged since `e6735690`) |
| Security | ✅ |
| Documentation | ⚠️ |
| PR Scope & Description | — not re-reviewed (unchanged since `e6735690`) |

Sourced `parent (follow-up)`.

docs/observability.md
solace-aross: Fixed: added certcontent/keytab/passphrase to the doc's transform-2 list (marked audit-only, so the seven-pattern list stays a byte-identical mirror of cmd/server's redactedKeys) and replaced the placeholder comment in the Python snippet with a real redact() covering both lists. Addressed in 1f78808.

bczoma (approved):
## PR Review: #381 — SOL-152090 / SOL-152096: enforce the audit-record schema, and emit one for every destructive tool call (Stories 21 and 22)

**Author:** solace-aross | **Branch:** `aross/sol-152090-audit-event-constructor` → `main`
**Follow-up review** at `1f78808` — one commit since my last pass, which `9d62255` is a direct ancestor of, so this reviews `docs/observability.md` (+19/-3) rather than the PR again.

---

### Verdict

**Recommendation:** ✅ Approve

The one condition of the last verdict is closed, and so is the 🟢 that turned out to be its other half.

### Summary

Transform 2 now names the three audit-only terms and says why they are a separate list, and the Python snippet ships a real `redact()` instead of a comment. I checked the snippet against the Go it has to reproduce rather than just reading it: the ten-term `SENSITIVE` tuple matches `sensitiveKeySubstrings` plus `auditOnlySensitiveKeySubstrings` exactly, `any(s in k.lower() ...)` is the same case-insensitive substring test as `isSensitiveKey`, and the dict/list recursion mirrors `redactValue` including the case where a sensitive key's value is itself a map — the whole value becomes the placeholder on both sides. Popping `broker` before redacting is order-independent, since `broker` matches no pattern. The claim that the seven-pattern list stays a byte-identical mirror also still holds: `hash.go:36` and `cmd/server/main.go:74` are the same line.

*Two settled this pass; the rest are unchanged from the previous review and are not re-listed.*

|---|---|---|---|
| 🟡 | The recipe's transform 2 lists seven patterns while the code redacts ten | Addressed | `docs/observability.md:610-614` — the three audit-only terms named, with the mirror rationale |
| 🟢 | The Python recipe implements transform 1 only | Addressed | `:626-634` — `redact()` shipped and verified equivalent to `redactValue`/`isSensitiveKey` |

The 🟢 notes still open from the previous pass (the 14-row `error_type` table, the three-vs-four drop causes, the unstated fail-open policy, the ungarded hand-copy of the redaction list, the `panic_recovered` gap in the 1.1 delta note and the fingerprint, and `RedactSensitive`'s doc comment) are unchanged and remain optional. None of them holds this verdict.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ✅ |
| Correctness, Code Quality, Test Coverage, Architecture, Security, PR Scope | — not re-reviewed (no code changed since `9d62255`) |

Sourced `parent (follow-up)`.

PR #382, opened 2026-09-04
SOL-153341: classify write-tool errors by desired-state outcome

solace-lootii:
## Summary

Write tools now classify certain SEMP failures by what they mean for the caller's *desired state*, not just the raw protocol outcome. Creating an object that already exists, or deleting one that's already gone, no longer surfaces as a hard error — and a genuine "parent object missing" error now names the parent's type in plain language instead of relaying the broker's internal CLI-mode wording.

Implements SOL-153341.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

An agent driving idempotent workflows — replaying its own prior success after a slow response, or issuing a batch of create/delete calls it doesn't know the current state of — had no way to tell "already done" from "actually broken" without a separate read to reconcile every `isError:true`. `ALREADY_EXISTS` on a create and `NOT_FOUND` on a delete both mean the desired state already holds; treating them as hard failures forces every caller to build the same reconciliation logic this server could do once, centrally.

Separately, a `NOT_FOUND` on a **create** doesn't mean the object being created is missing — it can't be "not found" before it exists — it means a *parent* in the path is missing (e.g. creating a queue in a VPN that doesn't exist). The broker reports this as CLI-mode jargon ("Cannot enter mode for message-vpn no-such-vpn: not found."), which an agent would otherwise have to relay to a user verbatim.

## Changes Made

- **`classifyDesiredStateOutcome`** (`internal/tools/errors.go`) — classifies a SEMP error as a desired-state outcome purely from its `SEMPStatus`/`SEMPCode` plus the operation ID's `create`/`delete` prefix. Tool-agnostic: every write tool inherits this with zero `tools.yaml` changes, including tools added after this PR (live-verified against SOL-153868's `create-queue-subscription`/`delete-queue-subscription` with no code changes needed on that side). A delete under a missing *parent* (e.g. `delete-queue` against a VPN that's already gone) classifies the same as an ordinary target-missing delete — standard idempotent-delete semantics, and the safe failure direction (nothing is destroyed that shouldn't be) — documented explicitly in the function's own doc comment rather than left as an unstated side effect.
- **`desiredStateStructuredContent`** — converts a classified outcome into the `isError:false` structured result an agent sees: `outcome` (`"already_exists"` | `"already_absent"`) and `changed:false` alongside the broker's own message. `already_exists` additionally carries `attributes_verified:false` — classification is existence-only (never compares the pre-existing object's attributes against the request; see Additional Notes), and this field says so explicitly instead of letting the value's name imply more than the code checks. `Outcome` is a defined `DesiredState` string type with two constants, not a bare string repeated at four call sites.
- **Output schema widening** — every write tool's declared `outputSchema` now admits this shape via `oneOf` alongside its normal step-keyed result, scoped structurally to tools whose operation is create-/delete-prefixed (`hasDesiredStateEligibleStep`, `internal/tools/composite_handler.go`) — not a hardcoded tool list, so a future write tool is covered automatically. The noop payload is routed through the same `validateAgainstCompiledSchema` path a handler's real result takes (`buildValidatedResult`), so the declared schema and the actual shape can't drift apart unnoticed. (Found in review: without this, the noop result violated every write tool's declared schema outright — missing the required step key, carrying properties `additionalProperties:false` forbids — and a validating MCP client rejects the whole result on any `isError:false` response that mismatches its schema.)
- **`manager.go`** hooks the classifier into `CallTool` right after `handler.Handle` fails, before the generic `execution_error` branch. Dispatch in `logToolResult` and the per-tool metrics/audit recorders keys on `toolErr`'s nilness / a `desiredOutcome` parameter, not on a sentinel `errorType` value — a non-nil `toolErr` means "the call failed" to every consumer that infers from it (the RED metrics from SOL-152086 and the destructive-call audit records from SOL-152090/152096, both merged into `main` during this review), so `errorType` is never touched for this case at all, keeping it inside `metrics.ErrorType`'s closed vocabulary by construction rather than by accident. These two cases now log at `INFO` with a `desired_state` field (`already_exists`/`already_absent`) instead of `ERROR` — named differently from the tool result's own `outcome` field to avoid colliding with the shared three-value `outcome` vocabulary metrics/audit/traces use (`docs/observability.md`).
- **Parent-naming translation** (`translateParentNotFound`, `parentModePattern`, `parentModeFriendlyNames`) — on a create- **or delete**-prefixed operation's `NOT_FOUND`, translates the broker's "Cannot enter X mode: not found" wording into plain language naming the missing parent's type, instead of showing the raw CLI-mode text (delete-prefixed added in review: a delete under a missing parent was classifying correctly but showing raw CLI jargon next to `outcome:"already_absent"`). Two structurally distinct message shapes exist (with/without an inline instance name), and **which shape a command uses depends on the specific command, not the parent's object type** — live-verified that `create-queue`/`create-topic-endpoint` and `create-rdp` against the *same* missing VPN produce different shapes — so both are always tried, never dispatched on the mode word first.
- **AC4 (`MAX_NUM_EXCEEDED` scope + count), deliberately descoped** — see Additional Notes.
- The desired-state-outcome category was already reachable via `errors.As()` (`sempv2.SEMPError`, since SOL-148434) — this PR is the classification built on top of that existing signal, not new access to it (AC5 was already true before this PR).
- Documentation (docs/tools-reference.md's return-shape paragraph, a new "Desired State Already Held" section in docs/user-guide.md, a namespace-separation note in docs/observability.md, and the eight create-*/delete-* tool descriptions in tools.yaml) updated to match — none of these mentioned the new non-failure shape before this pass.

## Testing

**Test Environment:**
- Go version: per `go.mod`
- OS: Linux
- Broker version: Dockerized broker pair for the gating suite (SEMPv2 10.26.5)

**Tests Performed:**
- [x] Integration tests added/updated (e2e-management)
- [x] Tested with real Solace broker (via Dockerized e2e-management)

**Test Coverage:**
- `internal/tools/errors_test.go` / `manager_test.go`: one fixture per AC — `ALREADY_EXISTS`+create, `NOT_FOUND`+delete (target missing and parent missing, the latter added in review), `NOT_FOUND`+create naming a parent (both message shapes), `NOT_FOUND`+**get** stays a hard error (extends the existing `TestCallTool_SEMPErrorWrapped` regression guard rather than duplicating it), the `MAX_NUM_SUBSCRIPTIONS_EXCEEDED` (403) hint, the `attributes_verified` field's presence/absence per outcome, the real-catalog write-operation-prefix invariant, and the real-catalog output-schema widening (validates the actual noop payload against the actual compiled schema for a create and a delete tool).
- `test/e2e-management/test-management-tools.sh`: `test_error_translation` asserts the non-failure shape (`isError:false`, `outcome:"already_exists"`, `attributes_verified:false` for a duplicate `create-message-vpn`); `test_delete_nonexistent_is_noop` (AC2) and `test_create_missing_parent_is_real_error` (AC3, `create-rdp` against a nonexistent VPN — the harder "bare" message shape). **Run for real against Dockerized brokers**: `make e2e-management-all` → **18/18 passing**.
- `internal/composite/executor_vpn_test.go`'s `TestExecute_CreateMessageVPN_AlreadyExists` is **deliberately unchanged** — it asserts the raw, unclassified `SEMPError` still propagates through `Execute()`. Classification lives one layer up, in `internal/tools`, not in the executor; this existing test is the drift guard that keeps it there.
- `golangci-lint run`: clean. Logging surface reviewed against `docs/internal/secure-logging-rules.md`.

## Breaking Changes

**Contract change: a duplicate create or delete-of-missing-object, which previously came back `isError:true` (HTTP 400/404), now comes back `isError:false` with `outcome`/`changed`(/`attributes_verified`) fields instead.** A caller branching only on `isError` without reading `outcome` will now treat these as ordinary successes — the intended behavior change, but confirm nothing downstream depends on the old `isError:true` shape for exactly these two cases. Marked `**BREAKING**` in the CHANGELOG with a `Migration:` line (the `outcome` value itself also renamed from an earlier review round: `"exists_unchanged"` → `"already_exists"`). Operator-visible effect: these two cases move from `ERROR`-level `status=error` log lines to `INFO`-level with a `desired_state` field — a monitor keyed on the old volume should re-key on `desired_state` instead (see CHANGELOG and `docs/observability.md`).

### Code Quality
- [x] Self-review completed
- [x] Code is well-commented

### Security
- [x] No credentials in code

### Testing
- [x] All tests pass locally: `go test -race ./...` (via `make check`)
- [x] Edge cases covered (both parent-naming message shapes on both create and delete, NOT_FOUND-on-get staying a hard error, MAX_NUM_SUBSCRIPTIONS_EXCEEDED hint, attributes_verified presence/absence, output-schema widening)
- [x] Error paths tested (all live-verified assertions run against real Dockerized brokers)

### Documentation
- [x] README.md — no change needed (no new tools, no tool-count change)
- [x] docs/ — updated (tools-reference.md, user-guide.md, observability.md — see Changes Made)
- [x] CHANGELOG.md updated (marked `**BREAKING**` with a `Migration:` line)

### Git & CI
- [x] Commit is signed off (DCO)
- [x] Merged with `main` three times during review (SOL-152086's per-tool metrics; SOL-152090/152096's destructive-call audit records; SOL-152421's request-path tracing + SOL-152097's audit-auth events) — all resolved deliberately, not mechanically: the third merge combined the tracing span wrapping with this branch's own `nil` desiredOutcome argument at each `register.go` dispatch site, matching the pattern the `list-brokers` site already had post-merge without conflict
- [x] No merge conflicts (resolved as of this update; `mergeable: MERGEABLE` on GitHub)
- [x] CI checks passing — `make check` (build/vet/lint/race tests, 89.6% coverage) and `make e2e-management-all` (18/18) both replicated locally against the merged tree and pass, see Testing above

## Additional Notes

**AC4 (`MAX_NUM_EXCEEDED` scope + count) — deliberately descoped, agreed with the ticket owner.** The original ask was to report *which* limit was hit (per-object vs. broker-wide) and its *current count*. Two findings against the actual broker C++ source, verified before writing any code, made that non-trivial:

1. **Subscription limits can't be attributed to a scope after the fact.** Creating a queue subscription is gated by `numQendptSubs_m >= adQueueTopics_g` (broker-wide) `|| topicSet_p->size() >= adPerQueueTopics_g` (per-queue) — an `||`, and the broker doesn't retain which side tripped once the check fails. `MAX_NUM_EXCEEDED` (135) and `MAX_NUM_SUBSCRIPTIONS_EXCEEDED` (403) get identical handling in `adQueueCommand.cpp`. A confident "this was per-queue" or "this was broker-wide" claim here would be a guess dressed up as fact.
2. **An accurate broker-wide count is expensive.** Queues/endpoints share one pool (`adEndpoints_g`); RDPs are gated by their own broker-wide global. Neither has a single SEMP call returning a broker-wide total — an exact count means enumerating every VPN and summing, the same manual exercise as the incident that raised this ticket, just automated (100+ calls on a large broker).

Raised with Balázs Czoma (ticket owner) before implementing either the scope-guessing or the expensive-count version. Agreed outcome: ship a hint that says plainly the tool can't tell, rather than a confident scope/count claim it can't back up — `MAX_NUM_SUBSCRIPTIONS_EXCEEDED` (403) and `MAX_NUM_EXCEEDED` (135) each say so, naming the actual scope pair their own broker-source finding supports (135: broker-wide vs. per-VPN; 403: broker-wide vs. per-queue — the two differ because the underlying ambiguity differs per code, per finding 1 above) — the richer, scope-aware and count-aware version is tracked as its own follow-up spike, **SOL-154047**.

**AC1 ruling, from a later review round.** `already_exists` (originally named `exists_unchanged`) never compares the pre-existing object's attributes against the request — a GET-and-diff per duplicate create would double the call count on exactly the batch-provisioning path this ticket exists to make cheaper, and "identical" is not well-defined against a broker that returns defaulted attributes the caller never sent. Ticket owner's ruling: ship existence-only, but say so explicitly — hence the rename (the earlier name paired "unchanged" with `changed:false` while nothing was compared, which reads as a stronger claim than the code makes) and the new `attributes_verified:false` field. A future ticket that adds a real attribute diff flips it to `true` with no shape change.

## Review Round 2 (aross, bczoma)

Both findings verified independently before being addressed — one by mutation, one by direct code/schema inspection — not taken on the review text alone.

- **aross: nothing pinned that a desired-state noop actually routes through `buildValidatedResult`'s real schema check inside `CallTool`**, rather than a hand-built result that happens to look right. Reproduced the exact mutation aross described (reverted the noop branch to build its own `CallToolResult` directly) and confirmed the full `internal/tools` suite — including both existing noop tests — stayed green; both use a fully permissive output schema specifically because they're about classification/logging, not schema shape, so neither would have noticed. Added `TestCallTool_DesiredStateOutcome_RoutesThroughSchemaValidation`, which installs a schema that would reject the noop's own payload (a required field the noop never sets, `additionalProperties:false`) and asserts `IsError:true` — only possible if the noop path genuinely validates against it. Re-ran the same mutation against the new test: it fails under the bypass, passes once restored.
- **bczoma: a desired-state noop against a destructive tool still writes an `operation` audit record (SOL-152090/152096) with `outcome:success`, with nothing distinguishing it from a real change.** Confirmed directly: `audit.Fields` (`internal/observability/audit/event.go`) has no `desired_state`/`changed` field, and `emitOperationAudit` branches only on whether `toolErr` is nil — which the noop path deliberately makes true, so `desiredOutcome` never reaches it. Per bczoma's own stated preference (the cheaper of his two proposed fixes), this is **documented, not schema-changed**: a new paragraph in `docs/observability.md`'s Outcome Vocabulary section names the gap and the affected code, and gives the join-to-the-log-line workaround; the CHANGELOG's SOL-153341 entry gets the equivalent operator-visible sentence. A real field is left to a future schema-version bump.
- bczoma's three remaining non-blocking notes — `parentModePattern` staying fully anchored rather than a substring match, classification not cross-checking HTTP status against `SEMPStatus`/`SEMPCode`, and `errors.go`'s size (presentation + policy mixed) — are left as-is. All three were explicitly framed as optional or "fair to leave for a follow-up" in the review itself, not addressed here.

## Related Issues/PRs

- Related to SOL-153868 (queue-subscription tools) — inherits this PR's classification with zero code changes on that side; verified via a temporary merge of both branches before either shipped
- Related to SOL-148434 — the underlying `SEMPStatus`/`SEMPCode` structural access this PR classifies on top of
- Related to SOL-152086 (#379, per-tool RED metrics) and SOL-152090/SOL-152096 (#381, destructive-call audit records) — both merged into `main` during this review; this PR's noop dispatch was redesigned so its signal (`desiredOutcome`, `toolErr` nilness) stays outside both features' closed `error_type` vocabularies rather than requiring either to special-case it
- Related to SOL-152421 (request-path tracing) and SOL-152097 (audit auth events) — both merged into `main` during this review; resolved the resulting `register.go` conflict by combining the tracing span wrapping with this branch's own dispatch changes at each call site
- Follow-up spike filed: **SOL-154047** — scope-aware, count-aware `MAX_NUM_EXCEEDED` reporting (AC4's original ask)

---

**For Reviewers:**

**Focus Areas**: the output-schema widening (`oneOf`, scoped via `hasDesiredStateEligibleStep`) and the merge resolution (dispatch on `desiredOutcome`/`toolErr` rather than a sentinel value) are the two changes from the review round most worth a second look, since they touch shared infrastructure (every write tool's schema; the metrics/audit dispatch every tool call goes through) rather than being purely additive.

**Questions**: none blocking.


internal/tools/manager.go
```diff
@@ -261,8 +261,23 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
...
-		errorType = "execution_error"
 		toolErr = fmt.Errorf("executing tool %q: %w", name, handleErr)
+		// SOL-153341: ALREADY_EXISTS on a create and NOT_FOUND on a delete
+		// both mean the desired state already holds — for an idempotent
+		// workflow that's success, not failure. Checked before the generic
+		// execution_error branch so these two cases never reach it; toolErr
+		// stays set to the real underlying error (not nil) so logToolResult
+		// can still classify and log the SEMP fields, just at INFO instead of
+		// ERROR — see its outcome-keyed branch below.
+		if outcome := classifyDesiredStateOutcome(handleErr); outcome != nil {
```
solace-aross, line 384: This keeps `toolErr` non-nil for a case that isn't a failure (per the comment below: "toolErr stays set to the real underlying error... so logToolResult can still classify and log the SEMP fields"). That breaks this codebase's implicit rule that a non-nil `toolErr` means the call failed — and I can show it actually breaking something, not just in theory.

PR #379 (SOL-152086, approved, not yet merged) adds `internal/tools/metrics.go`'s `recordToolInvocation`, which uses exactly that signal to decide the new `mcp_tool_invocation_total` metric's outcome:
```go
outcome := metrics.OutcomeSuccess
if toolErr != nil {
    outcome = metrics.OutcomeError
    et = errorType
}
```
I merged the two branches (`git merge-tree`) and ran it: a duplicate `create-queue-subscription` produces `mcp_tool_invocation_total{outcome="error", error_type="other"}` (the `noop_already_exists` value isn't in #379's closed error-type set, so it collapses to `other`) — while the MCP result is `isError:false` and the log line right next to it says `status:"success"`. That's precisely the "pages dashboards on a false error" problem this ticket exists to fix, just showing up in the metric instead of the log. Neither PR's own test suite catches it, since neither references the other's code; the full merged suite is green.

This is worth fixing here rather than leaving it for whoever rebases second: the root issue is that a non-error case shouldn't be signaled by a non-nil error value at all. Suggest either clearing `toolErr` to nil after building the desired-state result (and passing the SEMP detail `logToolResult` needs through the `desiredStateOutcome` struct instead of threading it through `toolErr`), or adding an explicit signal (e.g. a bool on the result, or a third case alongside "success"/"error") that any current or future consumer of this deferred block can check instead of inferring from `toolErr`'s nilness.

internal/tools/manager_test.go
```diff
@@ -646,6 +646,153 @@ func TestCallTool_SEMPErrorWrapped(t *testing.T) {
...
+	var foundInfo, foundError bool
+	for _, line := range strings.Split(strings.TrimSpace(buf.String()), "\n") {
+		var fields map[string]any
+		if json.Unmarshal([]byte(line), &fields) != nil || fields["msg"] != "tool invoked" {
+			continue
+		}
+		if fields["status"] == "error" {
+			foundError = true
+		}
+		if fields["status"] == "success" && fields["outcome"] == "exists_unchanged" {
```
solace-aross: This checks `fields["status"]` and `fields["outcome"]` but never `fields["level"]` — the key `slog.NewJSONHandler` actually writes for the log level. Changing the noop branch's `slog.LevelInfo` to `slog.LevelError` (leaving the `status`/`outcome` string values untouched) leaves this test green, and the whole package suite green with it. That's the exact claim this test exists to prove (the design note above says logging at INFO instead of ERROR is "per the design that this must not page dashboards keyed on error level") going unverified.

`_AlreadyAbsent` doesn't attempt a logging assertion at all.

Suggest adding `if fields["level"] != "INFO" { t.Errorf(...) }` to both tests — one line each.

solace-aross (changes requested):
Well-designed core idea and unusually thorough investigation (the broker C++ source citations for the AC4 descope are the kind of homework that saves a future reviewer real time). Two issues need a fix before merge; several more are worth a look but don't need to block.

**Blocking**

1. Keeping `toolErr` non-nil for a non-error outcome breaks this codebase's implicit contract that a non-nil `toolErr` means the call failed — and it's not just a style concern. See inline comment on manager.go: this collides with the (separate, approved, not yet merged) PR #379's new tool-invocation metric, which uses exactly that signal.
2. The two new `TestCallTool_DesiredStateOutcome_*` tests don't actually prove the log line lands at INFO rather than ERROR — the one claim the PR's own design note says matters most (must not page dashboards keyed on error level). See inline comment on manager_test.go.

**Notes (non-blocking)**

- The new `outcome` field on the `"tool invoked"` log line (`"exists_unchanged"`/`"already_absent"`) reuses a name `docs/observability.md` documents as a closed, shared 3-value vocabulary (`success`/`error`/`cancelled`) across metrics, audit, and traces. Nothing consumes it live yet (Metrics and Audit Trail are both `[Planned]`), so this isn't a runtime collision today, but whoever implements those stories will find two different things both wanting the same field name. Worth a rename (e.g. `desired_state`) or a docs/observability.md note now, while it's cheap.
- `classifyDesiredStateOutcome`'s correctness depends entirely on every write `operationId` in `tools.yaml` starting with `create`/`update`/`delete`/`do` — verified true today, but only by manual grep, not a test. This codebase already has the pattern for pinning exactly this kind of catalog invariant (`loader_embedded_test.go`, `write_tool_drift_test.go` both load the real embedded catalog and assert structural properties against it). A future non-conforming operation ID would silently stop getting classified, no test failure, no crash — worth one test in that same style.
- Relatedly: this codebase already has an established, explicit mechanism for "is this operation safe to treat as already-applied" — `annotations.idempotent`, enforced at load time in `loader.go`, chosen deliberately over inference ("forces a decision, does not presume the answer"). This PR takes the opposite approach for a closely related question, inferring silently from an operation-ID prefix with no per-tool opt-out. Not asking for a rework, but worth being consistent with the precedent the codebase already set, especially since a future write tool where `ALREADY_EXISTS`/`NOT_FOUND` doesn't mean "already done" would get this silently with no way to say otherwise.
- `exists_unchanged` is reported without checking that the pre-existing object's config actually matches the request (disclosed deliberately in the code comment, so not raising as a defect) — for a security-relevant `create-*` (ACL profile, client username), this hands back a structured "success" signal even if the object that already exists was configured completely differently by someone else. Worth confirming that's the intended read for those tools specifically, since "already exists" and "matches what I asked for" are different claims.
- The PR's own "Breaking Changes" section says "None for a well-behaved caller," immediately followed by a description of exactly the kind of caller-visible change (`isError:true` → `isError:false`) that section exists to flag. The CHANGELOG entry has the same gap: thorough on mechanism, but it never states the one sentence that matters to a caller checking only `isError` — that they'll now see these two cases as ordinary successes. Worth adding since the CHANGELOG is the durable record here (no other doc in the repo mentions `isError` semantics at all).
- `translateParentNotFound`'s formatted output skips `sanitizeBrokerText`, the one gate every other broker-derived agent-facing string in this file goes through. Low risk today (the `instance` value is just an echo of a caller-supplied name, not broker-internal data), but it's a silent exception to an otherwise-universal invariant — worth routing through the same sanitizer for consistency.

Happy to re-review once the two blocking items are addressed.

internal/tools/errors.go
```diff
@@ -176,6 +235,110 @@ func (m *ToolManager) buildErrorResult(err error, brokerAlias string) *mcp.CallT
...
+// behind "reported as a non-failure": an agent branching on IsError sees a
+// completed step, while the outcome/changed fields still let it report
+// accurately that nothing new happened rather than claiming a fresh success.
+func buildDesiredStateResult(outcome *desiredStateOutcome) *mcp.CallToolResult {
+	structured := map[string]any{
+		"outcome": outcome.Outcome,
+		"changed": false,
+		"message": outcome.Message,
+	}
+	return &mcp.CallToolResult{
```
bczoma: 🔴 **The noop result's `structuredContent` is forbidden by the schema this tool advertises**

Flagged by good:code + good:architecture, independently. Challenged, and it got worse on inspection.

**The problem.** `buildDesiredStateResult` returns `StructuredContent = {outcome, changed, message}` with `IsError:false`. Every composite write tool's declared `OutputSchema` comes from `composite.BuildStrictOutputSchema` (`internal/composite/schema.go:39-60`, routed at `internal/tools/composite_handler.go:110-114`), which unconditionally emits `required: [<stepID>]` and `additionalProperties: false`. The noop payload satisfies neither clause: it omits the required step key and adds three properties the schema forbids. And because `manager.go:287` returns inside the classify branch, it never reaches the output-validation gate at `manager.go:308` — so the server's own validator never sees it either.

Reproduced against the real embedded catalog, not reasoned about: `create-queue` gives `(root): createQueue is required; (root): Additional property changed is not allowed; (root): Additional property message is not allowed; (root): Additional property outcome is not allowed`, and `delete-queue` the same with `deleteQueue is required`.

**Why it matters.** The "we already ship an off-schema payload in `buildErrorResult`" defence — which I went looking for — does not hold, and the reason is precise rather than theoretical. Claude Code's MCP client, the primary consumer here, validates `structuredContent` against the advertised `outputSchema` and throws `InvalidParams` on mismatch, with the check explicitly gated on `!isError`. So `buildErrorResult`'s shape has been exempt since SOL-148425 for exactly the reason it carries `isError:true`; flipping these two cases to `isError:false` moves them into the one bucket that *is* checked. The result is that on every duplicate create and every delete of a missing object, a validating agent gets a hard protocol error in place of the result and never sees `outcome`, `changed`, or the message — which is worse than the pre-PR behaviour, where it at least received a readable `isError:true` result it could reason about. That is precisely the AC6 failure mode this ticket exists to remove.

Nothing recovers it: the `TextContent` half is set (`errors.go:337`), but the client throws on the whole `CallToolResult` before that text is surfaced. No retry, no fallback, and the e2e suites read with `jq` and never validate, so the gating suite cannot catch this.

**Recommended fix.** Widen the *declared* contract, not just the behaviour — a `oneOf`/`anyOf` at the top level admitting both the step-keyed envelope and `{outcome, changed, message}`, or declaring the three fields optional and dropping the step key from `required`, in `outputSchema()` / `writeToolIdentifierFields`. Then route `buildDesiredStateResult` through `validateAgainstCompiledSchema` so the mismatch cannot recur unnoticed. Note that validating *without* widening is not sufficient on its own — it would turn every noop into an `output_validation_error`. Pin it with a test in the `composite_handler_test.go` output-schema family that validates the noop payload against the real catalog's compiled schema, in the style of `TestCompositeToolHandler_OutputSchema_RealCatalogWiring`.

This is also the reversibility argument: a field that `tools/list` never declared, but that callers come to depend on, is a silent break when withdrawn. Declaring it now is what makes it deprecable later.

internal/tools/manager.go
```diff
@@ -171,6 +171,10 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
 	start := time.Now()
 	var brokerAlias, errorType string
```
bczoma: 🟡 **This no longer merges, and the conflict is semantic rather than textual**

Flagged by the parent review (verified live, not from the PR body).

`gh pr view 382` now reports `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`, and `git merge-tree --write-tree origin/main d336e71` conflicts in three files: `internal/tools/manager.go`, `internal/tools/register.go`, `internal/tools/describe_semp_schema.go`.

The checklist's claim — "No conflicts with `main`; checked with `git merge-tree --write-tree` (clean)" — was true when you wrote it. #379 (SOL-152086) merged as `398010d` at 20:05:24Z, three minutes after `d336e71`. Nothing was wrong with the verification; it just has a very short shelf life on this repo.

**Why it is worth a thread rather than a "please rebase".** #379 retyped this exact line. On `main`, `errorType` is no longer a `string` — it is `metrics.ErrorType` (`internal/observability/metrics/instruments.go:42`), a closed set of thirteen validated at emit time, where `Record` coerces anything outside `knownErrorTypes` to `other` (`instruments.go:118-119`). This PR writes two values into that variable which are not in the set (`noop_already_exists`, `noop_already_absent`) — and, worse for the concept, `errorType` is documented there as "the failure cause on an error outcome", which a desired-state noop is not.

Today's behaviour after a mechanical rebase is *accidentally* fine: `recordToolInvocation` keys `outcome` off `toolErr`, which you now correctly clear, so `et` stays empty and the noop labels never reach `Record`. The safety rests entirely on the untyped-const assignment compiling and on nobody ever recording `error_type` independently of `toolErr`'s nilness. That is a coincidence, not a design — and it is the same class of coupling @solace-aross's blocking comment was about, one layer over.

**Recommended fix.** Rebase onto `main`, then resolve it deliberately rather than mechanically: keep the noop signal out of `metrics.ErrorType` entirely. The thread on `manager.go:281` proposes the shape that makes this fall out for free — dispatch on `desiredOutcome != nil` and leave `errorType` empty for a noop — after which the closed metrics vocabulary keeps exactly its documented thirteen values and there is nothing to reconcile.

Since #379 is now on `main`, its `recordToolInvocation` is also finally testable against this branch — worth adding the merged-suite check @solace-aross ran by hand as a real test while you are in there.

internal/tools/errors.go
```diff
@@ -176,6 +235,110 @@ func (m *ToolManager) buildErrorResult(err error, brokerAlias string) *mcp.CallT
...
+	case isSEMPStatus(sempErr, "ALREADY_EXISTS", 10) && strings.HasPrefix(sempErr.Operation, "create"):
+		return &desiredStateOutcome{
+			Outcome:    "exists_unchanged",
+			Message:    buildSEMPv2Message(sempErr),
+			SEMPCode:   sempErr.SEMPCode,
+			SEMPStatus: sempErr.SEMPStatus,
+			Operation:  sempErr.Operation,
+			Detail:     sempErr.Error(),
+		}
+	case isSEMPStatus(sempErr, "NOT_FOUND", 6) && strings.HasPrefix(sempErr.Operation, "delete"):
```
bczoma, line 343: 🟡 **A delete under a *missing parent* classifies as `already_absent`, and that choice is recorded nowhere**

Flagged by good:code + good:architecture at 🔴, independently; reduced to 🟡 by challenge. I am reporting the reduction as well as the finding, because the fix both lenses proposed would have made things worse.

**The mechanism, confirmed.** This branch is `isSEMPStatus(sempErr, "NOT_FOUND", 6) && strings.HasPrefix(sempErr.Operation, "delete")` — `Description` is copied into `Detail` but never inspected. Three of the four classified delete tools are child resources parented on `msgVpnName` (`deleteMsgVpnQueue`, `deleteMsgVpnTopicEndpoint`, `deleteMsgVpnRestDeliveryPoint`); `delete-message-vpn` is the only unparented one, and it is the one the new e2e noop test uses. `delete-queue-messages` is `do`-prefixed and never classified. There is no pre-flight parent read anywhere on the path, so SEMP's status is the only thing between the broker and `already_absent`.

**Why it is not 🔴.** `already_absent` asserts that the object at the addressed path does not exist — and if the parent VPN does not exist, the child at that path genuinely does not exist. That is conventional idempotent-delete semantics and it is what AC2 asks for. The harm case needs a *second* condition on top of the trigger: the caller must have meant a different, existing VPN. In the common teardown case — the VPN was already removed by an earlier step or a peer agent — the current behaviour is correct, and falling through to a hard error (the originally proposed fix) would regress it. The failure direction is also the safe one: nothing is destroyed that shouldn't be, objects merely survive.

**What actually survives, and it is three cheap things.**

1. The message is left as raw CLI jargon. The AC3 translation at `errors.go:585` is gated to `create`-prefixed operations, so a parented delete falls through to `sanitizeBrokerText(err.Description)` — the agent gets `outcome: "already_absent"` sitting next to `message: "Problem with DELETE: Cannot enter message-vpn mode: not found."`. Self-contradictory, and it is the only signal the typo case ever produces, visible only to a caller reading the prose AC6 tells it it can skip.
2. The choice is undocumented. The doc comment at `errors.go:262-286` reasons carefully about attribute matching and about why reads are excluded, and never asks the parent question — which is the one asymmetry with the create side that a reader will trip over.
3. No test pins it. `errors_test.go:245` uses the target-missing shape (`"Could not find match for queue …"`); no delete case uses a parent-mode description, so a future change to this branch has nothing to fail against.

**Recommended fix.** Keep `already_absent`. Extend the message translation at `errors.go:585` to `delete`-prefixed operations so the result reads `outcome: "already_absent"` with `message: The Message VPN "x" does not exist.` — truthful, plain language, and it makes the typo case legible without breaking the legitimate case. Record the parent-vs-target reasoning in the classifier comment. Add a classify test with `Operation: "deleteMsgVpnQueue"`, `SEMPStatus: "NOT_FOUND"`, `Description: "Problem with DELETE: Cannot enter message-vpn mode: not found."` asserting the intended outcome and message. And add the missing CHANGELOG sentence: a delete under a nonexistent parent also reports `already_absent`.

internal/tools/manager.go
```diff
@@ -261,8 +265,28 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
...
+		// execution_error branch so these two cases never reach it. toolErr
+		// is cleared back to nil here, not just left set: a non-nil toolErr
+		// means "the call failed" to every other consumer in this codebase
+		// that infers from it — including SOL-152086's per-tool metrics,
+		// which key outcome=error off exactly that signal — and this isn't a
+		// failure. The SEMP fields logToolResult still needs travel via
+		// desiredOutcome instead of toolErr; see its outcome-keyed branch.
+		if outcome := classifyDesiredStateOutcome(handleErr); outcome != nil {
+			if outcome.Outcome == "exists_unchanged" {
+				errorType = errorTypeNoopExists
```
bczoma: 🟡 **The replacement for `toolErr` is two signals that must agree, plus a round-trip through a third**

Flagged by good:code + good:architecture + good:security, independently — three lenses arriving at the same seam from different directions is why it is here rather than in the minor notes.

The fix for @solace-aross's blocking comment is right in substance: `toolErr` is genuinely cleared, and the SEMP fields travel in a purpose-built value instead of being smuggled through the error. The shape it left behind is what this thread is about.

**Two costs.**

1. **The discriminator got weaker.** `logToolResult`'s success/failure test moved from `*toolErr == nil` to `*errorType == ""`. The old test was structurally guaranteed — every error return set `toolErr`. The new one is a convention asserted in a comment across four files (`manager.go`, `register.go` ×2, `describe_semp_schema.go`); today every path honours it (I checked all four), but `manager.go:268` genuinely does set `toolErr` before `errorType`, so a `return` inserted in that window logs a real failure as `status:"success"` at INFO. A silently lost audit line is the hardest kind of regression to notice, and nothing tests the invariant.
2. **The round-trip is dead weight.** `outcome.Outcome` is mapped *to* two `errorType` sentinels here at :281-283 and mapped straight *back* to the same two strings at :408-410. Two independent mappings that must agree, and the `if outcome != nil` guard at :422 is unreachable-false on every real path.

**Recommended fix.** Key the noop branch on the parameter that actually carries the signal, and keep `toolErr`-nilness as the failure test:

```go
if outcome != nil {           // desired-state noop → INFO + desired_state
    ...
    return
}
if *toolErr == nil {          // ordinary success → INFO
    ...
    return
}
// failure → ERROR
```

That deletes both `errorTypeNoop*` constants, both mappings and the unreachable nil guard; it restores a structural discriminator; and it keeps the noop values out of `errorType` altogether, which is what the `manager.go:172` thread needs once you rebase onto #379's typed `metrics.ErrorType`. One change settles three findings.

If you prefer to go further, good:architecture's suggestion is to encode the tri-state as one `invocationOutcome` value (`kind: success | desiredState | failure`, plus err/errorType/SEMP fields/id) built in `CallTool` and switched on once — which also collapses the eight-parameter, three-out-param signature and gives #379's `recordToolInvocation` a single thing to read. Worth it if you are touching this again anyway; the guard-clause reorder above is the minimum.

internal/tools/errors.go
```diff
@@ -176,6 +235,110 @@ func (m *ToolManager) buildErrorResult(err error, brokerAlias string) *mcp.CallT
...
+// SEMPStatus/SEMPCode, and a new create-*/delete-* tool inherits the
+// classification automatically with no tools.yaml change.
+//
+// Matches SEMPStatus first (already-parsed, self-documenting); falls back
+// to the pinned SEMPCode (6=NOT_FOUND, 10=ALREADY_EXISTS — see
+// TestCallTool_SEMPErrorWrapped and executor_vpn_test.go's
+// TestExecute_CreateMessageVPN_AlreadyExists) in case a broker version ever
+// omits the status string.
+//
+// Deliberately does not verify the existing/absent object's attributes
```
bczoma, line 304: 🟡 **`exists_unchanged` asserts an equivalence the code never checks — and as ticket owner, here is my ruling on AC1**

Flagged by good:code + good:security. @solace-aross raised it as non-blocking; this thread is not a restatement, it is the decision, since the comment here argues with an acceptance criterion I wrote and that is mine to settle.

**The disagreement in the comment.** It says the attribute diff is out of scope because "the ticket's own Proposal section never mentions an attribute diff, only the two SEMP status codes". The Proposal is a sketch; the acceptance criteria are the contract, and AC1 reads "Creating an object that already exists, **with identical attributes**, is reported as a non-failure distinguishable from a real error." Arguing from the sketch against the criterion is the wrong way round.

**But you are right that the check is not worth building here,** and I would rather say so than have you build it. A GET-and-diff per duplicate create doubles the call count on exactly the batch-provisioning path this ticket exists to make cheaper, and "identical" is not well-defined against a broker that returns defaulted attributes the caller never sent.

**So the ruling is: ship existence-only, and stop claiming more than that.** What is not acceptable is the pairing of the word "unchanged" with `changed:false` while nothing was compared — that is a structured assertion the code cannot support, and an agent acting on AC6 will skip precisely the read that would have caught the divergence. good:security's concrete case: `create-queue`'s `queueConfig` carries `permission` and `accessType`, so a caller asking for `permission: no-access` against a pre-existing queue with `permission: delete` is told `exists_unchanged`, `changed:false` — a structured claim that a security-relevant configuration is in force when it may not be. No privilege is gained (the caller could not have created it either), so this is a false assurance rather than an authz bypass — which is why it is 🟡 and not 🔴.

**Recommended fix**, either way round:

- Rename the outcome to `already_exists` and add `attributes_verified: false` to the structured result, so the agent knows reconciliation is still its job for a case it cares about; or
- keep `exists_unchanged` and make the same statement in the field documentation and the CHANGELOG — that it means "an object of this name exists", not "the object matches your request".

Whichever you pick, add a test that pins the chosen semantics, and update the four `create-*` descriptions in the same pass as the `tools/list` wording fix in the review body — the description is where an agent would look for this caveat.

I will amend AC1 on SOL-153341 to say existence-only with the successor noted, so verification does not later fail against text we have both agreed to move past. That amendment is mine to make, not a task for this PR.

internal/tools/manager.go
```diff
@@ -358,6 +394,46 @@ func logToolResult(ctx context.Context, tool string, broker *string, start time.
...
+			desiredState = "already_absent"
+		}
+		attrs := make([]slog.Attr, 0, 8)
+		attrs = append(attrs, slog.String("tool", tool))
+		if *broker != "" {
+			attrs = append(attrs, slog.String("broker", *broker))
+		}
+		attrs = append(attrs,
+			slog.String("status", "success"),
+			slog.String("desired_state", desiredState),
```
bczoma: 🟡 **The operator-facing half of this change is documented nowhere an SRE will look**

Flagged by good:security (Response & Monitoring) + docs.

The caller-facing half of this change is now well documented — the CHANGELOG's "one sentence that matters to a caller checking only `isError`" is exactly right, and it was the right response to the earlier review. The operator-facing half has no equivalent, and it is a real change in what the log stream says.

**What changed for a monitor.** Before this PR, a sweep of `delete-queue`/`delete-message-vpn` across guessed names produced a burst of ERROR lines with `status:"error"` — a legible enumeration signal, and the natural thing to alert on. After it, the same sweep is a burst of `status:"success"` INFO lines. The signal is not *lost* — it is in `desired_state` and `semp_status` — but a SIEM rule keyed on the documented `status=error` convention will not see it at all, and one keyed on `status=success` cannot separate it from genuine deletes. The same applies to existence probing via duplicate `create-*`.

**Why it is undiscoverable.** `docs/observability.md` documents no `desired_state` field, and nothing in `docs/` mentions the value strings. So an SRE upgrading keeps their ERROR-level dashboard for duplicate-create volume and loses it without knowing it moved; and one who reads the CHANGELOG and greps the log stream for `outcome=exists_unchanged` gets zero hits, because in the log the key is `desired_state` while on the wire it is `outcome`. The rename was the right call for the vocabulary collision, but it means the CHANGELOG's field name is not the field name an operator will grep for, and nothing says so.

**Recommended fix.** Add `desired_state` to `docs/observability.md`'s log-field table, with the explicit note that a no-op create/delete is a `status=success` INFO line, and name the rule an operator should now write instead (a per-principal no-op-delete rate). Add two sentences to the CHANGELOG entry: the level and `status` change for these two cases, and that the log field is `desired_state` while the tool-result field is `outcome`. One further sentence under `docs/observability.md:600` would close @solace-aross's vocabulary point on the doc side as well as the code side — that the write-tool result `outcome` and the log's `desired_state` are a separate namespace and never appear as telemetry `outcome` values.

Optional, and it would make the two surfaces uniform: emit `changed:false` on the log line too, so the same discriminator works in the log and in the result.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
... The category was already available structurally via `errors.As()` (`sempv2.SEMPError`, since SOL-148434) — this change is the classification built on top of that signal, not new access to it. `MAX_NUM_SUBSCRIPTIONS_EXCEEDED` (403) now gets a curated hint alongside `MAX_NUM_EXCEEDED` (135), and both say plainly that the tool can't tell which of two scopes was hit rather than staying silent about it. A richer version of AC4 — naming which limit was hit (per-object vs. broker-wide) and its current count — was scoped out after review with the ticket owner: subscription creation (403) checks a broker-wide total and a per-queue total with `||`, and the broker itself doesn't retain which side tripped, so a confident claim there would be a guess between broker-wide and per-queue; the other limit families 135 covers (queues, endpoints, REST delivery points) are broker-wide totals allocated per-VPN, and an accurate broker-wide count has no single SEMP call, only a full VPN enumeration, so the guess there is between broker-wide and per-VPN. That richer scoping/counting is tracked as its own follow-up spike rather than guessed at or built at real per-error cost here. Tracked under SOL-153341.
```
bczoma: 🟡 **The durable record covers the caller and is silent on the operator and on the delete-parent case**

Flagged by docs.

This entry is genuinely good on the contract change — the explicit "**The one sentence that matters to a caller checking only `isError`:**" clause is the right shape and it closed the earlier review's complaint properly. Three gaps remain, and the CHANGELOG is the only durable record of any of them once this PR closes:

1. **The observability change is absent.** These two cases moved from ERROR to INFO and from `status:"error"` to `status:"success"`, and gained a `desired_state` field. See the thread on `manager.go:419` — the entry names the field `outcome`, which is correct for the tool result and wrong for the log stream an SRE will grep.
2. **The delete-parent case is not stated.** The entry explains the create-side parent reasoning explicitly and well ("`NOT_FOUND` on a create — meaning a *parent* is missing, since the object being created can't itself be 'not found' yet"), but describes the delete side only as "deleting one that is already gone". A delete under a *nonexistent parent* also reports `already_absent` (thread on `errors.go:302`), which is the sentence a teardown-loop author actually needs.
3. **It is not marked BREAKING and carries no `Migration:` line.** Raised as a 🟢 in the body rather than here, but it belongs in the same edit: this repo's own `.claude/skills/changelog/SKILL.md:82-91` defines a changed result contract as BREAKING, and `:105-110` requires a migration note on one. Either mark it, or record why the "well-behaved caller" reading overrides the criterion — the point is that the release-time SemVer decision should be deliberate rather than inherited from an unmarked entry in a ~2,000-character bullet.

bczoma (changes requested):
## PR Review: #382 — SOL-153341: classify write-tool errors by desired-state outcome

**Author:** solace-lootii | **Branch:** `lkiri/sol-153341-error-classification` → `main` | **Changes:** +786 / -25 across 8 files
**Jira:** `SOL-153341` — Write tools report operation outcomes where an agent needs desired-state outcomes (Verifying)

---

### Verdict

**Recommendation:** 🚫 Request changes

One finding blocks: the new `isError:false` result violates the output schema this server advertises for the tool in `tools/list`, and the client that drives this server rejects such a result outright. Everything else here is a 🟡 or a note.

1. 🔴 `internal/tools/errors.go:335` — the noop result's `structuredContent` is forbidden by the write tool's own declared `OutputSchema`; widen the schema (and pin it with a test) rather than only widening the behaviour
2. 🟡 `internal/tools/manager.go:172` — the PR no longer merges: #379 landed on `main` three minutes after your last push and retyped the very `errorType` variable this code writes two new values into
3. 🟡 `internal/tools/errors.go:302` — a `delete-*` under a *missing parent* classifies as `already_absent`; the classification is defensible but undocumented, untested, and its message is left as raw CLI jargon
4. 🟡 `internal/tools/manager.go:281` — the replacement for `toolErr` is two signals that must agree, plus a string round-trip that maps `outcome` → `errorType` → `outcome`
5. 🟡 `internal/tools/errors.go:282` — `exists_unchanged` + `changed:false` asserts an equivalence the code never checks; my ruling as ticket owner on AC1 is in the thread
6. 🟡 `internal/tools/manager.go:419` — the operator-facing half of this change (ERROR → INFO, and `desired_state` vs `outcome`) is documented nowhere an SRE will look
7. 🟡 `CHANGELOG.md:36` — same gap in the durable record: the entry covers the caller and is silent on the operator

- 🟡 `internal/composite/definitions/tools.yaml:806, :896-897, :1067-1068, :1183` — all four `create-*` descriptions still say "Fails if … already exists", the four `delete-*` descriptions say nothing about the new non-failure, and `docs/tools-reference.md:906-908` still states write tools return "a step-keyed envelope … an empty object for `delete-*`". These are the served contract and the published reference; the new result shape is documented in neither, nor in `docs/user-guide.md`, whose NOT_FOUND bullet at `:257` ("Detect this from `sempCode`/`sempStatus`") is now false for a delete. Reword the eight descriptions, fix the return-shape paragraph, and add a short "desired state already held" subsection to `docs/user-guide.md` after "Tool Returns an Error".

**On the existing blocking review:** @solace-aross's two blocking items are genuinely addressed at `d336e71`, at the seam she asked for — I checked both in code, not in the commit message. `toolErr` is cleared to nil and the SEMP fields travel in `desiredOutcome` (`manager.go:279-289`), and both noop tests now assert `fields["level"] == "INFO"` (`manager_test.go:712`, `:786`) in a branch that survives a `LevelInfo → LevelError` mutation, so the assertion can actually fail. Her non-blocking `sanitizeBrokerText`, log-field-name and prefix-invariant points are addressed too. My 🚫 is on new ground, not a restatement of hers.

---

### Summary

This is a well-designed change with an unusually well-evidenced descope, and the review-fixup commit fixed the right things at the right seams rather than papering over them. What blocks it is one seam nobody had reason to look at: making these two cases `isError:false` moves them from a bucket where an off-schema `structuredContent` is expected and exempt into the one bucket a validating client actually checks — and the write tools' declared schema is strict (`required:[<stepID>]`, `additionalProperties:false`), so it positively forbids the `{outcome, changed, message}` body the new path returns. Three of the four most alarming findings in the first pass were reduced by measurement: a claimed data-loss-grade delete bug turned out to be a defensible semantic choice whose only real fault is being undocumented, and a claimed agent-misleading description turned out to be corrected by the result's own prose on the same turn. The one that survived got worse on inspection, not better — it was reproduced against the real embedded catalog, and the rejecting validator was found in the client.

---

### Minor notes

- 🟢 `internal/tools/errors.go:333` vs `manager.go:419` — one concept, two names: `outcome` on the wire, `desired_state` in the log. Rename the wire key too (still pre-GA) or record the asymmetry in `docs/observability.md`.
- 🟢 `internal/tools/errors.go:292, :301` — `"exists_unchanged"`/`"already_absent"` are bare literals in four production places; make `Outcome` a defined string type with two constants so `manager.go:280`'s equality test can't silently fall through on a typo.
- 🟢 `internal/tools/errors.go:103` — `parentModePattern` is anchored to the exact literal `^Problem with \w+: ` … `: not found\.$`; any broker wording drift silently stops AC3 holding with no test failure. Match the `Cannot enter …: not found` substring instead and keep the mode-word map as the guard.
- 🟢 `internal/tools/errors.go:287-315` — classification now trusts broker-supplied `SEMPStatus`/`SEMPCode` to decide success vs. failure, where before those fields only shaped text. Requiring the HTTP status to agree (400/409 for `ALREADY_EXISTS`, 404 for `NOT_FOUND`) keeps a mismatched pair on the error path.
- 🟢 `CHANGELOG.md:36` — not prefixed `- **BREAKING**:` and carries no `Migration:` line, although `.claude/skills/changelog/SKILL.md:82-91` defines a changed result contract as BREAKING. Either mark it or record why the "well-behaved caller" argument overrides the criterion, so the release-time SemVer decision is deliberate.
- 🟢 PR body — the AC4 note still quotes the superseded flat hint ("A configured maximum was reached.") that `cad2668` replaced with the two per-code scope-ambiguity hints, and still says the follow-up spike is "to be filed" when it is already SOL-154047, split from SOL-153341. Also: "Breaking Changes: None for a well-behaved caller" immediately above a description of a caller-visible contract flip reads as a contradiction — lead with the flip.
- 🟢 `test/e2e-management/README.md:53` — the test-catalog row still documents `duplicate create → isError=true`, and the two new tests have no rows; a maintainer debugging a red assertion may "fix" it back. `test-management-tools.sh:290-292` also leaves the old contract asserted in the present tense directly above the new paragraph that contradicts it.
- 🟢 `internal/tools/errors.go` (682 lines) — the file now holds both presentation (sanitizing, message building, the hint table) and policy (is this even an error). A pure move of `desiredStateOutcome`/`classifyDesiredStateOutcome`/`buildDesiredStateResult` into `internal/tools/desired_state.go` would keep the policy findable.

…and 3 further minor items (the SEMPv2-only scope of the classifier being worth one sentence for whoever writes the first SEMPv1 write tool; the `SubexpNames()` loop and per-call re-capitalization in `translateParentNotFound`; the assign-then-undo of `toolErr` at `manager.go:267-286`) — ask if you want them.

---

### Summary Verdict

|---|---|---|
| Correctness | ❌ | good:code |
| Documentation | ⚠️ (filed ❌; reduced after the served-description finding was measured down to 🟡, but four documentation gaps stand) | docs |

solace-aross (dismissed):
Both blocking items are fixed, and verified concretely rather than by reading the diff:

1. `toolErr` is now cleared to nil for a desired-state noop, and dispatch in `logToolResult` keys off `*errorType` instead of `*toolErr`'s nilness — exactly closing the invariant break, and doing it better than I asked (a real fix in the shared contract, not a workaround). Since #379 has now merged, I re-ran the actual cross-PR check for real this time: merged this branch onto current `main` in a throwaway worktree, hand-resolved the four resulting conflicts (all mechanical — combining `recordToolInvocation`'s call sites with `desiredOutcome`'s threading), built, ran the full suite (green), then wrote a scratch test calling `create-queue-subscription` through the merged code and scraping `/metrics`. Result: `mcp_tool_invocation_total{outcome="success", error_type=""}`, matching the log line's `status=success` and the API's `isError=false`. The bug is gone in the actual combined codebase, not just in theory.
2. Both `TestCallTool_DesiredStateOutcome_*` tests now assert `fields["level"] == "INFO"`, and `_AlreadyAbsent` gained the logging assertion it was missing entirely.

Also folded in, cleanly: renamed the log field to `desired_state` so it doesn't collide with docs/observability.md's reserved `outcome` vocabulary (confirmed this was worth doing now — it would have collided with #379's own `outcome` field on the very same log line the moment both merged, not just a future risk); a new test pinning the create/update/delete/do operation-ID prefix invariant against the real embedded catalog, same pattern as `write_tool_drift_test.go`; `translateParentNotFound` now routes through `sanitizeBrokerText`; and the CHANGELOG states the one sentence that matters to an `isError`-only caller.

Left open, correctly: the `annotations.idempotent` precedent and attribute-diff gating for security-relevant `create-*` tools are flagged as judgment calls for the ticket owner rather than folded in here, which is the right call for a PR that's already grown substantially fixing genuine defects.

CI green. Approved.

solace-aross (commented):
(test review comment retracted immediately — full review follows)

internal/tools/manager.go
```diff
@@ -343,8 +347,40 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
...
+		// not on errorType or toolErr's value — so neither is ever set to a
+		// sentinel outside metrics.ErrorType's closed set (SOL-152086):
+		// toolErr goes back to nil (a non-nil toolErr means "the call
+		// failed" to every other consumer that infers from it, including
+		// SOL-152086's per-tool metrics) and errorType is simply never
+		// touched, staying at its zero value.
+		if outcome := classifyDesiredStateOutcome(handleErr); outcome != nil {
+			desiredOutcome = outcome
+			toolErr = nil
+			result, valErr := m.buildValidatedResult(rt, name, desiredStateStructuredContent(outcome), false)
```
solace-aross, line 387: This now routes through `buildValidatedResult`, same as a real handler result — good, that's the fix for the blocking finding. But nothing currently pins the *routing* itself: I temporarily made this branch build the `CallToolResult` directly, skipping `buildValidatedResult`/`validateAgainstCompiledSchema` entirely (reproducing the pre-fix bypass), and the full `internal/tools` suite still passed. The two existing noop tests install a fully permissive stub schema (their own comment: "this test isn't about schema shape"), and the schema-regression test calls `ValidateOutput` directly without going through `CallTool` at all — so today's payload happens to satisfy the schema, but nothing would catch it if payload and schema ever drift apart again.

Worth a follow-up test that drives a noop through the real `CallTool` path with a schema that would reject a malformed noop payload, asserting an error result — that's the one test that actually pins routing rather than shape.

solace-aross (dismissed):
Verified bczoma's blocking finding and all 7 non-blocking findings by injection in a throwaway worktree, not by reading the fixup commit messages.

**Blocking finding — fixed.** `hasDesiredStateEligibleStep` + the `oneOf` schema widening are real; `TestCompositeToolHandler_OutputSchema_RealCatalogAdmitsDesiredStateNoop` validates the actual noop payload against the real compiled `create-queue`/`delete-queue` schemas. Sabotaged the predicate to always return false — test failed with the right error, not incidentally. A real MCP client no longer gets a hard protocol error on a duplicate create or an absent delete.

**All 7 non-blocking findings addressed.** Sabotage-tested three: the metrics-vocabulary/drift-guard fix (reverted `buildValidatedResult` to return `metrics.ErrorType` directly — `TestErrorTypeVocabularyMatchesAuditConstructor` failed with the exact diagnostic the commit message describes), the delete-under-missing-parent message translation (narrowed the gate back to `create`-only — exactly one subtest failed), and confirmed the schema-widening predicate above. Read-verified the rest (outcome rename, `attributes_verified` field, `DesiredState` type, docs, e2e README/script updates) — all present, nothing claimed but missing.

CI green; `make e2e-management-all` couldn't be run in this environment (no Docker), so the author's 18/18 claim is unverified by me — noted, not held against the PR.

Two follow-up notes below, neither blocking — the actual bugs are fixed and well-tested; these are missing regression guards against future drift.

Approved.

One more note, not anchorable to this diff (the code in question is unchanged from main, so GitHub won't let me thread it inline): Fresh finding, not in bczoma's list: the second fixup commit's message claims `emitOperationAudit`'s own `toolErr != nil` check means a desired-state noop correctly produces an `audit.OutcomeSuccess` record — I wrote a throwaway test exercising this directly (destructive `delete-queue` returning NOT_FOUND, audit enabled, via `CallTool`) and confirmed it: exactly one `operation` record, `outcome=success`, no `error_type`. But nothing in the repo pins this — it's correct-by-code-reading only. This is exactly the kind of cross-feature seam (this PR's noop reclassification meeting SOL-152090/152096's audit records, merged in from main mid-review) that regresses silently on a later refactor. Worth adding as a permanent case in `destructive_audit_test.go` alongside its siblings.

internal/tools/manager.go
```diff
@@ -343,8 +347,40 @@ func (m *ToolManager) CallTool(ctx context.Context, name string, params map[stri
...
+		// Dispatch here and in logToolResult keys on desiredOutcome != nil,
+		// not on errorType or toolErr's value — so neither is ever set to a
+		// sentinel outside metrics.ErrorType's closed set (SOL-152086):
+		// toolErr goes back to nil (a non-nil toolErr means "the call
+		// failed" to every other consumer that infers from it, including
+		// SOL-152086's per-tool metrics) and errorType is simply never
+		// touched, staying at its zero value.
+		if outcome := classifyDesiredStateOutcome(handleErr); outcome != nil {
+			desiredOutcome = outcome
+			toolErr = nil
```
bczoma, line 386: 🟡 **A desired-state noop audits as an ordinary successful destructive operation**

Raised by this follow-up pass. It is not a regression in your fix — it is this branch meeting `48ca806` (SOL-152090/152096, the audit-record story) which landed on `main` after my last review, so neither change could have seen it.

**What happens.** This branch nils `toolErr` for a noop, which is correct and is what makes the caller-facing contract and #379's metrics right. `emitOperationAudit` then reads that same nilness — `if toolErr != nil { fields.Outcome = audit.OutcomeError }` — so a duplicate `create-queue` or a `delete-message-vpn` against a VPN that is already gone emits an `operation` record with `outcome: success`, and `desiredOutcome` is not passed to it. Before this PR both cases audited as `outcome: error`.

**Why it matters.** The log line got the distinction (`desired_state`) and the tool result got it (`outcome`, `changed:false`), but the audit record — the durable, schema-enforced one, and the artifact SOL-152096 exists to make complete for destructive calls — did not. An auditor reconstructing what an agent actually changed reads `delete-message-vpn` / `outcome=success` and concludes this call deleted that VPN. It did not: the VPN was already gone, possibly removed by someone else, possibly never created. For a destructive-action trail, "I deleted it" and "it was already gone" are materially different claims, and the record now makes them identical.

**Two ways out, and I would take the second unless you disagree.**

1. Add the signal to the record — a `desired_state` (or `changed`) field on the `operation` event. `audit.Fields` (`internal/observability/audit/event.go:292-323`) has nothing that can express it today, so this means the struct plus the field-applicability table at `event.go:232`, and it is a schema change to a surface the doc describes as constructor-enforced.
2. Document the ambiguity instead: one paragraph in `docs/observability.md`'s audit section saying that a successful `operation` record does not distinguish a real change from a desired-state noop, and that a reader who needs that must join to the `tool invoked` log line for the same `correlation_id`. Cheap, honest, and it puts the decision on the record before the schema freezes.

Either is fine by me; leaving the reader to discover it is the thing I would not ship. If you take (2), the CHANGELOG's operator-visible sentence is the natural second home for it.

bczoma (dismissed):
## PR Review: #382 — SOL-153341: classify write-tool errors by desired-state outcome

**Author:** solace-lootii | **Branch:** `lkiri/sol-153341-error-classification` → `main` | **Changes:** +1228 / -64 across 15 files
**Jira:** `SOL-153341`
Follow-up review — parent-only pass over the two commits pushed since my review at `d336e71` (both merges from `main`, carrying the fixes). No lenses dispatched.

---

### Verdict

My 🚫 lifts. The 🔴 is properly fixed and all six 🟡s are addressed — several of them better than the fix I suggested. One new item, from this branch meeting the audit story that landed on `main` in the meantime:

1. 🟡 `internal/tools/manager.go:365` — a desired-state noop now audits as an ordinary successful destructive operation, with nothing in the record saying nothing changed

Verified at this head: `go build ./...` clean, `go test ./internal/tools/` passes, `mergeable: MERGEABLE`.

**Not an unblock of @solace-aross's thread:** their comment at this head on `manager.go:366` — the noop branch now correctly routes through `buildValidatedResult`, but nothing pins that routing, and they mutation-proved the whole `internal/tools` suite stays green with the bypass reinstated — is a separate gap from mine and I am not re-filing or answering it here. It is worth taking; it is theirs.

---

### Summary

The 🔴 was fixed at the right seam. Rather than loosening `required`/`additionalProperties` on the strict step-keyed schema — which would also have admitted a malformed hybrid carrying both shapes or neither — the widening is a `oneOf` gated by a structural `hasDesiredStateEligibleStep` test, so no per-tool list has to stay in sync and a future write tool is covered by construction. Two tests pin it: one asserts eligible tools declare a `oneOf`-widened schema, the other asserts the declared branch actually admits what `desiredStateStructuredContent` produces for both outcomes.

The 🟡 on the weakened discriminator came back stronger than I asked for. I suggested keeping the two signals from disagreeing; the fix removed the disagreement instead — a purpose-built `desiredOutcome *desiredStateOutcome` threaded to `logToolResult`, `toolErr` genuinely back to `nil`, `errorType` never touched, and dispatch keyed on the outcome value plus `*toolErr`'s nilness rather than on an `errorType` string round-trip. That also leaves #379's `recordToolInvocation` reading the case correctly with no coordination, which was the substance of @solace-aross's original blocking comment. On AC1, the ruling was implemented as ruled and then recorded where it will survive: `already_exists` no longer asserts anything about attributes, `attributes_verified: false` states the existence-only limit explicitly instead of leaving it implied, and `classifyDesiredStateOutcome`'s doc comment carries the reasoning and the cost argument rather than the ticket alone.

---

*12 of 15 prior findings settled — the 🔴 and all six 🟡s addressed, plus five of eight 🟢 notes.*

Closed, with the evidence, since these were the substance of the pass:

|---|---|---|---|
| 🔴 | Noop `structuredContent` forbidden by the write tool's own declared `OutputSchema` | Addressed | `internal/tools/composite_handler.go:152-165` — `oneOf` widening gated by `hasDesiredStateEligibleStep`; schema at `internal/tools/errors.go:406-421`; pinned by `composite_handler_test.go:457` and `errors_test.go:426-441` |
| 🟡 | No longer merges — #379 retyped the `errorType` this writes two new values into | Addressed | `mergeable: MERGEABLE` at `8701422`; `internal/tools/manager.go:213-240` threads `desiredOutcome` separately and leaves `errorType` at its zero value, so `recordToolInvocation` classifies correctly |
| 🟡 | `delete-*` under a missing parent classifies `already_absent`, undocumented and untested | Addressed | `internal/tools/errors.go:315-329` records the decision, the AC2 basis and the safe failure direction; CHANGELOG names the case; message translated by `buildSEMPv2Message`'s widened parent-naming gate |
| 🟡 | Two signals that must agree, plus a string round-trip `outcome`→`errorType`→`outcome` | Addressed | `internal/tools/manager.go:218`, `:363-365`, `:569-570` — purpose-built third value; dispatch keyed on outcome + `*toolErr` nilness, not on `errorType` |
| 🟡 | `exists_unchanged` + `changed:false` asserts an equivalence the code never checks | Addressed | Renamed `already_exists`; `internal/tools/errors.go:262-269` adds `AttributesVerified`, `:305-313` records the AC1 ruling and why a GET-and-diff was rejected |
| 🟡 | The operator-facing half documented nowhere an SRE will look | Addressed | `docs/observability.md:1049-1062` — `outcome=success` plus a separate `desired_state`, why it is not a fourth `outcome` value, and the re-keying advice for a monitor |
| 🟡 | The durable record covers the caller and is silent on the operator | Addressed | `CHANGELOG.md` — now `**BREAKING**`, carries the operator-visible effect and a `Migration:` line naming both the value rename and the log re-key |

Still open:

|---|---|---|---|
| 🟢 | `parentModePattern` anchored to exact broker wording; any drift silently stops AC3 holding | Partially addressed | `internal/tools/errors.go:94` — now covers both observed message shapes with named captures, and `:672-677` makes the mode-word map the stated safety net. Still anchored `^Problem with \w+: ` … `: not found\.$`, so the substring match I suggested is not there |
| 🟢 | Classification trusts broker `SEMPStatus`/`SEMPCode` alone to decide success vs. failure | Not addressed | `internal/tools/errors.go:335-348` still keys only on status/code + operation prefix; requiring the HTTP status to agree (400/409 for `ALREADY_EXISTS`, 404 for `NOT_FOUND`) would keep a mismatched pair on the error path. Offered as an option, not a condition |
| 🟢 | `errors.go` now holds both presentation and policy | Not addressed | 777 lines, up from 682; no `desired_state.go`. Pure-move refactor, entirely fair to leave for a follow-up |

---

### Minor notes

- 🟢 `internal/observability/audit/event.go:292-323` — `audit.Fields` has no field that can carry "nothing changed", which is why the thread below proposes the doc route as the cheaper of the two fixes. Worth deciding deliberately before the audit schema freezes rather than after.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ — `go build` clean, `go test ./internal/tools/` passes at this head |
| Code Quality | ✅ — the discriminator fix is better than the one proposed |
| Test Coverage | ⚠️ — the widened schema is pinned; @solace-aross's separate routing gap stands |
| Architecture | ✅ — structural gating, no per-tool list to keep in sync |
| Security | — not re-reviewed (unchanged since `d336e71`) |
| Documentation | ⚠️ — operator-facing gap closed; the audit-record ambiguity below is new |
| PR Scope & Description | ✅ — AC4 note corrected, SOL-154047 named |

Sourced: parent (follow-up).

solace-lootii:
Both items addressed, verified independently rather than taken on the review text — details in the new "Review Round 2" section in the PR body, short version:

**aross** — reproduced your exact mutation (bypassed `buildValidatedResult`, confirmed the suite stayed green including both existing noop tests) and added `TestCallTool_DesiredStateOutcome_RoutesThroughSchemaValidation`, which uses a schema that would reject the noop's own payload and asserts `IsError:true` — only passes if the noop path genuinely validates against it. Confirmed it fails under the same bypass, passes restored.

**bczoma** — confirmed directly that `audit.Fields` has no field for this and `emitOperationAudit` only branches on `toolErr`'s nilness, so your finding stands. Took your stated preference (document, don't schema-change): added a paragraph to `docs/observability.md`'s Outcome Vocabulary section plus the equivalent CHANGELOG sentence, naming the gap and the join-to-the-log-line workaround. A real `audit.Fields` addition is left for a future schema-version bump. Your three remaining non-blocking notes are left as-is, per your own framing of them as optional/follow-up.

Also resolved: `main` picked up SOL-152421 (request-path tracing) and SOL-152097 (audit auth events) since the last merge, both touching `register.go`'s dispatch signatures — merged and resolved by combining the tracing span wrapping with this branch's own dispatch changes at each site. `make check` (89.6% coverage) and `make e2e-management-all` (18/18) both green against the merged tree.

Re-requesting review since the push reset both approvals.

internal/tools/manager.go
solace-lootii: Reproduced this exactly: bypassed `buildValidatedResult` for the noop path, confirmed `internal/tools` stays green (both existing noop tests use a fully permissive schema, so neither would catch it). Added `TestCallTool_DesiredStateOutcome_RoutesThroughSchemaValidation`, which uses a schema that would reject the noop's own payload (a required field it never sets, `additionalProperties:false`) and asserts `IsError:true` — only possible if the noop path genuinely validates against it. Re-ran your mutation against the new test: fails under the bypass, passes restored.

internal/tools/manager.go
solace-lootii: Confirmed directly — `audit.Fields` has no `desired_state`/`changed` field, and `emitOperationAudit` only branches on `toolErr`'s nilness, so this stands exactly as described. Took your stated preference: documented rather than schema-changed. Added a paragraph to `docs/observability.md`'s Outcome Vocabulary section naming the gap, the affected code, and the join-to-the-log-line workaround, plus the equivalent CHANGELOG operator-visible sentence. Left a real `audit.Fields` addition for a future schema-version bump rather than freezing one in under review pressure.

solace-aross (approved):
Both open items from the last round are resolved. Verified by injection, not by reading the commit message:

**My finding (validation-routing untested)** — fixed. `TestCallTool_DesiredStateOutcome_RoutesThroughSchemaValidation` installs a schema that rejects the noop's own payload and asserts `IsError:true`. Reproduced the exact bypass I sabotaged before (`buildValidatedResult` swapped for a hand-built `CallToolResult`) — the two pre-existing noop tests stayed green (their permissive schema still doesn't notice), but the new test failed with the precise diagnostic it's designed to give. Reverted, all three pass.

**bczoma's finding (noop audit record indistinguishable from a real change)** — documented per his own stated preference rather than schema-changed. Confirmed the `docs/observability.md` paragraph and the CHANGELOG's "Known gap, documented rather than fixed here" sentence are both present and accurate: `audit.Fields` genuinely has no `desired_state`/`changed` field today, and the workaround (join to the "tool invoked" log line's `desired_state` field via `correlation_id`) is stated plainly.

His three remaining non-blocking notes are left as-is, correctly — he already framed all three as optional in his own review, not mine to re-litigate.

Approved.

bczoma (approved):
## PR Review: #382 — SOL-153341: classify write-tool errors by desired-state outcome

**Author:** solace-lootii | **Branch:** `lkiri/sol-153341-error-classification` → `main` | **Changes:** +1305 / -64 across 15 files
**Follow-up review** — parent-only pass over the one commit since my last look (`8701422`…`928293c`). No lenses dispatched.
**Jira:** `SOL-153341` — classify write-tool errors by desired-state outcome

---

### Verdict

**Recommendation:** ✅ Approve

Nothing outstanding. The one 🟡 I left last round is addressed, the new commit's own work is 351 lines across four files once `main`'s merge content is factored out, and I checked all four of them.

---

### Summary

This round is a merge of `main` plus two review responses, and the merge is the part that could have gone wrong. `main` brought SOL-152421 (request-path spans) and SOL-152097 (audit auth events), both of which retyped the same `register.go` dispatch call sites this branch was already editing. The resolution combines both changes at each site rather than picking one, and I verified that across every dispatch site rather than the one that conflicted: all four now carry the `tracer.Start` / `logToolResult` / `recordToolInvocation` / `endDispatchSpan` sequence, with `nil` passed for `desiredOutcome` at the three sites that cannot produce one and the real value only at `manager.go:260`. `go build ./...` is clean and `go test ./internal/tools/` is green at the new head.

The audit-record gap is closed the way I asked for it. I offered two fixes and named the documentation route as the cheaper one; that is the route taken, and it is done properly rather than minimally — the new paragraph names the two code sites responsible (`audit.Fields` having no such field, `emitOperationAudit` branching only on error), states the consequence in compliance terms rather than mechanical ones, gives the join-to-the-log-line workaround, and records that a real field is a future schema-version bump. That last clause also settles the 🟢 I raised about deciding this deliberately before the audit schema freezes: it is now a recorded decision rather than an unexamined omission. I checked the workaround actually works — the audit record takes `correlation_id` from the same source the correlation slog handler injects into every log record, so the two really are joinable.

The other reviewer's finding produced the more interesting change. `TestCallTool_DesiredStateOutcome_RoutesThroughSchemaValidation` installs an output schema the noop's own payload cannot satisfy — a required field it never sets, plus `additionalProperties: false` — and asserts `IsError: true`, which is only reachable if the noop genuinely validates against the registered schema. That is a test that can only pass for the right reason, and it closes a real hole: the two existing noop tests install a permissive schema because they are about classification, so neither would have noticed a regression to the pre-fix bypass.

My three remaining non-blocking notes — leaving `parentModePattern` fully anchored, not cross-checking HTTP status against `SEMPStatus`/`SEMPCode`, and `errors.go`'s size — are consciously left as-is, which is the right call: I framed all three as optional or fair to defer, and the commit message says so explicitly rather than quietly skipping them.

---

*Six prior findings settled — 2 addressed, 4 accepted as-is. No open objections.*

|---|---|---|---|
| 🟡 | A desired-state noop against a destructive tool audits as an ordinary successful operation, with nothing in the record saying nothing changed | Addressed | `docs/observability.md:1208-1220` — new Outcome Vocabulary paragraph naming `audit.Fields` and `emitOperationAudit`, the compliance consequence, and the `correlation_id` join; equivalent operator-visible sentence added to the `CHANGELOG.md` SOL-153341 entry. Join key verified present on both surfaces |
| 🟢 | `audit.Fields` has no field that can carry "nothing changed" — decide deliberately before the audit schema freezes | Addressed | `docs/observability.md:1219-1220` — recorded as tracked-not-fixed, with the future field named as a schema-version bump |
| 🟢 | `parentModePattern` fully anchored, so broker wording drift silently stops AC3 holding | Accepted as-is | Offered as optional last round; deliberately left, and said so in `928293c`'s message |
| 🟢 | Classification trusts `SEMPStatus`/`SEMPCode` without requiring the HTTP status to agree | Accepted as-is | Same — framed as fair to leave for a follow-up |
| 🟢 | `errors.go` (682 lines) now holds both presentation and policy | Accepted as-is | Same |

---

### Minor notes

🟢 **Minor** — worth fixing, not a condition of the verdict:

- 🟢 `docs/observability.md:1217` — the join-to-the-log-line workaround silently depends on `OBS_CORRELATION_ID_ENABLED`. It is on by default so the advice holds for almost everyone, but a deployment that turned it off gets no `correlation_id` on either surface and the documented workaround has no join key — with nothing saying why. Half a clause: "join to the … log line for the same `correlation_id` (present when `OBS_CORRELATION_ID_ENABLED` is on, the default)".

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ (parent, follow-up — merge resolution verified at all four dispatch sites; build and `internal/tools` tests green at `928293c`) |
| Test Coverage | ✅ (parent, follow-up — the one new test asserts an outcome only reachable if the noop path validates) |
| Documentation | ✅ (parent, follow-up — the new audit-gap paragraph and CHANGELOG sentence check out against the code) |
| Code Quality | — not re-reviewed (unchanged since `8701422`) |
| Architecture | — not re-reviewed (unchanged since `8701422`) |
| Security | — not re-reviewed (unchanged since `8701422`) |
| PR Scope & Description | — not re-reviewed (unchanged since `8701422`) |

PR #385, opened 2026-09-08
SOL-152421: create request-path spans at every layer

solace-awaheed:
## Summary

Adds OpenTelemetry spans at every layer of the request path — HTTP boundary, tool dispatcher, composite executor, and each SEMP call — so a failed MCP call reads as one trace instead of scattered log lines. Also installs the W3C trace-context propagator, which was never called before, so a trace started by an AI agent now continues unbroken into this server.

Jira: **SOL-152421** (Story 26). No GitHub issue.

## Type of Change

- [x] Documentation update
- [x] Dependency update

## Motivation and Context

A trace is what turns "the call failed" into "the SEMP call against broker B failed after the composite executor's second step". Two concrete gaps this closes:

1. **`otel.SetTextMapPropagator` was never called.** OTel's global default propagator is a no-op, so an inbound W3C `traceparent` was silently discarded and every entry span began a new root. An agent-initiated trace broke at this server's edge with no error and no log line — the failure was invisible to inspection.
2. **`outcome`/`error_type` existed on metrics and audit records but not on traces**, so the shared vocabulary ADR-009 exists for couldn't actually be carried from a dashboard into a trace backend.

## Changes Made

- Spans at four layers, each with its own named tracer so a backend attributes them to their source: `POST /mcp` (SpanKind Server, named per the OTel `{method} {route}` convention), `tools.CallTool`, `composite.Execute`, `semp.request` (SpanKind Client).
- Installs `propagation.TraceContext` as the global propagator — deliberately **not** composed with `Baggage`, which would carry arbitrary upstream key-values this server has no redaction policy for.
- `outcome` and `error_type` use the same closed vocabulary as the metric labels and the audit record, **enforced by test** rather than by convention (see Testing).
- All **four** tool-dispatch sites emit the span, not just `ToolManager.CallTool`. `list-brokers` and `describe-semp-schema` are registered directly against the MCP server, and an unparseable `arguments` payload is rejected before dispatch — each already emitted its own audit record and metric for exactly that reason. Extracted `internal/tools/spans.go:endDispatchSpan` as the shared helper.
- A panicked call is never exported as a success. Go leaves named returns unpopulated when a panic unwinds, so a span deriving `outcome` from the returned error would close as `success` moments before the process died. Each span recovers, records the error, and re-panics — the pattern review established on `tokenexchange.Exchange`.
- The SSE notification stream is not traced. The streamable transport holds a `GET /mcp` open for a whole session; spanning it produced a span reported only at session end, lost entirely if the pod died first, and long enough to swamp any latency view built on entry-span duration. `GET` is filtered; `POST` carries the requests, and a short-lived `DELETE` teardown is still traced.
- New direct dependency: `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`.

## Testing

**Test Environment:**
- Go version: 1.26.5 (module targets 1.25.0)
- OS: macOS 26.6.2 (darwin/arm64)
- Broker version: n/a — no real broker involved (see below)

**Tests Performed:**
- [ ] E2E tests added/updated — not applicable, no tool-surface change; E2E suites not run locally
- [x] Tested locally with `go test -race ./...` (via `make check`: build, vet, lint, race tests, coverage gate — exit 0, 0 lint issues, 89.3% coverage against the 85% floor)
- [ ] Tested with real Solace broker — used `httptest` SEMP fakes; no real broker exercised

**Test Coverage:**

24 new tests. The load-bearing ones:

- **Span hierarchy** asserted edge by edge, not by shared trace ID — a flat trace of four sibling roots is exactly the shape a stray `context.Background()` produces, and it would pass a trace-ID check.
- **Propagation both ways**: inbound `traceparent` continues the upstream trace; absent, the entry span is a new root. Also asserts `baggage` is *not* in the propagator's fields.
- **Cross-signal join**: one real call driven end to end, comparing all four shared keys (`tool`, `broker`, `outcome`, `error_type`) between the span and the `mcp_tool_invocation_total` series that same call produced — across success, tool failure, broker-resolution failure, and panic.
- **Chain composition** driven through the real `buildMCPEndpoint`, pinning tracing inside correlation and outside cross-origin, plus the tool-RED gauge's outermost position from both sides (a 413 and a 403 are both still counted).
- **Parallel-step context threading**, which is where the request context is easiest to lose silently.
- **`http.Flusher`/`Hijacker` preservation** across GET/POST/DELETE against a real server, so SSE notifications can't silently start buffering.

Each fix was verified by mutation — reverting it makes the corresponding test fail with the diagnostic that names the defect. Three bugs were caught this way and fixed:

1. The span carried the **raw caller-supplied** broker string where the metric carried the canonical label — a join that broke on caller casing alone, and an unresolvable broker put arbitrary caller input into an attribute bound for the collector. Both now use `canonicalBrokerLabel`.
2. `composite.Execute` and both SEMP clients reported `outcome: success` on a panic.
3. Three of the four dispatch sites had no span at all — `list-brokers` appeared in every dashboard and in no trace, and `bad_request`/`not_found` were metric-only values of a vocabulary documented as shared by all three signals.

## Breaking Changes

None. With `OBS_TRACING_ENABLED` off (the default) the assembled middleware chain is byte-identical to before this change, and no spans are created.

### Code Quality
- [x] Self-review completed

### Security
- [x] Followed secure logging rules — `/check-logs` on the staged diff: 21 files, 2343 added lines, **0 violations** at every severity. All nine new attribute keys clear the `ReplaceAttr` substring net.
- [x] Credential-carrying types implement `slog.LogValuer` — no new credential-carrying types added
- [x] No security vulnerabilities introduced (golangci-lint: 0 issues)

Two deliberate egress decisions worth a reviewer's eye:
- A failed span carries status `codes.Error` with an **empty** description and no recorded exception. Error text can quote a broker response body, and spans travel to whatever collector `OTEL_EXPORTER_OTLP_ENDPOINT` names; `error_type` carries the classification and the detail stays in the audit record.
- `semp.operation` is the spec operationId, never the resolved URL, whose path parameters interpolate customer topology.

### Testing
- [x] No race conditions detected (also clean at `-race -count=3` across all touched packages)
- [x] Error paths tested

### Documentation
- [ ] README.md updated — not applicable; `OBS_TRACING_ENABLED` was already documented
- [x] docs/ updated — `docs/observability.md`: span table, attributes, the SSE exclusion, the panic-fidelity rule, and the client-IP data-flow note
- [x] CHANGELOG.md updated
- [ ] Examples updated — not applicable

### Git & CI
- [x] Branch is up to date with main (0 behind)
- [x] No merge conflicts
- [ ] CI checks passing — not yet run; local `make check` (CI's build/lint/test jobs) is green, E2E jobs unverified locally

## Additional Notes

**Data-flow note for review.** Enabling `OBS_TRACING_ENABLED` exports the caller's IP address per request, via otelhttp's standard `client.address` and `network.peer.address` attributes, alongside `user_agent.original`. An IP address is personal data under GDPR and comparable regimes, so it is named here rather than left to be discovered; drop those attributes at your collector if a data-flow review would rather not hold them. Verified by test as **not** exported: the `Authorization` header, cookies, and the URL query string (`url.path` is recorded, not `url.query`).

**Known limitation — `semp.request` is per call, not per attempt.** Retries happen below this method inside `resilience.Sender.Do`, so one span covers a whole retry chain: a call that succeeded on its third attempt is one span, not three. The reserved `semp.attempt` name is deliberately left unused for Story 27 (SOL-152422); naming this span `semp.attempt` would have mislabelled a retry chain and left that story no name.

**Not emitted:** the `cancelled` `outcome` value, which `docs/observability.md` reserves for a later release (Story 42). Tool spans report `success` or `error` only.

**A hop-1 denial produces only the entry span.** Authorization is composed outside the dispatch instrumentation and short-circuits without dispatching, so there is no `tools.CallTool` span; the denial is recorded as an `authz_denied` audit event instead (Story 23). This is by design, not a gap.

## Related Issues/PRs

- Depends on #379 (SOL-152086, per-tool RED metrics) — merged; unblocked the span-to-metric join test and the tool-RED composition test
- Follows SOL-152420 (Story 25, tracer provider) and SOL-153333 (Story 50, token-exchange span)
- Unblocks SOL-152422 (Story 27, per-attempt SEMP spans) and SOL-152419 (Story 47, trace exemplars)
- Relates to SOL-154036 (cross-signal vocabulary test) — the span-to-metric leg is now covered here, so that ticket reduces to the audit leg

---

**For Reviewers:**

**Focus Areas**
1. **`endDispatchSpan` call sites** (`internal/tools/spans.go` and its four callers). The deferred call must be registered *before* each site's audit/metric defer so it runs *after* it (LIFO) — that defer rewrites `errorType` for a recovered panic. Reversed, the span silently reports nothing while the metric and audit record both say `panic`. Pinned by the `panic` subtest of the cross-signal test.
2. **The SSE `GET` filter** in `internal/observability/tracing/httpmw.go`. If the streamable transport ever moves request traffic to `GET`, this filter would stop tracing it.
3. **The propagator choice** — `TraceContext` alone, no `Baggage`. Worth confirming no downstream consumer expects baggage to survive this hop.

**Questions**
1. AC 7 states the chain order as `panic, correlation, tracing, tool-RED, auth, SDK`. The order actually assembled puts the tool-RED gauge **outermost** (outside `limitRequestBody`, where #379 composes it) and additionally includes `limitRequestBody` and `crossOriginProtection`. The tests pin the real order — should the AC text be corrected?
2. Is `semp.request` (per call) vs `semp.attempt` (per attempt, Story 27) the intended split? The story's own work-breakdown step 3 says "per-attempt", which would collide with Story 27.


test/integration/request_path_spans_test.go
```diff
@@ -0,0 +1,804 @@
...
+	if _, ok := spanAttr(dispatch, "error_type"); ok {
+		t.Errorf("tools.CallTool set error_type on a successful call; it is documented as present on outcome=error only")
+	}
+}
+
+// documentedErrorTypes is the closed error_type vocabulary, built from the
+// metrics.ErrorType consts so that this test and the metric label can never
+// disagree about what the set is. ErrorTypeOther is omitted on purpose: it is
+// the coercion sentinel, never a classification a span should report.
+var documentedErrorTypes = map[metrics.ErrorType]bool{
```
bczoma: 🟡 **The shared `error_type` vocabulary has four hand-maintained copies and no test that any two agree**

Flagged by good:code and good:architecture.

**The problem.** `documentedErrorTypes` is a hand-written map literal, and the only assertion is membership — `documentedErrorTypes[metrics.ErrorType(gotType)]` — of whatever single value one exercised path produced. The vocabulary therefore exists as four independent copies with no derivation between them: the `metrics.ErrorType` consts (`internal/observability/metrics/instruments.go:45-59`), `knownErrorTypes` (same file, `:64-70`), this map, and the 12-row table in `docs/observability.md:922-935`.

Two comments in this file promise otherwise. At `:388-391`: *"built from the metrics.ErrorType consts so that this test and the metric label can never disagree about what the set is"*, and at `:420-425`: *"adding a value in one place without the other fails here"*. The keys being consts rather than string literals does buy something real — a const **rename** propagates — but the quoted clause about adding a value is verifiably false.

**Measured, not asserted.** In an out-of-tree copy of this head (your worktree untouched) a thirteenth const `ErrorTypeBrokerPermissionDenied` was added to the type and *deliberately not* to `knownErrorTypes`, this map, or the doc table. `go build ./...`, `go vet`, and `go test ./...` across the whole module all passed clean. The mechanism the comment describes is provably absent.

**Why it matters, and it is wider than a doc-review gap.** `internal/tools/spans.go:82` writes the attribute raw — `attribute.String("error_type", string(errorType))` — while `ToolMetrics.Record` (`instruments.go:117-119`) coerces anything absent from `knownErrorTypes` to `ErrorTypeOther`. So a const added to the type but missed in `knownErrorTypes` yields `error_type="other"` on the metric and `error_type="broker_permission_denied"` on the span **for the same call** — silently defeating the exact pivot `docs/observability.md:698-701` sells as the point of the whole change ("filter a dashboard by `error_type=…` and you can carry that predicate into the trace backend and the SIEM unchanged, with no translation table"). An SRE sees a rising `other` bucket with no matching span filter, nothing fails in CI, and it persists from the merge of the adding story until someone reads the metric and the span for one call side by side.

The false comments make this *less* likely to get fixed, not more: the next implementer sees a guard that appears to exist.

**Recommended fix.** Make the consts the single source — export `func AllErrorTypes() []ErrorType` (or a package-level slice) and build `knownErrorTypes` from it, so the coercion set can no longer lag the type. Then two set-equality assertions: one in the metrics package between `AllErrorTypes()` and the coercion set, one here between `documentedErrorTypes` and `AllErrorTypes()` minus `ErrorTypeOther`. A new const then fails immediately, which is what SOL-152421's own note is counting on when it says this test "will fail when Story 49 lands — by design". Either delete both overstated comments or narrow them to what membership actually proves.

**Two notes on the ticket, separately.** Its 2026-09-04 correction says the committed vocabulary is **eleven** values; the consts, `docs/observability.md` and this test all agree on **twelve**, consistently, and `broker_permission_denied` is absent from all three. The repo is right and the ticket note is the stale artifact — worth correcting so the next alignment audit does not "fix" a correct doc back down to eleven.

internal/tools/register.go
```diff
@@ -248,6 +249,18 @@ func RegisterWithServer(mgr *ToolManager, server *mcp.Server, pool *semp.BrokerP
...
+
+					// And the span, for the same reason (SOL-152421): this
+					// return never reaches CallTool, so without it
+					// `bad_request` is a metric-only value of a vocabulary
+					// documented as shared by all three signals. Started
+					// with the request's own start time so the span covers
+					// the dispatch rather than reading as instantaneous;
+					// created only on this branch, since the success path
+					// below gets its span from CallTool itself.
+					_, span := tracer.Start(ctx, dispatchSpanName, trace.WithTimestamp(start))
```
bczoma: 🟡 **The dispatch seam owns the body of the instrumentation but not the ordering it depends on — and this fourth site is inverted**

Flagged by good:architecture, with good:code and good:security independently confirming the concrete case.

**The problem.** `endDispatchSpan` shares the *body* of the dispatch instrumentation across four sites but not its ordering invariant. That invariant — "register this defer **before** the audit/metric defer so LIFO runs it **after**, once `errorType` has been rewritten for a recovered panic" — is enforced only by a repeated comment at `manager.go:189`, `register.go:357` and `describe_semp_schema.go:337`. Those three are correct.

This fourth site cannot express the pattern at all. In the `arguments` unmarshal-failure branch, `logToolResult` and `recordToolInvocation` run at `:251-252`, *then* `tracer.Start` at `:261` discards the context it returns (`_, span :=`), and the helper is called inline rather than deferred. So the seam presents four "dispatch sites" as uniform when they have two genuinely different shapes: three defer-ordered, one straight-line and inverted.

**Copilot's comment on this line is correct, and its consequence is narrower than stated.** The ordering is as Copilot describes. But with tracing on, the request context at that point still carries the otelhttp **entry** span, so a Story 47 exemplar attaches to the entry span rather than to nothing. What is actually lost is dispatch-span *granularity* for the `bad_request` path — a fidelity gap on one error branch, not the silent-zero-exemplars defect the comment implies. Worth knowing so the fix gets scoped correctly rather than triggering a restructure.

**Why it matters.** An author adding a fifth dispatch site reads the helper's doc comment, gets a shared body, and still has to get LIFO right by hand — and this branch is the existing proof that getting it wrong is easy and invisible. Story 27 adds that fifth site.

**Recommended fix.** Make the seam own both halves: `beginDispatch(ctx, tool, …) (context.Context, func())` that starts the span, returns the span-carrying context, and returns one closure emitting the audit record, the metric and the span close in the correct order. No site can then invert it, this branch fits the same shape as the other three, and Copilot's point is resolved structurally rather than by moving two lines. If you would rather not reshape the seam in this PR, the minimum is to start the span first here and thread its context into both calls — but then say in the helper's comment that the four sites are not uniform.

docs/observability.md
```diff
@@ -652,8 +700,42 @@ audit record**, which is the point of a single vocabulary: filter a dashboard by
...
 
+This is enforced, not merely intended: a test drives one real tool call and asserts that all
+four shared keys — `tool`, `broker`, `outcome`, `error_type` — hold identical values on the
+span and on the `mcp_tool_invocation_total` series that call produced. Nothing in the type
+system couples the two, since the span writes an attribute and the metric writes a Prometheus
+label from a different call site, so the two can drift while each surface still looks healthy
+on its own. `correlation_id` is span-only by design: it is per-request, and as a metric label
+it would be unbounded.
+
+**Which spans carry which.** Every span in the table above carries `outcome`. Only
```
bczoma: 🟡 **"Every span carries `outcome`" is false for the entry span — which is the only span a denied request produces**

Flagged by docs.

**The problem.** The new sentence reads *"**Which spans carry which.** Every span in the table above carries `outcome`."* The entry span does not. `internal/observability/tracing/httpmw.go` sets only `correlation_id`; everything else on `POST /mcp` comes from otelhttp's HTTP semconv, which has no `outcome` attribute. No test asserts otherwise — `test/integration/request_path_spans_test.go:487` checks the entry span but not that key.

**Why it matters.** `outcome` being one predicate across metrics, audit and traces is this document's whole selling point. An SRE told every span carries it writes a trace-backend filter `outcome = "error"` to find failed requests — and it silently drops every failure that produced *only* an entry span: the 403 cross-origin rejection, the 413 body-limit rejection, and the hop-1 `authz_denied` case. Those are exactly the failures where there is no `tools.CallTool` span, and this same document tells the reader two paragraphs later that a lone `POST /mcp` span is the normal shape for a denial. So the reader is told both that the filter is safe and that the case it misses is routine, and the failures read as zero.

**Recommended fix.** Either scope the sentence — "Every span in the table above **except the entry span** carries `outcome`; on `POST /mcp` the result is expressed by the span status and otelhttp's `http.response.status_code`" — or set `outcome` on the entry span in `HTTPMiddleware` and pin it by test. Either is fine; the doc has to match whichever you pick.

docs/observability.md
```diff
@@ -667,6 +749,17 @@ values match OTel's `error.type` semantics. If your trace backend or trace-based
 the dotted `error.type`, tell us in your feedback, because this is the kind of thing that is
 cheap to change now and expensive after the freeze.
 
+**Enabling `OBS_TRACING_ENABLED` exports client network identity to your collector.** The entry
+span carries `otelhttp`'s standard OTel HTTP server attributes, and two of them —
+`client.address` and `network.peer.address` — are the **caller's IP address**, alongside
+`user_agent.original`. An IP address is personal data under GDPR and comparable regimes, so this
+is a category of export worth naming rather than discovering: it now travels to wherever
+`OTEL_EXPORTER_OTLP_ENDPOINT` points, on a per-request basis, and your collector's retention
+becomes its retention. What is **not** exported, verified by test rather than by assumption: the
```
bczoma: 🟡 **"Verified by test rather than by assumption" names a test that does not exist**

Flagged by good:security.

**The problem.** The line reads: *"What is **not** exported, verified by test rather than by assumption: the `Authorization` header, cookies, and the URL query string."* No such test exists. Nothing in the tree references a `url.query` or `Authorization` span attribute, nothing asserts on `client.address`, and nothing enumerates the entry span's attribute key set. The 24 new tests cover hierarchy, propagation, vocabulary, panic fidelity, SSE filtering and writer-interface preservation.

**The substance is true today.** Verified independently by reading otelhttp v0.71.0's `internal/semconv/httpconv.go`: `CurrentHTTPServer.RequestTraceAttrs` emits `server.address`/`server.port`, `http.request.method` (+`_original`), `url.scheme`, `network.peer.address`/`port`, `user_agent.original`, `client.address`, `url.path`, `network.protocol.*` and `http.route` — no headers, no cookies, no query string. So the claim is accurate; only its stated provenance is wrong.

**Why it matters.** This is a customer-facing security assurance in shipped documentation, about a third-party dependency that Dependabot bumps in this repo, and it is guarded by prose alone. OTel HTTP semconv has opt-in query capture and header-capture options; an otelhttp minor bump or an `OTEL_SEMCONV_STABILITY_OPT_IN` change could start exporting query strings or headers with no test failing and the document still asserting they are not exported. A documented negative that nothing enforces is worse than no documentation, because operators plan data-flow reviews around it — and here the same paragraph is what a reviewer would rely on to sign off enabling tracing.

**Recommended fix.** Add a test that drives one traced request through `HTTPMiddleware` and asserts the entry span's attribute keys against an explicit allowlist, failing on any key not on it. That earns the sentence as written and catches a future otelhttp bump that widens the set. Failing that, soften the wording to "verified by inspection against otelhttp v0.71.0" so the assurance matches its actual basis.

bczoma (dismissed):
## PR Review: #385 — SOL-152421: create request-path spans at every layer

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-152421` → `main` | **Changes:** +2490 / -54 across 26 files
**Jira:** `SOL-152421` — Create spans at MCP entry, the composite executor, and each SEMP attempt so a failed call reads as one trace (Story 26) (Verifying)

---

### Verdict

1. 🟡 `test/integration/request_path_spans_test.go:392` — the `error_type` vocabulary lives in four hand-maintained copies with nothing asserting any two agree; two comments promise a guarantee that provably does not exist, and the `knownErrorTypes` copy silently splits the metric from the span
2. 🟡 `internal/tools/register.go:261` — the arg-parse dispatch site is inverted relative to the other three; the shared seam owns the body of the instrumentation but not the LIFO ordering it depends on
3. 🟡 `docs/observability.md:711` — "Every span in the table above carries `outcome`" is false for the entry span, and the entry span is the only span a denied or rejected request produces
4. 🟡 `docs/observability.md:758` — "verified by test rather than by assumption" names a test that does not exist; the claim is true today by inspection only

- 🟡 SOL-152421 AC 7 — the chain order in the AC (`panic, correlation, tracing, tool-RED, auth, SDK`) is not what ships, and you are the ticket assignee. Both your reviewer questions resolve in favour of the code: tracing *must* sit inside correlation to stamp `correlation_id`, tool-RED outermost is #379's deliberate placement so 413s and 403s are still counted, and `semp.request` per call is right because a span should bound the unit of work its function owns while retries live below in `resilience.Sender.Do`. Correct the AC text and the work-breakdown's "per-attempt" wording rather than the code, and the questions close.

### Summary

This lands spans at four layers plus the W3C propagator that was never installed, and the engineering underneath it is careful in the places that are easy to get wrong: the panic path is designed rather than assumed (named returns, `recover` → rewrite `err` → `defer panic(r)` so the re-panic runs after `span.End()`, and `sempv2.Execute` reading `op.ID`/`op.Method` into locals before `tracer.Start` so the defer cannot dereference `op` mid-unwind), the request context is threaded intact through the parallel-step path the ticket flags as easiest to lose, the authorization frame is untouched as the ticket requires, and the hierarchy test asserts real parent/child edges rather than a shared trace ID — the assertion a flat four-root trace would slip past. Cardinality is bounded at every attribute the change introduces, and disclosing the client-IP data flow unprompted is the right instinct.

Three findings were filed as severe and two were reduced by measurement: an unbounded-attribute exhaustion claim did not survive contact with the numbers (feature off by default, two caller-controlled values against a fixed 13-key set, a 2048-span queue that drops, a collector receive cap that rejects precisely the pathological batches, and an existing operator lever), and the duplicated panic-fidelity defer turned out to be test-guarded at every site. What remains is one real code-shape issue, one test that cannot fail for the reason it claims, and three documentation claims that outrun their evidence.

### Minor notes

- 🟢 `CHANGELOG.md:28` and `docs/observability.md:1144` — both still say the `error_type` vocabulary has **ten** values while this PR's own entry says twelve; both blocks are `[Unreleased]`, so today's release notes would contradict themselves one paragraph apart. Editable now, frozen at release.
- 🟢 `docs/observability.md:752-761` — the data-flow note calls `client.address` "the caller's IP address". otelhttp derives it from the first comma-separated element of `X-Forwarded-For` verbatim, with no validation and no trusted-proxy handling; `network.peer.address` is the attested value. Point forensic use at the latter, and say which is which.
- 🟢 `internal/composite/executor.go:128-158`, `internal/semp/sempv1/client.go:190-219`, `internal/semp/sempv2/client.go:196-226` — the recover/outcome/`SetStatus`/`End`/re-panic defer is byte-identical across three sites (~24 lines each, including a 9-line comment), and Story 27 adds a fourth. Hoisting it is mechanically viable, but put it in an **API-only** leaf: `internal/observability/tracing`'s transitive deps include `otel/sdk/trace`, `otel/sdk/metric` and `otlptracegrpc`, which would enter `internal/semp`'s build graph.
- 🟢 `internal/observability/tracing/httpmw.go:60-62` — the SSE exclusion filters on `r.Method != http.MethodGet`, a proxy for "is this the notification stream". If the streamable transport ever carries requests over `GET`, tracing goes dark with no error and no failing test. Filter on `GET` **and** `Accept: text/event-stream`, or pin the SDK's request method by test.
- 🟢 `internal/semp/sempv2/client.go:216-217`, `internal/composite/executor.go:149` — `semp.operation`, `http.request.method` and `composite.steps` are asserted by no test, and `semp.request`'s `SpanKindClient` is pinned for v2 only. `semp.operation` is the attribute that distinguishes one SEMP span from another, and the "operationId, never the resolved URL" security argument has no regression guard.
- 🟢 `test/integration/request_path_spans_test.go:555-584` — `TestRequestPathSpans_SSEStreamIsNotSpanned` asserts every started span has ended, sampled at one instant with the session open and nothing draining in-flight notification POSTs. A latent flake whose diagnostic would read as "a long-lived stream is being spanned" — the most misleading possible message for a timing race.
- 🟢 `docs/internal/architecture.md:43`, `:370`, `:502` — still reads `Capability gate only — OTel tracer not yet implemented (skeleton)`, and `CLAUDE.md` names this file as *the* architecture reference. Staleness predates this PR, but this is the PR that makes it plainly wrong.
- 🟢 `internal/observability/tracing/provider.go:132` — the provider takes SDK defaults, so `AttributeValueLengthLimit` is unlimited, and `srv.MaxHeaderBytes` is unset anywhere in the repo. Not a blocker (operators already have `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`), but `WithSpanLimits` at 4096 and a 64 KB header cap are one line each and would align the outermost span with the care taken on `broker` and `tool`.

*and 8 further minor items — ask if you want them.*

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

internal/tools/register.go
solace-aross: This branch discards the span's context (`_, span := ...`) instead of threading it. The other three dispatch sites — `ToolManager.CallTool`, `list-brokers`, `describe-semp-schema` — all start the span first and reassign `ctx` before the audit/metric calls run. Here, `logToolResult`/`recordToolInvocation` two lines above run before this span exists, and `endDispatchSpan` right below still uses the un-enriched `ctx`.

Confirmed by injection: printed the span ID present in `ctx` at record time versus this dispatch span's own ID — they differ. The metric/log exemplar for this one failure mode (malformed `arguments` JSON) links to the parent `POST /mcp` span, not the dispatch span whose `tool`/`outcome`/`error_type` it's describing — narrower than `docs/observability.md`'s stated exemplar-correlation guarantee.

This is the only one of four dispatch sites that gets the ordering wrong. No log/metric value changes, only context linkage, and `TestRequestPathSpans_BypassDispatchSitesSpan`'s argument-parse-failure case passes whether this is fixed or not — it only checks attribute values agree, never which span is actually in context. Worth a fast follow-up: start the span first and thread `ctx` through like the other three sites, and strengthen that test to assert the in-context span ID matches the dispatch span's.

solace-aross (dismissed):
Verified the safety-critical claims by injection rather than reading:

- Panic recover-and-re-panic in `composite.Execute`, `sempv1.Client.Execute`, `sempv2.Client.Execute` — sabotaged each to swallow instead of re-panic; each layer's own `TestExecute_PanicMarksSpanAsErrorAndRepanics` went red, reverted clean.
- The `outcome`/`error_type` closed vocabulary "enforced by test" claim — sabotaged `endDispatchSpan` to emit an out-of-vocabulary value; four tests across `TestRequestPathSpans_*` caught it. Real, not aspirational.
- No-Baggage / no-redaction-policy claim — confirmed only `TraceContext` is installed, no `baggage` import anywhere outside its own test, and no span attribute in the diff derives from unredacted request/response bodies (SEMP client spans carry only `op.ID`, method, outcome — not the resolved URL, per the code's own comment).
- SSE stream correctly unspanned via the `otelhttp` GET filter, with its own passing test.
- Full suite green across every touched package at HEAD.

One real gap below — narrow, not a blocker.

Approved.

amitmorade (dismissed):
Re-reviewed `295a8aa` against the four 🟡s from the earlier pass.

1. **`error_type` copies** — `AllErrorTypes` is now the source; coercion, span vocabulary, and the docs table derive from it. A new const without that list fails CI. Agreed, and this closes it.

2. **Arg-parse dispatch site** — span starts first, context is threaded into log/metric, and a test pins the in-context span ID. That was the production bug. The shared `beginDispatch` reshape was not done; the four sites still own their own ordering. Fine for this PR.

3. **Entry span / `outcome`** — docs now exclude the entry span, warn that `outcome="error"` misses 403/413/denial, and point at `http.response.status_code` (4xx status is Unset). Agreed; they scoped the sentence rather than adding `outcome` on `POST /mcp`.

4. **“Verified by test”** — allowlist on entry-span keys, driven with a bearer, a cookie, and a query string. Agreed, and this closes it.

The rest of the follow-up (SSE filter, `client.address` / `X-Forwarded-For`, ten→twelve, architecture.md, SEMP attribute pins) matches the 🟢s that were cheap to land. Hoisting the panic defer and span/header limits were left for later, as noted in the commit.

Approving on that basis.

internal/tools/register.go
```diff
@@ -335,6 +354,15 @@ func RegisterListBrokers(server *mcp.Server, pool *semp.BrokerPool, tm *metrics.
 			var errorType metrics.ErrorType
 			var toolErr error
 			id := NewIdentityFromPrincipal(auth.PrincipalFrom(ctx))
+
+			// ...and its own dispatch span, for the same reason: this tool
+			// would otherwise appear in every dashboard and in no trace
+			// (SOL-152421). dispatch.finish emits all three signals in order.
```
bczoma: 🟡 **The ordering invariant's statement was replaced by a reference to something that does not exist**

Raised by this follow-up pass — it is a side effect of the fix for my earlier `register.go` finding, not a pre-existing issue.

**What changed.** Two of the four dispatch sites used to carry the invariant at the site:

```
// Registered before the audit defer so it runs after it and sees the
// panic rewrite below.
```

`295a8aa` replaced that, here and at `internal/tools/describe_semp_schema.go:340`, with:

```
// dispatch.finish emits all three signals in order.
```

There is no `dispatch.finish`. Grepping the tree finds the phrase in exactly three comments — this one, `describe_semp_schema.go:340`, and `spans_test.go:163` — and nothing else: no function, no method, no span name. It reads as a leftover from the consolidated seam that `endDispatchSpan`'s own doc comment says was "tried and reverted".

**Why it matters.** The code is correct: the `endDispatchSpan` defer is still registered before the emitting defer, LIFO still closes the span last, and the panic reclassification is still seen. What was lost is the statement of *why the order matters*, at two of the four sites that have to preserve it. A maintainer editing `list-brokers`' or `describe-semp-schema`'s dispatch now reads a pointer to a helper they cannot find, and has nothing at the site telling them the two defers are order-dependent — so swapping them, or hoisting one, looks free. That is the failure mode my earlier finding was about, and this is the sentence that was guarding against it.

The mitigation is real, which is why this is 🟡 and not more: `endDispatchSpan`'s doc comment now states the invariant in bold and names the two tests that enforce it behaviourally, so a maintainer who follows the call finds it. The sites just no longer point there.

**Fix.** One line at each of the two sites — either restore the original sentence, or point at the real owner, e.g.:

```
// Registered before the emitting defer so LIFO closes the span last —
// see endDispatchSpan's doc comment for the ordering invariant.
```

And drop or reword the same phrase in `spans_test.go:163` so a grep for `dispatch.finish` stops returning three hits for a thing that isn't there.

bczoma (dismissed):
## PR Review: #385 — SOL-152421: create request-path spans at every layer

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-152421` → `main` | **Changes:** +3232 / -65 across 29 files
**Jira:** `SOL-152421` (Story 26)
Follow-up review at `6a2161e` — parent-only pass over `295a8aa` (the review-feedback commit) plus the `main` merge above it. My earlier review at `e48540c` was dismissed by those pushes, not withdrawn; its findings carry forward and are checked below. No lenses dispatched.

---

### Verdict

All four 🟡s from the last pass are genuinely closed — I verified each in code rather than taking the commit message for it. One new item, introduced by the fix for the second of them:

1. 🟡 `internal/tools/register.go:360` (and `internal/tools/describe_semp_schema.go:340`) — the comment that stated the LIFO ordering invariant was replaced by a reference to a `dispatch.finish` that does not exist

The 🟢 note below is not a condition of this verdict.

Verified at this head: `go build ./...` clean; `go test ./internal/observability/... ./internal/tools/` all green.

**Note on @amitmorade's review:** they re-reviewed `295a8aa` against the same four 🟡s and approved; the `main` merge then dismissed it, so it needs re-issuing to count. I reached the same conclusion on all four independently and am not restating their points here. The one item below is new since their pass.

---

### Summary

The vocabulary fix is the one I'd single out. `allErrorTypes` is now the sole enumeration, `knownErrorTypes` is derived from it, `AllErrorTypes()` exports it, the span-vocabulary test drives off it, the `docs/observability.md` table is cross-checked against it, and an AST-walking test parses the const block so a value added to the type but missed in the list fails in the PR that adds it. That closes the exact failure my finding described — the same call reading `other` on the metric and its real value on its span — at the level of "cannot drift" rather than "is currently consistent".

The other three closed on their substance too: the arg-parse site now starts its span first and threads the context, with a test pinning the in-context span ID; the entry-span sentence is scoped and now tells an operator to select 4xx failures on `http.response.status_code` rather than on `outcome` or span status; and the "verified by test" claim is backed by an actual test that drives a bearer, a cookie and a query string and asserts on an **allowlist** of entry-span keys — the right shape, since a denylist would keep passing the day an otelhttp bump starts exporting `url.query`. On the seam reshape I proposed, the answer is better than the proposal: `endDispatchSpan`'s doc comment now carries the ordering invariant in full and explains that folding the four sites into one seam was tried and reverted because `audit_error_type_drift_test.go`'s static scanner needs the `logToolResult(..., &errorType, ...)` funnel visible. That is a good reason, and it settles the point.

---

*11 of 12 prior findings settled — all four 🟡s addressed, plus six 🟢s; one 🟢 accepted as-is.*

|---|---|---|---|
| 🟡 | `error_type` vocabulary in four hand-maintained copies, nothing asserting any two agree | Addressed | `internal/observability/metrics/instruments.go:62-110` — single `allErrorTypes`, `knownErrorTypes` derived, `AllErrorTypes()` exported; `error_type_vocabulary_test.go` parses the const block; `test/integration/request_path_spans_test.go:445` drives the span vocabulary off it and `:942` holds the docs table to it |
| 🟡 | The arg-parse dispatch site is inverted relative to the other three | Addressed | `internal/tools/register.go:257` — `ctx, span := tracer.Start(...)` now precedes `logToolResult`/`recordToolInvocation` and the context is threaded; pinned by `TestDispatch_AuditAndMetricSeeTheDispatchSpanInContext` |
| 🟡 | "Every span carries `outcome`" is false for the entry span | Addressed | `docs/observability.md:865-873` — scoped to "except the entry span", plus the concrete filter guidance (403/413/hop-1 denial; select on `http.response.status_code`, not span status, since otelhttp sets `Error` only for 5xx) |
| 🟡 | "Verified by test rather than by assumption" named a test that did not exist | Addressed | `internal/observability/tracing/httpmw_test.go:127-194` — allowlist on entry-span keys, driven with `Authorization`, `Cookie` and a query string; `docs/observability.md:920` reworded |

Still open:

|---|---|---|---|
| 🟢 | Provider takes SDK defaults, so `AttributeValueLengthLimit` is unlimited, and `srv.MaxHeaderBytes` is unset | Not addressed | No `WithSpanLimits` in `internal/observability/tracing/provider.go` and no `MaxHeaderBytes` in `cmd/server/main.go`. I filed it as explicitly not a blocker; entirely fair to leave |

Settled and not re-listed: ten→twelve corrected in both blocks; `client.address` vs `network.peer.address` now documented with the `X-Forwarded-For` caveat and the forensic pointer; the SSE filter now matches `GET` **and** `Accept: text/event-stream`; `semp.operation` (`"getMsgVpn"`, an operationId) and `composite.steps` now asserted, with `SpanKindClient` pinned for v1 as well as v2; the started-minus-ended snapshot dropped from `SSEStreamIsNotSpanned` with the race written down; `docs/internal/architecture.md`'s tracing rows updated. The panic-defer hoist was offered with caveats and left with a note in the commit — accepted as-is.

---

### Minor notes

- 🟢 `internal/observability/metrics/error_type_vocabulary_test.go:44` — the guard parses a single file, `parser.ParseFile(fset, "instruments.go", …)`. Every `ErrorType` const lives there today, so it holds; the day one is declared in another file of `package metrics` the test passes while `allErrorTypes` lags, which is precisely the divergence this test exists to prevent. `parser.ParseDir` over the package directory closes it and costs nothing.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ — build clean, `internal/observability/...` and `internal/tools` green at this head |
| Code Quality | ✅ — the arg-parse ordering fix is right, and the reverted-seam explanation is better than my proposal |
| Test Coverage | ✅ — vocabulary, span/metric join, entry-span allowlist and the SSE cases all pinned |
| Architecture | — not re-reviewed (span topology unchanged since `e48540c`) |
| Security | ✅ — the entry-span allowlist test is the finding I most wanted covered, and it is |
| Documentation | ⚠️ — the four doc findings closed; the comment drift below is new |
| PR Scope & Description | — not re-reviewed (unchanged since `e48540c`) |

Sourced: parent (follow-up).

solace-aross (dismissed):
Re-reviewed at this head, focused on the fix for my own finding plus the headline vocabulary fix bczoma raised — both sabotage-tested in a throwaway worktree, not read.

**My finding (argument-parse dispatch span ordering) — fixed and now pinned for real.** `register.go`'s arg-parse branch starts the span first and threads `ctx` into `logToolResult`/`recordToolInvocation`, matching the other three sites. Reverted the ordering back to the pre-fix shape — `TestDispatch_AuditAndMetricSeeTheDispatchSpanInContext` failed with exactly the right diagnostic (captured span ID `0000...0` vs. the dispatch span's real ID). This is the regression guard I asked for last pass: it compares actual span IDs, not just attribute values, so it can't pass regardless of the fix.

**`allErrorTypes` single-source-of-truth — fixed and enforced.** Added an unregistered `ErrorType` const (`ErrorTypeSabotageTest`) directly to the const block without touching `allErrorTypes`. `TestAllErrorTypes_CoversEveryDeclaredConst` failed immediately with a precise, actionable message. `knownErrorTypes` is genuinely derived from `allErrorTypes` now, not a second hand-maintained map — the exact drift bczoma's experiment demonstrated can't recur silently.

Build and the touched-package suites clean at this head, matching bczoma's independent check.

One open item — bczoma's still-outstanding comment-drift thread (a reference to a `dispatch.finish` that doesn't exist) — I agree it's real and don't have anything to add to it.

Approved.

amitmorade (dismissed):
LGTM

bczoma (dismissed):
## PR Review: #385 — SOL-152421: create request-path spans at every layer

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-152421` → `main` | **Changes:** +3272 / -65 across 29 files
**Follow-up pass:** incremental review of `6a2161e..cb27f18` (1 commit, 4 files, +50/-10)
**Jira:** `SOL-152421` — request-path spans at every layer

---

### Verdict

**Recommendation:** ✅ Approve

Both items still open after my last pass are closed, and I checked each in code at `cb27f18` rather than taking the commit message for it. No new findings, so there are no threads on this pass.

- The 🟡 comment regression is fixed at both sites. `internal/tools/register.go:359-367` and `internal/tools/describe_semp_schema.go:337-346` now state the LIFO invariant in full — registered before the emission defer, so it runs after it, once a recovered panic has been reclassified — and point at `endDispatchSpan` for the rationale. The dangling `dispatch.finish` reference is gone repo-wide (`grep` over `*.go`/`*.md`: no hits), including the stale one in `internal/tools/spans_test.go:161-166`, which I had not flagged. Defer registration order verified as described at both sites.
- The 🟢 vocabulary guard now covers the package, not one file. `internal/observability/metrics/error_type_vocabulary_test.go:44-76` globs every non-test `.go` file and parses each, with an explicit `t.Fatal` if the glob matches nothing. I mutation-probed it: adding `const ErrorTypeProbeOnly ErrorType = "probe_only"` in a second file of `package metrics` fails the test with the diagnostic that names the omission, where the previous single-file parse would have passed. Choosing `filepath.Glob` over `parser.ParseDir` is right — the latter is deprecated as of Go 1.25 along with `ast.Package` — and the reasoning is in the comment.

Verified at this head: `go build ./...` clean; `go test ./internal/observability/metrics/ ./internal/tools/` green. CI reports no failures at this head; the five E2E jobs were still pending when I looked, which is not a review finding.

Nothing carried over holds this verdict: the one remaining prior item (`WithSpanLimits` / `srv.MaxHeaderBytes`) I filed as explicitly not a blocker and raised across two passes, so it is settled as declined and not re-argued here.

---

### Summary

This push does exactly the two things it says and nothing else. The comment fix is the better outcome of the two: my finding was that the invariant's *statement* had been replaced by a reference to a symbol that never existed, and the fix restores the statement at both sites instead of just deleting the bad reference — which is what a fix aimed at the grep rather than the reader would have done. The vocabulary guard widening closes the last way the four-copy `error_type` drift could have come back through a file boundary, and it is the kind of test change worth probing rather than reading, so I probed it.

---

*3 prior findings settled this pass — 2 addressed, 1 declined.*

---

### Summary Verdict

|---|---|---|
| Correctness | ✅ — defer ordering verified at both sites; build clean, `internal/observability/metrics` and `internal/tools` green | parent (follow-up) |
| Test Coverage | ✅ — widened guard mutation-probed: a const in a second package file now fails the test | parent (follow-up) |
| Code Quality | ✅ — both comments state the invariant and explain why, no stale references left | parent (follow-up) |
| Architecture | — not re-reviewed (unchanged since `6a2161e`) | |
| Security | — not re-reviewed (unchanged since `6a2161e`) | |
| Documentation | — not re-reviewed (unchanged since `6a2161e`) | |
| PR Scope & Description | — not re-reviewed (unchanged since `6a2161e`) | |

solace-aross (dismissed):
Sabotage-tested both fixes rather than trusting the commit message.

**Vocabulary drift-guard widened to the whole package.** The test previously parsed only `instruments.go`; now it globs every non-test file in the package. Added an `ErrorType` const to a different file (`metrics.go`) — caught immediately with the same actionable diagnostic. Also confirmed a const declared in a `_test.go` file is correctly excluded (production vocabulary only). This closes a real gap my own earlier sabotage didn't exercise, since I'd only tested within `instruments.go`.

**Stale `dispatch.finish` comment — fixed for real, not just deleted.** `register.go` and `describe_semp_schema.go` each now state the LIFO ordering invariant directly (span defer registered before the emission defer, so it runs after) instead of pointing at a helper that doesn't exist. Read both sites; the replacement text is accurate.

Approved.

bczoma (approved):
## PR Review: #385 — SOL-152421: create request-path spans at every layer

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-152421` → `main` | **Changes:** +3272 / -65 across 29 files
**Follow-up pass:** `cb27f18..f681397` — one commit, and it is a `main` merge, not new work
**Jira:** `SOL-152421` — request-path spans at every layer

---

### Verdict

**Recommendation:** ✅ Approve

Re-issuing my approval: it was dismissed by the merge of `main`, not by anything the author changed. `f681397`'s first parent is `cb27f18`, the commit I approved an hour ago, and the second is `0a63259` (`main`, carrying #388 and #386) — so there is no authored delta on this pass. What I checked instead is that the merge is sound, since #388 lands audit emission in `internal/tools/authorization.go` and `internal/semp/resilience` while this PR instruments the same request path:

- **Two files needed conflict resolution, and both were resolved correctly.** `CHANGELOG.md` keeps both entries at top level under `[Unreleased]` — this PR's spans entry and #388's three-audit-record-types entry — with nothing dropped or re-nested. `docs/observability.md`'s capability table keeps `main`'s updated Audit trail row *and* this PR's rewritten Distributed tracing row; both rows' anchors resolve to the headings that actually exist in the merged file (`:452` and `:768`).
- **No semantic conflict between the two instrumentation stories.** `go build ./...` and `go vet ./...` clean; `./internal/tools/`, `./internal/auth/`, `./internal/observability/...`, `./internal/semp/...`, `./cmd/server/` and `./test/integration/` all green at `f681397`. The chain-composition and cross-signal join tests are the ones that would have caught a collision between #388's emit sites and this PR's dispatch spans, and they pass.

Nothing from earlier passes is outstanding. The one prior item still open — `WithSpanLimits` / `srv.MaxHeaderBytes` — I filed as explicitly not a blocker and it is settled as declined; it is not re-argued here.

---

### Summary

A merge-only pass. The merge is clean where it matters and the two conflict resolutions are the ones a careful author would have made, which is worth saying because this is exactly the kind of merge that quietly loses a CHANGELOG entry or leaves a status table describing two different releases at once. Neither happened.

---

*1 prior finding settled — declined (raised twice, not a blocker).*

---

### Summary Verdict

|---|---|---|
| Correctness | ✅ — build, vet, and every affected package green at the merged head | parent (follow-up) |
| Documentation | ✅ — CHANGELOG and observability status table resolved without loss; anchors verified | parent (follow-up) |
| Code Quality | — not re-reviewed (no authored change since `cb27f18`) | |
| Test Coverage | — not re-reviewed (no authored change since `cb27f18`) | |
| Architecture | — not re-reviewed (no authored change since `cb27f18`) | |
| Security | — not re-reviewed (no authored change since `cb27f18`) | |
| PR Scope & Description | — not re-reviewed (no authored change since `cb27f18`) | |

solace-aross (approved):
This head is a plain merge from `main` (picking up #388/Story 23's auth-audit work) with no new commits of its own on this branch — confirmed via the compare diff, every changed file traces to the merge, nothing authored fresh here. CI green. bczoma's approval already stands at this exact head. Re-approving to keep the record current; nothing new to verify.

PR #386, opened 2026-09-08
SOL-154045: test coverage for describe-semp-schema's untested error_type branches

solace-aross:
## What

Test coverage only. Adds the missing assertions for `describe-semp-schema`'s four `error_type` branches that reach `logToolResult`/`recordToolInvocation` but weren't tested at that call site:

- `describe_semp_schema.go:348` — unparseable JSON arguments → `bad_request`
- `describe_semp_schema.go:355` — missing required `operation` parameter → `bad_request`
- `describe_semp_schema.go:364` — invalid `view` parameter → `bad_request`
- `describe_semp_schema.go:371` — unknown operation (`reg.describe` not-found) → `not_found`

Each new test drives the handler through a real `mcp.NewServer` + `session.CallTool` dispatch (not a direct handler call), matching the existing `TestDescribeSempSchema_EmitsAuditLog` pattern, and asserts both:
- the audit line's `error_type` field, and
- the `mcp_tool_invocation_total` Prometheus counter's `error_type` label

— the same two surfaces `register.go`'s `bad_request` branch is already covered against (`register_test.go`, `principal_emit_sites_test.go`).

**No behavior change.** `describe_semp_schema.go` is untouched — only `describe_semp_schema_test.go` changed.

## Why the unparseable-JSON branch needed a new middleware

A normal client call can never produce genuinely unparseable JSON on the wire — the SDK always marshals a well-formed JSON object for `Arguments`, and omitting the field entirely lands on the missing-`operation` branch instead, not the JSON-parse-failure branch. Added `forceMalformedArguments`, a client-sending middleware analogous to the existing `forceOmitArguments` (`register_test.go`), that overwrites `Arguments` with a syntactically valid but wrong-typed JSON literal (`json.RawMessage(`"not an object"`)`) immediately before send, so the wire genuinely carries `"arguments":"not an object"` — something `json.Unmarshal(..., &args)` fails on with a real type-mismatch error, not a parse/omission edge case already covered elsewhere.

## Verification

- Mutation-tested each of the four branches by hand: temporarily removed each `errorType = metrics.ErrorTypeX` assignment one at a time (leaving the sibling `toolErr = ...` line intact), confirmed the corresponding new test fails, then reverted and confirmed the suite is green again. All four failed as expected when neutered.
- `make check` green: build, vet, lint, race-enabled tests, 85% coverage gate (89.7% actual).
- Independently re-verified by a second skeptical review pass (mutation tests re-run from scratch, byte-level trace of the `forceMalformedArguments` payload through the SDK's marshal/unmarshal path, completeness check that no two tests silently exercise the same branch).

## Scope notes

- **Does not touch `docs/observability.md`.** The vocabulary-count/table AC from the ticket is already shipped in #381 (merged), which rewrote those exact docs hunks. This PR is test-coverage only, so it stays out of that file entirely.
- **No CHANGELOG entry** — test-only, no behavior change, per CLAUDE.md's own rule that entries are for production-surface changes.
- **Distinct from `audit_error_type_drift_test.go` (also from #381, now on main).** That test is a static AST scan proving the `error_type` vocabulary this package computes matches the audit constructor's declared set — a *declaration*-level guard. This PR's tests are *behavioral*: they prove a specific value actually reaches the audit line and the metric label for four specific code branches, through a real dispatch. Neither test makes the other redundant.
- Branch is tip-of-main, not stacked on any other open PR.

Jira: https://sol-jira.atlassian.net/browse/SOL-154045

### DIFF
--- a/internal/tools/describe_semp_schema_test.go
+++ b/internal/tools/describe_semp_schema_test.go
@@ -17,12 +17,18 @@ package tools
 import (
 	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
 	"log/slog"
+	"net/http"
+	"net/http/httptest"
 	"strings"
 	"testing"
 
+	"github.com/SolaceProducts/solace-broker-mcp/internal/observability/metrics"
 	"github.com/SolaceProducts/solace-broker-mcp/internal/semp/sempv2/specs"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
+	sdkresource "go.opentelemetry.io/otel/sdk/resource"
 )
 
 // Smoke tests against the embedded specs: trimmed POST, trimmed PATCH,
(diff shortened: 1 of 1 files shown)


bczoma (approved):
## PR Review: #386 — SOL-154045: test coverage for describe-semp-schema's untested error_type branches

**Author:** solace-aross | **Branch:** `aross/sol-154045-describe-semp-schema-error-type-tests` → `main` | **Changes:** +182 / -0 across 1 file
**Jira:** `SOL-154045` — The error_type freeze isn't enforced everywhere it claims, and the schema doc undercounts by two (Verifying)

---

### Verdict

The test work is sound and I verified it rather than taking its word for it. The one condition is about the *ticket*, not this diff:

Not anchorable to this diff, so no thread — the file is not part of this PR:

- 🟡 `docs/observability.md:988`, `:993`, `:995`, `:998` — **AC3 is not satisfied, so this PR does not close SOL-154045.** The `error_type` value table now holds **14 rows for 12 distinct values** under a header reading "a closed set of twelve values" (`:983`): `bad_request` appears at both `:988` and `:993`, and `not_found` at both `:995` and `:998`. #381 added both values to a table that already contained them. Worse, the two `bad_request` rows carry *different* definitions — "The tool arguments could not be parsed as a JSON object" versus "The request itself was malformed — unparseable `arguments`, or a missing or invalid parameter on a tool that validates its own input" — so an operator reconciling their alert vocabulary against the table counts 14 and cannot tell which is authoritative. The ticket is in **Verifying** with all five ACs unticked; AC3 asks for "the two missing values added to the value table", and a verifier will see them present and tick it. Fix is deleting `:993` and `:998` while keeping `:993`'s broader wording on the surviving `bad_request` row — a two-line edit that would close AC3 inside this PR without touching its test scope.

This is inherited from `main`, not introduced here, and nothing in it faults the test file. AC4 **is** satisfied — the cardinality passage at `:188-193` uses the corrected count and correctly names `bad_request` and `not_found` as pre-resolution-only. The 8 🟢 notes below are **not** conditions of this verdict.

### Summary

Four branches, four tests, and the claims in the PR body hold up under checking. The mutation-testing claim is not just true but stronger than stated: all four `errorType` assignments (`describe_semp_schema.go:348`, `:355`, `:364`, `:371`) were neutralised in turn and produced a clean 1:1 mapping — exactly one test failed each time, and always the intended one. That also settles branch distinctness: no two tests land on the same branch, and the missing-`operation` test does not reach the unmarshal branch. The metric assertion pins every label rather than just the family, the audit assertion compares `error_type` for equality rather than substring, isolation is genuinely sound (`metrics.New` builds its own registry, so the four providers cannot collide), and `-race -count=2` is clean. The "complementary, not redundant" argument against `audit_error_type_drift_test.go` holds — a static AST scan of declared vocabulary cannot catch what these behavioural dispatches catch.

One finding was filed at 🟡 and reduced by measurement. A reviewer established that `forceMalformedArguments` is unnecessary at the pinned SDK — correctly, as it turns out, and the SDK reading reproduced exactly. But a challenger found the middleware still buys something nothing else would catch, so only the false comment survives, at 🟢. Details in the first note below.

### Minor notes

- 🟢 `internal/tools/describe_semp_schema_test.go:253-262` and the PR body's "Why the unparseable-JSON branch needed a new middleware" — both are factually wrong at the pinned go-sdk **v1.7.0**: `CallToolParams.Arguments` is typed `any` (`mcp/protocol.go:227`) and `ClientSession.CallTool` substitutes a map **only when it is nil** (`mcp/client.go:1281-1284`), so a plain `Arguments: "not an object"` with no middleware at all reaches `describe_semp_schema.go:349` with a byte-identical error and audit line (verified twice, independently). So "the only way to make `json.Unmarshal` itself fail" and "a normal client call can never produce genuinely unparseable JSON on the wire" are both untrue. **Keep the middleware** — it pins the payload at the wire, which is real insurance: if a future SDK coerced a non-object `Arguments`, the direct form would silently land on the missing-`operation` branch instead, which also yields one audit line, `outcome=error` and `error_type=bad_request`, i.e. all three things the assertion checks, and neither the 85% gate (89.7% actual) nor hand-run mutation testing would notice. Just rewrite the comment to say *that* instead.
- 🟢 `internal/tools/describe_semp_schema_test.go:295-300` — the Prometheus check is `strings.Contains(body, "…{…} 1")`, so the counter value is unanchored and the assertion also passes for `10`, `11`, `100`, `1.5`. The harness comment at `:212-217` justifies the per-test provider precisely so a bare `== 1` can be asserted, but `Contains` cannot distinguish 1 from 10. `activeRequests` in `internal/tools/metrics_test.go:48-62` already does this properly — split lines, `strings.CutPrefix`, compare the parsed value.
- 🟢 `internal/tools/describe_semp_schema_test.go:219-251` — the new harness duplicates three existing blocks rather than sharing them: `TestDescribeSempSchema_EmitsAuditLog` (`:162-181`) has the same slog-swap + `NewServer` + register + in-memory-transport + connect sequence almost line for line and was not refactored onto the harness, so the file now carries two copies; `:238-249` duplicates `callToolTestHarness`'s connect block (`register_test.go:458-483`); `:225-233` duplicates `newMetricsManager`'s provider block (`metrics_test.go:32-45`). Three copies of session bring-up in one package makes the next SDK API change a three-site edit. That older test also uses the literal `"describe-semp-schema"` where the new ones correctly use `describeSempSchemaToolName`.
- 🟢 `docs/observability.md:184`, `:188-189` — the metric-label domain is documented as twelve values, but `internal/observability/metrics/instruments.go:59` defines a thirteenth member of `knownErrorTypes`: the `other` coercion sentinel, which appears nowhere in `docs/`. An SRE building the allow-list from this section drops the one series that would appear if an emit site ever drifted outside the closed set. Keep the *audit* vocabulary at twelve (`audit/event.go:114-127` correctly excludes `other`) and note the metric's extra value.
- 💡 `internal/tools/describe_semp_schema_test.go:310`, `:327`, `:345` — the three `bad_request` tests assert byte-identical expectations, so the only thing tying each to its branch is its input; the returned `CallTool` error is checked for `err == nil` only, never for its message. Distinctness is real today but unguarded — reordering the guards so `view` validates before `operation` would leave all three green with one branch dead. Assert the error text (`parsing tool arguments` is unique to `:349`), which also closes the drift gap in the first note.
- 💡 `internal/tools/describe_semp_schema_test.go:225`, `:237` — teardown is partial: the `*metrics.Provider` is never `Shutdown` (it exists, `metrics/provider.go:202`) and the `server.Run` goroutine gets a `context.Background()` that is never cancelled; `t.Cleanup` closes only the client session. Harmless with a pull-based reader and consistent with existing tests, so polish only.
- 💡 SOL-154045 AC1/AC2 line numbers — **the PR body is right and the ticket is stale by three lines.** `grep -n "errorType = metrics" describe_semp_schema.go` gives 348/355/364 (`bad_request`) and 371 (`not_found`); the ticket's 345/352/361/368 no longer resolve to those assignments. Worth updating the AC text, or dropping the line numbers for branch names, which do not rot.
- 💡 `CHANGELOG.md:31` — says the closed set is "twelve, not the ten `docs/observability.md` listed", but the doc's table and count sentence already said twelve before #381; only the "Open Items" prose did. Optional: "not the ten one section of `docs/observability.md` still claimed".

### Summary Verdict

|---|---|---|
| Architecture | — not reviewed (tests-only diff, no architecture surface) | — |
| Security | — not reviewed (tests-only diff, no security surface) | — |
| Documentation | ⚠️ | docs |
| PR Scope & Description | — not reviewed (lens not dispatched for a tests-only diff) | — |

One process note: the five e2e jobs were still `pending` when this review ran, so they are unverified here. Everything else at this head is green.

PR #387, opened 2026-09-08
SOL-152093: Per-attempt SEMP request and latency metrics so SREs can spot retry storms and broker latency

solace-wmaryam:
## Summary

Emit per-attempt SEMP request metrics on `/metrics` — `mcp_semp_request_total` and `mcp_semp_request_duration_seconds` — so SREs can see retry storms and per-broker latency directly. Story 16.

Jira: [SOL-152093](https://sol-jira.atlassian.net/browse/SOL-152093)

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

Per-call metrics hide retry behaviour: a slow broker and a retrying client look the same. Observing once per HTTP attempt makes a retry storm visible as climbing `attempt` counts, and lets an SRE separate a slow broker from a client that keeps retrying. HTTP labels follow OpenTelemetry semantic conventions so the vocabulary is portable across the rest of a customer's estate.

## Changes Made

- New instruments `mcp.semp.request` (counter) and `mcp.semp.request.duration` (histogram) with the committed buckets `0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10` seconds — coarser at the low end than the tool histogram because a SEMP call is a broker round-trip. `Provider.SEMPMetrics()` accessor, registered once.
- Labels: OTel HTTP semconv `http.request.method`, `http.response.status_code`, `server.address` (the exporter translates dots to underscores) plus Solace `broker`, `api` (`v1`/`v2`), `operation`, `attempt` (1-based integer string). A no-response attempt records an empty status.
- Recording sits in an `http.RoundTripper` wrapped around the shared resilience `Sender`. The attempt number is tracked on the per-request retry state and bumped once per round-trip.
- Wiring: the metrics provider is built before the broker pool; the recorder is threaded via a new variadic `semp.WithSEMPMetrics` pool option, and each protocol client tags its own `api`.
- Golden file, unit tests, `docs/observability.md`, and `CHANGELOG.md` updated.

## Testing

**Test Environment:**
- Go version: go1.26.3
- OS: darwin/arm64
- Broker version (if applicable): n/a (httptest servers)

**Tests Performed:**

**Test Coverage:**
- `make check` passes; total coverage 88.7% (≥ 85% floor).
- Metrics package: label set, the eleven histogram buckets, and the empty-status no-response case pinned via `GatherAndCompare`; golden file extended with both SEMP families.
- sempv2 client: per-attempt recording across a 503→200 retry (`attempt="1"`/`status="503"`, `attempt="2"`/`status="200"`), no-response empty status, and the disabled-metrics off-gate.
- sempv1 client: recording tagged `api="v1"`, `operation="SEMPv1"`.

## Breaking Changes

None. The capability is gated by `OBS_METRICS_ENABLED` and is off by default; when off (or the recorder is nil) no transport wrapper is installed and nothing is emitted.

## Additional Notes

**Deviation from the story's work breakdown (step 1).** The story says to instrument `sempv2/client.go:Execute` and the SEMPv1 equivalent. `Execute` calls `sender.Do` once and only sees the final result — the per-attempt retry loop is owned by `retryablehttp`, one layer below `Execute`, where each try is a single `Transport.RoundTrip`. So per-attempt observation cannot happen in `Execute`. Recording is done in a `RoundTripper` wrapped around the tuned transport instead, which is the only point that sees every attempt and is the idiomatic Go spot for per-attempt HTTP metrics. `Execute` is unchanged; `operation` already rides the context via `OperationIDKey`, and `api`/`broker` are constants captured when the client is built.

The `docs/observability.md` schema for these metrics was finalised ahead of this change (pilot review, Story 41), so this implements to that committed spec; the golden file asserts the published names and labels against real `/metrics` output.


solace-aross (dismissed):
Verified the core claim by injection: sabotaged `nextAttempt` in `retry.go` to always return `1` and re-ran `TestClient_Execute_RecordsPerAttempt` (drives a 503 then a 200) — it went red, collapsing both samples onto `attempt="1"` instead of `1`/`2`. Reverted, clean. The retry-storm-visibility claim is genuinely tested end-to-end against a real retryablehttp retry loop, not just a label-formatting test.

Also checked the shared-state risk this kind of transport-wrapping usually carries: `resilience.New` mutates `httpClient.Transport` in place, but `sempv1.NewHTTPClient`/`sempv2.NewClient` each construct a fresh `*http.Client` per broker per protocol version, so there's no cross-broker aliasing. `pool.go`'s `append(p.senderOptions(), ...)` is safe too — `senderOptions()` returns a fresh literal (or nil) every call, never a shared backing array. `SEMPMetrics.Record` is nil-safe and the transport wrapper is only installed when the recorder is non-nil, so there's no redundant nil-checking masking a real gap. Label cardinality is bounded (operation IDs come from the embedded catalog / a v1 constant, broker is the canonicalized alias, attempt is capped by the existing retry limits).

One thing still open: Copilot's comment on the CHANGELOG entry (wrong Jira number, SOL-152091 instead of SOL-152093, in the parenthetical) is correct and still unaddressed — worth a one-line fix before merge, not a blocker.

CI green across the board (Guardian gate shows `skipping`, which is the known repo-wide pattern, not this PR's issue). Approved.

bczoma (dismissed):
## PR Review: #387 — SOL-152093: Per-attempt SEMP request and latency metrics

**Author:** solace-wmaryam | **Branch:** `SOL-152093-semp-per-attempt-metrics` → `main` | **Changes:** +594 / -40 across 16 files
**Jira:** `SOL-152093` — Per-attempt SEMP request and latency metrics so SREs can spot retry storms and broker latency (Story 16) — status *Verifying*

---

### Verdict

**Recommendation:** ✅ Approve

Nothing here is a condition of merge. The eight 🟢 notes below are worth fixing but do not hold the verdict, and they get no threads — this note is the only record of them.

One item is still outstanding from the automated round and is **not** re-filed here: Copilot's comment on `CHANGELOG.md:30` — the new bullet's parenthetical reads `(SOL-152091)` where it should read `SOL-152093`, while the same bullet correctly ends "Tracked under SOL-152093". Verified still unfixed at `9f714fd`.

---

### Summary

This adds two customer-visible metric families to `/metrics` and does it at the right seam: the story's work breakdown said to instrument `Execute`, but `Execute` only ever sees the final result because `retryablehttp` owns the retry loop one layer down, so per-attempt observation there was impossible — the `RoundTripper` is the only point that sees every try, including the ones that never return a response. The deviation is argued rather than asserted, the doc was written ahead of the code and the code implements it exactly (names, all seven label keys, the eleven bucket boundaries, the empty-status no-response rule all line up across `docs/observability.md`, `instruments.go`, and the golden file), and the off-gate is enforced at two independent levels with a test that asserts the observable consequence rather than the internal flag.

Two claims were filed at 🟡 and both came down on measurement. The transport double-wrap was challenged directly: the harm needs a deliberate three-file signature change to share an `*http.Client` between the two protocol clients, whereas the actual connection-pool optimisation — sharing the `*http.Transport`, which is where the pool lives — is already safe under this code, so it reduces to an undocumented ownership invariant. The "riskiest logic is untested" claim was contradicted by a mutation probe from another lens: replacing `nextAttempt`'s body with `return 1` does fail `TestClient_Execute_RecordsPerAttempt`, so the behaviour is covered — just two packages away from where it lives. What is left is a set of semantic and documentation details that are cheap now and expensive after the schema freeze.

---

### Minor notes

- 🟢 `docs/observability.md:223` — cardinality is documented as bounded, but `attempt` on the *histogram* multiplies through: with `DefaultRetries = 10` (11 attempt values) × 14 histogram series per combination, three brokers, ~40 v2 operations and five observed statuses gives ~90k series from one instrument. Bounded is not the same as affordable, and label sets are described in this doc as effectively frozen after GA. Consider carrying `attempt` on the counter only — that is what retry-storm detection actually needs — or at minimum put the worst-case arithmetic in the doc so the freeze is an informed decision. *(good:architecture)*
- 🟢 `internal/semp/resilience/metrics.go:38` — `attempt` counts round trips, not retryablehttp tries, and neither client sets `CheckRedirect`, so `net/http`'s default follows up to 10 redirects and each hop is a fresh `RoundTrip` on the same `retryState`. A broker that 301/302s (http→https, path normalisation) renders one try as `attempt="1"` then `attempt="2"` — a false retry signal in the exact metric the story exists to make readable, and it contradicts both `nextAttempt`'s "1-based try counter" comment and `docs/observability.md:212` ("the retry attempt"). Either set an explicit `CheckRedirect` on the SEMP clients (arguably right anyway, given auth headers on redirect) or make the label's documented meaning honest. *(flagged by good:architecture + good:code)*
- 🟢 `internal/semp/resilience/metrics.go:40-42` — the histogram times `base.RoundTrip` only, so it ends at response *headers*: it excludes the semaphore and rate-limiter waits in `Sender.Do` and excludes the body read, where SEMPv1 `<show>` envelopes and large SEMPv2 pages actually stream. Excluding backoff is correct and deliberate; the boundary just is not written down. One sentence in the metric's section — "time to response headers; excludes MCP-side pacing/queueing and body transfer" — makes the freeze lock a documented semantic instead of an inferred one. *(flagged by good:architecture + good:code)*
- 🟢 `docs/observability.md:99` + `internal/observability/schema/version.go:24` — `metrics_schema` stays at `1.0` while two families land, although this doc's own rule at lines 104-107 says a minor bump is additive ("a new metric") and tells operators to pin dashboards to `mcp_schema_version`. An operator who followed that advice gets no drift signal. Either bump to `1.1` (doc line 99 + the golden) or add one sentence saying the version is held at 1.0 until the schema freezes at GA. *(docs)*
- 🟢 `docs/observability.md:160-165` — the `/metrics` **Exposure** paragraph still enumerates what is exposed as "build version, schema versions, and — once tools run — tool names, broker aliases, and usage timing". This change adds `server_address`, i.e. broker hostnames/IPs from `req.URL.Hostname()`, to an unauthenticated endpoint that defaults to a wildcard bind. The data is fine to publish to an authorised scraper; the gap is that the risk statement operators read to decide whether `:9091` can be left reachable is now incomplete. *(good:security)*
- 🟢 `docs/observability.md:207-208` — `operation` is documented only as "(the SEMP operation)", but `operationID` returns the literal sentinel `unknown` when no operation ID is on the context, so `operation="unknown"` is a real emitted value. The analogous `none`/`unknown` sentinels *are* documented for the tool metrics' `broker` label. An SRE filtering a per-operation panel by the names they know silently drops the one bucket that shows a call path which lost its operation tag. *(docs)*
- 🟢 `docs/observability.md:419-421` — the Trace Exemplars section says `mcp_semp_request_duration_seconds` carries exemplars "when both metrics and tracing are enabled", but no SEMP span exists yet (the CHANGELOG itself defers it to Story 26), so the histogram will carry none. The section already carries two "because both look like bugs otherwise" caveats; this is a third, and without it an operator debugs their sampler config instead. *(docs)*
- 🟢 `internal/semp/resilience/sender.go:355-368` — `New` installs the wrapper by rewriting the caller's `httpClient.Transport` in place, non-idempotently, and no caller shares a client today so nothing misbehaves. The residue is that the resulting "hand me a fresh, exclusively-owned client" requirement is recorded nowhere — conspicuous in a doc comment that spells out the sharing contracts of `sem` and `limiter` and panics to enforce them. A shallow copy (`wrapped := *httpClient; wrapped.Transport = &metricsTransport{...}`, assigned after line 296) makes double-wrapping unrepresentable for two lines; a doc sentence is the cheaper alternative. *(flagged by good:architecture + good:code; reduced from 🟡 under challenge)*

...and 12 further minor items — the `nextAttempt` silent-fallback when no retry state is on the context, folding `WithAPI` into `WithMetrics` so `api=""` cannot compile, `retryState`'s doc comment not naming `attempt`, the degenerate `operation="SEMPv1"` label on every v1 series, the missing `[Implemented]` tag on the SEMP doc heading, the two now-decided Open Items, the ticket sitting in *Verifying* while this PR is unreviewed, and a few smaller ones. Ask if you want them.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

The ⚠️ marks reflect the 🟢 notes above, none of which blocks: the categories are marked on the state of the code, and every finding that would have justified a stronger mark was reduced by measurement.

amitmorade (dismissed):
## Approval

Reviewed across the spec, metrics package, retry transport, construction/lifecycle, tests, and client wiring. The important invariant holds: **one sample per HTTP attempt**, not per `Execute`.

Recording in a `RoundTripper` around the shared resilience `Sender` is the right layer. `Execute` only sees the final result; `retryablehttp` is what actually loops. Attempt state is attached once in `Sender.Do` and survives request copies, so retries climb `attempt` instead of collapsing into a single call. Duration is RoundTrip-only, so broker RTT is not mixed with admission or backoff. No-response tries record an empty status. Metrics stay off when the recorder is nil.

Provider build is moved before the broker pool, `WithSEMPMetrics` / `WithAPI` land before the transport wrap, and a failed or disabled SEMP instrument leaves clients working with recording inert. Names, labels, buckets, and the golden `/metrics` scrape match the committed observability schema.

The 503→200 client test plus the empty-status and nil-recorder cases are enough to prove the feature for this story.

**Approved.**

solace-wmaryam:
All review comments have been addressed in the latest push. Summary:

**Fixed:**
- CHANGELOG typo: removed the erroneous `(SOL-152091)` parenthetical from the SEMP bullet (Copilot, @solace-aross)
- `attempt` label removed from the histogram — counter only, so bucket series are not multiplied by the retry cap (🟢1)
- Redirect hops no longer counted as retry attempts — `metricsTransport.RoundTrip` checks `req.Response != nil` and passes through without recording; redirects themselves are still followed transparently (🟢2)
- Duration measurement boundary documented: time to first response byte, excludes admission wait and body read (🟢3)
- `metrics_schema` bumped to `1.1` (🟢4)
- `server_address` added to the `/metrics` exposure security note (🟢5)
- `operation="unknown"` sentinel documented in the label description (🟢6)
- `nextAttempt` bypass case made explicit — no longer falls through `getRetryState`'s fallback silently (minor #1)
- `retryState` struct doc comment updated to mention `attempt` (minor #3)
- `operation="SEMPv1"` documented as a constant for all v1 calls, with the reason (minor #4)
- `[Implemented]` tag added to the SEMP section heading (minor #5)
- Open Items 1 and 2 (histogram buckets, no-response behaviour) marked decided (minor #6)

**Not addressed — two items:**

- **Trace exemplars caveat** (🟢7): the exemplars section lists `mcp_semp_request_duration_seconds` as carrying exemplars, but no SEMP span exists yet (deferred to Story 26). Left as-is intentionally — the section will be accurate once Story 26 lands, and adding a caveat now that we then have to remove adds churn for no operator benefit.

- **`WithAPI` / `WithMetrics` merge** (minor #2): agreed the pitfall is real. The fix requires moving `WithMetrics` from the pool to `NewBrokerClient`, adding a new parameter there, and updating ~10 test call sites. The current code already works correctly — the client constructors always set `WithAPI` — so the risk is documentation-only. The refactor cost is disproportionate to a minor safety improvement on an internal option with a single production wiring path. Happy to revisit at a later story if it causes a real problem.

internal/semp/resilience/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+	api         string
+	brokerAlias string
+}
+
+// RoundTrip times one try and records its method, host, status, and attempt
+// number. A try that gets no response records an empty status. Redirect hops
+// (req.Response != nil) are passed through without recording — they are not
+// new SEMP attempts.
+func (t *metricsTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Response != nil {
```
solace-aross, line 40: Verified by injection (real 302→200 redirect through the SEMPv2 client, not just reading): this correctly stops the redirect hop from inflating `attempt` to 2 — confirmed the server sees 2 real HTTP requests but only one `mcp_semp_request_total` series is emitted. The side effect: that one series carries the *first* hop's status and duration, not the resolved one. A broker that always 302s would show `http_response_status_code="302"` on every scrape, at the (fast) redirect-response latency — never the 200 the caller actually got, and never the true round-trip time including the followed hop. The client's own return value is correct (`Execute` gets the real body); only the metric is stuck on the intermediate response.

Not a blocker: this is the same hypothetical scenario bczoma filed as non-blocking (SEMP is an internal admin API, not known to redirect), and it's strictly less harmful than the false-retry-count problem this fix solves. Worth a one-line addition if it's cheap: record on the *last* hop's status/timing instead of skipping it outright — thread `start` and the counted `attempt` forward via context so the final `RoundTrip` call records once, using the original attempt number and start time. Not asking for it now, just flagging it while it's fresh from injection rather than leaving it to be rediscovered from a real broker redirect later.

solace-aross (approved):
This fixup addresses bczoma's and amitmorade's earlier review thoroughly — CHANGELOG typo fixed, `attempt` moved off the histogram (closing the ~90k-series cardinality concern), `metrics_schema` bumped to 1.1, docs updated with the duration-boundary and operation-sentinel semantics, and both open items marked decided. Read all of it; it's accurate.

Verified the one claim that changes runtime behavior (not just docs): sabotage-tested by writing a throwaway redirect scenario (a real `httptest.Server` returning 302 then 200, driven through the real `SEMPv2` client). Confirmed the redirect hop no longer inflates `attempt` to 2 — that part of the fix works. But the same test surfaced a real, unaddressed side effect, which the review below explains.

Approved — this is a clean fixup and CI is green. One follow-up note, not a blocker.

bczoma (approved):
## PR Review: #387 — SOL-152093: Per-attempt SEMP request and latency metrics

**Author:** solace-wmaryam | **Branch:** `SOL-152093-semp-per-attempt-metrics` → `main` | **Changes:** +646 / -78 across 17 files
**Jira:** `SOL-152093` — Per-attempt SEMP request and latency metrics so SREs can spot retry storms and broker latency (Story 16)

Follow-up review: since my review at `9f714fd` the branch carries one authored commit, `3ac2adf` ("Addressing review comments" — 8 files, +81/−60), plus a merge of `origin/main` at `3f1d37b`. The merge is what makes the raw range look like 74 files; none of that is this PR's work and none of it is re-reviewed here. No lenses re-dispatched: the new commit adds no files and touches metrics, docs, and the retry-state helper only.

---

### Verdict

**Recommendation:** ✅ Approve

Nothing is a condition of this approval. The 🟢 notes below hold nothing.

---

### Summary

Six of the eight prior notes are closed, and closed at the right seam rather than papered over in prose. `attempt` now rides the counter only, which is the change that actually retires the ~90k-series arithmetic — the histogram's bucket series are no longer multiplied by the retry cap, the doc's own cardinality paragraph says so, and both the unit golden and `metrics_golden.txt` were regenerated to prove the label is gone from the histogram rather than merely undocumented. `metrics_schema` moved to `1.1` in the constant, its doc comment, `docs/observability.md:99` and the golden together, so an operator who pinned to `mcp_schema_version` as that doc instructs now gets the drift signal the doc promised. The duration's boundary, the `operation="unknown"` and `operation="SEMPv1"` sentinels, and `server_address` in the `/metrics` Exposure paragraph are all written down now, and the two decided Open Items were struck through rather than left as live questions.

The redirect note was answered with a different fix than either option I offered — skipping the record when `req.Response != nil` rather than setting `CheckRedirect` — and the mechanism is right: `net/http` sets `Request.Response` only on a request it generated to follow a redirect, and `retryablehttp` reuses the original request across tries without ever setting it, so no genuine retry is swallowed. It does change the failure mode rather than remove it, which is the one 🟢 below.

One prior note is now moot rather than fixed: the Trace Exemplars section claimed `mcp_semp_request_duration_seconds` would carry exemplars when no SEMP span existed. `semp.request` landed in `internal/semp/sempv2/client.go:199` on `main` and came in with the merge, so the doc is now simply true.

*8 prior findings settled — 6 addressed, 1 overtaken by the `main` merge, 1 below.*

|---|---|---|---|
| 🟢 | `resilience.New` rewrites the caller's `httpClient.Transport` in place, non-idempotently; the resulting exclusive-ownership requirement is recorded nowhere | Not addressed | `internal/semp/resilience/sender.go:382-386` — unchanged; still no doc sentence and no shallow copy. Still not a blocker, still cheapest as one sentence beside the `sem`/`limiter` contracts |

---

### Minor notes

- 🟢 `internal/semp/resilience/metrics.go:38-41` — the redirect guard drops the hop entirely, not just its `attempt` number, and the hop is where the real response lives. On a broker that 301s (http→https, path normalisation), `RoundTrip` records attempt 1 with `http_response_status_code="301"` and the following hop — the one that returns the 200 and carries the actual broker latency — is never recorded at all. An SRE reading `mcp_semp_request_total` for that broker sees a call path that is 100% 3xx and no successes, and the latency histogram measures only the redirect. Recording the hop while leaving `attempt` at its current value would keep the false-retry fix and keep the terminal status; setting `CheckRedirect` to refuse redirects outright is the other end of it and is arguably right anyway with auth headers in play.
- 🟢 `docs/observability.md:213` + `CHANGELOG.md:30` — both now say redirects are "followed transparently and not counted as retry attempts", which is accurate about `attempt` and silent about the consequence above: the redirected response is not counted at all. Whichever way the point above lands, this sentence should say which of the two it means.
- 🟢 `internal/semp/resilience/` — the `req.Response != nil` branch is new logic on the recording path with no test. A two-line `httptest` server that 302s to a second handler, asserting how many series the recorder saw and with which status, would pin the intended behaviour before the freeze; today the guard's contract lives only in its comment.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ — sourced: parent (follow-up); the label split and the redirect guard both do what they claim |
| Code Quality | ✅ — sourced: parent (follow-up) |
| Test Coverage | ⚠️ — sourced: parent (follow-up); goldens regenerated for the label split, but the redirect branch is untested |
| Documentation | ✅ — sourced: parent (follow-up); six doc notes closed, one wording point above |
| Architecture | — not re-reviewed (unchanged since `9f714fd`) |
| Security | — not re-reviewed (unchanged since `9f714fd`) |
| PR Scope & Description | — not re-reviewed (unchanged since `9f714fd`) |

internal/semp/resilience/metrics.go
solace-wmaryam: Thanks for the injection test — good catch. Leaving this for now: SEMP is a directly-configured admin API and redirects are not a known production scenario, so this is a hypothetical edge case. The current behaviour (recording the 302) is strictly less harmful than the problem the fix solved (counting the redirect as a false retry attempt). The proposed fix — threading `start` and `attempt` forward through context so the final hop records — adds real complexity for a case we have no evidence of hitting. If a real broker redirect surfaces in a pilot, we can revisit then.

PR #388, opened 2026-09-08
SOL-152097: Record auth success, failure, and refresh in the audit stream

solace-aross:
## What

Story 23 (SOL-152097): "Record auth success, failure, and refresh in the audit stream to detect anomalous access."

Three new audit-record emission sites, all behind the existing `OBS_AUDIT_LOG_ENABLED` flag (off by default, door-closing policy) — no behavior change with the flag off.

1. **`internal/auth/middleware.go`** — `auth_success` / `auth_failure` after `TokenVerifier` resolves, on both the static dev-token and OIDC paths. Also covers a request presenting no bearer token at all (`auth_failure{reason=missing}`), which never reaches our `TokenVerifier` — a thin wrapper that only observes the SDK's own `RequireBearerToken` chain, never changes its 401 decision.
2. **`internal/semp/resilience`** — `broker_auth_retry` on a 401/cookie-clear or OAuth re-auth retry. `outcome` (`success`/`error`) is decided exactly once, in `Sender.Do`, after the whole retry chain concludes.
3. **`internal/tools/authorization.go`** — `authz_denied` on both hop-1 deny branches (`missing_claim`, `not_permitted`), alongside the existing `"tool authorization"` WARN, reusing its `decision_reason` literals verbatim.

## The auth-failure classifier decision (read this if reviewing)

The ticket asks `auth_failure`'s `reason` to use "the same closed set as `mcp_auth_failure_total{reason}` from Story 24 (SOL-152099) ... taken from the same classification." **Story 24 has not landed** (still `To Do`, assigned to a colleague) and is only a Jira "coordinate with," not a blocking dependency of Story 23 — no such classifier exists anywhere on `main` today.

So this PR builds it now: `auth.ClassifyAuthFailure(err error) string` in `internal/auth/classify.go`, mapping to a closed 5-value vocabulary (`invalid_token`, `expired`, `audience_mismatch`, `signature_invalid`, `missing`). It lives in `internal/auth`, not the audit package, specifically so Story 24 can call the exact same function for its counter labels — `auth.AuthFailureReasons()` exposes the vocabulary itself for label pre-registration, and a new drift-guard test (`TestAuthFailureReasons_MatchesEventSchemaVocabulary`) keeps this package's copy and the audit schema's own closed set from silently diverging (mirrors `internal/tools/audit_error_type_drift_test.go`'s guard for `error_type`).

**Not scope creep** — this is the classifier a not-yet-landed sibling story needs, built here because nothing else defines it yet, documented so Story 24's author isn't left to reverse-engineer the shape or duplicate it.

## Design note: the auth↔audit import cycle

`internal/observability/audit` already imports `internal/auth` for identity (`auth.PrincipalFrom`). Triggering audit emission directly from inside `internal/auth`'s `TokenVerifier` closures would create the reverse edge — a real cycle, not just bad layering. Fixed with `AuthAuditHook`, an interface *defined* in `internal/auth` (`audit_emit.go`) and *implemented* in `internal/observability/audit` (`auth_hook.go`, `audit.NewAuthHook`), injected as a plain parameter (nil = off) rather than a variadic slot — a variadic hook would silently drop a second one with no error, the wrong failure mode for an audit observer. cmd/server wires the real implementation.

## Correctness note: `broker_auth_retry`'s outcome, decided once

An earlier version of this PR decided the `broker_auth_retry` outcome from inside `checkRetry`, at the point the 401 first recovers. That's wrong: `checkRetry` can run many more times afterward (a transient 503/500 retry, a second 401) or, on some paths, never run again at all (a transport error, context cancellation during backoff, the re-auth itself failing in `prepareRetry`). Deciding it mid-chain either double-emitted contradictory records for one request (a `401→503→401` chain produced both `outcome=success` and `outcome=error`) or silently dropped the record on the paths that skip `checkRetry`. Fixed by deciding it exactly once, in `Sender.Do`, after the whole chain concludes — covered by dedicated regression tests (`TestBrokerAuthRetry_401Then503Then401`, `TestBrokerAuthRetry_401Then503ThenSuccess`, `TestBrokerAuthRetry_TransportErrorAfterRetry`, `TestBrokerAuthRetry_ContextCancelledDuringBackoff`).

`outcome` answers "did the credential problem get resolved?", not "did the call succeed?" — a 401 that recovers into a broker overload which then exhausts its own retry cap is `outcome=success` here even though the call itself ultimately fails. Documented in `docs/observability.md`.

## Docs

`docs/observability.md`: flipped two stale claims this work makes wrong — the "authentication and denial record types are not emitted yet" note (three of the four now are; `broker_authz_denied`, SOL-153332, still is not), and a claim that an expired or audience-mismatched token carries `principal.sub` on `auth_failure` (it does not — those fail before any claim is readable server-side). Story 35 (SOL-152094) still owns the full documentation pass; this is the narrower "don't ship a doc that's now factually wrong" fix.

## Verification

- `make check` green: build, vet, `golangci-lint` (0 issues), full race-enabled test suite, 89.5% coverage (floor 85%).
- `/check-logs`: no violations. The only two new production log statements (`retry.go`, `authorization.go`) log only `audit.NewEvent`'s own contract-safe error text, matching the existing `emitOperationAudit` pattern.
- Reviewed by a skeptical Opus pass focused on field-applicability, hop-1 test scoping, credential/group-membership leakage, and the classifier's reusability. Findings addressed: the `broker_auth_retry` double-emission/missing-emission bug above, a doc contradiction, a false "pins against go-oidc" test claim, the vacuous hop-1 "no operation record" assertion (now backed by a positive control proving an operation record really was reachable from the same call shape), the missing-bearer-token gap, and the single-slot-hook design smell.

Tracked under SOL-152097.


internal/semp/resilience/retry.go
```diff
@@ -320,3 +332,66 @@ func (d *Sender) checkRetry(ctx context.Context, resp *http.Response, err error)
...
+// separate question the caller's returned error answers — and error
+// otherwise: a persisted 401, a transport failure, a context cancellation, or
+// the re-auth attempt itself failing inside prepareRetry.
+func (d *Sender) auditBrokerAuthRetryOutcome(ctx context.Context, resp *http.Response, err error) {
+	state := getRetryState(ctx)
+	if !state.auth401Retried {
+		return
+	}
+	outcome := audit.OutcomeError
+	if err == nil && resp != nil && resp.StatusCode != http.StatusUnauthorized {
```
bczoma: 🟡 **The published `outcome` contract and this predicate disagree on a chain the doc names explicitly**

Flagged by good:architecture (at 🔴) and good:code independently. Challenged, and the challenge settled a disagreement between reviewers as well as reducing the severity — so the measurements are below rather than asserted.

**The disagreement first.** A third lens read the same doc region and reported that it matched the code. It does not. `docs/observability.md:667-674` says:

> `outcome` answers "did the credential problem get resolved?", not "did the call succeed?" — a 401 that recovers into a broker overload (503) which then exhausts its own retry cap is `outcome: success` here even though the call itself ultimately fails; join on `correlation_id` to the call's own `operation` record … to see the call's actual disposition.

This function's own doc agrees with that reading — "success if the chain ended in a non-401 response … whatever this call's own final disposition turns out to be". The predicate is `err == nil && resp != nil && resp.StatusCode != http.StatusUnauthorized`.

**Why the two cannot both be true.** Every retries-exhausted terminal routes through `errorHandler` (wired at `sender.go:308`), which returns a **nil** `*http.Response` on all of its populated paths (`sender.go:835`, `:842`, `:851`). So on an exhausted chain `resp != nil` is unsatisfiable, and the documented `success` case emits `error`. Confirmed at runtime, not by inspection — a throwaway probe (401, then 503 on every subsequent attempt, `retries=10`) produced:

```
Do() resp=<nil> err=request failed after 5 attempts: transient-error (429/503) retry cap reached
PROBE outcome=error
```

Note `RetriesExhaustedError.StatusCode` **is** set to 503 on that path (`sender.go:836`) — the information the audit field discards is right there.

**Why this is 🟡 and not 🔴.** The divergence is one-directional and fail-safe. `success` requires a real non-401 response, so no unresolved credential is ever labelled `success`; the only failure mode is over-reporting credential problems. And the affected chain needs a coincidence: `maxTransientRetries = 3` (`retry.go:37`), so it takes a credential expiring *inside* a broker overload episode lasting four or more attempts of one call.

What does not survive is the doc's own remedy. It tells a reviewer to join on `correlation_id` to recover the call's disposition — but the two WARNs that actually disambiguate resolved-from-unresolved (`retry.go:248` and `:255`) use `slog.Warn`, not `WarnContext`, so they carry no `correlation_id` at all. A reviewer can reconstruct the truth by reading logs for that broker; they cannot do it with the join the doc offers.

**Recommended fix — the code, not the doc.** Three separate texts (this doc, `CHANGELOG.md:35`, and the function doc at `retry.go:349-355`) all describe the credential-scoped meaning, and it is the more useful of the two: a call-scoped `outcome` is redundant with the same call's `operation.outcome`. So carry the fact the record is about:

```go
// in checkRetry, once state.auth401Retried is true:
if resp != nil && resp.StatusCode != http.StatusUnauthorized {
    state.authRecovered = true
}
```

and derive `outcome` from `state.authRecovered` alone here. (`errors.As` on `*RetriesExhaustedError.StatusCode` is a partial substitute but reads 0 on transport errors.)

Either way, please add the pin — `TestBrokerAuthRetry_401Then503Exhausted`. The file has nine tests including `_401Then503Then401` and `_401Then503ThenSuccess`; the exhausted variant, which is the one case where the two readings diverge, is the one that is missing.

docs/observability.md
```diff
@@ -47,7 +47,7 @@ capability headings carry the same tag:
...
-| Audit trail | **[Interim — `operation` records only]** | Destructive tool calls emit an `operation` record behind `OBS_AUDIT_LOG_ENABLED` (default off). The authentication and denial record types, and the `mcp_audit_events_dropped_total` counter, are not emitted yet. See [Audit Trail](#audit-trail--interim--operation-records-only). |
+| Audit trail | **[Interim — `operation` records only]** | Destructive tool calls emit an `operation` record behind `OBS_AUDIT_LOG_ENABLED` (default off). `auth_success`, `auth_failure`, `authz_denied`, and `broker_auth_retry` also emit today (SOL-152097). `broker_authz_denied` and the `mcp_audit_events_dropped_total` counter are not emitted yet. See [Audit Trail](#audit-trail--interim--operation-records-only). |
```
bczoma: 🟡 **The Audit Trail status tag contradicts the prose directly beneath it**

Flagged by docs.

The capability table here and the section heading at `:445` both still read `[Interim — `operation` records only]`, while the prose under each now says four more record types emit today. The tag and its own body disagree.

**Why it matters more than a stale label.** This table is the one the doc tells a reader to consult first. A compliance reviewer scanning it sees "operation records only", concludes the authentication and denial records are not available yet, and does not build the `auth_failure` / `authz_denied` detection rule — which is the entire outcome this story exists to enable. The records would be flowing and nobody would be looking at them.

**Fix.** Retag both to something true, e.g. `[Interim — all record types except `broker_authz_denied`]`, and update the heading to match. Note the anchor `#audit-trail--interim--operation-records-only` is referenced from this file (line 50 and the intra-doc links) and from `docs/user-guide.md:175`, so both files need to change in the same commit or the links break.

internal/auth/middleware.go
```diff
@@ -88,31 +104,60 @@ func NewAuthMiddleware(cfg *config.ServerConfig, httpClient *http.Client, next h
...
+// independent copy of that parsing — the one thing this must never disagree
+// with. This wrapper only ever adds a record alongside the SDK's own 401,
+// never changes whether a request is accepted: it always calls next
+// regardless of what it observes.
+func auditMissingBearerToken(hook AuthAuditHook, next http.Handler) http.Handler {
+	if hook == nil {
+		return next
+	}
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if _, ok := parseBearerToken(r.Header.Get("Authorization")); !ok {
```
bczoma: 🟡 **`reason=missing` now conflates three structurally different causes, one of them an IdP misconfiguration**

Flagged by docs and good:security, independently and from different directions.

Three distinct conditions land in the same bucket:

1. **No `Authorization` header at all** — `auditMissingBearerToken` here calls `ClassifyAuthFailure(nil)` → `missing`. This is the unauthenticated-probe case.
2. **A header present but not a parseable bearer token** — `Basic …`, a wrong scheme, `Bearer` with no value. `parseBearerToken` fails, so this is also recorded as `missing`, indistinguishable from having sent nothing.
3. **A fully verified, validly signed token that omits `sub`** — `errors.Is(err, errNoSubject)` in `internal/auth/classify.go` → `missing`. This is a *successfully authenticated* caller whose IdP is misconfigured.

**Why it matters.** An operator builds "unauthenticated access attempt" alerting on `reason=missing`. Case 3 then lands in the same bucket: the probe count is inflated by a caller who authenticated fine, and the real IdP misconfiguration never surfaces as its own signal because it never appears as anything but a probe. Case 2 loses the distinction between "client misconfigured" and "someone is presenting credentials of the wrong kind" — and `invalid_token` exists in the closed set precisely for that shape.

**Fix — two parts, both cheap, no schema change.**

- Classify case 2 as `invalid_token` rather than `missing`: a present-but-unparseable `Authorization` header is a token problem, not an absent one. Both values are already in the vocabulary.
- For case 3, either give it its own value or say plainly that `missing` covers both meanings. `docs/observability.md:259-262` is the only place the five values are enumerated and only `signature_invalid` carries a gloss — a five-row table like the `authz_denied` one at `:686-689` would fix that and is where a reader will look.

The first is the one I'd take: it makes the audit stream answer the question an operator is actually asking, rather than documenting why it can't.

docs/observability.md
```diff
@@ -661,7 +667,12 @@ field:
 
 - `auth_success`, carrying `principal` and `agent_client_id`.
 - `auth_failure`, carrying `reason` (same closed set as `mcp_auth_failure_total`).
```
bczoma, line 676: 🟡 **The audit stream's new volume is undocumented, and one of the new record types is anonymous-triggerable**

Flagged by docs and good:security, independently.

**The cadence is not stated.** The volume expectation this doc sets is "one JSON event per state-changing operation … read-only calls are not audited". `auth_success` does not follow that rule: the hook fires inside the `TokenVerifier`, so it emits once per successful token verification — i.e. once per authenticated HTTP request to `/mcp`, every JSON-RPC POST, not only the destructive ones. `auth_failure` likewise, once per rejected request.

An operator sizing SIEM ingest before turning `OBS_AUDIT_LOG_ENABLED` on budgets from the documented per-destructive-call rate and gets orders of magnitude more. They then either blow the ingest budget or hit a rate limit that silently drops part of the audit stream — which is the worst outcome available here, because the dropped part is the evidence channel.

**And one record type is driven by unauthenticated callers.** `auth_failure{reason=missing}` is emitted per tokenless request, before any admission control — the wrapper sits inside `crossOriginProtection` but outside any rate limiter — with no throttle, dedup or per-source cap. A load-balancer probe, an SSE reconnect, or a scripted unauthenticated POST loop each produce a record. This is the one record type in the schema an anonymous party controls the volume of. It is only reachable with the flag on, but "on" is exactly the compliance configuration.

Neither is new *in kind* — the verification path already carries unbounded WARNs — but the previously-silent "no token at all" case now writes to the audit stream, and that is new.

**Fix.** The doc part is one or two sentences here: state that `auth_success`/`auth_failure` are emitted once per authenticated/rejected request rather than per session or per destructive call, and note that `reason=missing` is anonymous-triggerable so retention is sized with that in mind. If the audit stream ever gets a volume budget, a coarse cap on that one `reason` value is the place to spend it.

bczoma (dismissed):
## PR Review: #388 — SOL-152097: Record auth success, failure, and refresh in the audit stream

**Author:** solace-aross | **Branch:** `sol-152097-audit-auth-events` → `main` | **Changes:** +2120 / -101 across 24 files
**Jira:** `SOL-152097` — Record auth success, failure, and refresh in the audit stream to detect anomalous access (Story 23) — status *Verifying*

---

### Verdict

Nothing blocks. One finding was filed at 🔴 and came down to 🟡 on measurement; another at 🟡 came down to 💡.

1. 🟡 `internal/semp/resilience/retry.go:363` — the published `outcome` contract and the code disagree on a named chain, and no test pins it
2. 🟡 `docs/observability.md:50` — the Audit Trail status tag still reads `[Interim — operation records only]` directly above prose saying four more types now emit
3. 🟡 `internal/auth/middleware.go:135` — `reason=missing` now conflates three structurally different causes, one of which is an IdP misconfiguration
4. 🟡 `docs/observability.md:669` — the audit stream's new volume is undocumented, and one record type is anonymous-triggerable and unthrottled

---

### Summary

This is careful work on the surface where careless work is most expensive. The import cycle is genuinely inverted rather than deferred — the interface is declared in the consumer and implemented in the dependency, which is the correct side of the boundary in Go — and the security lens traced every branch to confirm the observer cannot alter an authentication decision: the hook methods return nothing, every rejection branch still returns its original sentinel, a constructor rejection is handled locally and never propagates, and the flag-off path is inert. Field applicability is left to the Story 21 constructor at all four sites, hop-1 scoping is respected with no hop-2 emission anywhere, and the per-request retry state is correctly scoped — a fresh state per `Sender.Do`, read on the caller's own goroutine, so a shared Sender cannot cross requests.

The tests earned their claims under mutation rather than by assertion. Breaking the two string-matching branches of `ClassifyAuthFailure` in turn fails tests that drive a real `oidc.Verifier` through a mock IdP, so the documented go-oidc-bump drift guard is real; setting `Outcome` on the `authz_denied` record fails both deny tests; and the hop-1 positive control drives the *identical* `next` through an allowing policy, so the "zero operation records" assertions mean something. That last one was called out in the PR body as a fix to an earlier vacuous assertion, and it is.

Two claims were reduced by measurement, and both reductions are worth reading. The `outcome` divergence is real — a probe drove a 401 into a cap-exhausted 503 and observed `outcome=error` where the doc promises `success` — but it is one-directional: `success` requires a genuine non-401 response, so the mislabel over-reports credential failures and can never hide one. And the option-coupling concern turned out unreachable: all three `resilience.New` sites either inherit the alias unconditionally or set no audit flag, config validation forbids an empty alias, and a future caller who reached it would get an ERROR log plus an emitted `audit_drop` on the first 401 rather than a silent gap.

---

### Minor notes

- 🟢 **Jira SOL-152097's field-applicability table is wrong, and the code is right.** The ticket says "an expired or audience-mismatched token still parses far enough to yield `sub` and the client id"; in code `bestEffortIdentity` is reached only after `Verify` succeeded *and* claims decoded, so those two cases take the `reportAuthFailure(ctx, hook, …, "", "")` path with no identity. Attributing a pre-verification rejection would mean trusting unverified claim content, so the implementation is the safer choice and `docs/observability.md:548-552` now matches it. The ticket is in *Verifying*: amend that table, or whoever verifies against it will either sign off on an unrecorded deviation or file a bug against correct code. *(docs + good:security, independently)*
- 🟢 **Coordinate the classifier with SOL-152099's assignee before the enums ship.** Building it here is defensible — AC 2 cannot be met without one, none exists on `main`, and the drift-guard test makes divergence a CI failure — but the ticket says "coordinate both `reason` enums with Story 24", and defining another engineer's label vocabulary is the part that needs a handshake rather than an argument. A comment on SOL-152099 naming the function, the five values and the guard test, plus adding that assignee as a reviewer, turns an assertion into evidence. *(good:pr)*
- 🟢 `internal/observability/audit/emit.go` — the build-or-drop block (`NewEvent` → on error `slog.ErrorContext` + `EmitDrop` → else `Emit`) is now copy-pasted three times by this PR on top of the pre-existing copy in `emitOperationAudit`, each with its own message string and its own choice of `DropContext` fields. `emitAuthEvent` already *is* the generic version — exporting it as `EmitOrDrop(ctx, Fields)` and having all four sites keep only their gate check would mean a change to drop policy lands in one place instead of four. *(good:code)*
- 🟢 `internal/auth/audit_emit.go:59-70` — the single-slot hook is defended against variadics, but the alternative worth considering is a composite: `type AuthAuditHooks []AuthAuditHook` fanning out `Success`/`Failure` and skipping nils, ~5 lines in `internal/auth`. Story 24 then wires `auth.AuthAuditHooks{metricsHook, audit.NewAuthHook(...)}` with no new dependency edge and each capability keeping its own flag, instead of the combined hook the interface doc currently prescribes — which has to live somewhere importing both packages. *(good:architecture, reduced from 🟡 — the harm lands on a future story, not this one)*
- 🟢 `internal/semp/resilience/auth_retry_audit_test.go:440-480` — `TestBrokerAuthRetry_ContextCancelledDuringBackoff` needs a 30 ms goroutine sleep to land inside a 200 ms backoff window. On a loaded runner the cancel arrives after the retry has fired and the test fails on `requestCount != 1` — a flake in a `-race` suite. Signal from the handler and cancel once `checkRetry` has run, or raise `RetryWaitMin` to seconds so skew cannot close the window. *(good:code)*
- 🟢 `internal/auth/middleware.go:130-140` — `auditMissingBearerToken` reimplements the SDK's unexported bearer extraction, so "at most one `auth_failure` per rejected request" rests on two independent parsers agreeing across an SDK upgrade. A per-request marker on `r.Context()` that `reportAuthFailure` checks would make the property structural instead of parse-dependent, and would survive any future change to the SDK's header handling. *(good:architecture)*
- 🟢 `internal/tools/authorization.go:93` / `internal/tools/register.go:261` — `withAuthorization` now takes five positional parameters including a bare `bool`, and the call sites read `"groups", true,`. Grouping the static configuration into a small struct, or passing the emitter as a function value so "off" is a no-op rather than a flag, removes the argument-order hazard the test files now all carry. *(good:code)*
- 🟢 `CHANGELOG.md:35` — the entry opens "**Three** more audit-record types now have an emitter" and then names four. (The later "all three emit sites" is correct — three packages.) Everything else in the entry checks out against the code. *(docs)*

…and 12 further minor items — `docs/internal/architecture.md` still calling the audit package a skeleton in three places, the five `reason` literals existing twice, `broker.go`'s defensive double-append lacking the one line saying why, wrapping the two `report*` bodies in their own `recover` rather than relying on `audit.write`'s, a one-line `New` panic guard for the `auditLog`/`brokerAlias` pair (reduced from 🟡 — unreachable today, and its neighbour `WithScheduler` holds a stronger invariant the same conventional way), narrowing `Success` to `(sub, clientID)` instead of passing an SDK type, passing a defensive copy of `*TokenInfo`, emitting with `context.WithoutCancel`, splitting the single commit, an AC→test map, and the Verification section's framing. Ask if you want them.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

Security is a ✅ on evidence, not on absence: every leakage vector was traced to its value source — no token, header, cookie, client secret or `err.Error()` reaches any record, `authz_denied` has no field slot for groups anywhere, and claims are read only from an IdP-signed token.

internal/semp/resilience/retry.go
solace-aross: Fixed: auditBrokerAuthRetryOutcome now derives outcome from a retryState flag (authRecovered) maintained through the chain — cleared on any 401, set on a non-401 that follows one — instead of Sender.Do's own final resp/err. That flag correctly reports success on a 401 that recovers into a 503 which then exhausts maxTransientRetries (errorHandler nils the response on that path, which is what the old code misread), and still correctly reports error if a later 401 recurs after a transient recovery. New test: TestBrokerAuthRetry_401Then503Exhausted. Addressed in b876eab.

docs/observability.md
solace-aross: Fixed: retagged both to [Interim — all record types except `broker_authz_denied`] and updated the anchor here and in docs/user-guide.md:175. Addressed in b876eab.

internal/auth/middleware.go
solace-aross: Took the fix you'd take: a header present but not bearer-shaped now classifies as invalid_token instead of missing. Also expanded the five-value reason list into a gloss table (docs/observability.md), noting that missing covers both an absent header and a verified token with no sub, per your second option. Addressed in b876eab.

docs/observability.md
solace-aross: Added both notes: auth_success/auth_failure emit once per authenticated/rejected request rather than per destructive call, and reason=missing is anonymous-triggerable so retention should be sized with that in mind. Addressed in b876eab.

bczoma (approved):
## PR Review: #388 — SOL-152097: Record auth success, failure, and refresh in the audit stream

**Author:** solace-aross | **Branch:** `sol-152097-audit-auth-events` → `main` | **Changes:** +2241 / -106 across 25 files
**Jira:** `SOL-152097` (Story 23) — status *Verifying*
Follow-up review at `b876eab`. My approval at `e37adf9` was dismissed by this push, not withdrawn. The branch was rebased, so `git cherry` recovered the authored delta by patch-id: one new commit, 330 lines across 8 files, all of them already reviewed. Parent-only pass — no lenses re-dispatched.

---

### Verdict

**Recommendation:** ✅ Approve

All four 🟡 conditions are closed, and I checked each in code at this head rather than taking the commit message for it. Nothing new blocks; two 🟢 notes below, neither a condition.

Verified at this head: `go build ./...` clean; `go test` green on `internal/auth`, `internal/semp/resilience`, `internal/tools`, `internal/observability/audit`.

---

### Summary

The `broker_auth_retry` fix is the right one and goes further than the thread asked. Rather than rewriting the doc to match the code, `checkRetry` now maintains `state.authRecovered` as the credential status of the most recently observed response, and `auditBrokerAuthRetryOutcome` reads that instead of re-deriving it from `Do`'s final `resp`/`err`. The doc's worked example — a 401 recovering into a cap-exhausted 503 reporting `outcome: success` — is now true, and `TestBrokerAuthRetry_401Then503Exhausted` pins it. The part I did not ask for and would have missed: `authRecovered` is *cleared* on every 401, not merely set once, so a chain that recovers and then hits a fresh 401 reports the persisted failure it actually ended in rather than the transient recovery partway through. I traced the terminal shapes — persisted 401, recover-then-exhausted, recover-then-transport-error, re-auth failing in `prepareRetry` — and each now yields the credential-scoped answer the three published texts describe. The function's doc comment also now records *why* `resp`/`err` cannot carry the fact, naming `errorHandler`'s nil response, so the next reader does not have to rediscover it.

The `invalid_token` split landed with both halves of the thread. A present-but-unparseable `Authorization` header now classifies as `invalid_token` and an absent one as `missing`, each pinned by its own test — and the reason vocabulary gained a full five-row gloss table, where the `missing` row states outright that it covers both an absent header and a fully-verified token missing its `sub`, with the reasoning for sharing one value. That is more than the thread asked for. The status retag caught the anchor dependency I flagged, updating `docs/user-guide.md:175` in the same commit so the cross-reference does not break, and the volume note covers both the per-request cadence and the anonymous-triggerable amplification.

---

*All four 🟡 conditions addressed.*

|---|---|---|---|
| 🟡 | The published `outcome` contract and the code disagreed on a named chain; no test pinned it | Addressed | `internal/semp/resilience/retry.go:68` adds `authRecovered`; set at `:270-279` on a non-401 after a 401, cleared at `:243` on any 401; `auditBrokerAuthRetryOutcome` reads it at `:394` and its `resp`/`err` parameters are gone. Pinned by `auth_retry_audit_test.go:374` `TestBrokerAuthRetry_401Then503Exhausted` |
| 🟡 | The Audit Trail status tag contradicted the prose beneath it | Addressed | `docs/observability.md:50` and the `:452` heading both now read `[Interim — all record types except broker_authz_denied]`, and the anchor was updated in `docs/user-guide.md:175` in the same commit |
| 🟡 | `reason=missing` conflated three causes, one an IdP misconfiguration | Addressed | `internal/auth/middleware.go:143-149` classifies a present-but-unparseable header as `invalid_token`; tests at `auth_audit_test.go:419` (present, unparseable) and `:387` (absent). The dual meaning of `missing` is now documented at `docs/observability.md:267`, in a five-row gloss table |
| 🟡 | The audit stream's new volume was undocumented and one type is anonymous-triggerable | Addressed | `docs/observability.md:684-691` — the per-authenticated-request cadence, the "size ingest against request volume, not destructive-call volume" instruction, and the unauthenticated-amplification note for `reason=missing` |

The eight 🟢 notes from the last pass were not taken. They were offered as non-conditions and I am not re-arguing them; two are worth a moment when convenient — the `CHANGELOG.md` entry still opens "**Three** more audit-record types" while naming four, and SOL-152097's field-applicability table still contradicts the code on whether an expired token yields `sub` (the code is right; the ticket is in *Verifying*, so amending it stops a reviewer signing off on an unrecorded deviation).

---

### Minor notes

- 🟢 `internal/auth/middleware.go:147` — `reason = "invalid_token"` is a bare string literal at the call site rather than the classifier or a constant, making a third hand-written copy of a vocabulary value that already exists in `authFailureReasonVocabulary` and in `ClassifyAuthFailure`'s switch returns. `auth_audit_test.go:419` does pin it, so a rename breaks a test rather than shipping silently — but the prior 🟢 about those literals now has one more site to keep in step. A `reasonInvalidToken` constant used in all three places closes it.
- 🟢 `docs/observability.md:688` — the volume note justifies the amplification with "`auth_failure{reason=missing}` in particular is emitted before any authentication succeeds", which is now true of only one of the two meanings the `:267` gloss just established: the verified-token-without-`sub` case is emitted *after* full verification. An operator reading the volume note alone would conclude every `missing` is anonymous. Scoping the clause to the absent-header case ("when it means no header was presented") reconciles the two paragraphs.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ — the credential-scoped `outcome` now holds on every terminal path I traced, and the recover-then-fresh-401 case is handled |
| Code Quality | ✅ — the fix is at the right seam, and the doc comment records why the obvious implementation does not work |
| Test Coverage | ✅ — the missing exhausted-retry case is pinned, and both classification branches have their own test |
| Security | — not re-reviewed (the classification split does not change any decision: `next` is still called unconditionally on every branch) |
| Documentation | ✅ — the retag, the anchor, the volume note and a five-row reason gloss |
| Architecture | — not re-reviewed (unchanged since `e37adf9`) |
| PR Scope & Description | — not re-reviewed (unchanged since `e37adf9`) |

Sourced: parent (follow-up).

amitmorade:
Thanks for the thorough write-up — the `AuthAuditHook` inversion and classifying the rich `Verify` error *before* `sanitizeTokenError` both make sense. The client sentinels and the audit `reason` set really do look like different contracts; folding OIDC failures into `errVerificationFailed` first would lose the distinctions this story is trying to record.

One thing I am not quite following, posting after merge in case it is useful for Story 24 or a small follow-up.

The comment on `ClassifyAuthFailure` says audit cannot share this vocabulary because of the import cycle. `internal/observability/audit` already imports `internal/auth` for identity, though, and `AuthFailureReasons()` is already exported. From the outside it looks like the cycle only blocks **auth → audit**, not the other way — so I may be missing why `event.go` still needs its own `authFailureReasons` map.

As it stands, the five `reason` values seem to live in two places:

- constructor set in `audit.authFailureReasons`
- producer set in `auth.authFailureReasonVocabulary`

…with a drift test holding them together, and `AuthAuditHook.Failure(reason string)` taking any string until `NewEvent` rejects it. Story 24 is then asked to call `AuthFailureReasons()` so it does not grow a third copy.

If we own both packages, is a CI equality check the seam we want here? It reads a bit like a compensating control for two sources of truth. Pinning go-oidc’s error text feels different — that library is not ours — but these two slices are.

Would it work to have one owner? For example a named `AuthFailureReason` type that the classifier, the hook, and `NewEvent` all use, or the smaller step of having `event.go` consume `AuthFailureReasons()` and dropping the duplicate map. Totally possible I am missing a constraint that makes the copy necessary — happy to be walked through it.

solace-aross:
Agreed, the comment's reasoning is wrong. `internal/observability/audit` already imports `internal/auth` (`event.go`, `auth_hook.go`), so the cycle only blocks `auth` importing `audit`, not the direction you're asking about.

There's still a reason the literal exists, just not the one the comment gives. Everywhere else in this package, audit owns the closed vocabulary and producers get checked against it: `error_type` lives in audit and `internal/tools` has a drift test against `audit.ErrorTypes()`, same pattern for the authz reasons. `auth_failure` is the one place that's backwards, a producer package exports and audit copies, only because `auth` can't import `audit` back.

So I don't think the fix is just having `event.go` call `auth.AuthFailureReasons()` in place. That keeps the inversion and leaves Story 24 pointed at `internal/auth` for the same five values, which is the wrong direction for an observability package to depend on.

Proposal: move the vocabulary itself into `internal/observability/schema`. It's a leaf package, audit and metrics both already import it, and auth can import it too with no cycle. A named `AuthFailureReason` type and the five consts live there. `ClassifyAuthFailure` returns that type. `event.go` builds its set from it instead of the hand copy. The drift test goes away since there's nothing left to drift. `AuthAuditHook.Failure` keeps a plain `string` argument, so Story 24's hook shape doesn't move at all.

Timing-wise, no Story 24 branch exists yet, so this is cheap now and gets more expensive once metrics starts importing from `auth` directly.

Take a look and let me know what you think. If you're happy with the shape, I'll file a small cleanup ticket and get it built as a fast follow-up rather than reopening this PR.

amitmorade:
That makes sense — thanks for walking through it.

Agreed that having `event.go` call `auth.AuthFailureReasons()` would point the wrong way. I had treated “audit already imports auth” as enough to share the table; you are right that observability owning the closed set is the pattern everywhere else, and putting the five values in `internal/auth` would make metrics depend on auth for a label set. `schema` as a leaf that audit, metrics, and auth can all import is the cleaner owner.

Follow-up ticket rather than reopening this PR sounds right to me.

PR #390, opened 2026-09-08
SOL-153965: make shutdown-hook registration observable and add a wiring test

solace-lootii:
## Summary

SOL-153884 wired the metrics meter-provider flush into the shared shutdown-hook registry (SOL-152449). Its own unit test proved the provider's `Shutdown` satisfies the hook contract, but not that `main()` actually registers it — `hooks.Registry` held its hooks in an unexported slice with no accessor, so a registration couldn't be asserted. A reviewer confirmed the gap by deleting the `Register` call in `main.go`: the existing test still passed.

This PR closes that gap:

- `hooks.Registry` gains `Names()`, returning every registered hook's name in registration order (nil-safe, matching `RunAll`).
- `cmd/server/main.go`'s two `Register` call sites — metrics meter provider and tracer provider, each previously sitting next to its own provider's construction — are consolidated into one `registerShutdownHooks(reg, metricsProvider, tracerProvider)` call. This is the seam a test can invoke directly, rather than exercising the rest of `main()`'s startup (network binds, OTLP exporter construction, route wiring) — exactly the limitation `resource_wiring_test.go`'s own doc comment names.
- New `cmd/server/shutdown_hooks_wiring_test.go` builds both providers for real (same pattern as `resource_wiring_test.go`) and asserts both flushes land in the registry under the names `main()` has always used, plus the nil-provider and metrics-only cases.

**No behavior change**: registration order and hook names are unchanged.

## Note on ticket scope vs. what's implemented

The ticket's trigger condition was "once the tracing and OTLP-metrics provider flushes land" (SOL-152420 and SOL-152418). SOL-152420 (tracing) is already closed; SOL-152418 (Story 46, OTLP metrics egress) is still open. This PR covers the two registrations that exist today (metrics, tracer) — extending `registerShutdownHooks` with a third `Register` call for Story 46's flush, once it lands, is then a one-line, low-risk addition to the seam this PR creates, rather than a second wiring-test rewrite.

The ticket also sketched a `registerShutdownHooks(cfg, readiness, reg)` signature that would rebuild both providers from config. The actual code doesn't support that shape usefully — the providers are already built earlier in `main()` for other reasons (starting the metrics listener, deriving the tracing meter provider) — so this PR takes the narrower, lower-risk seam: register the already-built providers' shutdown hooks, which is the part SOL-153965 is actually about.

## Testing

- `go build ./...`, `go vet ./...` clean
- `golangci-lint run` clean on touched packages
- New tests: `cmd/server/shutdown_hooks_wiring_test.go`, additions to `internal/observability/hooks/hooks_test.go`
- Full suite: `go test ./...` — 2852 passed

Epic: SOL-150253. Milestone/FD/KA: SOL-150251.


bczoma (approved):
## PR Review: #390 — SOL-153965: make shutdown-hook registration observable and add a wiring test

**Author:** solace-lootii | **Branch:** `lkiri/sol-153965-shutdown-hook-wiring` → `main` | **Changes:** +217 / -6 across 5 files
**Jira:** `SOL-153965` — Unit-test the shutdown-hook wiring once tracing and audit flushes land (Verifying)

---

### Verdict

The code is right and the seam is the correct shape. Five lenses ran; three findings came back at 🟡 and all three were reduced by measurement — the details ride with the notes they became, because two of them were reduced by mutations worth knowing about.

Conditioned on one item, which is not in the code:

1. 🟡 `SOL-153965` / `SOL-152418` — the ticket text still specifies a seam signature and a three-flush assertion that the delivered code does not meet, and cannot. See the "Not anchorable" note below; it is a Jira edit, not a code change.

- 🟡 **The ticket's acceptance criteria now specify something unbuildable, and nothing tracks the correction.** SOL-153965 still asks for `registerShutdownHooks(cfg, readiness, reg)` and "a unit test asserting all three provider flushes are registered". The PR body explains why the signature deviates — correctly; that shape would have to rebuild providers whose constructors mutate process globals (`otel.SetTracerProvider`, `registry.MustRegister`), so it would clobber the tracer provider or panic on a duplicate collector. But the deviation lives only in the PR body, and the third flush is not merely "not landed yet": it will never exist. PR #391 (SOL-152418, open now) adds the OTLP reader to the *same* meter provider, and `internal/observability/metrics/provider.go:278-296` on that branch says so in as many words — one `Shutdown` shuts every reader, so a second hook would be redundant. Two consequences: SOL-153965 is in **Verifying** and would close against criteria the merged code does not satisfy, and whoever verifies SOL-152418 against its own AC ("the OTLP reader's flush is registered as a shutdown hook against the Story 48 registry") will look for a registration that is deliberately absent. Comment on both tickets and amend SOL-153965's Task section to the delivered two-hook shape.

The 8 notes below are **not** conditions of this verdict.

Verified at this head: `make check` — build, vet, lint, race tests — exit 0, **89.8%** total coverage against the 85% floor. That is the repo's own gate rather than the narrower `go build` / `go vet` / `go test ./...` the PR body reports, so the gate question is settled rather than taken on trust.

---

### Summary

A small, honest change: `Names()` on the hook registry, two scattered `Register` calls consolidated into one `registerShutdownHooks` seam, and a test over the seam covering both providers, the nil cases and the metrics-only case. The "no behavior change" claim holds under inspection rather than merely being asserted — `tracing.New` returns a nil provider on every error and disabled path, so rewriting `else if tracerProvider != nil` as an unconditional nil check is exactly equivalent, and moving the registration later in startup opens no signal window because `RunAll` is only reachable past `main()`'s `select`.

Three findings were filed at 🟡 and all three came down under challenge. Two are worth stating because the measurement went further than the finding did. One lens reported that the test still does not fail when the registration is deleted; the mutation that actually matters — deleting a `Register` call inside the seam, which is the reproduction the CHANGELOG names — turns two tests **red**, so the PR's central claim stands and only the one-line call site in `main()` remains unasserted, which `docs/internal/unit-test-coverage.md:47-51` explicitly assigns to the E2E suites. In the other direction, a challenge meant to shrink a finding found the hole slightly wider than filed: registering `metricsProvider.Shutdown` under **both** names, dropping the tracer flush outright, keeps all three tests green and the compiler silent, because both method values are `func(context.Context) error`. That is note 1 below, and it is the one note I would actually spend ten minutes on.

---

### Minor notes

🟢 **Minor** — worth fixing, not a condition of the verdict. 💡 is folded in.

- 🟢 `cmd/server/shutdown_hooks_wiring_test.go:80` (and `:117`) — `reflect.DeepEqual(reg.Names(), want)` pins labels only. Verified by mutation: registering `metricsProvider.Shutdown` under both names — dropping the tracer flush, and with it the batch processor's trailing spans at every exit — keeps all three tests green. A pure *swap* of the two handlers is harmless in effect (`RunAll` uses the name only for a log field), so the failure mode worth guarding is a duplicated or dropped handler, not a crossed one. Cheapest sufficient fix: `RunAll` the registry after `registerShutdownHooks` and assert each provider is actually shut; at minimum stop the doc comments claiming "both flushes land" when the assertion covers name and order.
- 🟢 `cmd/server/shutdown_hooks_wiring_test.go:14-27` — the rationale block has no blank line before `package main`, so it becomes the package doc comment for `package main` in the test binary, opening "registerShutdownHooks actually registers…" rather than "Package main …". No sibling file in `cmd/server` does this. Add a blank line, or move the text onto the first test.
- 🟢 `cmd/server/shutdown_hooks_wiring_test.go:56-73`, `:110-117` vs `cmd/server/resource_wiring_test.go:50-70` — the `metrics.New` + 2s-cleanup block is now copy-pasted three times in one package (and the `tracing.New` block twice), so the shutdown budget has to be kept in step in three places. Same package, so one `newTestProviders(t)` helper would do.
- 🟢 `internal/observability/hooks/hooks.go:54-58` — `Names()`' doc covers nil-safety and ordering but omits the concurrency caveat `Register` states three lines up; the registry has no mutex. One sentence: "Like `Register`, not safe for concurrent use with `Register` or `RunAll`."
- 💡 `CHANGELOG.md` (this entry) and `cmd/server/shutdown_hooks_wiring_test.go:15` — "closing a gap" and "actually registers what main() thinks it registers" read absolute, while `main()`'s single call to the seam is unasserted. The bodies of both comments, and `registerShutdownHooks`' own doc comment, are already correctly hedged — this is one clause in each summary line.
- 💡 PR title, and `internal/observability/hooks/hooks.go:54` — "observable" is true of a test, not of an operator; the CHANGELOG says as much ("Internal Go API only"). Also, the accessor's doc opens with a disclaimer ("Production code has no need for this") where the house style one package over (`internal/observability/correlation/middleware.go:52-55`) states the positive contract first. Worth noting that Go left no alternative here — `export_test.go` is package-local, so a `cmd/server` test cannot reach `hooks`' unexported slice — and this repo already ships the same pattern in three places, so the accessor itself is fine.
- 💡 `cmd/server/shutdown_hooks_wiring_test.go:82`, `:122`, `internal/observability/hooks/hooks_test.go:162` — `slices.Equal` over `reflect.DeepEqual`: type-safe, and it does not treat `nil` and `[]string{}` as unequal.
- 💡 `docs/internal/unit-test-coverage.md:47-49` — pins `cmd/server` at 37.2% "rather than unit tests"; measured 54.0% at this head (52.9% without this PR's test file), so the figure was already stale before this change. Pre-existing, optional to refresh here.

…and 3 further minor items (`resourceForTest`'s split error contract, the 20-line doc comment on a 6-line function, the ~110-word CHANGELOG paragraph) — ask if you want them.

---

### Summary Verdict

|---|---|---|
| Correctness | ✅ — no-behavior-change claim verified against every `tracing.New` return path; `make check` green | good:code |
| Code Quality | ✅ — 6-line seam, flat guards, hook names string-identical to before | good:code |
| Test Coverage | ⚠️ — the registration logic is now covered, but a duplicated/dropped handler passes (see note 1) | good:code |
| Architecture | ✅ — no new import edge, one-way layering, near-total reversibility; the deviation from the ticket's seam shape is the right call | good:architecture |
| Security | ✅ — `Names()` copies rather than aliases, hook names are compile-time literals, no new log line, no teardown flush lost | good:security |
| Documentation | ✅ — correctly classified internal-only; CHANGELOG accurate and in the right bucket; stale in-code cross-references refreshed | docs |
| PR Scope & Description | ⚠️ — description is unusually good, but the ticket was never brought in line with what shipped | good:pr |

solace-aross (approved):
bczoma's review already covers this thoroughly with real mutation evidence; independently reran their two load-bearing claims from scratch rather than trusting the write-up.

Deleted the `metrics_provider` `Register` call in `registerShutdownHooks` — two of the three wiring tests went red, confirming the central claim (and the CHANGELOG's own reproduction). Then reproduced bczoma's actual finding: registered `tracer_provider`'s name against `metricsProvider.Shutdown` instead of `tracerProvider.Shutdown` (keeping the nil-guard structure intact) — all three tests stayed green, silently dropping the tracer's flush (and with it, the batch processor's trailing spans on every exit) while asserting nothing wrong. Confirms bczoma's note 1 exactly: the tests pin names and order, not that each provider's own handler actually landed.

Agree with the verdict: approve with minor notes, none blocking. Nothing to add beyond bczoma's list.

PR #392, opened 2026-09-08
SOL-154163: give the auth_failure reason vocabulary one owner in schema

solace-aross:
Cleanup follow-up to #388 (SOL-152097). Not new capability, no behavior change.

Jira: SOL-154163
Agreed design: the [comment thread on #388](https://github.com/SolaceProducts/solace-broker-mcp/pull/388#issuecomment-5590575482) — @amitmorade's question, the proposal, and his agreement. This PR implements that shape.

## Before / after ownership

| | Before | After |
|---|---|---|
| The five reason strings | **Two copies.** `internal/auth/classify.go`'s `authFailureReasonVocabulary` (the real source, exported as `auth.AuthFailureReasons()`) **and** `internal/observability/audit/event.go`'s hand-written `authFailureReasons` map literal | **One declaration**, in `internal/observability/schema`, as the named `AuthFailureReason` type and its five consts |
| Kept in sync by | A CI drift test (`TestAuthFailureReasons_MatchesEventSchemaVocabulary`) | Nothing to sync — `event.go` derives its set from the schema value |
| Direction of dependency | Inverted: a producer package (`internal/auth`) exported the vocabulary and observability copied it | Right way up: observability owns the closed set, `internal/auth` reads it like every other producer |
| `ClassifyAuthFailure` returns | `string` | `schema.AuthFailureReason` |
| `AuthAuditHook.Failure` | `(ctx, reason, sub, clientID string)` | **unchanged, byte for byte** |

## Why `schema` and not just `event.go` calling `auth.AuthFailureReasons()`

The old comment in `classify.go` blamed an import cycle. That was only true in the direction nobody was asking about: `internal/observability/audit` already imports `internal/auth` for identity, so the cycle blocks `auth` importing `audit`, not the reverse. It never blocked importing `internal/observability/schema`, which imports nothing of ours.

But the smaller fix (have `event.go` call `auth.AuthFailureReasons()`) would have removed the duplication and kept the inversion. Everywhere else here observability owns the closed vocabulary and producers are checked against it — `error_type`, the two authz-denied reason sets. `auth_failure` was the one place backwards. Leaving it that way would also point Story 24's `mcp_auth_failure_total{reason}` counter at `internal/auth` for a metric label set, which would drag `oidc`, `config`, `authz` and `idpclient` into `internal/observability/metrics` for five strings.

`schema` is a true leaf — its only import is `slices` — so the producer and both renderers can all read one declaration.

## What changed

- **`internal/observability/schema/auth_failure_reason.go`** (new): the type, the five consts, and `AuthFailureReasons()` (sorted, fresh slice per call). The package doc now states the rule for what belongs there, so `error_type` and the authz reason sets correctly stay with `audit`.
- **`internal/auth/classify.go`**: returns the named type. The classifier itself stays here — it is the only package holding the rich pre-sanitization verifier error it needs. `authFailureReasonVocabulary` and `AuthFailureReasons()` deleted rather than kept as wrappers; the drift test was their only caller, and a wrapper would preserve exactly the inversion this ticket removes. The one-direction cycle claim is rewritten to name which package imports which.
- **`internal/auth/audit_emit.go`**: `reportAuthFailure` takes the named type and is the single place that widens it to the plain string `Failure` accepts, so all five call sites stay typed. The interface doc now says why a plain string is deliberate: an implementor records the value it is handed, it does not re-derive it — and Story 24 (SOL-152099) will implement the same interface.
- **`internal/auth/middleware.go`**: the last hand-written `"invalid_token"` literal is now the const.
- **`internal/observability/audit/event.go`**: `authFailureReasons` derived from the schema value.
- **`CHANGELOG.md`**: the unreleased SOL-152097 entry described the old ownership in detail, naming the deleted function and the deleted drift test. Corrected in place, since it has not shipped. No new entry — this touches none of `internal/config/`, `tools.yaml`, or `internal/tools/`, and there is no behavior change.

## Tests

Net +1 test function (119 → 120 across the three packages). One deleted, two added:

- **Deleted** `TestAuthFailureReasons_MatchesEventSchemaVocabulary` — genuinely tautological now: `event.go` has one derivation and nothing on the other side to compare against.
- **Kept, and strengthened**: `TestClassifyAuthFailure_ReturnsOnlyClosedVocabulary` now reads the set from `schema` instead of restating it. The named return type does not make it vacuous — `AuthFailureReason` is a string type, so a case could return `AuthFailureReason("weird")` and compile; this test is what forbids that.
- **Kept, untouched**: `TestNewEvent_reasonVocabulariesDoNotBleed` (all nine cross-pairings) and `TestNewAuthHook_ConstructorRejection_EmitsDrop`. The latter is now load-bearing proof that `Failure`'s string parameter is deliberate: it calls `hook.Failure(ctx, "not-a-real-reason", ...)`, which is only writable because the parameter is a plain string.
- **New**, in `schema`: the five wire strings (the form that leaves the process, where a rename breaks a customer's saved query) and the accessor's sorted/fresh-slice contract. Nothing asserted either before.

`internal/auth/auth_audit_test.go` is untouched, and its `fakeAuthHook.Failure(_ context.Context, reason, sub, clientID string)` still compiles — compiler-level proof the interface did not move.

## Verification

- `make check` green: build, vet, lint, race tests, coverage **89.9%** against the 85% floor.
- Four injected defects each caught by the new/changed tests: renaming a const's wire value (fails in all three packages), dropping a value from the closed set, the classifier returning an off-vocabulary value, and the accessor sharing its backing array.
- `TestApplicabilityByType_isVersioned` still passes, so the audit record shape is unchanged and no `audit_schema_version` bump is due.
- No behavior change: wire strings, validation semantics, and the constructor's rejection message (already sorted via `sortedKeys`) are identical.
- `docs/observability.md`'s five-row gloss table and the `:1167` note still list exactly the same five strings — nothing here falsifies a doc claim, so no doc change.

Reviewed by a skeptical Opus pass before opening, which is what caught the stale CHANGELOG entry.


bczoma (approved):
## PR Review: #392 — SOL-154163: give the auth_failure reason vocabulary one owner in schema

**Author:** solace-aross | **Branch:** `sol-154163-auth-failure-reason-schema` → `main` | **Changes:** +261 / -104 across 10 files
**Jira:** `SOL-154163` — Give the auth_failure reason vocabulary one owner in internal/observability/schema (Verifying)

---

### Verdict

**Recommendation:** ✅ Approve

Nothing here is a condition of merging. The eight 🟢 notes below are optional.

### Summary

This lands the ticket's design exactly: the five `auth_failure` reason strings now have one declaration, in a leaf package both the producer and both renderers can import, and `event.go` *derives* its validation set rather than restating it — so the deleted drift test is replaced by a structural impossibility, not by an unguarded gap. I verified the claims rather than taking them: the leaf claim holds (`schema` imports only `slices`), the five wire strings are byte-identical, `AuthAuditHook.Failure`'s signature is unchanged, all six `ClassifyAuthFailure` branches still return a vocabulary member, the constructor's rejection path and message are untouched, and no reference to the deleted `auth.AuthFailureReasons()` survives anywhere in the tree. The new `TestAuthFailureReasons_WireValues` is a compliance guard the repo did not have before — it pins the customer-visible strings independently of the Go identifier names.

One finding was filed at 🟡 by two lenses independently and reduced to 🟢 by measurement: the new placement rule in the package doc really is wrong on its examples, but it has no CI or runtime consequence, and the fix the reviewers proposed turned out to be unsound in its own right — see the first note below.

### Minor notes

- 🟢 `internal/observability/schema/version.go:22-28` — the new placement rule is wrong on **both** its examples, and this is the sentence the PR exists to write down. It cites `error_type` as a vocabulary "only one signal renders", owned by `audit`; in fact it is declared twice (`metrics/instruments.go:73` as `allErrorTypes`, `audit/event.go:114` as `errorTypeVocabulary` — each comment claiming to be the single source of truth), rendered by three signals (metric label, `tools.CallTool` span attribute, audit field), and primarily owned by `metrics`. Separately the rule's discriminator is unstated: `mcp_auth_failure_total` is described in the present tense as *sharing* `AuthFailureReason` but does not exist yet (SOL-152099 is still To Do), and the authz-denied set has exactly the same documented-but-unbuilt second renderer (`mcp_authz_denied_total`, `docs/observability.md:280`) yet is cited as staying put — so the two examples are indistinguishable on the criterion the rule states. Suggest: drop `error_type` from the parenthetical entirely, state the criterion in terms of *documented* renderers rather than shipped ones, and either move the authz-denied sets too or say why they are deferred. Mark the counter as forthcoming.
- 🟢 `docs/internal/architecture.md:506` — the package-responsibility row still describes `observability/schema` as "Metrics/audit output schema-version constants" against a single file. This PR widens the charter and adds a second file; a contributor checking the map before adding a vocabulary sees a version-constants-only package.
- 🟢 `internal/observability/schema/auth_failure_reason.go:31` — `AuthFailureReason` is a defined string type, so an untyped constant still converts implicitly: `reportAuthFailure(ctx, hook, "probe_flood", "", "")` compiles today. The named type narrows intent rather than the type system. Containment is real but runtime/test-time (the retained closed-vocabulary test, and `NewEvent`'s check turning an off-vocabulary value into a rejected record plus `audit_drop`) — worth a line only if a non-classifier producer ever appears.
- 🟢 `internal/observability/schema/auth_failure_reason.go:64-78` — the package var is declared in doc-gloss order and the accessor sorts a copy on every call, so two orderings exist but only the sorted one is observable, and the sort re-runs for a value that can never change. Declaring it sorted and dropping `slices.Sort` (keeping the copy) satisfies the same documented contract and the same test.
- 🟢 `internal/auth/audit_emit.go:60-72` — eight lines of design rationale, including a forward reference to SOL-152099, sit on the interface doc above the contract itself. One sentence of contract on the interface, with the ADR-shaped argument left in the CHANGELOG entry that already carries it, would age better.
- 🟢 `internal/observability/audit/event.go:169-181` — a nine-line comment pre-empting a "why not generic over `~string`" refactor nobody proposed, on a helper with one call site. The helper is right; the comment is heavier than what it guards.
- 🟢 `internal/auth/classify_test.go:70-90` — `TestClassifyAuthFailure_ReturnsOnlyClosedVocabulary` re-drives the error list `TestClassifyAuthFailure` already covers, now that the latter pins the exact const per branch. Its unique value is guarding a future off-vocabulary case, but a future case arrives with a new error value this test's hardcoded list would not exercise. Driving both from one shared table would close that.
- 🟢 PR-level — @amitmorade raised the question this ticket answers and agreed the shape on #388, but only the `dax-developers` team is requested; the one person holding the design context is not guaranteed to see the implementation.

and 3 further minor items (PR-description length, a review-order steer, the `internal/auth` → `observability/` import-path optics) — ask if you want them.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ — two inaccurate statements, both 🟢, neither holding the approval | docs |

amitmorade (approved):
This addresses the vocabulary-ownership issue from #388.

One owner in `schema`, `ClassifyAuthFailure` returns the named type, `event.go` derives its set from that instead of a second copy, and the drift test is gone. Observability owning the closed set, with `auth` producing into it, is the right direction.

LGTM.

PR #393, opened 2026-09-08
fix: SOL-154164 stop a newer broker turning 9 successful write tools into reported failures

solace-aross:
Fixes [SOL-154164](https://sol-jira.atlassian.net/browse/SOL-154164) — a newer broker turns 9 successful write tools into reported failures.

## Root cause

The strict output schema added in SOL-152947 derives each write tool's response shape from the SEMPv2 spec embedded at build time (10.26.5), and applied `additionalProperties: false` at every level — including the two levels the broker itself supplies: the SEMP `{data, meta, links}` envelope, and the resource inside it.

`ToolManager.CallTool` validates a tool's output *after* the handler has already applied the mutation. So on a broker newer than 10.26.5, an attribute that release added to an affected object came back legitimately, validation failed with `Additional property <x> is not allowed`, and the agent was told the create/update had failed (`isError: true`, `retryable: false`) although the broker had applied it. That is the one outcome a destructive tool must never produce, and the hazard SOL-152980 exists to prevent.

Nine tools were affected: `create-message-vpn`, `update-message-vpn`, `create-queue`, `update-queue`, `create-queue-subscription`, `create-topic-endpoint`, `update-topic-endpoint`, `create-rdp`, `update-rdp`. The small resources were the most exposed — `create-queue-subscription` resolves only 3 response fields, `create/update-rdp` only 6, so a single added attribute broke those tools outright.

The request path already tolerates the same spec/broker mismatch (`constructRequestBody` in `executor.go`, with a versioned-server-aware message). The response path had no equivalent.

## The change

Two lines of production code: drop `additionalProperties: false` from `fieldPropertiesSchema` and `envelopeSchema` in `internal/composite/schema.go`. Everything else in the diff is tests, comments, and the changelog.

The rule the fix draws, and documents in the code: **strict where the shape is ours, tolerant of additions where the payload is the broker's.**

| Level | Assembled by | Strict? |
|---|---|---|
| Step-keyed top level | `collectSteps`, from this tool's own step IDs | yes |
| Paginated wrapper (`data`/`truncated`) | `fetchPaginated` | yes |
| Fan-out wrapper (`byKey`/`skipped`) | `fetchFanOut` | yes |
| SEMP response envelope (`data`/`meta`/`links`) | the broker | no |
| The resource inside `data` | the broker | no |

## What still gets caught

- Every field the spec declares is still type-checked (a retyped field fails).
- The resource's own identifier field(s) are still required — a response that doesn't name what it just created fails.
- A response missing its `data` payload, or a step result that isn't an object at all, fails.
- An unexpected key on any level this server assembles itself fails.

One detection is deliberately given up: a **renamed** non-identifier response field, previously noticed only because its new name arrived as an unknown key. Stated plainly in the CHANGELOG rather than papered over. From a single response an added attribute and a renamed one are indistinguishable, and output validation is the wrong instrument for the difference — it charges the cost to a caller whose mutation already succeeded instead of alerting a maintainer. A CI-time spec diff is the right home for that; worth a follow-up ticket.

## Verification

TDD: tests written first and confirmed red against the exact reported symptom, then the fix.

The regression test that reproduces the ticket end to end, through the real `CallTool` path (compiled-schema validation, real generated `create-queue` schema, destructive annotation) — its failure message on `origin/main` is the ticket verbatim:

- `TestCallTool_NewerBrokerAttributeIsNotReportedAsFailure` (`internal/tools/write_tool_drift_test.go`)

```
a successful create was reported to the agent as a failure: map[error:tool "create-queue"
output validation: output validation failed: createQueue.data: Additional property
aBrandNewAttribute is not allowed retryable:false]
```

Schema-level, against the real embedded spec and synthetic fixtures:

- `TestWriteToolDrift_NewerBrokerAttributeAccepted`, `TestWriteToolDrift_NewerBrokerEnvelopeKeyAccepted` — the two broker-supplied levels
- `TestWriteToolDrift_NonObjectStepRejected`, `TestWriteToolDrift_UnknownStepKeyRejected` — new floor under the added tolerance
- `TestWriteToolDrift_MissingIdentifierRejected`, `TestWriteToolDrift_TypeChangedFieldRejected`, `TestWriteToolDrift_MissingEnvelopeDataRejected` — unchanged, still pass
- `TestBuildStrictOutputSchema_FlatStep`, `_PaginatedStep`, `_FanOutStep`, `_RealCreateQueue` — extended to pin both sides of the boundary, including a new `retypedRealField` case proving real-spec typing still bites

Two existing guards were rewritten rather than relaxed, because their old assertions keyed on the presence of `additionalProperties` and would have become permanent no-ops:

- `loader_embedded_test.go` write-tool-schema-resolves now asserts the spec-derived `data` item actually carries properties, across all 9 tools
- `composite_handler_test.go` `DeleteToolStaysUsable` now asserts the permissive fallback is exactly `{"type": "object"}`, which is a **stronger** check than before — the old one would have passed for the spec-derived envelope

`make check` (build, vet, lint, race-enabled tests, coverage gate): green, exit 0, **89.8% coverage** against the 85% floor.

No new log statements, so no `docs/internal/secure-logging-rules.md` exposure; diff scanned for CRITICAL/HIGH, none found.

## Independent review

Reviewed cold by a separate Opus agent with no context on the work, given only the diff and the ticket. Verdict: sound, approve. It independently confirmed all 9 tools are fixed (18 of 18 cases fail on `origin/main`, pass here), checked each level of the strictness boundary against `runSingle`/`fetchFanOut`/`fetchPaginated`/`collectSteps` and found none misclassified, and ran mutation testing — reverting `schema.go` fails 8 tests, and separately dropping the `required` clause fails 4, so the identifier guard is multiply pinned.

Both of its "worth fixing" findings are addressed in this PR:

1. A stale `t.Errorf` message still explained a failure in terms of `additionalProperties: false` — the one text a future engineer reads when that test fires. Rewritten.
2. The CHANGELOG claimed the caught drift was "unaffected". It isn't quite; the rename case is gone. Now stated explicitly, including that 0.8.0's "removes" claim never held for non-identifier fields.

Its one nit was skipped, as pre-existing and made marginally less wrong by this PR: `docs/tools-reference.md:65` says five tools depart from the generic envelope, when 18 config/action tools now get a generated schema. That drift arrived with SOL-152947 and belongs in a docs pass, not here.

## Not verified here

`test/e2e-management` was not run locally (needs Dockerized brokers; CI covers it). The reviewer checked the suite's assertions and confirmed it only asserts `isError=false` on valid calls, so nothing there depends on the removed strictness and no e2e update is needed.


bczoma (approved):
## PR Review: #393 — fix: SOL-154164 stop a newer broker turning 9 successful write tools into reported failures

**Author:** solace-aross | **Branch:** `sol-154164-write-tool-output-validation` → `main` | **Changes:** +267 / -38 across 6 files
**Jira:** `SOL-154164` — A newer broker turns 9 successful write tools into reported failures (Verifying)

---

### Verdict

**Recommendation:** ✅ Approve

Nothing is a condition of this approval. The 8 🟢 notes below are worth a look but hold nothing.

---

### Summary

Two production lines that remove a real hazard: a successful destructive mutation being reported to the agent as `isError: true, retryable: false` because the broker echoed an attribute the build-time-embedded 10.26.5 spec doesn't declare. The strictness boundary the change draws — strict on the levels this server assembles, tolerant of additions on the levels the broker supplies — was checked level by level against `collectSteps`, `runSingle`, `fetchPaginated` and `fetchFanOut`, and none is misclassified in either direction. The tests were mutation-verified rather than taken on trust: restoring the two deleted lines fails nine tests including the end-to-end `CallTool` reproduction, and both rewritten guards were separately confirmed still able to fail rather than having become no-ops.

Two findings were filed at 🟡 and both were reduced by measurement. The claim that a post-mutation `isError` can still occur is true but now requires SEMP to break its own compatibility contract — the generated schema carries only `string`/`boolean`/`integer` with no formats or bounds, and the `required` identifiers are the resource's own URI keys, so neither residual trigger has precedent in the 2203-attribute spec. The claim that the new comment misdescribes the request path is correct, but self-corrects the moment a reader follows its own pointer.

---

### Minor notes

- 🟢 `internal/composite/schema.go:189-191` (and the `[Unreleased]`/Fixed CHANGELOG entry, and PR body line 11) — all three say the request side "already tolerates" this mismatch. `constructRequestBody` (`executor.go:840-850`) **rejects** an undeclared body field; it only phrases the rejection in a version-aware way, and `BodyFields` is non-nil for all nine operations so that branch is the live path. The ticket's own wording ("gives a versioned-server-aware *error message*") was accurate and got upgraded in transit. The CHANGELOG line is the one that matters — it's user-facing and has no `executor.go:829` next to it to self-correct. Better premise, same conclusion: the request side can reject because nothing has been mutated yet; the response side can't, because the mutation already landed.
- 🟢 `internal/tools/write_tool_drift_test.go:29-35` — `writeToolDriftTestSetup`'s doc comment is the one comment in the diff that wasn't updated. It still says the file feeds "an unexpected field … and asserts each is rejected … proving that this schema isn't accidentally permissive", which is now the opposite of what the file asserts, at the top of the file that asserts it.
- 🟢 `internal/tools/composite_handler_test.go:283` — `TestCompositeToolHandler_OutputSchema_WriteToolIsStrict` now opens by asserting the envelope is *not* strict. The block comment says it right; the name reads as self-contradictory in a failure report. `..._WriteToolStrictnessBoundary` or similar.
- 🟢 Follow-up worth filing now, not after merge: the CI-time spec diff the body calls "worth a follow-up ticket". Nothing is filed or linked, and it's the mechanism that would cover both the renamed-field case given up here and the removed-field case the CHANGELOG now admits 0.8.0 never really caught — and the precondition for ever reintroducing strictness safely.
- 🟢 `internal/tools/manager.go:384-392` — output validation can still return `isError: true, retryable: false` post-mutation, via a retyped spec-declared field or a dropped identifier. Measured as requiring a SEMP compatibility break rather than a routine release, so not a defect in this change. Worth carrying on the same follow-up: for a destructive tool, an output-shape mismatch arguably belongs as a warning annotation plus an `ErrorTypeOutputValidationError` metric rather than `isError` — though whether an agent should be told the shape was unrecognised is a real design question, not an obvious win.
- 🟢 `internal/composite/schema.go` (`fieldPropertiesSchema` comment) — the ticket's second option, gating strictness on the broker's reported version, is silently dropped. It's the right rejection and for a structural reason: output schemas compile once per tool at `Register` (`manager.go:154`) and are shared across brokers whose versions differ and none of which is known then. Two sentences recording that stops it being relitigated.
- 🟢 `internal/composite/schema.go:174-216` — tolerated drift now leaves no trace: `CallTool` only reacts to a validation *failure*, so an unknown key at the envelope/resource level produces no log, metric or audit field. A DEBUG line or a counter labelled by tool would mean a spec bump gets prompted by data rather than by a bug report.
- 🟢 `internal/tools/write_tool_drift_test.go:213-256` — behavioural coverage of the fix is `create-queue` only; the other eight are covered at schema-shape level. Defensible, since all nine reach the same `BuildStrictOutputSchema` path, but "nine tools fixed" is inferred from shared-code reachability rather than asserted. The setup helper is already parameterisable on tool name if table-driving it is cheap.

…and 7 further minor items — ask if you want them.

---

### Summary Verdict

|---|---|---|
| Documentation | ✅ | docs |

PR #394, opened 2026-09-08
SOL-154166: skip the list-vpns real-client probe on zero-connection VPNs

solace-aross:
Ticket: https://sol-jira.atlassian.net/browse/SOL-154166

## Root cause

`list-vpns` runs two steps. The `vpns` step lists every Message VPN and already selects `msgVpnConnections`. The `real-clients` fan-out then issues one `getMsgVpnClients` request per VPN to answer a single question: "is any non-reserved client connected?" Its `forEachIf` gated only on `enabled` and `state == "up"`, ignoring the connection count the first step had just fetched.

At `msgVpnConnections == 0` the probe's answer is already settled. Zero connections means no clients of any kind, so the probe returns no rows, and `postprocess/handlers/list_vpns.go` reads an absent `byKey` entry exactly as it reads an empty one: no real client, counted under `zeroConnectionCount`. The request was pure cost.

## Change

```yaml
forEachIf: '{{and .Item.enabled (eq .Item.state "up") (ne (index .Item "msgVpnConnections") 0.0)}}'
```

Two deliberate details:

- `index .Item "msgVpnConnections"` rather than `.Item.msgVpnConnections`. These templates run under `missingkey=error`, where the field form fails the entire call on a row that omits the key; `index` yields the zero value, so such a row falls back to probing. `and` short-circuits, so the term is never evaluated for disabled, down, or standby rows.
- `0.0`, not `0`. SEMP counts arrive as JSON numbers decoded into `map[string]any`, i.e. `float64`, and `text/template` refuses a `float64`-vs-`int` comparison.

A count of `1` is still probed on purpose: it cannot be told apart from one real client, which is what the probe exists to resolve.

## Honest scope of the win

The saving applies only to VPNs a broker reports at exactly 0 connections. On the brokers this project has observed, an idle enabled+up VPN reports 1, the reserved `#client` (recorded against the `test-vpn-empty` fixture in `test/e2e-monitoring`, and the reason the old `msgVpnConnections <= 1` heuristic was replaced by this fan-out in the first place). So on those brokers nothing changes and no request is saved. I have no live broker in this environment to produce a 0-connection observation; if you want that evidence before merging, say so and I will get it. Correctness is unaffected either way, and the guard costs one template term.

## Why zero connections is safe to trust

The skip depends on `msgVpnConnections` being the VPN's total client-connection count across every service rather than a subset. AMQP is the service worth checking, because it is the one client-facing service whose SEMP per-service counter sits outside the `msgVpnConnections*` family (it is `connectionsServiceAmqp`), so an AMQP-only VPN is where a real client could hide behind a `0`.

Evidence: the SEMP monitor spec describes `msgVpnConnections` as "The current number of Client connections to the Message VPN", with every per-service field naming its service explicitly. Solace Insights documents the same counter as `vpn.connections`, "the current number of connections to the Message VPN", alongside six per-service siblings: `connections_service_smf`, `_web`, `_mqtt`, `_rest_incoming`, `_rest_outgoing`, `_amqp`. AMQP sits inside that family as a breakdown of the total, not outside it.

One legacy doc reads the other way and is worth naming: the CLI VPN `connections thresholds` page describes the aggregate as "SMF, Web transport, and incoming REST connections". That enumeration also omits MQTT and REST-outgoing, both of which have their own `msgVpnConnections*` counters, so it looks like stale pre-MQTT/AMQP text about the event threshold rather than a statement about the counter.

## Verification

- `make check` green: build, vet, lint, race-enabled tests, total coverage 89.8% against the 85% floor.
- New `TestExecute_ListVPNs_SkipsRealClientProbeOnZeroConnectionVPNs` executes the shipped embedded definition against a stub broker that models the invariant in question (a VPN reporting 0 connections returns no clients). It asserts which VPNs are probed (zero-connection skipped; count 1, count 4 and count-absent all still probed) and requires the summary to equal what the pre-fix predicate produced over the same fixtures.
- Fails on `origin/main` with `probed VPNs = [busy count-absent idle one-real reserved-only], want [busy count-absent one-real reserved-only]`, and nothing else in the package fails, so this is the only guard on the behavior.
- The fixture set carries a pair at count 1, one with a real client and one without, so a predicate that widened the skip to `<= 1` fails the summary assertions rather than only the probe list. Confirmed by mutation: `zeroConnectionCount` goes to 4 against an expected 3.
- `msgVpnConnections` is now pinned in `TestLoadTools_ListVPNs_RetainsDiscoveryFields`, which owns the "select field no framework validation covers" concern. The loader performs no static cross-check of `forEachIf`-referenced fields against the parent step's `select`.
- E2E: `test/e2e-monitoring`'s recompute for `zeroConnectionCount` already treats a missing `byKey` entry as zero-connection, so it passes whether the fixture VPN reports 0 or 1. `e2e-basic-mcp` does not call `list-vpns`; `e2e-oauth` and `e2e-llm` only read `.vpns.data`.

## Second-order effects

- The raw `real-clients` block now carries one fewer `byKey` entry and a correspondingly larger `skipped` count per skipped VPN. Both already varied with the broker's disabled and standby VPNs, and `internal/composite/schema.go` models `skipped` as optional.
- The VPN list and the probes are separate SEMP calls, so a client that connects between the two is no longer seen for a VPN listed at zero connections. That VPN's verdict now reflects the list snapshot rather than the later probe.

Comments in `list_vpns.go` that asserted a missing `byKey` key never reaches `hasRealClient` were corrected, since that path is now reachable and `false` is the right answer there. The stale `<=1 predicate` line in `test/e2e-monitoring/README.md` was refreshed, since it is where the `msgVpnConnections == 1` observation lives.

CHANGELOG entry added under `[Unreleased]` / `Changed`. `Changed` rather than `Fixed`: no output was ever wrong, the tool just paid for a request whose answer it already had.

## Review gate

An independent Opus reviewer was given only the diff and the ticket. It verified the template behavior empirically, found no correctness break, and raised four things, all addressed above: the AMQP question (evidence now in the YAML comment and here), a test gap where an over-wide skip survived the summary assertions (fixture pair added), two CHANGELOG overclaims (qualified), and the stale e2e README line (updated). Its standing recommendation is that a human confirm a real 0-connection VPN exists before merging, since the measured saving on observed brokers is zero.


bczoma (dismissed):
## PR Review: #394 — SOL-154166: skip the list-vpns real-client probe on zero-connection VPNs

**Author:** solace-aross | **Branch:** `sol-154166-list-vpns-redundant-query` → `main` | **Changes:** +311 / -21 across 6 files
**Jira:** `SOL-154166` — list-vpns queries the broker for clients on VPNs we already know have none (Verifying)

---

### Verdict

**Recommendation:** ✅ Approve

Nothing is a condition of this approval. The 🟢 notes below are worth a look but hold nothing.

---

### Summary

One template term, correctly written, well tested, and honestly scoped. The `ne` / `index` / `0.0` triple was reproduced against `text/template` under `missingkey=error` across nine row shapes: an absent key and an explicit nil both probe, a negative count probes, `0` skips, and `and` short-circuits so the term is never reached for disabled/down rows. The divergence from the ticket's suggested `gt … 0.0` is a correctness improvement rather than a stylistic one — `gt` errors outright on the nil that `index` yields for an absent key, which would abort the whole call instead of failing open. The new test is not vacuous: reverting the predicate adds `idle` to the probe list and fails, and widening it to `<= 1` fails the *summary* assertion rather than only the probe list.

**The open question in your "Honest scope of the win" section now has an answer, and it agrees with you.** This review sampled two live lab brokers. On `lab-129-78`, all 98 enabled+up VPNs report `msgVpnConnections == 1`; the only two rows at `0` are `enabled=false, state=down` and were already excluded by the pre-existing terms. So the new term saved 0 of 98 probes there — your `==1` observation generalises beyond the Dockerized fixture. On `lab-128-16` (appliance-class) the single `default` VPN reports 121, all attributed to `msgVpnConnectionsServiceSmf`, so the reserved client is service-attributed rather than a floating extra.

Three findings were filed at 🟡 and all three were reduced by measurement. In particular, the AMQP premise your body defends at length holds up on structural evidence you did not cite: `MsgVpn.maxConnectionCount` ("the maximum number of client connections to the Message VPN") sits over six per-service sub-caps **including `serviceAmqpMaxConnectionCount`**, and the `eventConnectionCountThreshold` family mirrors it exactly. For AMQP to sit outside the VPN-wide total, the VPN-wide cap would have to not cap AMQP — which would make the limit unenforceable. The `connectionsServiceAmqp` naming is a wart, not a signal.

---

### Minor notes

- 🟢 **Before closing SOL-154166, amend it.** The ticket still reads "a few hundred avoidable probes cost on the order of 30 seconds of the broker's entire management-plane budget", and it is in Verifying. That premise is now contradicted on every broker anyone has sampled. The CHANGELOG and this PR body are both honest about the zero measured saving, but the ticket is the permanent record and is the artefact someone will read in a year to ask whether SOL-154166 delivered. Close it as a guard against a broker shape not yet seen, not as a realised performance win.
- 🟢 `internal/composite/definitions/tools.yaml:314` — the AMQP premise is still argued rather than observed, and `executor_vpn_probe_test.go` fences it with a fixture guard ("a broker cannot report that") rather than testing it. The structural evidence above makes it very likely right and the branch is unreachable on sampled brokers, so this is opportunistic: next time an AMQP client is to hand, connect it to an otherwise-idle VPN, confirm `getMsgVpns` reports `msgVpnConnections >= 1`, and record it in `test/e2e-monitoring/README.md` beside the `#client` observation. Note there is no AMQP-capable client in the usual local toolset, so this is a small task rather than a five-minute one.
- 🟢 `internal/composite/definitions/tools.yaml:306` — the bullet justifying `index` ends "`and` short-circuits, so the term is not even reached for disabled/down/standby rows." That is true, but it is the fact that would make the *field* form safe for those rows, so as written it reads as an argument against the choice it is defending. The real reason `index` is needed is the enabled+up row that omits the field. Suggested: "`and` short-circuits, so the field form would be safe for disabled/down/standby rows — but not for an enabled+up row that omits the field, which is why `index` is used."
- 🟢 `internal/composite/definitions/tools.yaml` — the comment justifies `index` and `0.0` but never `ne` over `gt`, which is the form the ticket asked for and the one a future reader will find more natural. Since `gt` on the nil from `index` aborts the entire `list-vpns` call, "simplifying" it back would silently convert the documented fail-open into a total outage on any broker omitting the field. One line naming that is cheap insurance.
- 🟢 The Copilot thread on the tool description appears unaddressed at the current head, and its anchor commit `97865f6` no longer exists in the branch after the force-push to `b7bcfff` — so it is orphaned on a dead commit and will not auto-resolve. Worth either amending the description or replying to say why the wording stands, rather than letting it vanish.
- 💡 `internal/composite/loader_embedded_test.go:294-334` — the predicate's dependency on `vpns.select` retaining `msgVpnConnections` is pinned by a hand-maintained list, and this is the first entry whose absence changes *behaviour* rather than output content. SOL-153764 already compiles every `forEachIf`/`args` template at `LoadTools`, so the parsed tree is sitting right there: walking it for `.Item.X` and `index .Item "X"` references and cross-checking against the parent step's `select` would retire the pin list and cover the next fan-out tool too. Follow-up, not this PR — the failure mode is benign, since the predicate falls back to probing.
- 💡 `internal/composite/definitions/tools.yaml:314` — the term depends on `client.go:279` decoding SEMP numbers as `float64`; an `int`, `int64` or `json.Number` makes `ne` error, and `fetchFanOut` treats a `forEachIf` error as a whole-step abort. The dependency is documented in prose in both the YAML comment and the test header, and it is unreachable today (one decode site, never changed in 15 commits, `UseNumber` proposed nowhere, and the failure would be instant and universal rather than silent). Cheapest guard if you want one: a single assertion in the SEMPv2 client test round-tripping a numeric field and asserting `float64`, with a comment naming the dependent — it fails at the layer that would actually change.
- 💡 `internal/composite/executor_vpn_probe_test.go:83-104` — `brokerStub.Execute` ignores the step's `where` args, so `realClients` is already post-`clientUsername != #*` filtering. Reasonable, but unstated, and the `reserved-only` fixture looks like it exercises the reserved-prefix filter when that filter is not under test. One line on the stub would prevent a later reader assuming coverage that is not there.

---

### Summary Verdict

|---|---|---|
| Architecture | ⚠️ — sound, but the skip rests on a premise the repo cannot verify and whose first real execution will be on an uncharacterised broker | good:architecture |
| Documentation | ✅ | docs |

internal/composite/definitions/tools.yaml
```diff
@@ -223,8 +223,10 @@ tools:
       that are down or have no active connections. The response summary counts
       enabled VPNs in state == "down" under downCount and enabled+up VPNs with
       no non-reserved client connected under zeroConnectionCount — derived by
-      probing each enabled+up VPN for a client whose clientUsername does not
-      match Solace's reserved `#*` prefix. Standby VPNs are excluded from both.
+      probing each enabled+up VPN that reports at least one connection for a
+      client whose clientUsername does not match Solace's reserved `#*` prefix;
+      a VPN reporting zero connections has no clients and is counted without a
+      probe. Standby VPNs are excluded from both.
```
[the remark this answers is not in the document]
solace-aross: Fixed: the description now says the probe fires unless a VPN reports exactly zero connections, and calls out that a missing count also probes (fails open) rather than being assumed empty. Addressed in 4ee3531.

internal/composite/definitions/tools.yaml
```diff
@@ -223,8 +223,11 @@ tools:
       that are down or have no active connections. The response summary counts
       enabled VPNs in state == "down" under downCount and enabled+up VPNs with
       no non-reserved client connected under zeroConnectionCount — derived by
-      probing each enabled+up VPN for a client whose clientUsername does not
-      match Solace's reserved `#*` prefix. Standby VPNs are excluded from both.
+      probing each enabled+up VPN unless it reports exactly zero connections
+      for a client whose clientUsername does not match Solace's reserved `#*`
```
bczoma: 🟡 **The new clause splits "probing … for a client whose clientUsername …", and changes what the sentence claims**

Flagged by: docs (follow-up pass)

**What the problem is.** The pre-`4ee3531` text read "probing each enabled+up VPN **for a client whose clientUsername does not match** Solace's reserved `#*` prefix". The new clause was inserted between the verb and its object, so the sentence now reads:

> probing each enabled+up VPN **unless it reports exactly zero connections for a client whose clientUsername does not match** Solace's reserved `#*` prefix

The natural parse attaches "for a client whose clientUsername does not match `#*`" to "reports exactly zero connections", not to "probing". So the description now states that the skip fires when a VPN reports zero *non-reserved* connections.

**Why it matters.** That is precisely the claim the predicate does not make, and precisely the wrong belief this PR's history is about. `msgVpnConnections` is the total across all clients including the reserved `#client`, which is why an idle VPN reports 1 and why a count of 1 is still probed on purpose. A reader who takes the sentence at face value concludes that a reserved-only VPN is counted into `zeroConnectionCount` without a probe — the behaviour of the `<= 1` heuristic this fan-out was introduced to replace. Two concrete consequences: an agent explaining the tool's output to an operator will describe the wrong mechanism, and a maintainer reconciling the predicate to its documented contract would "correct" `ne … 0.0` back toward `<= 1`, silently reintroducing the bug. The YAML comment twelve lines below is the only thing that currently contradicts them, and a tool consumer never sees it.

**Recommended fix.** Keep the verb and its object together and put the exception at the end:

```yaml
      probing each enabled+up VPN for a client whose clientUsername does not
      match Solace's reserved `#*` prefix, unless the VPN reports exactly zero
      connections; a VPN with no connection count at all is probed rather than
      assumed empty.
```

This keeps both facts the commit set out to add — only an explicit zero skips, and a missing count fails open — while leaving "zero connections" unqualified, which is what the predicate actually tests.

bczoma (dismissed):
## PR Review: #394 — SOL-154166: skip the list-vpns real-client probe on zero-connection VPNs

**Author:** solace-aross | **Branch:** `sol-154166-list-vpns-redundant-query` → `main` | **Changes:** +312 / -21 across 6 files
**Jira:** `SOL-154166` — list-vpns queries the broker for clients on VPNs we already know have none

Follow-up review: one new commit (`4ee3531`) since my review at `b7bcfff`, which GitHub dismissed on the push. Doc-only — the `list-vpns` tool description, answering Copilot's thread. No lenses re-dispatched; the code the prior pass measured is byte-identical at this head.

---

### Verdict

Conditioned on this 1 — it has an inline thread on the lines in question:

1. 🟡 `internal/composite/definitions/tools.yaml:226-227` — the new clause lands between "probing each enabled+up VPN" and "for a client whose clientUsername...", so the zero-connections test now reads as scoped to non-reserved clients; move it to the end of the clause

The 🟢 note below is **not** a condition of this verdict.

---

### Summary

The commit does what Copilot asked: the description now says only an explicit zero skips the probe, and that a missing count probes rather than being assumed empty. Both facts are correct and both were previously absent, so this is a real improvement to a string the LLM client actually reads. The problem is where the new clause was spliced in — it separated a verb from its object, and the resulting sentence asserts something narrower and wrong. That is worth one more pass precisely because the sentence is the artefact being fixed; everything else in the PR stands exactly as it did last review.

*7 prior findings settled — 1 addressed, 6 accepted as-is (all were offered as optional or as follow-up work, and the verdict was an unconditional approve).*

|---|---|---|---|
| 🟢 | The Copilot thread on the tool description is unaddressed and orphaned on a dead commit | Addressed | `internal/composite/definitions/tools.yaml:226-230` — description rewritten in `4ee3531`; author replied on the thread |

---

### Minor notes

- 🟢 `internal/composite/definitions/tools.yaml:226-230` — with the fix above applied, "a VPN reporting zero connections has no clients and is counted without a probe" restates the `unless` clause that now precedes it. Dropping the first half of that sentence and keeping "a VPN with no connection count at all is probed rather than assumed empty" would say everything the paragraph needs in one clause fewer. Description length is not free on a tool an agent loads on every session.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ⚠️ — the new sentence is more complete than the one it replaced, and misparses |
| Correctness | — not re-reviewed (unchanged since `b7bcfff`) |
| Code Quality | — not re-reviewed (unchanged since `b7bcfff`) |
| Test Coverage | — not re-reviewed (unchanged since `b7bcfff`) |
| Architecture | — not re-reviewed (unchanged since `b7bcfff`) |
| Security | — not re-reviewed (unchanged since `b7bcfff`) |
| PR Scope & Description | ✅ — sourced: parent (follow-up); the commit is doc-only and says so |

internal/composite/definitions/tools.yaml
solace-aross: Fixed exactly as suggested: kept the verb and object together, moved the exception to the end. Addressed in ba7a2cf.

bczoma (approved):
## PR Review: #394 — SOL-154166: skip the list-vpns real-client probe on zero-connection VPNs

**Author:** solace-aross | **Branch:** `sol-154166-list-vpns-redundant-query` → `main` | **Changes:** +310 / -20 across 6 files
**Jira:** `SOL-154166` — list-vpns queries the broker for clients on VPNs we already know have none

Follow-up review: one new commit (`ba7a2cf`) since my review at `4ee3531`, which GitHub dismissed on the push. Doc-only again — the same `list-vpns` description string. No lenses re-dispatched; the code both earlier passes measured is byte-identical at this head.

---

### Verdict

**Recommendation:** ✅ Approve

Nothing is a condition of this approval.

---

### Summary

The clause was moved exactly where the last pass asked: `probing each enabled+up VPN for a client whose clientUsername does not match Solace's reserved '#*' prefix, unless the VPN reports exactly zero connections`. Verb and object are back together, so the zero-connections test now reads as a property of the VPN rather than of the client being looked for, which is what the predicate actually does. The commit also took the optional note and dropped the restating half-sentence, so the paragraph is one clause shorter than before `4ee3531` while carrying strictly more information — the fail-open on a missing count is still stated. Both of the last pass's items are closed and the description now matches `ne (index .Item "msgVpnConnections") 0.0` clause for clause.

One housekeeping note that is not a review finding: my 🟡 thread on `tools.yaml:227` is still open and this repo requires conversation resolution to merge, so it needs resolving (yours or mine) before the merge button unblocks.

*2 prior findings settled — both addressed.*

|---|---|---|---|
| 🟡 | The new clause splits "probing … for a client whose clientUsername …" and narrows what the sentence claims | Addressed | `internal/composite/definitions/tools.yaml:225-228` — exception relocated to the end of the clause |
| 🟢 | "a VPN reporting zero connections has no clients and is counted without a probe" restates the `unless` clause preceding it | Addressed | `internal/composite/definitions/tools.yaml:227-228` — half-sentence dropped; net −1 line |

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ✅ — sourced: parent (follow-up) |
| Correctness | — not re-reviewed (unchanged since `b7bcfff`) |
| Code Quality | — not re-reviewed (unchanged since `b7bcfff`) |
| Test Coverage | — not re-reviewed (unchanged since `b7bcfff`) |
| Architecture | — not re-reviewed (unchanged since `b7bcfff`) |
| Security | — not re-reviewed (unchanged since `b7bcfff`) |
| PR Scope & Description | ✅ — sourced: parent (follow-up); the commit is doc-only and says so |

PR #399, opened 2026-09-09
SOL-152422: show retry attempts as sibling spans (SEMP and OAuth token exchange)

solace-aross:
## Summary

A retried broker call collapsed into one `semp.request` span, so a retry storm was invisible in a trace and nothing said which attempt succeeded. This adds one span per attempt — `semp.attempt` on the SEMP path and `tokenexchange.attempt` on the OAuth token-exchange path — each carrying the retry decision the policy actually made.

Jira: [SOL-152422](https://sol-jira.atlassian.net/browse/SOL-152422) (Story 27)

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

`semp.request` (Story 26) covers a whole retry chain, so a call that succeeded on its third try is one span, not three. The identical gap existed on the IdP side: three attempts, one aggregate count in a debug log line, no per-attempt tracing.

Both halves ship in one PR by design. SOL-152422's 2026-08-14 revision (ISSUE-031) extended the story to cover OAuth token-exchange retries alongside SEMP, and that scope was confirmed with the milestone owner.

| Span | Kind | Parent | One per |
|---|---|---|---|
| `semp.attempt` | Client | `semp.request` | SEMP request attempt |
| `tokenexchange.attempt` | Client | the winner's `tokenexchange.Exchange` | HTTP attempt of a live IdP exchange |

## Changes Made

**The decision attribute is the real decision.** `retryablehttp` owns the retry loop, so there is no call-site loop to hook, and `CheckRetry` runs *after* the transport returns. So the span is opened in a transport wrapper and closed from `checkRetry` — converted to named returns with one `defer`, which covers all fourteen of its exits without touching the policy logic, including the `return retryablehttp.DefaultRetryPolicy(...)` tail call.

The difference is observable and pinned: a `POST` answered with `503` reports `retry.decision: false`, because the non-idempotent-method guard refuses the replay. A status-derived attribute would claim a retry that never happened.

- **SEMP** (`internal/semp/resilience/attempt_span.go`, new): an always-installed `attemptTransport` opens the span and owns the attempt counter. `Sender.Do` already builds the per-request `retryState`, so there is no new plumbing.
- **Token exchange** (`internal/idpclient/retrying.go`): an `attemptSpanRecorder` sibling of the existing `attemptsRecorder`, plus an `attemptSpanSeeder` mounted *outside* retryablehttp's own RoundTripper. That outer position is the only layer that can seed per-request state where both `CheckRetry` and the inner transport will see it, and the only one that sees the whole chain finish.

Attempt spans parent to the singleflight-winning caller's `tokenexchange.Exchange` span through the `parentSpanCtx` that Story 50 (SOL-153333) already threaded past the `context.Background()` detach. That plumbing existed and was unused; this uses it.

**`retry.exhausted` means one thing:** a retry allowance the policy enforces was already spent. It covers all four on the SEMP path — `semp.retries`, the 429/503 sub-cap, the once-only non-429/503 5xx replay, and the once-only 401 re-auth. It is absent where the call failed but nothing ran out: a refused non-idempotent replay, a context that ended, a status never retried, or an auth mode that could not recover the first 401. Those want a different remedy from a bigger budget.

**One counter, two readers.** The attempt counter was bumped by `metricsTransport`, installed only when metrics are on — so with metrics off and tracing on, every attempt would have reported as attempt 1. Ownership moved to `attemptTransport`; `metricsTransport` now reads it, so the `attempt` span attribute and the `attempt` label on `mcp_semp_request_total` come from one source.

**Correlation ID.** Every attempt of one SEMP request carries the byte-identical `correlation_id`, resolved once per request and reused across retries. That identity is what lets an operator carry one ID from a retry storm in the trace into the broker-side command log.

**No `outcome`, span status left `Unset`** — deliberately, unlike every other span this server creates. A `503` that was retried and then succeeded is a normal step of a healthy call; tagging it `outcome: error` would put an error span under a successful parent on every retried call and inflate any error view a backend builds from that filter.

**Two pieces of connection hygiene that came with making the wrapper unconditional:**

- Every wrapper in the chain now forwards `CloseIdleConnections`. `net/http` relays it by type assertion on an unexported interface, so a wrapper without the method makes it a silent no-op — and retryablehttp calls it on all three of its "this chain went wrong, do not reuse these connections" paths. Metrics default to off, so before this change the transport was the bare `*http.Transport` and those calls worked; leaving it out would have disabled them for every broker on the default configuration.
- The SEMP wrappers compose onto a shallow copy of the caller's `*http.Client` rather than mutating it. Two Senders over one client each get a complete chain, instead of the second silently inheriting the first's — which would have dropped that Sender's SEMP metrics if only it carried `WithMetrics`, the same shape of defect this story fixes.

**Docs:** `docs/observability.md` moves per-attempt spans out of "pending", documents both new spans and all five attributes, and states the two token-exchange limitations below.

## Testing

**Test Environment:**
- Go version: go1.26.5
- OS: darwin/arm64
- Broker version (if applicable): n/a (httptest servers, in-memory span recorder)

**Tests Performed:**

**Test Coverage:**

- `make check` green; total coverage 89.8% (≥ 85% floor). `-race` clean on `internal/semp/...`, `internal/idpclient`, `internal/tokenexchange`, and `test/integration`; no flakes over `-count=2`.
- Covered on both paths: one span per attempt with the right parent, trace, kind and numbering; correlation-ID identity and its omission when absent; the not-re-derived decision; every exhaustion case and every non-exhaustion case; connection errors carrying no synthetic status; redirect hops not counting as attempts; the bypassed-state passthrough; the dangling-span backstop; the `CloseIdleConnections` relay; the transport-copy invariant; and the deduped-caller case.
- Span assertions are scoped by trace ID, not by span name alone, so a span from another test cannot make them pass or fail by accident.
- **Mutation-verified**, not just read. Each was injected and confirmed to fail with a useful message:

  | Injected defect | Caught by |
  |---|---|
  | `retry.decision` re-derived from the status code | the POST/503 test, plus both exhaustion tests |
  | `correlation_id` regenerated per attempt | the identity test and the omit-when-absent test |
  | token-exchange span parented from `context.Background()` | the nesting test and the deduped-caller test |
  | exhaustion predicate's false branch collapsed to `!retry` | both non-exhaustion tests and the predicate table |
  | `CloseIdleConnections` relay removed from the outer wrapper | the relay test |
  | `>` boundary in the exhaustion formula changed to `>=` | the RetryMax-exhaustion tests on both paths |

- One thing deliberately not pinned, and said so in the code: the `closeDanglingAttemptSpan` call site in `Sender.Do`. It is unreachable on today's retryablehttp, where `CheckRetry` runs immediately after every dispatch, so no behavioural test can hold it there. The method itself is tested directly; the defer bounds the failure mode if a future library version grows an exit between the dispatch and the check.

## Breaking Changes

None. Both spans are gated by `OBS_TRACING_ENABLED`; with tracing off the added cost is one integer increment and one non-recording span per attempt.

## Additional Notes

**Two documented limitations on the token-exchange path.** Both follow from the exchange running detached from any one caller, so that a cancellation cannot abort work other callers are waiting on. Neither is a bug to solve here, and both are asserted by test rather than only described — either would look healthy in a single-caller trace.

- A **deduped caller sees no attempt spans.** Concurrent identical exchanges collapse into one IdP round trip and the attempts hang off the winner's span. A follower's own `tokenexchange.Exchange` span has no attempt children; its existing `singleflight_role: "follower"`, `winner_trace_id` / `winner_span_id`, and span `Link` are the route to the trace that holds them.
- **`correlation_id` there is the winner's**, so it names the request that triggered the exchange rather than necessarily the one being looked at. Weaker than the SEMP path, where every attempt carries that request's own ID.

**Span-attribute egress.** The attempt spans carry only server-generated scalars — `attempt`, `retry.decision`, `retry.exhausted`, `http.response.status_code` — plus `correlation_id`, which every other span already exports and which the correlation middleware sanitizes and length-caps. No broker URL, no resolved SEMP path, no operationId, no token material, no caller-supplied strings. Span names are compile-time constants.

## Related Issues/PRs

- Builds on #379 / SOL-152421 (Story 26, `semp.request`) and SOL-153333 (Story 50, `tokenexchange.Exchange`)
- Related to SOL-152093 (Story 16, the per-attempt SEMP metrics whose `attempt` label now shares this counter)

---

**For Reviewers:**

**Focus Areas**: the span lifecycle crossing two layers (opened in a RoundTripper, closed in `checkRetry`) — in particular whether a span can be left unended, double-ended or overwritten; the `retry.exhausted` semantics, since which cases count is a judgement call worth a second opinion; and the `*http.Client` shallow copy in `resilience.New`, which changes who owns the transport chain.

**Questions**: `retry.exhausted` now covers all four retry allowances rather than only the two chain-wide ones. If you would rather it named only `semp.retries` and the 429/503 sub-cap, say so — the predicate is one function and the doc row is one line.


solace-aross:
Heads up for whichever of #399 / #400 merges second: both rewrite the same two regions of docs/observability.md (the tracing-status paragraph and the semp.attempt bullet) with opposite claims — #400 says Story 27's spans are still pending, #399 says they landed. This will conflict on merge, or ship stale if it resolves silently. Whoever rebases second, please reconcile that paragraph to reflect both landing. Caught by an independent fable review pass (git merge-tree) before either merged.

solace-aross:
Independent fable review pass (separate from the PR's own Opus gate). Verdict: approve, safe to merge as-is. Two small things worth a follow-up commit before merge, not blockers:

1. `internal/idpclient/attempt_span_test.go:380-429` doesn't test what it claims. The comment says it drives a cancelled-context-then-500 case to prove `retry.decision` isn't status-derived, but cancelling the request context mid-flight aborts the transport (resp==nil, err!=nil) rather than producing an intact 500. Replacing the decision logic with a status-derived expression leaves this test green. The SEMP twin (`resilience/attempt_span_test.go:1550-1578`, POST+503) is a real test and doesn't have this problem. Fix: drive the recorder at the transport level so an intact 500 actually arrives after cancel.

2. The stated rationale for which cases set `retry.exhausted` ("those want a different remedy from a bigger budget") doesn't actually discriminate — a persisted 401 and a 500,500 chain also don't want a bigger budget, yet both are marked exhausted, correctly. All four cases (semp.retries, the 429/503 sub-cap, the once-only 5xx replay, the once-only 401 re-auth) are the right scope, but the real discriminator is "the policy retried this failure class until an allowance it enforces was used up, and would have retried again with more" — worth restating that in the PR body, CHANGELOG, docs/observability.md, and the code comment at attempt_span.go:278-290, since the current phrasing implies a different (wrong) split.

Lower-priority notes if useful later: the log line "request failed after retries exhausted" and `retry.exhausted=true` are not the same predicate (worth a doc callout); `attemptSpanSeeder.CloseIdleConnections` in retrying.go can never actually relay (its inner has no such method) — dead code, not a live bug; the "backstop is untestable" claim in the PR body is false, a panicking base transport reaches it on both paths and a ~30-line test would pin it.

internal/semp/resilience/attempt_span.go
```diff
@@ -0,0 +1,269 @@
...
+// method would disable that hygiene on the default configuration for every
+// broker, invisibly.
+func (t *attemptTransport) CloseIdleConnections() {
+	forwardCloseIdleConnections(t.base)
+}
+
+// forwardCloseIdleConnections calls CloseIdleConnections on rt when it has one,
+// so a chain of wrappers relays it down to the real *http.Transport. Every
+// wrapper in the chain has to relay, or the first one that does not breaks it.
+func forwardCloseIdleConnections(rt http.RoundTripper) {
```
bczoma: 🟡 **The shared half of the span lifecycle is duplicated across two packages with no shared home**

*Flagged by: good:architecture + good:code*

**What the problem is.** `internal/semp/resilience` and `internal/idpclient` now carry six near-identical constructs: `forwardCloseIdleConnections` (byte-for-byte, doc comment included), `endAttemptSpan`, the whole `IsRecording`-guarded attribute block (same five attributes, same order, same omit-when-empty rule), `retriesExhausted`'s `retry && attempt > retryMax` core, the dangling-span closer, and the span-opening RoundTripper.

The *mechanism* divergence is genuinely forced and I am not asking you to unify it: `idpclient` hands back a bare `*http.Client` with no `Do` seam, so it must seed state from an outer RoundTripper, while SEMP already had `Sender.Do` and a `retryState` to hang it on. That difference is real. The ~150 lines of lifecycle logic that depend on neither are not.

**Why it matters.** Today `semp.attempt` and `tokenexchange.attempt` carry an identical attribute set, and `docs/observability.md` documents them in one table as though that were guaranteed. Nothing structural keeps them aligned, so the next attribute added to one silently forks a schema operators query across both — and a fix to one copy of the lifecycle rule lands in one package and not the other, invisibly, because each copy looks correct in isolation. The `retry.exhausted` divergence already in this PR (SEMP has four allowances, the IdP path has one) is a preview of exactly that drift, and it has already produced one documentation defect.

**Recommended fix.** Extract the shared half into one internal package — e.g. `internal/observability/attemptspan` — holding the state struct, the span-opening RoundTripper, `ForwardCloseIdleConnections`, and one `Finish(span, attempt, retry, resp, exhausted, corrID)` that writes the attributes. The two call sites then differ only in tracer name, span name, and how state is seeded, which is the part that really is different. Leave the exhaustion *derivation* forked — SEMP has sub-caps, the IdP path does not.

internal/semp/resilience/attempt_span.go
```diff
@@ -0,0 +1,269 @@
...
+		c.CloseIdleConnections()
+	}
+}
+
+// endAttemptSpan closes the attempt span opened by attemptTransport, tagging it
+// with the retry decision checkRetry is about to return. Called from
+// checkRetry's own defer, so it fires on every one of that function's exits
+// without any of them having to remember to.
+//
+// retry and checkErr are checkRetry's actual return values, read through named
```
bczoma: 🟡 **Three invariant comments describe a signature and a coverage gap that no longer exist**

*Flagged by: good:code, confirmed while measuring the lifecycle invariant*

**What the problem is.** These are the comments a reader consults to answer "is `retry.decision` really `checkRetry`'s own decision?", and three of them are inaccurate:

1. This doc says "`retry` and `checkErr` are checkRetry's actual return values, read through named returns" — but the signature at :149 is `(ctx, retryMax, resp, retry, allowanceSpent bool)`. There is no `checkErr`. The mirror comment in `idpclient` (`retriesExhausted`: "this takes no checkRetry error, where the SEMP version does") is wrong the same way.
2. :204-217 says "No behavioural test can therefore pin the deferred call site in `Sender.Do`". True of the backstop `defer` itself, but it reads as though the `RoundTrip`→`CheckRetry` *ordering* is unpinned. It is not: the `retry.decision` value assertions in `attempt_span_test.go` (154, 267, 313, 345 and peers) compare against a `bool` while driving the real `retryablehttp` client, so an absent attribute is an untyped nil that matches neither `true` nor `false` and the test fails.
3. :88 and `sender.go:777` describe the ordering as a property of "today's retryablehttp". It is the library's documented contract — `client.go:380-388` on the exported `CheckRetry` type: "It is called following each request with the response and error values returned by the http.Client." Changing that is major-version territory.

**Why it matters.** A reader auditing the story's core invariant is sent looking for a parameter that isn't there and may conclude the code drifted from the design. Worse, (2) and (3) together undersell the guarantee: a maintainer reading them concludes the ordering is unverified, and either adds a redundant guard test or hesitates over a routine dependency bump that CI would in fact have caught.

**Recommended fix.** Three sentence edits, no code change. Name `allowanceSpent` and say plainly that the sentinel errors are *not* consulted — that is the actual, and better, design. Narrow the "no behavioural test can pin this" claim to the backstop `defer`, noting the ordering is pinned by the decision-value assertions. Cite the `CheckRetry` type contract where the ordering is asserted, which turns a contingent-sounding note into a contract reference.

internal/idpclient/retrying.go
```diff
@@ -176,3 +241,211 @@ func (a *attemptsRecorder) RoundTrip(req *http.Request) (*http.Response, error)
...
+
+// endAttemptSpan closes the span attemptSpanRecorder opened, tagging it with
+// the decision checkRetry is about to return. Called from checkRetry's own
+// defer so it fires on every exit of that function.
+//
+// A nil parked span means the request never went through
+// attemptSpanRecorder, or checkRetry ran twice for one attempt (which
+// retryablehttp does when a Request carries a responseHandler — this package
+// sets none). Both are no-ops.
+func endAttemptSpan(ctx context.Context, resp *http.Response, retry bool, checkErr error) {
```
bczoma: 🟡 **`endAttemptSpan` takes an error it never reads, on the one path where an error must not be recorded**

*Flagged by: good:code + good:security*

**What the problem is.** The signature takes `checkErr error` as its last parameter and the body never reads it; `checkRetry`'s defer dutifully passes it. It is residue from an earlier design where exhaustion was derived from the sentinel error — the SEMP twin now takes `allowanceSpent bool` instead and needs no error at all.

**Why it matters.** Two reasons, and the second is why this is not a nit. A parameter that exists but is ignored tells the next reader the error influences the span attributes when it cannot, and `unparam` did not catch it, so nothing will remove it later. More importantly this is the **OAuth token-exchange** path: an unused `error` sitting in scope invites the obvious "use the available argument" edit, `span.RecordError(checkErr)` or `SetStatus(codes.Error, checkErr.Error())`. Today `checkErr` here is only `ctx.Err()`, but the function is one line from the transport error, and a Go `*url.Error` stringifies the full IdP token endpoint — which this package's own rules already treat as an unsanitised leak (SOL-153000, `Params.LogValue`'s `idp_endpoint`). Span attributes export off-box over OTLP and are covered by none of the `slog` redaction machinery.

That the current code records no error anywhere on either path is one of this PR's better properties. This parameter is the one loose thread in it.

**Recommended fix.** Drop the parameter and the argument: `endAttemptSpan(ctx, resp, retry)`. If it is deliberately kept as a placeholder for a future SEMP-style allowance, say so in the doc *and* say that no error may be written to this span and why. Dropping it is better — a comment can rot, a signature cannot.

internal/idpclient/retrying.go
```diff
@@ -176,3 +241,211 @@ func (a *attemptsRecorder) RoundTrip(req *http.Request) (*http.Response, error)
...
+// it and owns the retry loop — and does two things:
+//
+//   - Seeds the per-request attempt-span state onto the context, which is what
+//     makes it visible to CheckRetry and to the inner per-attempt transport.
+//   - Backstops the span lifecycle after the whole chain returns. On today's
+//     retryablehttp every dispatch is followed by a CheckRetry call, so nothing
+//     should be left open; an unended span is dropped entirely rather than
+//     exported, so being wrong here would cost the attempt its place in the
+//     trace.
+type attemptSpanSeeder struct {
```
bczoma: 🟡 **Two transports differing by one word, holding two independent counters that count different things**

*Flagged by: good:code*

**What the problem is.** `attemptsRecorder`, `attemptSpanRecorder`, `attemptSpanSeeder` and `attemptSpanState` all live in this one file, and the composition reads `&attemptSpanRecorder{inner: &attemptsRecorder{inner: inner.Transport}}`. The two `*Recorder` types differ by one word, and they hold **two independent attempt counters that do not count the same thing**: `attemptsRecorder` counts redirect hops, `attemptSpanState.attempt` deliberately does not.

**Why it matters.** This is the defect class the SEMP half of this PR exists to fix — "one counter, two readers", after the metrics transport's private counter made every attempt report as attempt 1 — reproduced on the other path. Here the two counters are legitimately separate, but nothing says they are *expected to disagree*, and the note at :279-282 says only that they are separate. A future author reconciling them "for consistency" would silently change what the span attribute means. The names compound it: grepping `attemptsRecorder` misses the span one, and that composition line reads like a typo on a skim, which is how a wrapper ends up inserted at the wrong depth.

**Recommended fix.** Rename by role rather than by noun — `attemptCountTransport` / `attemptSpanTransport` / `attemptSpanSeedTransport` — so the composition line states what each layer does. Separately, add one sentence at `attemptSpanState.attempt` naming which counter feeds the span attribute and stating that the two are expected to diverge on redirects.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `audit.HashArgs` computes the `arguments_hash` an auditor reproduces: a lowercase hex SHA-256 over the RFC 8785 (JSON Canonicalization Scheme) form of a tool call's arguments, via the new `github.com/gowebpki/jcs` dependency (Apache-2.0). A published canonicalisation spec rather than a hand-rolled one is the whole point — number serialisation, Unicode handling, and recursive key ordering are where a home-grown canonicaliser diverges silently, producing a digest a customer's own Python, Java, or Rust tooling cannot reproduce and our tests would not catch. Raw argument values never enter an audit record; the digest is the only representation of them the schema carries. Hash stability is pinned by goldens whose expected digests were computed outside Go from hand-derived canonical forms, so a change in canonicalisation output fails CI loudly rather than silently orphaning every hash emitted by an earlier build. `docs/observability.md` now carries the RFC reference, a reproduction recipe, and the additive-only schema commitment. Tracked under SOL-152090.
```
bczoma, line 36: 🟡 **"All of it is gated by `OBS_TRACING_ENABLED`" also covers two changes that are unconditional**

*Flagged by: good:pr*

**What the problem is.** The entry closes with "**All of it is gated by `OBS_TRACING_ENABLED`**; with tracing off the added cost is one integer increment and one non-recording span per attempt." That sentence follows, and appears to cover, the two connection-hygiene changes described two clauses earlier. Those are unconditional by design: `attemptTransport` is always installed where `metricsTransport` was not, so the `CloseIdleConnections` forwarding and the compose-onto-a-copy change take effect with tracing **and** metrics off — that is, on the default configuration.

**Why it matters.** A release manager or operator sizing this entry's blast radius for a tracing-off deployment reads "all of it is gated" and concludes the default configuration is unchanged. So the two behaviours that actually did change there — connection reuse after a failed retry chain, and two `Sender`s over one `*http.Client` each getting a complete chain — go un-regression-tested. It also reads as a claim that the story's DoD "inert when the flag is off" holds end to end, which for the hygiene half it deliberately does not.

**Recommended fix.** Narrow the sentence's scope, e.g.: "The spans and attributes are gated by `OBS_TRACING_ENABLED` …; the two transport-composition changes above apply regardless of either flag, including on the default configuration." The PR body's connection-hygiene section has the same ambiguity and is worth the same clause.

docs/observability.md
```diff
@@ -779,20 +779,21 @@ which you own.** The server does not itself persist or sign events.
 
 ## Distributed Tracing — [Interim: request-path spans wired]
```
bczoma: 🟡 **The Distributed Tracing heading still carries the old status tag, against this doc's own convention**

*Flagged by: docs*

**What the problem is.** This heading still reads `## Distributed Tracing — [Interim: request-path spans wired]`, while the capability table row at line 51 was correctly flipped to `[Interim — request-path and per-attempt spans wired, exemplars pending]`. Line 44 states the convention this doc holds itself to: "capability headings carry the same tag". The one place the tag was not flipped is the heading operators actually land on.

**Why it matters.** An operator or pilot who scans section headings for shipped status — the mechanism this document tells them to use — reads "request-path spans wired" here, concludes per-attempt spans are not in this build, and keeps inferring retries from `semp.request` duration instead of building the attempt-level view this PR shipped. Or files an "attempt spans missing" ticket against a feature that is present.

**Recommended fix.** Bring the heading in line with the table row: `## Distributed Tracing — [Interim: request-path and per-attempt spans wired, exemplars pending]`. Note that changes the anchor, so update both inbound links to the old slug `#distributed-tracing--interim-request-path-spans-wired` — at `docs/observability.md:51` and `docs/observability.md:386`.

docs/observability.md
```diff
@@ -881,8 +899,10 @@ distinction is the point, since they have different causes and different remedie
...
+| `retry.exhausted` | `true` on the final attempt when a retry allowance ran out; absent otherwise. It means one thing: the policy stopped because something it was counting was already spent. On `semp.attempt` that covers all four allowances — the configured `semp.retries`, the internal 429/503 sub-cap (which fires well below `semp.retries`, and is how a real broker-overload episode usually ends), the once-only replay of a non-429/503 5xx, and the once-only 401 re-auth. **Absent when nothing ran out**, even though the call still failed: a replay the policy refused because the caller declared the request non-idempotent, a context that ended, a status never retried at all, or an auth mode that could not recover the first 401. Those need a different remedy from a bigger budget, which is why they are distinguishable | Solace |
```
bczoma: 🟡 **The `retry.exhausted` row claims both spans but defines the attribute only for SEMP**

*Flagged by: docs*

**What the problem is.** The row says the attribute appears on both `semp.attempt` and `tokenexchange.attempt`, then defines it entirely in SEMP terms: the four allowances (`semp.retries`, the 429/503 sub-cap, the once-only non-429/503 5xx replay, the once-only 401 re-auth) and four SEMP-specific exclusions. On the token-exchange path the code is `retriesExhausted(retryMax, attempt, retry) => retry && attempt > retryMax` (`internal/idpclient/retrying.go:434-436`) against the IdP client's own `MaxRetries` (`retrying.go:72,128,148`) — no sub-cap, no non-idempotency guard, no 401 re-auth allowance. Both the inclusion list and the exclusion list are wrong for one of the two spans the row covers.

**Why it matters.** An SRE who sees `retry.exhausted: true` on `tokenexchange.attempt` follows the only remedy this row names and raises `semp.retries`, which has no effect on the IdP path — the token exchanges keep exhausting on an allowance the document never names. In the other direction, an SRE alerting on the attribute's *absence* on that path applies SEMP exclusions (a refused non-idempotent replay, an unrecoverable 401) that the path cannot produce, so the alert's semantics are wrong in a way nothing will reveal.

**Recommended fix.** Append to the row: "On `tokenexchange.attempt` it means only that the IdP client's configured retry count (`MaxRetries`) was spent — that path has no sub-cap, no non-idempotency guard, and no 401 re-auth allowance, so the four SEMP allowances and the four SEMP exclusions above do not apply. Raise the IdP retry setting, not `semp.retries`."

bczoma (dismissed):
## PR Review: #399 — SOL-152422: show retry attempts as sibling spans (SEMP and OAuth token exchange)

**Author:** solace-aross | **Branch:** `sol-152422-retry-attempt-spans` → `main` | **Changes:** +2391 / -80 across 13 files
**Jira:** `SOL-152422` — Show retry attempts as sibling spans so SREs can see which attempt actually succeeded (Story 27) (Verifying)

---

### Verdict

Conditioned on these 8 — each of the first 7 has an inline thread on the line in question:

1. 🟡 `internal/semp/resilience/attempt_span.go:124` — extract the span-lifecycle half the two paths share; the mechanism divergence is forced, the ~150 lines of identical attribute-writing and relay code are not
2. 🟡 `internal/semp/resilience/attempt_span.go:136` — three invariant comments describe a signature and a coverage gap that no longer exist; they undersell the guarantee they document
3. 🟡 `internal/idpclient/retrying.go:382` — `endAttemptSpan` takes `checkErr` and never reads it; on this path that parameter is one edit away from putting the IdP endpoint in a span
4. 🟡 `internal/idpclient/retrying.go:305` — `attemptsRecorder` / `attemptSpanRecorder` differ by one word and hold two independent counters that count different things
5. 🟡 `CHANGELOG.md:36` — "all of it is gated by `OBS_TRACING_ENABLED`" also covers the two transport changes, which are unconditional and land on the default configuration
6. 🟡 `docs/observability.md:780` — the Distributed Tracing heading still carries the old status tag, against this doc's own stated convention
7. 🟡 `docs/observability.md:905` — the `retry.exhausted` row claims both spans but defines the attribute only for SEMP; the token-exchange path has no sub-cap, no idempotency guard and no 401 allowance

Not anchorable to this diff (the metrics section is untouched), so no thread — please treat as part of the same list:

8. 🟡 `docs/observability.md:216-218` — the `attempt` label on `mcp_semp_request_total` previously read `1` for every attempt when metrics were on and tracing off. This release makes it real, so an existing dashboard or recording rule that assumes `attempt="1"` matches all series will silently under-count retried traffic and a retry-storm alert that could never fire will start firing — read as a broker regression. That behaviour change is recorded only in the CHANGELOG's `Added` bullet; the operator-facing metrics section says nothing, and line 218's "retry-storm detection reads the counter" was simply untrue before this fix. Add an upgrade note under line 218 naming the old behaviour and telling operators to revisit rules written against it.

---

### Summary

This does the hard thing correctly. `retry.decision` had to be the decision the policy actually made rather than one re-derived from the status code — AC 2 is explicit about it, and the `POST`+503 case makes the difference observable — and since `retryablehttp` owns the loop and calls `CheckRetry` only after the transport returns, the span has to be opened in a transport wrapper and closed from the policy callback. That seam is the whole design, and I checked it rather than trusting it: all fourteen `checkRetry` exits supply both named returns, the `DefaultRetryPolicy` tail call assigns through them before the defer runs, double-`End()` is prevented structurally by nilling the parked span before ending it, and both never-ended paths have a backstop. Tests are non-vacuous, every `range spans` is preceded by a length assertion, and both AC-mandated tests assert what the AC says rather than something adjacent. `-race` is green across the three packages.

Three findings were filed at 🟡 and measured down. The claim that the lifecycle invariant rests on an unpinned library ordering did not survive: the ordering is retryablehttp's documented `CheckRetry` contract, bumps are not auto-merged, and nine `retry.decision` value assertions in this PR drive the real client, so a version that skipped the call fails CI rather than degrading quietly. The `CloseIdleConnections` concern reduced too — all three of the library's failure-path calls target the inner client, whose chain is complete on both paths, and an unclosed idle connection self-evicts inside the 90s `IdleConnTimeout` with the request retried anyway. A suspected flake in the singleflight follower test was refuted outright by deleting the sleep and passing 500/500 under 16-way contention. What remains is mostly documentation and one duplication call.

The security surface deserves a word, because it is the part that could have gone badly: the attribute set on both spans is `attempt`, `retry.decision`, `correlation_id`, `http.response.status_code`, `retry.exhausted` and nothing else — no URL, no headers, no body, no error string — on a path whose requests carry `client_secret` and `subject_token`. That is stricter than default `otelhttp` instrumentation would have produced at the same position, and it holds on both paths.

---

### Minor notes

- 🟢 `docs/internal/secure-logging-rules.md` — the Never-Log List, `LogValuer`/`SanitizeURLString` and the `ReplaceAttr` safety net are all scoped to `slog`, and `/check-logs` scans logging calls, not `span.SetAttributes`. Today's attributes are clean, but this PR is the first to put attributes on the token-exchange path, so it is the change that exposes the gap. Worth a "Rule 7: span and metric attributes" clause and extending the scan to `SetAttributes`/`RecordError`/`SetStatus`.
- 🟢 `internal/idpclient/retrying.go:321` — "every wrapper in the chain now forwards `CloseIdleConnections`" is false as written: `attemptSpanSeeder` forwards to `*retryablehttp.RoundTripper`, which has only `init` and `RoundTrip` in v0.7.8, so the outer relay dead-ends and the returned client's `CloseIdleConnections()` is a no-op for an external caller. The hygiene the PR is about does work (the three library calls target `c.HTTPClient`). Reword the claim, and either drop the method as dead code or forward it to `rc.HTTPClient`.
- 🟢 `CHANGELOG.md:36` — the relay was also broken on the token-exchange path, where `attemptsRecorder` has always been installed unconditionally, so those three calls have been silent for as long as that wrapper existed. The entry credits the fix only to the SEMP default configuration; one clause would name the second place.
- 🟢 `internal/semp/resilience/sender.go:319` — `retryClient.HTTPClient = httpClient` is dead, unconditionally overwritten at :419, and its comment describes what no longer happens. If someone later guards the wrapping behind a condition, the fallback is a silently unwrapped client with no counter and no spans.
- 🟢 `internal/semp/resilience/sender.go:417-419` — the shallow copy isolates `Transport` but shares `Jar` and `CheckRedirect` by reference and snapshots `Timeout`, so `New` has quietly become a snapshot constructor: a caller that configures its client after `New` now gets no error and no effect. No live defect (both protocol clients set `Jar` before calling `New`), but say it in `New`'s doc comment.
- 🟢 `internal/semp/resilience/attempt_span.go:139` — when a `retryablehttp.Request` carries a `responseHandler` the library calls `CheckRetry` twice for one attempt and acts on the second decision, while the span closed on the first — the one path where this seam yields a *wrong* attribute rather than a missing one. Nothing sets one today and the comment says so; a constructor assertion or test would make adopting one a build failure instead.
- 🟢 `docs/observability.md:894` — the `correlation_id` row reads "the shared request ID" unqualified. The winner's-ID limitation is documented at line 861 but not in the table someone reads while writing a query, so they filter on it, get the winner's attempts or none, and conclude the IdP made no retries for their request.
- 🟢 `docs/observability.md:928` — "every span in the table above except the entry span carries `outcome`" now contradicts the new exception three lines above, in the very paragraph that exists to tell a reader what an `outcome="error"` filter misses. Add the two attempt spans to both the exception clause and the "misses" list.

*and 6 further minor items — the `### Fixed` vs `### Added` split for the metric-label fix, the untested declined-first-401 exhaustion boundary, the per-attempt `req.WithContext` clone the CHANGELOG's cost claim omits, a sampler note for attempt-span volume under retry storms, `attempt_span.go`'s 170-comment-lines-to-75-logic-lines ratio, and the follower audit-pivot completeness note — ask if you want them.*

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

internal/semp/resilience/attempt_span.go
solace-aross: Extracted into a new package, internal/observability/attemptspan: State, Transport, ForwardCloseIdleConnections, Finish. Both packages now call into it instead of carrying their own copies. Exhaustion derivation stays forked per your note, since the two policies really are different. New package has its own direct unit tests, 100% coverage. Addressed in d7b274d.

internal/semp/resilience/attempt_span.go
solace-aross: Rewrote the three comments. Named allowanceSpent instead of the nonexistent checkErr, narrowed the unpinned-ordering claim to the backstop defer alone (noting the RoundTrip-to-CheckRetry ordering is pinned by the decision-value assertions), and cited CheckRetry's documented contract instead of 'today's retryablehttp'. Addressed in d7b274d.

internal/idpclient/retrying.go
solace-aross: Dropped the parameter. endAttemptSpan is now endAttemptSpan(ctx, resp, retry). Addressed in d7b274d.

internal/idpclient/retrying.go
solace-aross: attemptSpanRecorder no longer exists as a distinct type (that half is now the shared attemptspan.Transport, built via newAttemptSpanTransport). Renamed attemptsRecorder to attemptCountTransport and attemptSpanSeeder to attemptSpanSeedTransport, and added a sentence on attemptSpanState naming which counter feeds the span attribute and that the two are expected to diverge on redirects. Addressed in d7b274d.

CHANGELOG.md
solace-aross: Narrowed the sentence: the spans and attributes are gated by OBS_TRACING_ENABLED, the two transport-composition changes are called out as unconditional and landing on the default configuration. Addressed in d7b274d.

docs/observability.md
solace-aross: Flipped the heading to match the table row and updated both inbound anchor links (lines 51 and 386). Addressed in d7b274d.

docs/observability.md
solace-aross: Appended the token-exchange clause. Addressed in d7b274d.

solace-aross:
All 7 anchored threads addressed and resolved. Also covers condition 8 (not anchorable, no thread): added an upgrade note under the SEMP request-metrics table (docs/observability.md) explaining that `attempt` previously read `"1"` for every attempt whenever metrics were on and tracing was off, and that a dashboard or alert rule tuned against that dead value should be revisited now that the counter is real.

Full extraction (`internal/observability/attemptspan`) plus all fixes are in d7b274d. `make check` green, 89.8% coverage. Branch was rebased twice onto current main during this pass (force-with-lease pushed) — re-requesting review since that dismisses the prior approval.

docs/observability.md
```diff
@@ -218,6 +218,15 @@ recorded per retry attempt so you can see retry storms and per-broker latency.
...
   **counter only**: the histogram omits it so its bucket series are not multiplied by the
   retry cap. Retry-storm detection reads the counter.
+- **Upgrade note (SOL-152422):** before this release, `attempt` read `"1"` for every
+  attempt whenever metrics were on and tracing was off — the metrics transport kept its
+  own counter that a bug left permanently at 1, so retried traffic was invisible to this
+  label. This release gives the counter a single owner shared with the trace spans, and
+  `attempt` now reflects the real try number. A dashboard or recording rule written
+  against the old behaviour — one that assumes `attempt="1"` matches all series, or that
+  never saw a retry-storm alert fire on this metric — should be revisited: it was reading
+  a value that could never change, and a rule tuned against that will now see real
```
bczoma: 🟡 **This upgrade note describes a regression that never shipped — and the ask was mine, so the correction is too**

*Flagged by: docs, then measured by an adversarial challenge that upheld it*

**What the problem is.** The note tells operators that before this release `attempt` read `"1"` for every attempt whenever metrics were on and tracing was off, "so retried traffic was invisible to this label". The pre-PR source says otherwise, and so does running it.

**How this was settled.** The merge-base tree at `0a537fa` was extracted and driven through a 503/503/200 retry chain with `WithMetrics` on and no tracer configured. Observed scrape:

```
mcp_semp_request_total{...,attempt="1",...,http_response_status_code="503"} 1
mcp_semp_request_total{...,attempt="2",...,http_response_status_code="503"} 1
mcp_semp_request_total{...,attempt="3",...,http_response_status_code="200"} 1
```

The mechanism matches: the metrics transport read the *shared* per-request state (`metrics.go:43` → `nextAttempt` → `retry.go:161` `s.attempt++`), and that `*retryState` pointer was attached once per `Do` at `sender.go:740` and carried through every retryablehttp try. Tracing owned no counter at all — `attempt_span.go` does not exist at that revision; the only pre-PR mentions of `semp.attempt` are comments reserving the name for Story 27. The one code path that returns a constant `1` is `retry.go:157-160`'s `if !ok { return 1 }`, reached only when the request bypasses `Sender.Do` — and the metrics transport is installed solely by `Sender.New`, so it is unreachable in shipped code. What this PR changed is counter *ownership*: `metrics.go` now calls `attemptNumber`, a pure read, while the attempt-span transport owns the bump. The emitted values are identical before and after.

**Why it matters.** `docs/observability.md` is the canonical operator doc in a public repo. This note tells an SRE that a metric they rely on was pinned at a constant, and directs them to revisit retry-storm dashboards and recording rules that were already correct and to distrust historical retry data that was in fact accurate. A false regression notice costs more than no notice.

**There is a second, independent problem in the same area.** `CHANGELOG.md:38` describes the *mirror-image* configuration — "with metrics off and tracing on every attempt would have reported as attempt 1". That one is accurate, but it is an in-PR design hazard for the *new* span attribute, which had no pre-PR existence. So the doc and the CHANGELOG in the same PR name opposite configurations for the same change, and at most one of them can be about released behaviour.

**Recommended fix.** Delete the note, or rewrite it as what actually happened and is a no-op for operators: the `attempt` label values are unchanged; the counter now has a single owner shared with the `semp.attempt` span attribute, so the label and the attribute cannot drift. Leave `CHANGELOG.md:38`'s framing where it is — it is a design note about the new attribute — and do not promote it into an upgrade note.

**And to be explicit: this one is on me.** My previous review asserted the stuck-at-`1` history and asked for the note. You implemented what was asked. Thanks for the round-trip.

internal/idpclient/attempt_span_test.go
```diff
@@ -0,0 +1,541 @@
...
+	attrs := spanAttrs(spans[0])
+	if got := attrs["retry.decision"]; got != false {
+		t.Errorf("retry.decision = %v, want false: the context was already done, so checkRetry refused to "+
+			"retry. A span reporting true is re-deriving the attribute from the response instead of "+
+			"reading the decision the client acted on", got)
+	}
+	if _, present := attrs["retry.exhausted"]; present {
+		t.Error("retry.exhausted is set on a call the caller cancelled; nothing ran out")
+	}
+}
```
bczoma: 🟡 **The IdP-side AC-2 proof cannot fail for the reason it states**

*Flagged by: good:code*

**What the problem is.** The comment says the context is already done when the response arrives, so `checkRetry` returns no-retry on a 500, and "an implementation that read the status code at the span site would report `retry.decision=true`". Instrumented and run three times, `client.Do` returns `resp=<nil>, err=Post "http://…": context canceled` every time, and the span carries no `http.response.status_code` at all. The 500 never reaches the span site, so a status-re-deriving implementation would also produce `retry.decision=false`.

**Why it matters.** This is the mutation proof for acceptance criterion 2 on the IdP path — `retry.decision` must come from the `checkRetry` decision itself and not be re-derived. The SEMP equivalent at `internal/semp/resilience/attempt_span_test.go:249` is genuine: a real POST plus a 503 reaches the span site, and the non-idempotency guard makes re-derivation observable, so a re-deriving implementation fails it. The IdP path currently has no test that distinguishes the decision from the status code, so the AC is covered on one of the two surfaces the ticket's 2026-08-14 extension put in scope.

**Recommended fix.** Make the response actually arrive before the decision — serve 500 and cancel from a `PrepareRetry`/`CheckRetry`-adjacent hook — or drop to a unit test that calls `checkRetry(doneCtx, &http.Response{StatusCode: 500}, nil)` directly and asserts both `retry.decision=false` and `http.response.status_code=500`. If it stays as it is, rewrite the comment so it claims only what it proves.

Noting for the record that an independent reviewer flagged a nearby range in this file for the same underlying reason, so this is a converged observation rather than a lone one.

bczoma (dismissed):
## PR Review: #399 — SOL-152422: show retry attempts as sibling spans (SEMP and OAuth token exchange)

**Author:** solace-aross | **Branch:** `sol-152422-retry-attempt-spans` → `main` | **Changes:** +2780 / -90 across 16 files
**This pass:** follow-up review of the 2 commits pushed since my review at `e5268c8` (recovered by patch-id across the rebase — 13 files, +1159/-394)
**Jira:** `SOL-152422` — Show retry attempts as sibling spans so SREs can see which attempt actually succeeded (Story 27) (Verifying)

---

### Verdict

1. 🟡 `docs/observability.md:221-228` — **the upgrade note I asked for describes a regression that never shipped; my prior finding was wrong.** Delete or rewrite it.
2. 🟡 `internal/idpclient/attempt_span_test.go:221-277` — the IdP-side AC-2 mutation proof cannot fail for the reason it claims; the 500 never reaches the span site.

**On finding 1, the correction is mine.** Last round I filed a 🟡 saying the `attempt` label on `mcp_semp_request_total` previously read `1` for every attempt whenever metrics were on and tracing was off, and asked for an operator upgrade note. You wrote the note on that instruction. It is wrong, and so was I. This pass rebuilt the pre-PR tree at the merge-base `0a537fa` and ran a 503/503/200 retry chain with `WithMetrics` on and no tracer: the scrape emitted `attempt="1"`, `attempt="2"`, `attempt="3"`. The metrics transport read the *shared* `*retryState` (`metrics.go:43` → `retry.go:161` `s.attempt++`) attached once per `Do` at `sender.go:740`, and tracing owned no counter at all — `attempt_span.go` did not exist at that revision. What this PR changes is counter *ownership*, not the emitted values. Sorry for the round-trip.

---

### Summary

The extraction landed well. `internal/observability/attemptspan` holds exactly the mechanism-independent half of the lifecycle — counter, parked span, attribute set, idempotent `End`, backstop — and leaves the two genuinely divergent exhaustion policies with their owners, with the package doc explaining why each split falls where it does. All seven anchored 🟡s from the previous round are verified fixed in code at this head, not merely claimed: the `checkErr` parameter is gone from `endAttemptSpan`, the three stale invariant comments now name the real `retry`/`allowanceSpent` returns, the near-identical transports are renamed and de-duplicated, and the docs and CHANGELOG corrections all check out. `Finish` clearing `state.Span` before `End()` makes double-close structurally impossible rather than conventionally avoided, and that is pinned by a test. Security re-verified the whole attribute surface on the token-exchange path — `attempt`, `retry.decision`, `correlation_id`, `http.response.status_code`, `retry.exhausted`, with zero `RecordError`/`SetStatus` calls anywhere in the new code — so nothing on a path carrying `client_secret` reaches a span. Two findings survive, one of which is a correction of my own previous instruction; both are cheap.

*16 prior findings, 10 settled — 7 addressed, 3 accepted as-is.* The exceptions:

|---|---|---|---|
| 🟡 | `docs/observability.md:216-218` — add an operator upgrade note for the `attempt` label's behaviour change | Partially addressed — **and the ask itself was wrong** | `docs/observability.md:221-228` — a note was added in the right place and the right shape, but its premise is false; see the thread and the correction above |
| 🟢 | `sender.go:319` — `retryClient.HTTPClient = httpClient` is dead, unconditionally overwritten at `:419` | Not addressed | `sender.go:318` still assigns it; `:420` unconditionally replaces it with `&wrapped`. This commit removed the `if _, wrapped := …` guard, so the assignment is now dead on *every* path rather than most |
| 🟢 | `observability.md:928` — "every span except the entry span carries `outcome`" contradicts the new exception | Not addressed | `docs/observability.md:949` still reads "Every span in the table above **except the entry span** carries `outcome`", seven lines after `:942` states the two attempt spans deliberately carry none |
| 🟢 | `CHANGELOG.md:36` — the `CloseIdleConnections` relay was broken on the token-exchange path too | Not addressed | The bullet still frames the fix as SEMP-only ("every broker on the default configuration"); no clause names the token-exchange wrapper chain |
| 🟢 | `retrying.go:321` — the "every wrapper forwards `CloseIdleConnections`" claim is false; the outer relay dead-ends | Partially addressed | The real hole is closed (`metrics.go:81` is new) and the false sentence left `retrying.go`, but it reappears in `attemptspan.go:114-119`, and `attemptSpanSeedTransport.CloseIdleConnections` (`retrying.go:322`) still forwards into `*retryablehttp.RoundTripper`, which has only `init`/`RoundTrip` in v0.7.8 |
| 🟢 | `sender.go:417-419` — `New` has quietly become a snapshot constructor | Partially addressed | The body comment at `sender.go:387-405` explains it and `TestNew_DoesNotMutateTheCallersHTTPClient` pins it, but `New`'s own doc comment (`sender.go:250-272`) — the only place a caller reads — still does not say `httpClient` is snapshotted |

---

### Minor notes

- 🟢 `internal/semp/resilience/sender.go:318` — delete the dead `retryClient.HTTPClient = httpClient` and its now-inaccurate comment, or move all `HTTPClient` wiring to one place at the end of `New`.
- 🟢 `internal/observability/attemptspan/attemptspan.go:114-119` — the doc claims the chain relays down to the real `*http.Transport`; narrow it to "every wrapper below retryablehttp's own RoundTripper", and either drop `attemptSpanSeedTransport.CloseIdleConnections` as dead or forward it to `rc.HTTPClient`.
- 🟢 `docs/observability.md:949` — add the two attempt spans to the `outcome` exception clause; the misses-list at `:952` needs no change, but say so.
- 🟢 `internal/semp/resilience/sender.go:250-272` — one clause in `New`'s doc comment saying `httpClient` is snapshotted, since a caller configuring it after `New` now gets no error and no effect.
- 🟢 `internal/observability/attemptspan/attemptspan.go:52-55` — `State.Attempt` and `State.Span` are exported mutable fields whose invariants are now enforced only by comment across a package boundary; unexporting them behind an `AttemptNumber()` accessor would put the compiler back in front of a double-`End()`.
- 🟢 `internal/observability/attemptspan/attemptspan.go:166` — `Finish(ctx, state, resp, retry, exhausted bool)` ends in two adjacent same-type bools meaning opposite-direction things; a transposition compiles clean. Named types or a small `Outcome` struct.
- 🟢 `internal/semp/resilience/retry.go:88-95` and `internal/idpclient/retrying.go:296-299` — `closeDanglingAttemptSpan()`/`closeDanglingSpan()` are one-line pass-throughs with 5-line comments; call `state.spanState.CloseDangling()` at the two defer sites and delete both.
- 🟢 `CHANGELOG.md` + PR body — the CHANGELOG credits the relay fix only to SEMP (the token-exchange wrapper was broken as long as it has existed), and the PR body's Changes Made / Focus Areas never mention `internal/observability/attemptspan`, the largest structural change since the last round.

**Merge-ordering hazard you flagged yourself, confirmed:** #399's side of `docs/observability.md` is internally self-consistent — the status row at `:51`, the cross-reference at `:397`, the heading at `:799` and the span bullets at `:865-883` all say Story 27's spans landed and exemplars are pending. #400 rewrites the same two regions with the opposite claim. Whoever rebases second must re-assert the "landed + exemplars pending" wording rather than accept either side wholesale.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ |
| Code Quality | ✅ |
| Test Coverage | ⚠️ |
| Architecture | ✅ |
| Security | ✅ |
| Documentation | ⚠️ |
| PR Scope & Description | ⚠️ |

docs/observability.md
solace-aross: Verified against the pre-PR tree (0a537fa): retryState.attempt already existed and was bumped correctly by the metrics transport on every attempt, independent of tracing, so the note described a regression that never shipped. Rewritten as what actually happened: the values are unchanged, only the counter's ownership moved to the new span. Thanks for catching the round-trip. Fixed in 221c772.

internal/idpclient/attempt_span_test.go
solace-aross: Confirmed: the transport aborts before the response is parsed in that scenario, so resp is nil and the ctx guard is the only thing exercised. Reworded that test's comment to claim only what it proves, and added TestCheckRetry_DoesNotRederiveDecisionFromStatus, a direct unit test that calls checkRetry with an already-done context and a real 500 response so the decision and the status can actually diverge. Confirmed it's a real proof by injecting a status-rederiving endAttemptSpan call: fails with retry.decision=true, want false, then passes clean reverted. Fixed in 221c772.

solace-aross:
Both threads addressed and resolved. On the first: you were right to catch the round-trip, verified against 0a537fa and rewrote the note as what actually happened. On the second: added a direct unit test for the actual AC-2 proof, confirmed it catches the mutation the end-to-end version couldn't. Both in 221c772.

Re-requesting review since the push dismissed the approval (this looks to dismiss on any push here, not just a rewrite).

bczoma (dismissed):
## PR Review: #399 — SOL-152422: show retry attempts as sibling spans (SEMP and OAuth token exchange)

**Author:** solace-aross | **Branch:** `sol-152422-retry-attempt-spans` → `main` | **Changes since my last review:** +73 / -21 across 2 files (`221c772`)
**Jira:** `SOL-152422` — per-attempt retry spans (Story 27)

---

### Verdict

**Recommendation:** ✅ Approve

Both conditions of my previous 🔄 are addressed, and I verified each in code at `221c772` rather than from the commit message. Nothing new in this push needs changing.

*Follow-up pass: prior review at `d7b274d` was dismissed by this push; its findings still applied, and the authored delta was recovered by patch-id — one commit, two modified files, no new files and no auth/crypto/API/config surface.*

### Summary

This push does exactly the two things it claims. The `attempt`-label upgrade note in `docs/observability.md` is rewritten from a regression narrative into an accurate "no operator-visible change" note, and the IdP-side AC-2 mutation proof is replaced by one that can actually fail. Both of my prior findings were about claims that were stronger than the evidence — one of them a claim I had made myself — and both corrections are right on the merits.

*2 prior conditions settled — both addressed. The 8 🟢 notes from the previous pass were offered as optional; those not taken are accepted as-is and are not restated here.*

|---|---|---|---|
| 🟡 | The `attempt` upgrade note describes a regression that never shipped (my prior ask was itself wrong) | Addressed | `docs/observability.md:221-228` — rewritten as "No operator-visible change". I re-checked the premise against the base branch: `origin/main:internal/semp/resilience/retry.go:156` `nextAttempt` bumps a per-request `retryState` counter that the recording transport calls once per try, so `attempt` already ran 1, 2, 3 with metrics on and tracing off. The new note's claim is accurate, including the tracing-only hazard it says the design avoids — at head the attempt-span transport is always installed and `internal/observability/attemptspan/attemptspan.go:97` is the single `state.Attempt++`, which `internal/semp/resilience/metrics.go:51` reads rather than bumps. |
| 🟡 | The IdP AC-2 mutation proof cannot fail for the reason it claims; the 500 never reaches the span site | Addressed | `internal/idpclient/attempt_span_test.go:283-330` — a new `TestCheckRetry_DoesNotRederiveDecisionFromStatus` calls `checkRetry` directly with a real `*http.Response` carrying 500 and an already-cancelled context, and asserts both `retry.decision=false` and `http.response.status_code=500`, so the response is provably at the span site while the decision is false. I confirmed it empirically: re-deriving the decision from `resp.StatusCode` inside `endAttemptSpan` (`internal/idpclient/retrying.go:355-362`) fails the new test (`retry.decision = true, want false`) and leaves the old end-to-end test green. The old test is correctly renamed to `TestAttemptSpans_CancelledDuringDispatchIsNeverRetried` with a comment stating plainly what it does and does not prove. |

### Minor notes

None from this push.

One piece of housekeeping that is already tracked in your own comment rather than a finding of mine: the `docs/observability.md` overlap with #400 still needs reconciling by whichever of the two rebases second.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ |
| Test Coverage | ✅ — the added test is mutation-verified, not just green |
| Documentation | ✅ |
| Code Quality | — not re-reviewed (no non-test source changed since `d7b274d`) |
| Architecture | — not re-reviewed (unchanged since `d7b274d`) |
| Security | — not re-reviewed (unchanged since `d7b274d`) |
| PR Scope & Description | — not re-reviewed (unchanged since `d7b274d`) |

*Assessed by the parent on the follow-up pass; no discipline lenses were dispatched.*

solace-aross:
Rebased onto main now that #400 merged. That PR touched the same docs/observability.md regions you'd flagged as needing manual reconciliation regardless of merge order — git didn't mark it as a conflict on the capability-table row and the Distributed Tracing heading (only the two prose blockquotes conflicted textually), so I checked those by hand: both still said "exemplars pending" after the rebase even though #400 shipped them. Fixed: dropped that clause from both, added a pointer to Trace Exemplars from the capability-table row, and updated the heading's anchor to match (no more "exemplars-pending" suffix).

make check green, 89.8% coverage. Force-with-lease pushed (8a379e9), re-requesting review since that dismisses the approval again.

solace-aross:
Correction to my last comment: the cross-PR doc reconciliation (dropping the stale "exemplars pending" clause from the capability-table row and heading) wasn't actually committed when I posted that — it was still sitting in the working tree. It's in c3dcbb8, not 8a379e9. Sorry for the noise.

bczoma (approved):
## PR Review: #399 — SOL-152422: show retry attempts as sibling spans (SEMP and OAuth token exchange)

**Author:** solace-aross | **Branch:** `sol-152422-retry-attempt-spans` → `main` | **Changes since my last review:** +3 / -3 in 1 file (`c3dcbb8`)
**Jira:** `SOL-152422` — per-attempt retry spans (Story 27)

---

### Verdict

**Recommendation:** ✅ Approve

Nothing outstanding. My previous pass closed both of its conditions, and this push does exactly one thing — the cross-PR doc reconciliation we'd both flagged as needing a manual catch — correctly and completely.

*Follow-up pass: my prior review at `221c772` was dismissed by the rebase onto main after #400 merged. Its findings still applied and none were outstanding. The authored delta was recovered by patch-id: three of the branch's commits are patch-equivalent to ones I'd already reviewed and only `c3dcbb8` is new — one doc file, no new files, no auth/crypto/API/config surface, so no lenses were dispatched.*

### Summary

`c3dcbb8` drops the stale "exemplars pending" clause from the Distributed Tracing capability-table row and heading, points the row at Trace Exemplars, and renames the heading anchor to match. This is the gap that git could not surface: #400 landed the exemplar work without touching these two lines, so the rebase carried them forward still claiming exemplars were pending after they'd shipped.

I verified the reconciliation rather than taking the commit message for it:

- **No dangling anchors.** The heading is now `## Distributed Tracing — [Interim: request-path and per-attempt spans wired]`, whose GitHub anchor is `#distributed-tracing--interim-request-path-and-per-attempt-spans-wired`; the one remaining inbound reference (`docs/observability.md:397`) was updated to match, and a repo-wide grep for the old `…-exemplars-pending` anchor returns nothing. The new outbound link `#trace-exemplars--implemented` resolves to the real `### Trace Exemplars — [Implemented]` heading at `docs/observability.md:447`.
- **No stale prose left.** A case-insensitive grep across all markdown for "exemplars pending" / "pending exemplars" is clean.
- **The rebase didn't clobber #400.** Diffing `origin/main...HEAD` on this file, every removed line is one this branch deliberately replaces with its shipped-state wording; #400's exemplar sentence and its `[Trace Exemplars]` link survive intact in the rewritten blockquote at `docs/observability.md:838-840`.

Keeping `[Interim]` on the row is still right — span names and span kinds remain open for pilot input, so the tag is not carrying the exemplar claim.

No need to act on the two comments correcting which commit carried the change; `c3dcbb8` is where it is, and that's what I reviewed.

*2 prior conditions settled in the previous pass — both addressed and re-confirmed still in place at this head. The 8 🟢 notes from the first pass were offered as optional and are accepted as-is.*

### Minor notes

None from this push.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ✅ — anchors, inbound references and cross-PR content all verified at this head |
| Correctness | — not re-reviewed (no source changed since `221c772`) |
| Test Coverage | — not re-reviewed (unchanged since `221c772`) |
| Code Quality | — not re-reviewed (unchanged since `221c772`) |
| Architecture | — not re-reviewed (unchanged since `221c772`) |
| Security | — not re-reviewed (unchanged since `221c772`) |
| PR Scope & Description | — not re-reviewed (unchanged since `221c772`) |

*Assessed by the parent on the follow-up pass; no discipline lenses were dispatched.*

bczoma (approved):
## PR Review: #399 — SOL-152422: show retry attempts as sibling spans (SEMP and OAuth token exchange)

**Author:** solace-aross | **Branch:** `sol-152422-retry-attempt-spans` → `main` | **Changes since my last review:** none authored
**Jira:** `SOL-152422` — per-attempt retry spans (Story 27)

---

### Verdict

**Recommendation:** ✅ Approve

Re-approving unchanged. My approval at `c3dcbb8` was dismissed by the rebase onto main after #343 merged, not by any new work.

*Follow-up pass: `git cherry c3dcbb8 HEAD origin/main` marks all five of the branch's commits `-` — every one is patch-equivalent to a commit I had already reviewed, under a new SHA. The only content difference between the two heads is #343's CI workflow changes arriving from main (`.github/workflows/guardian-scan.yaml`, `release-readiness-check.yaml`, `release.yml`, `THIRD_PARTY_BUILD_TEST.md`), which this PR did not author. The PR's own diff against main is unchanged at +2832/-93 across 16 files.*

Nothing outstanding: both conditions from my earlier passes were closed and re-confirmed at `c3dcbb8`, and the cross-PR doc reconciliation I verified there — anchors, inbound references, and #400's exemplar content surviving the rebase — is carried forward intact.

*All prior conditions settled in earlier passes; nothing re-opened by this rebase.*

### Minor notes

None.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| All categories | — not re-reviewed (no authored change since `c3dcbb8`; head moved by rebase only) |

*Assessed by the parent on the follow-up pass; no discipline lenses were dispatched.*

PR #400, opened 2026-09-09
SOL-152419: prove and pin trace exemplars on the two latency histograms (Story 47)

solace-aross:
## Summary

Link a slow latency bucket straight to the trace that produced it, so an SRE stops correlating by timestamp. Story 47.

The audit this story asked for found **no defect**: both histogram observation sites already thread the span-bearing request context, so exemplars already reach an OpenMetrics scrape. The deliverable is therefore the guarantee, not the bytes — three tests that fail when it breaks, plus the operator documentation.

Jira: [SOL-152419](https://sol-jira.atlassian.net/browse/SOL-152419)

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

## Motivation and Context

The epic already promises the experience: "a customer SRE clicks an exemplar link from a Grafana panel and lands on the failed trace in Tempo." Under ADR-007 it is nearly free, because metrics and traces share one SDK and the trace context is already on the request context at the observation site.

That cheapness is also the risk. The OTel metrics SDK attaches an exemplar from whatever span is active on the `context.Context` passed to `.Record()`. Substitute a bare context at either observation site and **spans still export, the histogram still records correctly, and the exemplars vanish with no error on any surface.** Nothing in the suite noticed. That is the gap this PR closes.

## Changes Made

**Audit — no production fix needed.** Both sites already carry the request-scoped context:

| Site | Path | Evidence |
|---|---|---|
| Tool RED (`mcp_tool_invocation_duration_seconds`) | `manager.go:227` reassigns `ctx` from `tracer.Start`; the deferred `recordToolInvocation(ctx, …)` at `manager.go:256` closes over that variable, which is never reassigned again in `CallTool` | `duration.Record(ctx, …)` at `instruments.go:172` |
| SEMP RED (`mcp_semp_request_duration_seconds`) | `sempv2/client.go:199` starts `semp.request` and reassigns `ctx`; `buildRequest(ctx, …)` and `req.WithContext(ctx)` carry it; `Sender.Do` derives from it and re-attaches; `retryablehttp` copies the request per attempt as a shallow struct copy (`client.go:795` `httpreq := *req.Request`), which carries the unexported `ctx` field | `t.recorder.Record(req.Context(), …)` at `resilience/metrics.go` |

The other three tool dispatch sites (`RegisterListBrokers`, `describe-semp-schema`, and the argument-parse failure closure) and the SEMPv1 `Execute` were audited by reading and thread the same span-bearing `ctx`; they share the single `ToolMetrics.Record` / `SEMPMetrics.Record` seam the tested paths exercise.

**Tests (the deliverable).** Each asserts the exemplar's `trace_id` **equals the trace ID of the span an in-memory recorder captured for that same call**, not that some exemplar is present.

- `internal/tools/exemplar_test.go` — drives `ToolManager.CallTool` end to end, plus the two dispatch sites that bypass it (`list-brokers`, `describe-semp-schema`) through the in-memory MCP transport. Those two each duplicate the whole start-span-then-defer mechanism rather than sharing `CallTool`'s, so each is an independently breakable copy. The fourth site (`register.go`'s argument-parse branch) is already pinned by `TestDispatch_AuditAndMetricSeeTheDispatchSpanInContext`.
- `internal/semp/sempv2/client_exemplar_test.go` — drives `Execute` through a **503-then-200 retry**, so a context dropped only on a retried attempt is caught, and requires both attempt series (`status="503"` and `status="200"`) to carry the trace.
- `internal/observability/metrics/exemplar_test.go` — the representation half of decision D4: the real recorders under a sampled span, an unsampled span, an explicit no-op provider, and the process default (which is what `tracing.New` leaves in place when the flag is off). Only the sampled case may carry an exemplar, and the **exemplar-stripped output must be identical across all four**, which asserts two acceptance criteria at once: metrics gained no dependency on tracing being enabled, and exemplars added no label key and no series. It is also the only assertion that would catch a label key added conditionally on a span being present. Covers **both `_total` counters as well as both histograms** — the exporter attaches exemplars to a monotonic sum too, and the docs now say so, so the claim is held to a test. It also pins that the plain-text exposition carries no exemplar — the premise that keeps `TestGoldenSchema` deterministic — and pins what `OTEL_METRICS_EXEMPLAR_FILTER` does (below).

**Two comments** mark the contexts as load-bearing where a future edit would break them silently, and name the test that catches it. The one in `manager.go` replaces an accidentally duplicated copy of itself.

**`docs/observability.md`** — `Trace Exemplars` moves to `[Implemented]`, with the two caveats that otherwise read as bugs (exemplars ride only the OpenMetrics exposition; an exemplar can only reference a *sampled* trace, so a low `OTEL_TRACES_SAMPLER_ARG` legitimately leaves most buckets without one) and the tracing-off behaviour. Removes the stale claim that exemplars were pending on a `[Planned]` SEMP histogram, and a nearby stale `[Planned]` marker on `mcp_semp_request_total` — which shipped in #387 and which that same paragraph points operators at for retry counts.

**`OTEL_METRICS_EXEMPLAR_FILTER` is now documented, not pinned.** `metrics.New` passes no `WithExemplarFilter`, so the SDK reads the standard env var and defaults to `trace_based`. Under `always_off` a fully sampled trace produces no exemplar at all; under `always_on` an exemplar is attached with **no** span on the context, carrying an empty `trace_id` that links nowhere. Honouring the variable rather than overriding it in code is deliberate — it is the standard OTel contract, the tracer provider treats `OTEL_TRACES_SAMPLER` the same way, and silently ignoring an operator's explicit `always_off` would be its own defect. So the doc's caveat list is reframed as *four* reasons exemplars can be missing from an otherwise healthy scrape, none of them visible from the scrape, and both behaviours are pinned by test.

**No `metrics_schema` bump.** The policy reserves a minor bump for a new metric, field, or `outcome` value. Exemplars are none of those: no new series, no new label keys, and they were already on the wire. The unchanged golden file — which pins `mcp_schema_version{metrics_schema="1.1"}` — is the confirmation.

## Testing

**Test Environment:**
- Go version: go1.26.5
- OS: darwin/arm64
- Broker version (if applicable): n/a (httptest servers, in-memory span recorder)

**Tests Performed:**

**Test Coverage:**
- `make check` green; total coverage 89.5% (≥ 85% floor). No flakes over `-race -count=3` on the three affected packages.
- **`testdata/metrics_golden.txt` is untouched** and deliberately still does not cover exemplars: a fixture over a live trace ID would flap every run, and its being unchanged is itself the confirmation that cardinality did not move.
- **Every negative case fails loudly if its own premise stops holding**, rather than passing on an absence: the span's sampled-ness is asserted before the exemplar assertion, an OpenMetrics `Content-Type` is required before any exemplar is looked for, and an empty histogram is a fatal error rather than a vacuous pass.
- **Mutation-verified**, not just read. Each of these was injected and confirmed to fail with a useful message:

  | Injected defect | Caught by |
  |---|---|
  | `ToolMetrics.Record` observes on `context.Background()` | metrics + `AbsentFromPlainTextScrape` |
  | `recordToolInvocation(context.Background(), …)` in `CallTool` | tools call-site test |
  | `metricsTransport.RoundTrip` observes on `context.Background()` | sempv2 call-site test |
  | `EnableOpenMetrics: false` on the handler | all (fails on `Content-Type`) |
  | histogram recorded only when the span is sampled | metrics test, on both the "nothing recorded" guard and the cross-case comparison |
  | a label key added only when a span is present | metrics test, cross-case comparison |
  | both `_total` counters observe `context.Background()` | metrics test (this one **escaped** the first round and is why the counters are now covered) |
  | `list-brokers` site observes `context.Background()` | brokerless-sites test (**escaped** the first round) |
  | `describe-semp-schema` site observes `context.Background()` | brokerless-sites test (**escaped** the first round) |

## Breaking Changes

None. Exemplars require both `OBS_METRICS_ENABLED` and `OBS_TRACING_ENABLED`; with tracing off the histograms are byte-identical apart from carrying no exemplars, which is asserted rather than asserted-about.

## Additional Notes

**Grafana panel wiring is deliberately out of scope** and is left to Story 37 (SOL-152092), which owns the importable reference dashboard and its graceful degradation to plain histograms when no trace data source is configured. Nothing here touches a dashboard.

**The full metrics schema doc pass is Story 35 (SOL-152094)**, and the operator runbook entry is Story 36 (SOL-152098). The `docs/observability.md` edits here are scoped to this story's own acceptance criterion — the sampling interaction and the exemplar section's status — not those passes.

**Not stacked on Story 27 (SOL-152422, per-attempt retry spans).** That work is in flight separately and this branch neither depends on nor duplicates it: no new spans are created here. When `semp.attempt` lands, the SEMP exemplar will point at the innermost sampled span on the observation context, which is the intended behaviour and needs no change on this side.

**Why a CHANGELOG entry despite no behaviour change.** The `/changelog` skill excludes pure test/docs churn, and by the letter of that rule this could carry `no-changelog`. I added an entry anyway because the *documented* surface moved: `docs/observability.md` previously told operators exemplars were pending, and an operator reading the changelog to decide whether they can wire exemplar links now gets an answer. The entry says explicitly that nothing new is emitted, so it does not overclaim.

## Related Issues/PRs

- Depends on (all merged): #387 (Story 16, SEMP RED), #379 (Story 15, tool RED), #357 (Story 14, meter provider + `EnableOpenMetrics`), #385 (Story 26, request-path spans), #355 (Story 45, exporter fidelity)
- Blocks: Story 37 (SOL-152092, reference dashboard)
- Not related to Story 27 (SOL-152422, per-attempt SEMP spans), which is in flight separately; this branch creates no spans

---

**For Reviewers:**

**Focus Areas**: whether the audit conclusion is right — that no production change was needed — and whether the negative cases in `exemplar_test.go` are genuinely load-bearing rather than absence-assertions that cannot fail. The mutation table above is the evidence I have; contradicting it is the most useful thing a reviewer can do.

**Questions**: `scrapeExemplarBearing` is duplicated across the two call-site test files rather than shared, because a non-test helper package reports 0% coverage against the 85% gate (see `internal/observability/panics/panicstest`). The scrape underneath it is *not* duplicated — sempv2's now reuses that package's own `scrapeMetricsAccepting`. Happy to reconsider if you would rather have the whole thing shared.

Second commit is the response to an internal review pass that injected ten defects; three escaped the first round (both `_total` counters, and the two brokerless dispatch sites) and one documented claim was falsifiable via `OTEL_METRICS_EXEMPLAR_FILTER`. All four are closed and re-verified by re-injection. Reading the commits separately shows what was wrong and why.


solace-aross:
Heads up for whichever of #399 / #400 merges second: both rewrite the same two regions of docs/observability.md (the tracing-status paragraph and the semp.attempt bullet) with opposite claims — #400 says Story 27's spans are still pending, #399 says they landed. This will conflict on merge, or ship stale if it resolves silently. Whoever rebases second, please reconcile that paragraph to reflect both landing. Caught by an independent fable review pass (git merge-tree) before either merged.

internal/tools/exemplar_test.go
```diff
@@ -0,0 +1,230 @@
...
+			"so this test would assert an absence and pass for the wrong reason")
+	}
+	wantTraceID := dispatch.SpanContext().TraceID().String()
+
+	withExemplar, all := scrapeExemplarBearing(t, p, "mcp_tool_invocation_duration_seconds")
+	if len(withExemplar) == 0 {
+		t.Fatalf("no bucket carries an exemplar, want one with trace_id %q — the observation "+
+			"site is not seeing the dispatch span's context:\n%s", wantTraceID, strings.Join(all, "\n"))
+	}
+	for _, line := range withExemplar {
```
bczoma, line 124: 🟡 **No test asserts the exemplar's `span_id` — an exemplar pointing at a never-exported span passes green**

*Flagged by: good:code (found by mutation, outside your table)*

**What the problem is.** Every exemplar assertion in this PR matches on `trace_id` against the recorded span and stops there — here, at `:216-222`, at `client_exemplar_test.go:152-156`, and in `internal/observability/metrics/exemplar_test.go` where `exemplarTraceIDRe` deliberately captures only `trace_id`. I injected an observation on a child span started but never ended, so never exported: same trace, a `span_id` that exists in no backend. The exemplar is emitted, the `trace_id` matches, and the whole suite stays green (`ok .../internal/tools`). This is the one mutation shape the table does not cover.

**Why it matters.** The epic promise is that an SRE clicks the exemplar and lands on the failed trace — and a Grafana→Tempo exemplar link resolves the **span**, not the trace. A valid `trace_id` with a dead `span_id` is precisely the broken link this story exists to prevent, and it is invisible from the scrape in the same way the missing-exemplar case is.

It is also not hypothetical for long. You note that when Story 27's `semp.attempt` spans land, the SEMP exemplar will point at the innermost sampled span instead — which is the intended behaviour, but nothing here would notice if that innermost span turned out to be unexported, filtered, or otherwise not the one the recorder captured. The guarantee would silently weaken from "links to the span we exported" to "links somewhere in the right trace".

**Recommended fix.** Extend each assertion to span identity, not just trace identity. In the two call-site files, require the exemplar line to contain `span_id="` + `dispatch.SpanContext().SpanID().String()` + `"` as well — one extra `strings.Contains` each. In `internal/observability/metrics/exemplar_test.go`, widen the regex to capture `span_id` by name (the same by-name, order-independent style already used there) and compare it to the recorded span. Cheap, and it turns "the trace exists" into "the linked span is the one we exported".

internal/tools/exemplar_test.go
```diff
@@ -0,0 +1,230 @@
...
+// covers the two tool dispatch sites that never reach ToolManager.CallTool.
+//
+// They are the weakest link in this story rather than an afterthought: each
+// duplicates the whole start-span-then-defer-the-emission mechanism instead of
+// sharing CallTool's, so the ctx threading is a separate, independently
+// breakable copy at each — and both tools appear on every dashboard, so an
+// operator carrying a `tool` label into a trace backend is exactly who a
+// missing exemplar strands.
+//
+// The fourth site, the argument-parse failure in register.go's instrumented
```
bczoma, line 141: 🟡 **The argument-parse dispatch site is the fourth breakable copy and the only one with no exemplar assertion**

*Flagged by: good:architecture*

**What the problem is.** There are four tool dispatch sites, each an independent copy of the start-span-then-defer mechanism. This PR adds an exemplar assertion for three of them — `CallTool`, `list-brokers`, `describe-semp-schema`. The fourth, `register.go`'s argument-parse failure branch (`register.go:265-268`), is claimed to be "already pinned by `TestDispatch_AuditAndMetricSeeTheDispatchSpanInContext`". That test (`spans_test.go:186`) asserts the **audit log line** carried the dispatch span's ID. It never looks at a metric, and never at an exemplar.

So it would catch the historical regression it was written for — the span being started after emission — but it would **not** catch `recordToolInvocation(context.Background(), …)` on line 267 while `logToolResult(ctx, …)` on line 266 stays correct. That is exactly the mutation class this PR exists to catch, and exactly the mutation that escaped your first round at the other two brokerless sites.

**Why it matters.** The PR's thesis is the guarantee rather than the bytes, and the guarantee is only as good as its weakest copy. Three of four sites are independently breakable with an assertion each; the fourth is independently breakable with none. It is also the only site whose emission is straight-line rather than deferred, so it has the least structural protection against a future edit that reorders or re-derives that block — and the audit-line proxy would keep passing throughout.

**Recommended fix.** Add a third case to the brokerless-sites table driving the parse-failure branch — valid-JSON-non-object args, the same input `spans_test.go` already uses — and assert the `trace_id` (and, per the other thread, the `span_id`). About 15 lines, and it removes the asymmetry entirely. If you would rather not, the alternative is to reword the comment here so it states plainly that this site's exemplar is *inferred* from the audit line's span rather than asserted — so the next reader does not count four pinned sites when there are three.

docs/observability.md
```diff
@@ -433,24 +433,42 @@ substituting the metric schema's label value into a field name (e.g. guessing
...
 `mcp_semp_request_duration_seconds`) carry **trace exemplars** when both metrics and tracing are
 enabled, so a slow bucket on a Grafana panel links straight to the trace that produced it and
-you skip correlating by timestamp.
+you skip correlating by timestamp. The matching `_total` counters carry them too.
 
-Two things to know, because both look like bugs otherwise:
+Four things to know. Each is a reason exemplars can be missing from a scrape that is
+otherwise perfectly healthy, and none of them is visible from the scrape itself:
 
 - **Your Prometheus must negotiate OpenMetrics to receive them.** Exemplars are not part of the
```
bczoma, line 448: 🟡 **The section requires OpenMetrics negotiation but the `Accept` header appears nowhere in the docs**

*Flagged by: docs*

**What the problem is.** This bullet tells the operator "your Prometheus must negotiate OpenMetrics", and that is correct and important — but neither this section nor anywhere else in `docs/` gives the header value. The only place the exact string exists in the repo is a test: `internal/observability/metrics/exemplar_test.go:107`, `application/openmetrics-text; version=1.0.0; charset=utf-8`.

**Why it matters.** This is the one verification step the section exists to enable, and it is the step an SRE will actually take. They do the obvious thing — `curl http://host:9091/metrics` — get the default plain-text exposition, see not a single `#` exemplar suffix, and conclude the `[Implemented]` status is wrong or the build is broken. The same gap hits anyone told to "confirm your scraper negotiates OpenMetrics": the doc gives them nothing to compare their scrape config against. Since exemplars are invisible from a default scrape *by design*, the doc is the only thing standing between correct behaviour and a bug report.

**Recommended fix.** One line in the bullet:

```
To confirm by hand:
curl -H 'Accept: application/openmetrics-text; version=1.0.0; charset=utf-8' \
  http://<host>:<metrics-port>/metrics
```

then say what success looks like — an exemplar appears as a `# {trace_id="…"} <value> <timestamp>` suffix on a bucket line — and that a plain `curl` returns the text exposition and shows none. That last clause is what stops the negative result being read as a defect.

bczoma (approved):
## PR Review: #400 — SOL-152419: prove and pin trace exemplars on the two latency histograms (Story 47)

**Author:** solace-aross | **Branch:** `andrea/SOL-152419-trace-exemplars` → `main` | **Changes:** +945 / -16 across 9 files
**Jira:** `SOL-152419` — Prove and pin trace exemplars on the two latency histograms (Story 47) (Verifying)

---

### Verdict

You asked for the mutation table to be contradicted. It largely could not be — four rows were re-injected here, including all three you disclose as escaping your first round, and every one reproduced as a failure with the message the test promises. One mutation outside the table did survive, and it is the first item below.

1. 🟡 `internal/tools/exemplar_test.go:124` — no test asserts the exemplar's `span_id`, so an exemplar pointing at a never-exported span passes green
2. 🟡 `internal/tools/exemplar_test.go:141` — `register.go`'s argument-parse site is the fourth independently breakable copy and the only one with no exemplar assertion of its own
3. 🟡 `docs/observability.md:446` — the section tells an operator to negotiate OpenMetrics but the `Accept` header appears nowhere outside a test

---

### Summary

The audit conclusion is right, and I checked it rather than accepting it: all seven paths that reach `ToolMetrics.Record` / `SEMPMetrics.Record` receive an uninterrupted descendant of a `tracer.Start` context — including SEMPv1 `Execute`, which your table covers by reading only, and including across retries, where the guarantee leaves your code and rests on `httpreq := *req.Request` carrying `http.Request`'s unexported `ctx` field in the pinned `go-retryablehttp@v0.7.8`. No production defect, so no production change. The two seams really are the only two, which is why the exemplar came for free.

The tests are load-bearing on every axis the table names. Re-injecting `context.Background()` at the `CallTool` site, at `metricsTransport.RoundTrip`, and at both brokerless dispatch sites reproduced four distinct failures; the conditional-label-key mutation fired only on the cross-comparison, as you claim, though it is the two *tracing-off* cases that carry it — an unsampled span still has a valid `SpanContext`, so `tracing_on_but_sampler_declined` does not catch that one. Running the whole metrics package under `OTEL_METRICS_EXEMPLAR_FILTER=always_on` made every negative case fail, which is the evidence that the absence assertions are live rather than vacuous. `-race -count=3` is clean on all three packages.

The gap is `span_id`. Every assertion matches `trace_id` against the recorded span and stops there, so an observation on a child span that is started but never ended — correct trace, span that exists in no backend — leaves the entire suite green. A Grafana exemplar link resolves the span, not the trace, so that is a dead link of exactly the kind this story exists to prevent, and it becomes live the moment Story 27's `semp.attempt` lands and the SEMP exemplar starts pointing at an innermost span instead.

Two things worth knowing that are not defects. Your `metrics_schema` call is right as the policy is written, and the reservoir cost is fixed per series rather than per request, so there is no volume concern at rate. And an exemplar *can* carry arbitrary span attributes on the wire — the exporter puts `FilteredAttributes` into the exemplar's label set — which is inert here only because the package declares no Views; that is the 🟢 below.

**On your open question about `scrapeExemplarBearing`:** your premise checks out — the gate is a single aggregate over `./...` minus `test/`, `panicstest` really does sit at 0%, and there is no exclusion mechanism you missed. Two independent looks split on the conclusion, one calling ~30 duplicated lines the right trade and one noting a `metricstest` package would move the aggregate by under 0.1pp against 4.5pp of headroom. The tiebreaker is that the two copies have already diverged — sempv2's delegates to `scrapeMetricsAccepting`, tools' inlines the scrape — under one name. Either extract it, or make each copy a five-line wrapper over its own package's scrape helper so the OpenMetrics guard exists once per package. Two full bodies under one name is the option I would not keep.

---

### Minor notes

- 🟢 `internal/observability/metrics/exemplar_test.go` — the four-way identity assertion compares *exemplar-stripped* output, so nothing pins what labels the exemplar itself carries. The exporter puts `FilteredAttributes` into that label set, populated exactly when a View filters an attribute off the timeseries — so the day someone removes a high-cardinality broker or VPN label from a series "for cardinality", it returns as an exemplar label and this guarantee does not notice. Asserting the label set is exactly `{trace_id, span_id}` is ~5 lines and converts "no View exists yet" into a test.
- 🟢 `internal/semp/resilience/metrics.go:51` — the comment says `req.Context()` is load-bearing and names its test, but not *why a retried attempt still has a span*: the shallow struct copy inside retryablehttp. The dependency is genuinely pinned by the 503→200 test, so this is diagnostic only — but when a bump turns that test red, the reviewer has no pointer from the failure to the cause and "flaky exemplar test" is the likely reading. One sentence fixes it.
- 🟢 `internal/observability/metrics/exemplar_test.go:190` — the four-case matrix inherits any ambient `OTEL_METRICS_EXEMPLAR_FILTER`; with `always_on` exported, three of the four fail for a purely environmental reason invisible in the failure text. `t.Setenv(..., "trace_based")` at the top makes the matrix's premise explicit.
- 🟢 `docs/internal/architecture.md:503` — the `metrics` row still lists "Trace exemplars (SOL-152419)" in its *not-owned* column, which now contradicts the `[Implemented]` status this PR establishes and the test that lives in that very package.
- 🟢 `docs/observability.md:450` — "raise the sampler argument" names only `OTEL_TRACES_SAMPLER_ARG`, which `tracing/provider.go:70-79` records is not read at all unless `OTEL_TRACES_SAMPLER` is also set. The advice bites in the scenario described (coverage is already low, so the sampler must already be set); the risk is a reader landing here first and generalising.
- 🟢 `docs/observability.md:49` — `Trace Exemplars` now carries `[Implemented]` but appears nowhere in the capability status table, while the parallel case (`Panic Recovery`) is pointed at from the Metrics row's Notes. An operator reading only the table sees no signal that a sub-capability under it went live.
- 🟢 PR-level — SOL-152419's own AC list contains a checkbox whose subject is Story 37: the dashboard wiring exemplar links. You correctly declare it out of scope, so nothing is deferred silently, but the ticket can now transition with the epic's "click through to the trace in Tempo" recorded as delivered while no panel link exists. Worth a line in the body, or striking that AC.
- 🟢 PR-level — the #399/#400 conflict on the same two `docs/observability.md` regions is disclosed in a comment, but nothing enforces the ordering: no label, no note in the body reviewers actually read. If #400 lands second and resolves toward the incoming hunk, the shipped doc says Story 27's spans are pending when they landed.

*and 5 further minor items — the `docs/observability.md:118` "text exposition format" line that now contradicts the OpenMetrics requirement, the "four things … missing" framing that does not cover the present-but-dead `always_on` exemplar, `Added` vs `Changed` for a changelog entry that ships no new bytes, the unjoined `server.Run` goroutine in the brokerless test, and `secure-logging-rules.md` having nothing to say about metric or exemplar labels — ask if you want them.*

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

internal/observability/metrics/exemplar_test.go
```diff
@@ -0,0 +1,474 @@
...
+				}
+
+				stripped[tc.name+"|"+family.name] = withoutExemplars(lines)
+			}
+		})
+	}
+
+	// "The histogram is otherwise unchanged": every case must have produced the
+	// same series, labels and values as the sampled one.
+	const reference = "tracing_on_and_sampled"
```
bczoma, line 325: 🟡 **The cross-case identity comparison can silently become a no-op**

Flagged by: good:code

**What the problem is.** `stripped` is populated inside the `t.Run` subtests above and read here, after the loop. That ordering is correct only because none of those subtests calls `t.Parallel()`. Add it — which is a natural thing for a future author to do to a table-driven test with four independent cases — and the parent function reaches this loop before any subtest body has run. `want` and `got` are then both `""`, `got != want` is false, and all twelve comparisons pass. The map writes also become unsynchronised.

**Why it matters.** This is the assertion the file's own doc comment calls "the only assertion here that catches a label key added conditionally on a span being present", and it carries two acceptance criteria at once — that metrics gained no dependency on tracing being enabled, and that exemplars added no label key and no series. I confirmed it has real teeth today by injecting a `traced="yes"` label key added only when the span is sampled: six errors, correctly. So the mechanism works; the concern is only that it degrades to silence rather than to a failure.

It is also the one place the file's `t.Parallel` discipline is not written down. `TestExemplars_ExemplarFilterEnvVarOverridesTheDefault` warns about it carefully, and so does `spans_test.go` — which is what makes the omission here worth a line rather than a shrug.

**Recommended fix.** A length guard before the loop turns the silent pass into a loud failure and costs nothing:

```go
if len(stripped) != len(cases)*len(exemplarFamilies) {
	t.Fatalf("stripped has %d entries, want %d — a subtest did not record, "+
		"so the comparison below is vacuous", len(stripped), len(cases)*len(exemplarFamilies))
}
```

It also fixes a smaller annoyance: if the `tracing_on_and_sampled` subtest `t.Fatalf`s early, the reference entries are missing and this loop currently emits twelve spurious "differs from the sampled case" diffs on top of the real failure, burying the cause.

bczoma (dismissed):
## PR Review: #400 — SOL-152419: prove and pin trace exemplars on the two latency histograms (Story 47)

**Author:** solace-aross | **Branch:** `andrea/SOL-152419-trace-exemplars` → `main` | **Changes:** +945 / -16 across 9 files
**Jira:** `SOL-152419` — Link a slow latency bucket straight to its trace with exemplars (Verifying)

---

### Verdict

1. 🟡 `internal/observability/metrics/exemplar_test.go:313` — guard the cross-case identity comparison against reading an unpopulated map, so it cannot silently become a no-op

The 4 🟢/💡 notes below are **not** conditions of this verdict.

---

### Summary

The audit conclusion holds. I checked both observation sites independently and the PR's reading is right: `manager.go:227` reassigns `ctx` from `tracer.Start` and the deferred `recordToolInvocation` closes over that variable, and the SEMP path threads the span-bearing context through `buildRequest` → `req.WithContext` → `retryablehttp`'s shallow request copy to `req.Context()` at the recorder. No production change was needed, and the PR says so plainly rather than manufacturing one.

The more important claim — that the new tests are load-bearing rather than absence-assertions — was checked by re-running four of the ten mutations from your table independently, not by reading. All four failed, and failed with the intended message: a bare context in `ToolMetrics.Record`, `EnableOpenMetrics: false` (which fails on the Content-Type fatal in both packages, exactly as designed), a bare context in the SEMP RoundTripper, and a label key added only when a span is sampled — that last one caught by six "differs from the sampled case" errors, which is real enforcement of the no-new-label-keys criterion rather than prose about it. `exemplarTraceIDRe` matching by label name with `*` rather than `+` on the hex is a good, subtle call: `trace_id=""` is reported as an empty exemplar instead of silently reading as "no exemplar".

Every documentation claim the PR adds was verified against the source and every one checks out — `provider.go:106` genuinely passes no `WithExemplarFilter`, `provider.go:186` genuinely sets `EnableOpenMetrics: true`, and `mcp_semp_request_total` really did ship in #387, so removing that stale `[Planned]` marker closes the last inconsistent copy rather than creating one.

The single 🟡 is the one place the file's own careful `t.Parallel` discipline is not applied — and it happens to be the assertion carrying two acceptance criteria.

On the `scrapeExemplarBearing` question you raised: your reasoning holds. The coverage gate is `grep -v '^$(MODULE)/test/'` (`Makefile:155`), so a shared `internal/...` helper package would be counted and would dilute. Leave it. If a fourth copy ever appears, the cheaper fix is one more term in that Makefile filter, not a redesign — worth a sentence in `docs/internal/unit-test-coverage.md` so the next author does not re-litigate it.

---

### Minor notes

- 🟢 `internal/tools/exemplar_test.go:188` — the MCP server goroutine is neither cancellable nor awaited, and `defer session.Close()` does not synchronise with `Run` returning, while `recordSpans(t)` resets a process-shared `SpanRecorder` for the next subtest. A span from subtest 1's still-running server arriving after subtest 2's `Reset` would be picked up by the `Ended()` loop, which takes the last match. `-race` passes today but nothing enforces it: `ctx, cancel := context.WithCancel(...)` plus a `done` channel in `t.Cleanup`.
- 🟢 `docs/internal/architecture.md:503` — the `metrics` row's third column is the "does not own" boundary column, and it still lists `Trace exemplars (SOL-152419)` there, while this PR moves them to `[Implemented]` and puts their test inside that very package. A developer using the table to find the exemplar surface is sent away from it.
- 🟢 PR body, Summary — says the deliverable is "three tests" while the CHANGELOG entry this PR adds says "four tests now hold it". The Summary sentence predates commit 2, which added two more.
- 💡 `internal/observability/metrics/exemplar_test.go:421` — the `always_on` case asserts `id != tc.wantTraceID` with `wantTraceID: ""`, so the equality check compares `""` to `""`. It is not vacuous — the `len(got) == 0` branch above gives the case teeth — but it is the exact tautology-through-a-shared-table-field shape you asked reviewers to hunt for, and a reader has to reconstruct why it is safe. One line of comment, or assert the presence directly.

---

### On the #399 collision you flagged

Not re-raising it — but it is one region wider than your note says. #399 also rewrites the Implementation Status table's `Distributed tracing` row at `docs/observability.md:52` to read `[Interim — … exemplars pending] … Trace exemplars are still pending (Story 47)`. #400 does not touch that row. In the #400-then-#399 merge order that lands text calling Story 47 pending directly above a section headed `### Trace Exemplars — [Implemented]`, and because #400 does not touch it, git will not flag it as a conflict. Whoever rebases second should reconcile that row too, not just the two regions you named.

---

### Summary Verdict

|---|---|---|
| Architecture | — not reviewed (production edits are comment-only; no architectural surface in the diff) | — |
| Security | — not reviewed (no auth, crypto, input-validation or dependency surface in the diff) | — |
| Documentation | ✅ | docs |

solace-wmaryam:
## Review

Approve. This is a test-and-docs change with comment-only production edits, so the risk is low. I read the existing threads and agree with them; I won't restate them. The audit conclusion holds (both sites thread a span-bearing `ctx` into `.Record()`, and the SEMP one survives the per-attempt request copy), and the tests are genuinely load-bearing rather than absence-assertions.

### One new point (Low, not a blocker)

The `always_on` doc bullet and its test case both miss the more common dead-link shape.

The doc describes only the *no span* case: an exemplar with an empty `trace_id`. But with `always_on` and an **unsampled** span active, the exemplar carries a real, valid-looking `trace_id` that the sampler dropped, so it exists in no backend. An empty ID renders as an obvious dead link; a real-looking one looks clickable and resolves to nothing, which is the harder support ticket.

`TestExemplars_ExemplarFilterEnvVarOverridesTheDefault` has the same gap: it covers `always_on` with `sampledSpan: false` only.

Fix: add a clause to the doc bullet, and optionally a third table case (`always_on` + unsampled span, expecting a non-empty but unexported `trace_id`).

### On the open conditions

The three 🟡 threads are correct and worth doing before merge: assert `span_id` (a live `trace_id` with a dead `span_id` is exactly the broken link this story prevents), add an exemplar assertion for the `register.go` argument-parse site, and put the OpenMetrics `Accept` header in the docs.

### Merge note

The #399/#400 conflict on `docs/observability.md` is real and already flagged, but nothing enforces the order. If this lands second and resolves toward the incoming hunk, the doc will call Story 27 spans "pending" after they landed. Reconcile that paragraph by hand at rebase.

internal/tools/exemplar_test.go
solace-aross: Added span_id assertions alongside the existing trace_id ones: the call-site tests now require the exemplar line to also carry the dispatch span's span_id, and the metrics-package regex widened to capture span_id by name and compare it to the recorded span. Addressed in 526c83a.

internal/tools/exemplar_test.go
solace-aross: Kept this as the audit-line proxy rather than adding a fourth direct assertion, and reworded the comment to say so plainly: register.go's argument-parse branch is inferred from TestDispatch_AuditAndMetricSeeTheDispatchSpanInContext's audit-line assertion, not independently pinned by an exemplar test of its own. Addressed in 526c83a.

docs/observability.md
solace-aross: Added the curl command with the Accept header, what a healthy exemplar suffix looks like, and that a plain curl showing none isn't itself a defect. This one was still open after the first round; fixed now in 46fc72b.

internal/observability/metrics/exemplar_test.go
solace-aross: Added the length guard before the cross-case loop: if a subtest didn't record (t.Parallel added later, most likely), it now fails loudly instead of comparing two empty strings. Addressed in 46fc72b.

solace-aross:
All 4 threads addressed and resolved. Two (span_id assertions, register.go audit-line reword) were already fixed in 526c83a from an earlier pass; the Accept-header doc snippet had been missed then and is fixed now, along with the t.Parallel guard from your second review. Both in 46fc72b.

Rebased onto current main (force-with-lease pushed), so re-requesting review since that dismisses the prior approval. make check green, 89.6% coverage.

bczoma (approved):
## PR Review: #400 — SOL-152419: prove and pin trace exemplars on the two latency histograms (Story 47)

**Author:** solace-aross | **Branch:** `andrea/SOL-152419-trace-exemplars` → `main` | **Changes:** +968 / -17 across 9 files
**This pass:** follow-up review of the 1 commit pushed since my review at `526c83a` (recovered by patch-id across the rebase — 2 files, +23/-1). Parent-only pass; no lenses dispatched.
**Jira:** `SOL-152419` — prove and pin trace exemplars on the two latency histograms (Story 47)

---

### Verdict

**Recommendation:** ✅ Approve

Both outstanding conditions are closed, verified in code at `46fc72b` rather than taken from the commit message. Nothing new in this push needs changing, and no threads are opened.

- **`internal/observability/metrics/exemplar_test.go:320` — the vacuous-comparison guard is correct and exact.** `stripped[tc.name+"|"+family.name]` is written unconditionally at the end of every family iteration (`:306`), so `len(cases)*len(exemplarFamilies)` is precisely the count a complete run produces — the guard cannot false-positive, and any `t.Fatalf` that aborts a subtest leaves the map short and trips it. Under the `t.Parallel` scenario the comment names, the parent reaches line 320 before any subtest body has run, `stripped` is empty, and it fails loudly instead of comparing `""` to `""`. `go test ./internal/observability/metrics/ -run TestExemplars -count=1` passes.
- **`docs/observability.md:451-461` — the `Accept` header gap is closed properly.** The snippet gives the exact `application/openmetrics-text; version=1.0.0; charset=utf-8` header, shows what a healthy exemplar suffix looks like, and — the part that makes it useful rather than decorative — states that a plain `curl` returning none is not itself a defect, only a scraper that has not asked for OpenMetrics. That is the sentence that stops a false bug report.

The 4 minor notes below remain open from the previous round. None is a condition of this verdict.

---

### Summary

This push is 23 lines and both of them are the right 23 lines. The test guard is the more interesting of the two: it pins a premise the file's own `t.Parallel` discipline otherwise left implicit, and it does so with an invariant that is exact rather than approximate, which is what makes it a guard and not another assertion that can rot. The docs addition closes the last operator-facing gap in the exemplar section. Nothing in the incremental change touches production code, adds a file, or moves auth, config or API surface, so this stayed a parent-only pass. @solace-wmaryam's independent approve stands and is consistent with what I read.

*8 prior findings, 4 settled — all 4 addressed.* The exceptions, all 🟢/💡 and all carried from the previous round:

|---|---|---|---|
| 🟢 | `internal/tools/exemplar_test.go:188` — the MCP server goroutine is neither cancellable nor awaited, while `recordSpans(t)` resets a process-shared `SpanRecorder` between subtests | Not addressed | `internal/tools/exemplar_test.go:188` still reads `go func() { _ = server.Run(ctx, serverTransport) }()` with a plain `context.Background()` and a `defer session.Close()` that does not synchronise with `Run` returning |
| 🟢 | `docs/internal/architecture.md:503` — the `metrics` row still lists trace exemplars in the "does not own" boundary column | Not addressed | `docs/internal/architecture.md:503` — the row is unchanged and still reads `Trace exemplars (SOL-152419)` in that column, while this PR moves them to `[Implemented]` and puts their test inside `internal/observability/metrics/` |
| 🟢 | PR body Summary says "three tests" while the CHANGELOG entry says four | Not addressed | The PR body's Summary still reads "three tests that fail when it breaks"; the count moved when commit 2 added two more |
| 💡 | `internal/observability/metrics/exemplar_test.go:421` — the `always_on` case compares `""` to `""` through a shared table field | Not addressed | Unchanged. Still safe — the `len(got) == 0` branch gives the case teeth — but it is the exact shape this PR asks reviewers to hunt for, so one line of comment would spare the next reader the reconstruction |

---

### Minor notes

- 🟢 `internal/tools/exemplar_test.go:188` — `ctx, cancel := context.WithCancel(...)` plus a `done` channel closed in `t.Cleanup`, so a subtest's server cannot outlive it and land a span after the next subtest's `Reset`.
- 🟢 `docs/internal/architecture.md:503` — move `Trace exemplars (SOL-152419)` out of the "does not own" column; a developer using that table to find the exemplar surface is currently sent away from the package that holds it.
- 🟢 PR body, Summary — "three tests" → four, to match the CHANGELOG entry this PR adds.
- 🟢 `internal/observability/metrics/exemplar_test.go:421` — one line saying why the `always_on` equality check is not vacuous, or assert the presence directly.

**Merge-ordering hazard, still live and now more urgent.** You flagged it yourself: this PR and #399 rewrite the same two regions of `docs/observability.md` with opposite claims — #400 says Story 27's spans are still pending, #399 says they landed. I have just approved #399, so if that merges first and this one rebases without reconciling that paragraph, the released doc tells operators a shipped feature is pending. Whoever rebases second needs to re-assert the combined wording — Story 27's spans landed, exemplars landed — rather than accepting either side wholesale. Worth doing deliberately rather than at conflict-resolution speed.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Test Coverage | ✅ (parent, follow-up) |
| Documentation | ✅ (parent, follow-up) |
| Correctness | — not re-reviewed (unchanged since `526c83a`) |
| Code Quality | — not re-reviewed (unchanged since `526c83a`) |
| Architecture | — not re-reviewed (unchanged since `526c83a`) |
| Security | — not re-reviewed (unchanged since `526c83a`) |
| PR Scope & Description | — not re-reviewed (unchanged since `526c83a`) |

PR #401, opened 2026-09-09
SOL-152099: emit SDK-independent auth-failure and authorization-denial counters

solace-awaheed:
## Summary

Adds two security counters the server increments itself, `mcp_auth_failure_total{reason}` and `mcp_authz_denied_total{tool,reason}`, so SRE alerting on rejected credentials and authorization denials no longer depends on parsing the MCP SDK's log format.

Tracked under SOL-152099 (Story 24).

## Type of Change

- [x] Documentation update

## Motivation and Context

Rejected credentials were visible only through the SDK's log lines, whose format can change underneath an alert. Authorization denials produced no metric at all: a denied call short-circuits above the instrumented handler, so it never reached `mcp_tool_invocation_total`. This closes both gaps with counters whose `reason` labels are drawn from the same closed vocabularies the `auth_failure` and `authz_denied` audit records already carry, so the metric and the record cannot disagree.

## Changes Made

- `internal/observability/metrics`: new `SecurityMetrics` type holding both counters, a `CountingAuthHook` decorator around the existing auth audit hook, and a once-only `Provider.SecurityMetrics()` accessor. The five `reason` series of `mcp_auth_failure_total` are seeded at zero so `increase()` fires on a process's first rejected token.
- `internal/tools`: `mcp_authz_denied_total` is incremented in both hop-1 deny branches of `withAuthorization`, using the same `decision_reason` constants as the WARN log and the audit record. A new `WithSecurityMetrics` manager option threads the recorder through `RegisterWithServer`.
- `cmd/server`: `buildSecurityMetrics` owns the gate. Both counters are on iff `OBS_AUTH_FAILURE_COUNTER_ENABLED` resolves true (derived from `OBS_METRICS_ENABLED` unless set explicitly) and a metrics provider exists. Forced on with metrics off logs a WARN and records nothing.
- `internal/observability/schema`: `metrics_schema` bumps to 1.2 for the two additive families, following the SOL-152093 precedent. The reason vocabulary is read from `schema.AuthFailureReasons()` (SOL-154163), so `metrics` stays free of `internal/auth`.
- `docs/observability.md` and `CHANGELOG.md`: both counters marked as emitted, the forced-on flag behaviour settled, and one `[Unreleased]` entry added.

## Testing

**Test Environment:**
- Go version: 1.26.5 (darwin/arm64)
- OS: macOS 26.6.2
- Broker version: N/A (no broker interaction)

**Tests Performed:**

**Test Coverage:**
- Every `reason` in the vocabulary increments only its own `mcp_auth_failure_total` series; all five are present at zero after registration.
- The hook decorator passes the identical reason to the counter and the delegated audit hook, and `Success` delegates without counting.
- Both authorization deny branches increment `mcp_authz_denied_total` with the right `reason` and `tool`; an allow records nothing; the family is absent until the first denial.
- Cross-signal test: for both deny branches, the counter's labels equal the `authz_denied` audit record's `reason` and `tool` from the same call.
- `RegisterWithServer` passes the recorder through to the wrapper, proven via a real in-memory SDK session.
- Flag matrix for `buildSecurityMetrics`: off with a provider, on with metrics off (WARN), on with a failed provider build (silent), on with a provider (samples reach `/metrics` and every real vocabulary value is seeded).
- Golden scrape fixture regenerated to pin both families and the schema bump.
- `make check` passes: build, vet, lint, race tests, 85% coverage gate.

## Breaking Changes

None. `metrics_schema` moves 1.1 → 1.2 as an additive minor bump.

### Code Quality

### Security
- [x] Credential-carrying types implement `slog.LogValuer` (none added)

### Testing
- [x] Error paths tested

### Documentation
- [ ] README.md updated (no user-facing change to the README surface)
- [x] docs/ updated
- [x] CHANGELOG.md updated
- [ ] Examples updated (not applicable)

### Git & CI
- [x] No merge conflicts
- [ ] CI checks passing (runs once the PR is opened)

## Additional Notes

- `reason="missing"` also counts credential-less requests of any method that reach `/mcp`, including CORS preflights and misrouted healthejects those with a 401 and the audit record already counts them. The docs state this so an `increase()` alert on `missing` is tuned withit in mind.
- The `tool` label is deliberately on the authorization counter and not on the auth-failure counter: authentication fails before any tool or broker is selected. Story 35 (SOL-152094) documents the contrast.
- `mcp_authz_denied_total` is not pre-seeded. Its tool dimension is only known at registration, and the series is documented as absent whoff.

## Related Issues/PRs

- Related to SOL-152097 (Story 23, shared `reason` vocabulary with the audit records)
- Related to SOL-154163 (moved the vocabulary into `internal/observability/schema`)
- Related to SOL-152093 (`metrics_schema` bump precedent)
- Blocks SOL-152092 (Story 37, reference Grafana dashboard)

---

**For Reviewers:**

**Focus Areas**: The decorator approach in `internal/observability/metrics/auth_hook.go` instead of wrapping `verifier.Verify` a second time, and the gate logic in `cmd/server/security_metrics.go`.

**Questions**: None.


solace-aross (dismissed):
Approving. Solid PR: both counters are gated symmetrically with the existing metrics flag, the reason/tool label values on `mcp_authz_denied_total` are the same Go constants the audit record and WARN log already use (no drift possible there), and the cross-signal test (`TestAuthzDenied_CounterAndAuditRecordAgreeOnReason`) and the flag matrix (`security_metrics_wiring_test.go`) both actually exercise what they claim.

One thing I want on record as a note, not a blocker (ran it past a second, adversarial pass before deciding that): `SecurityMetrics.RecordAuthFailure` (`instruments.go:299-304`, also flagged by Copilot) takes `reason string` and records it as a label with no validation against `schema.AuthFailureReasons()`. Confirmed by injection that an out-of-vocabulary reason mints a new series silently, no error. But the real issue isn't cardinality hygiene in the abstract, it's an asymmetry already in this diff: the *other* implementation of the same `AuthAuditHook.Failure` interface method, `audit.authHook.Failure`, rejects an unknown reason loudly (`NewEvent` → schema validation → ERROR log + `audit_drop` record). Two hooks fed the identical string from the identical call site disagree about what to do with garbage input, which is exactly the metric/audit disagreement this vocabulary was centralized to prevent, per `auth_failure_reason.go`'s own doc comment.

Not blocking because there's no live path today: `reportAuthFailure` is the only call site, typed to `schema.AuthFailureReason`, and every producer is `ClassifyAuthFailure` — a total function with a catch-all default, pinned by `TestClassifyAuthFailure_ReturnsOnlyClosedVocabulary`. So this is a boundary that's type-safe end to end right now, not a bug shipping today. I'd also push back gently on Copilot's specific fix (coerce to `invalid_token`): that constant has documented, load-bearing meaning (malformed JWT, issuer mismatch, static dev-token mismatch), so silently folding a future classifier bug into it would make a real failure mode indistinguishable from a genuine one, not safer.

If you want a cheap improvement now rather than deferring: `metrics` already imports `schema` (instruments.go:291), so typing `RecordAuthFailure(ctx, reason schema.AuthFailureReason)` costs nothing and moves the one unsafe `string` widening into `countingAuthHook.Failure`, where it's visible at the actual boundary instead of implicit in the recorder's own signature. Doesn't add validation, just makes the remaining gap honest. Your call whether that's worth doing here or as a follow-up alongside a real design discussion about a sixth vocabulary value.

bczoma (dismissed):
## PR Review: #401 — SOL-152099: emit SDK-independent auth-failure and authorization-denial counters

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-152099` → `main` | **Changes:** +784 / -56 across 21 files
**Jira:** `SOL-152099` — Emit SDK-independent auth-failure and authorization-denial counters (Story 24) — *Verifying*

---

### Verdict

Conditioned on one item, which is not anchorable to a diff line:

Not anchorable to this diff, so no thread — please treat as the one requested change:

- 🟡 **PR title is truncated mid-word and the remainder leaked into the PR body.** The title ends `…authorization-denia…`, and `…l counters` is the literal first line of the PR body. The commit itself (`379da5c`) is clean — subject and body are both intact — so this is purely a GitHub-side artefact. It still matters: this repo's `squash_merge_commit_title` is `COMMIT_OR_PR_TITLE`, and with two commits on the branch a squash resolves to the **PR title**, so `…denia…` would land on `main` permanently and flow into the release-notes extraction. Fix is an edit of the PR title plus deleting the stray leading `…l counters` line from the body — no amend or force-push needed.

---

### Summary

This adds `mcp_auth_failure_total{reason}` and `mcp_authz_denied_total{tool,reason}` so SRE alerting stops depending on the MCP SDK's log format, and it is unusually well built: the module boundary claim holds in the actual import graph, the gate degrades to a nil recorder on every failure mode with nil-receiver guards on both recorders, and `sync.Once` memoizes the accessor's value *and* its error. All nine Jira acceptance criteria are met, including AC 4's cross-signal requirement — `TestAuthzDenied_CounterAndAuditRecordAgreeOnReason` reads `reason` and `tool` out of the *emitted audit record* and asserts the scrape carries those extracted values, which is genuine coupling rather than two independent hardcoded assertions.

Two findings were filed at 🟡 and both were reduced by measurement. Worth stating explicitly, because it answers the question the author raised for reviewers: **the decorator placement is not merely an acceptable deviation from the ticket's work breakdown — it is strictly better than what the breakdown specified.** `hook.Failure` is called from exactly one place (`internal/auth/audit_emit.go:92`), reached from the five rejection sites in `internal/auth/middleware.go`, and `:150` (no/unparseable bearer token) sits *outside* `RequireBearerToken` because the SDK's own bearer-extraction reject never calls our verifier at all. Wrapping `verifier.Verify` as the ticket prescribed would therefore have missed `reason="missing"` — the most common rejection shape and the one the story exists to alert on.

**On Copilot's open comment** (`instruments.go:304`, unbounded `reason` label): the concern is sound in shape but **not reachable on any wired path**, so nothing further is needed to merge. `RecordAuthFailure` has one caller, `countingAuthHook.Failure`, reached only via `reportAuthFailure`, which is the single widening point from the typed `schema.AuthFailureReason` to `string`. All five call sites pass a `ClassifyAuthFailure` result, and that function is total — its `default:` arm returns `invalid_token`, and no branch propagates `err.Error()`, a claim value, an audience or a key ID into the return. Error text is only matched against, never emitted. Cardinality is hard-bounded at 5 series by the type system, exactly the seeded set in the golden fixture. What remains is type erasure at the module boundary — a latent invariant, not a live bug. Worth a reply on that thread recording the adjudication before it is resolved.

---

### Minor notes

- 🟢 `docs/observability.md:166` — the **Exposure** paragraph still calls the `/metrics` series "low-sensitivity" and enumerates build version, tool names and timing. This PR adds two *security* signals to that same unauthenticated, wildcard-bound listener: `mcp_auth_failure_total{reason="signature_invalid"}` is a readable key-rotation signal, and `mcp_authz_denied_total{tool}` tells an unauthenticated reader which tools RBAC is refusing. Add a sentence naming both and making listener restriction the recommended default rather than optional hardening.
- 🟢 `internal/observability/metrics/auth_hook.go:37` — no single test proves the whole composition `NewAuthMiddleware → reportAuthFailure → CountingAuthHook → mcp_auth_failure_total`; the guarantee is transitive across `auth_audit_test.go` (14 cases, middleware → hook), `security_test.go` (hook → counter) and compile-time interface satisfaction at `main.go:1384`. Correct, but reads as unproven, and the authz half of this same PR has `TestRegisterWithServer_PassesSecurityMetricsToAuthorization` for exactly this symmetry. One header-less-request case would document it. Filed at 🟡 initially and reduced: it is documentation-by-test, **not** a regression guard, since no existing-layer test catches a path that does not yet exist.
- 🟢 `internal/observability/metrics/auth_hook.go:41` — `CountingAuthHook(sm, nil)` with non-nil `sm` returns a decorator that nil-derefs on the first rejection. Unreachable today (`audit.NewAuthHook` always returns non-nil), but the doc comment covers only the nil-`sm` half of the matrix while `internal/auth/audit_emit.go` documents a nil hook as legal. Guard `next` or state that it must be non-nil.
- 🟢 Design-level (`cmd/server/main.go:1383`) — the hook covers every rejection *our* code makes, but not the four the SDK's `verify()` makes after the verifier returns (nil tokenInfo → 500, insufficient scope → 403, zero `Expiration` → 401, expiry re-check → 401), on which `Success` has already fired. I verified all four are genuinely unreachable here: `opts.Scopes` is never set, neither verifier can return `(nil, nil)`, and both always set a non-zero `Expiration`. The coverage argument rests on SDK internals nothing in this repo pins — a comment at the `RequireBearerToken` construction, or a test that fails if `Scopes` becomes non-empty, would make it durable.
- 🟢 `internal/tools/authorization.go:104` — `withAuthorization` now takes six positional parameters, and 16 call sites across five test files were mechanically edited to insert a bare `nil`. `"groups", false, nil` tells a reader nothing about which argument is which, and the next collaborator forces another repo-wide churn commit. Consider grouping the non-essential collaborators into a small `authzDeps` struct.
- 🟢 `docs/observability.md:295` — the bullet says `absent(mcp_auth_failure_total)` "means the counter was never wired", but absence has three causes, two deliberate (explicit `OBS_AUTH_FAILURE_COUNTER_ENABLED=false`, and forced-on-with-metrics-off). The next bullet does cross-link the flag, so the risk is modest, but an SRE reading only this section could page on-call for a wiring bug that does not exist.
- 🟢 Design-level — one flag now gates two independent capabilities. An operator suppressing authentication-failure noise silently also loses `mcp_authz_denied_total`, the more security-relevant of the two, from a flag whose name mentions only the other. Consider having the authz counter follow `OBS_METRICS_ENABLED` alone, or make the explicit-`false` case log a WARN naming both counters as the forced-on case already does.
- 🟢 `internal/observability/schema/version.go:36` — back-out is cheap in code but the `metrics_schema` 1.2 bump is not: versions are monotone and the schema is additive-only, so a revert leaves 1.2 advertising two absent families. Worth one line in the PR or CHANGELOG recording that a back-out keeps 1.2 and relies on the documented `absent()` semantics.

---

### Summary Verdict

|---|---|---|
| Documentation | ✅ | docs |

solace-aross (approved):
Confirmed the typed-reason fix and the new e2e coverage.

`RecordAuthFailure` is now `schema.AuthFailureReason`, with the one `string` widening pushed to `countingAuthHook.Failure` — exactly the narrowing I suggested, and the comment is honest that it's a type restatement, not validation (matches bczoma's independent read that this is a latent-invariant note, not a live bug — nothing further needed there).

Sabotage-verified the new `TestCountingAuthHook_WiredThroughAuthMiddleware`: dropped `metrics.CountingAuthHook` from the `NewAuthMiddleware` construction in a throwaway edit, confirmed the test fails cleanly on both the `missing` and `invalid_token` scrape assertions, reverted, green again. This closes bczoma's "no single test proves the whole composition" note with a real regression guard, not just documentation-by-test.

Spot-checked the rest of bczoma's list by reading rather than re-deriving: the `RequireBearerToken` construction now carries the durability comment he asked for (pinning that all four of the SDK's own post-verifier rejections are unreachable with this config); the Exposure paragraph in docs names both security counters and states listener restriction as the default posture, not optional; the `absent()` bullet lists all three causes instead of implying only one; CHANGELOG records the schema back-out posture (`1.2` stays on a revert, relies on `absent()` semantics). All accurate against the current code.

Full `go test` across the same four packages green. Approving.

bczoma (approved):
## PR Review: #401 — SOL-152099: emit SDK-independent auth-failure and authorization-denial counters

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-152099` → `main` | Follow-up review at `105366aa`
**Jira:** `SOL-152099` — Story 24

---

### Verdict

**Recommendation:** ✅ Approve

The one item my previous review conditioned on — the truncated PR title — is fixed, and the incremental commit closes five of the eight minor notes besides. Nothing outstanding blocks or conditions this. The 10 🟢 notes below are **not** conditions of the verdict.

### Summary

One commit since my last pass, and it is the right response to the review it answers. `SecurityMetrics.RecordAuthFailure` is now typed to `schema.AuthFailureReason`, which collapses the string-widening sites to exactly one and puts it at the SDK's `AuthHook` interface boundary, where the erasure actually originates — rather than bolting a runtime allow-list onto the recording site. That is the correct answer to Copilot's cardinality thread, and it is worth saying plainly that the bound was never actually breachable: I re-traced every producer independently at this head, and `auth.ClassifyAuthFailure` is total, its `default:` arm returns `invalid_token`, and no branch propagates `err.Error()`, a claim, an audience or a key ID into the label. The typing makes an existing invariant legible rather than fixing a live bug.

`TestCountingAuthHook_WiredThroughAuthMiddleware` is the composition test my last review asked for, and it picks the one input the layer-by-layer tests structurally cannot reach — the header-less request, which the SDK never routes to a verifier at all — while also asserting the seeded `expired 0` stays put. It is a genuine mutation guard: remove `CountingAuthHook` from the composition and both `1` assertions fail against the seeded zeros.

Two findings were filed at 🟡 in this pass and both were reduced by measurement rather than by argument, which is worth stating because both were mine to check. One claimed the *commit* subject carried the same truncation as the PR title; reading `379da5c` directly shows the subject is complete and clean, so it is dropped. The other claimed the new comment on `countingAuthHook.Failure` falsely assures a reader that something validates the reason; the analogy in it is indeed factually wrong — `audit.authHook.Failure` performs no boundary step, and the vocabulary check lives in `NewEvent`, which *rejects* an off-vocabulary value where this path would silently mint a series — but the very same sentence already says "a restatement of type rather than a validation" and names the real invariant. So it is a wrong clause in a comment, not a false assurance, and it sits at 🟢 below.

All checks pass, including build, lint and the five E2E suites.

*9 prior findings settled — 6 addressed, 2 accepted as-is.* The exception, plus the one conditioning item, below.

|---|---|---|---|
| 🟢 | The hook covers no rejection the SDK's own `verify()` makes after our verifier returns (nil `TokenInfo`, missing scope, zero `Expiration`, expired `Expiration`), and the coverage argument rests on SDK internals nothing pins | Partially addressed | `internal/auth/middleware.go:79-85` — all four named, each with why it is unreachable here and the "would under-count" consequence. I re-verified the claim at this head: `RequireBearerTokenOptions` is built with `ResourceMetadataURL` only, so `Scopes` is empty; neither verifier can return `(nil, nil)`; the static path sets `now+24h` and the OIDC path `idToken.Expiry`. Correct — but still a comment, not a pin. See the first two 🟢 notes. |
| 🟡 | PR title truncated mid-word (`…authorization-denia…`) with `…l counters` leaked as the PR body's first line, which a squash would land on `main` | Addressed | Title is now complete; the stray fragment is gone from the body. Commit `379da5c`'s own subject was and remains clean. |

---

### Minor notes

- 🟢 `internal/auth/middleware.go:79-85` — the strongest clause of the invariant is the cheap one to enforce: a three-line test constructing the same `RequireBearerTokenOptions` and failing if `Scopes` is non-empty turns prose into a gate. The SDK half is fairly left as prose, but `AllowMissingExpiration` and `ClockSkew` are recent `RequireBearerTokenOptions` additions, so the SDK's reject set is a moving target and Dependabot bumps it unattended. Under-counting a security counter is invisible by construction — the graph just stays flat, and `absent()` does not fire either.
- 🟢 `internal/auth/middleware.go:81-84` — the expired-`Expiration` re-check is not strictly unreachable: go-oidc checks `exp` inside `Verify`, our verifier fires `Success`, and the SDK re-checks with `ClockSkew` 0 microseconds later. A token expiring inside that window yields an `auth_success` record plus a 401 with no `auth_failure` record. Vanishingly rare and self-correcting on retry, but it is the one case where the record disagrees with the response, and the comment reads as absolute. Soften the wording, or set a small non-zero `ClockSkew`.
- 🟢 `internal/observability/metrics/instruments.go:310` — `RecordAuthzDenied(ctx, tool, reason string)` is left fully untyped beside its now-typed sibling, though its `reason` is also a closed two-value vocabulary. Two adjacent same-typed params are a silent swap hazard, and the next person adding a security counter has no rule to follow. Either lift the two `decision_reason` constants into `schema`, or add one line saying why this one stays a string.
- 🟢 `internal/observability/metrics/auth_hook.go:55-58` — the clause "the same untrusted-boundary step `audit.authHook.Failure` performs" is inaccurate: that function passes `reason` through untouched into a plain-`string` `Fields.Reason`, and the vocabulary check lives in `NewEvent`, which rejects an off-vocabulary value with a drop notice where this path would mint a series. The rest of the comment is correct, so nobody is falsely assured — but the analogy invites a maintainer to assume symmetric behaviour where the two paths behave oppositely. Delete the analogy and state the asymmetry.
- 🟢 `internal/observability/metrics/auth_hook.go:60` with `internal/auth/audit_emit.go:88` — the round trip is self-inflicted rather than imposed: `AuthAuditHook` is this repo's own interface, and only `*sdkauth.TokenInfo` on `Success` comes from the SDK. Widening `AuthAuditHook.Failure`'s `reason` to `schema.AuthFailureReason` would delete both conversions with no cycle risk (`schema` is a leaf). The interface's doc gives two reasons not to, and one of them — "would move an interface SOL-152099 is also going to implement" — has now expired. Fine to defer; worth recording that this commit relocated the erasure rather than removed it.
- 🟢 `CHANGELOG.md:38` — the sibling SOL-152097 entry in the same `[Unreleased]` block still forward-references this work as unlanded ("when SOL-152099's counter lands (a colleague's still-`To Do` story)"). Both ship in the same release, so one paragraph describes as future what the next describes as shipped.
- 🟢 PR description, "Additional Notes" — two sentences lost characters the same way the title did: `…including CORS preflights and misrouted health` **`ejects`** `those with a 401…` (a clause is missing) and `…documented as absent wh` **`off`**. These are the only place the `reason="missing"` alert-tuning caveat is explained. Description-only; nothing lands on `main` under this repo's squash settings.
- 🟢 `cmd/server/security_metrics_wiring_test.go:115` — the comment opens "The whole authentication composition, as main wires it", but the test builds that composition itself; deleting `metrics.CountingAuthHook` from `main.go`'s wiring leaves this test green. The authz half has `TestRegisterWithServer_PassesSecurityMetricsToAuthorization`, which does exercise production wiring, so the asymmetry is real. Reword to "the composition main wires", or extract main's hook construction into a helper the test calls.

*and 2 further minor items — the middleware `cfg` in the new test carries inert `Port`/`Observability` fields `NewAuthMiddleware` never reads, and its scrape assertions use `strings.Contains` on `… 1`, which also matches a rendered `10`. Ask if you want them written up.*

---

### Summary Verdict

|---|---|---|
| Documentation | ✅ | docs |
| PR Scope & Description | ⚠️ | good:pr — rests solely on the two corrupted description sentences, which are 🟢 and condition nothing |

PR #389, opened 2026-09-08
SOL-153332: emit a broker_authz_denied audit event and counter for hop-2 authorization denials

solace-lootii:
## Summary

A broker-side (SEMP) permission denial now gets recorded as its own audit event and counted separately from an MCP-server-side authorization denial, so a compliance reviewer or SRE can tell hop 1 ("the agent wasn't allowed to call this tool") apart from hop 2 ("the broker rejected the credentials we presented") without inferring it from a generic SEMP error.

Classified at the tools layer (`internal/tools/manager.go`), from signals the SEMP clients already expose — SEMPv1's `ErrorKindPermission` and SEMPv2's error code 72 — so no new broker-response parsing is needed.

On a match:
- `audit_event_type=broker_authz_denied` (`tool`, `broker`, `reason=permission_denied`)
- `mcp_broker_authz_denied_total{tool,broker,reason}` increments
- `error_type=broker_permission_denied` (a new, 13th value) is set on the coexisting Story 22 `operation` record for a destructive call

The two records coexist rather than one suppressing the other, because `Handle` had already run by the time the broker refused. A non-destructive call produces only the `broker_authz_denied` record, since it never reaches the destructive gate that produces an `operation` record at all.

Both new signals are gated by the existing `OBS_AUDIT_LOG_ENABLED` / `OBS_METRICS_ENABLED` flags and are byte-for-byte inert when off.

## Acceptance criteria (from the ticket)

- [x] New `audit_event_type` value `broker_authz_denied`
- [x] New counter `mcp_broker_authz_denied_total{tool,broker,reason}`
- [x] New `error_type` value `broker_permission_denied`, set on the coexisting operation record
- [x] Classified at the tools layer, not the protocol client layer
- [x] Test asserting both records for a destructive call, only the denial record for non-destructive
- [x] Story 21/23 closed-set text — already updated ahead of this PR per the ticket's alignment-audit note
- [x] Story 35 documentation (`docs/observability.md`) — updated in this PR
- [x] Story 22's "exactly one audit event per destructive call" scoped to the `operation` record type — already updated ahead of this PR; this PR's own test (`TestBrokerAuthzDenial_DestructiveCall_EmitsBothRecords`) asserts the `operation`-typed count rather than a total

## Testing

- `go build ./...`, `go vet ./...` clean
- `golangci-lint run` clean on touched packages
[...]

PR #402, opened 2026-09-09
SOL-152088: Dashboard broker reachability with classified failure reasons, without ejecting healthy MCP pods

solace-wmaryam:
## Summary

Adds passive per-broker reachability tracking to `/metrics`. The tracker
piggybacks on real SEMP calls — no heartbeat goroutine, no new probe path.
Two Prometheus gauges are exposed: `mcp_broker_reachable{broker}` and
`mcp_broker_unreachable_reason{broker,reason}`. The signal is deliberately
decoupled from `/readyz` (ADR-004 / ISSUE-026).

## Type of Change

## Motivation and Context

Operators have no way to observe broker reachability today. A broker outage
or bad credentials is invisible until an LLM tool call fails. This change
makes broker health dashboardable and alertable without affecting pod
readiness.

## Changes Made

- `internal/observability/health/tracker.go` — `BrokerTracker`,
  `RecordBrokerResult`, `Snapshot`, `Classify`. States: `reachable`,
  `unreachable`, `credential_invalid`, `broker_error_NNN`.
- `internal/semp/resilience/sender.go` — `ResultHook` type,
  `WithResultHook` option, named-return defer in `Do` that fires the hook
  after every real call, filtering admission failures and context
  cancellations.
- `internal/observability/metrics/broker.go` — `BrokerMetrics` with two
  OTel observable gauges backed by a scrape-time `brokerStates` callback.
- `internal/observability/metrics/provider.go` — `BrokerMetrics` lazy-init
  method following the `SEMPMetrics`/`ToolMetrics` pattern.
- `internal/semp/pool.go` — `WithBrokerResultHook` pool option, wired into
  `getOrCreate` alongside `WithMetrics`.
- `cmd/server/main.go` — tracker always created; gauges registered when
  metrics are enabled; hook passed to pool.

## Testing

Unit tests cover: `Classify` across all HTTP status codes, state
transitions, absent-until-first-call, snapshot copy safety, gauge
exposition for reachable/unreachable/credential_invalid, and the `/readyz`
decoupling assertion.

Coverage: 89.0% (gate: 85%). All tests pass with `-race`.

## Related Issues/PRs

- Closes SOL-152088
- Epic: SOL-150253


internal/observability/health/tracker.go
```diff
@@ -0,0 +1,90 @@
...
+// Classify maps an HTTP status code to a BrokerState. 0 means no HTTP response (transport failure).
+func Classify(httpStatus int) BrokerState {
+	switch {
+	case httpStatus == 0:
+		return StateUnreachable
+	case httpStatus == 401:
+		return StateCredentialInvalid
+	case httpStatus >= 200 && httpStatus < 300:
+		return StateReachable
+	default:
```
bczoma: 🟡 **A caller-induced 4xx marks a healthy broker unreachable**

Flagged by good:security. Measured under challenge.

`Classify` sends every non-2xx, non-401 status to `brokerErrorState`, so an ordinary 404/400/409 becomes `broker_error_NNN` — and the gauge callback maps anything that is not `reachable` to `0`. A 404 is positive proof the broker answered.

**This is routine, not a corner case.** `checkRetry` implements "all other status codes (4xx): no retry" (`retry.go:207`), so with `shouldRetry == false` and no error, go-retryablehttp returns `(resp, nil)` and never enters `ErrorHandler` — no `RetriesExhaustedError` is built, the defer's drop branch is not taken, and `d.resultHook(alias, 404)` fires. Your own `TestSender_NoRetry_404` pins the single-attempt behaviour. The reachable path is deliberate: `list-queue-subscriptions` runs a preflight `monitor/getMsgVpnQueue` on the LLM-supplied `queueName` *precisely so* a nonexistent queue errors rather than returning an empty list (`tools.yaml:1201-1235`), so every wrong-name guess is an intentional 404. No tool-layer existence check precedes it.

**What it costs.** Two things came out better than first feared, and are worth stating: the `reason` label does distinguish `broker_error_404` from a true `unreachable`, so a real outage is not masked; and the next successful call on that alias overwrites the state, so it self-heals with no human action. The residue is still real — state is last-write-wins with no TTL and no heartbeat (`docs/observability.md:267`: "not a heartbeat"), so a 404 that happens to be the *terminal* call before an idle period leaves `mcp_broker_reachable{broker}` at 0 indefinitely. That satisfies a `for: 5m` rule and pages on a healthy broker. Given the workload shape this server has — sporadic, interactive, LLM-driven calls where an agent probes a guessed name and the session then goes quiet — that is a plausible Tuesday, not a contrived one.

**Recommended fix.** Classify 4xx as `reachable`: the broker responded, so reachability is proven. Keep 401 → `credential_invalid`. Give 403 its own reason rather than `broker_error_403` — it is an access-control misconfiguration owned by the same team as a 401, not by the broker-ops team a `broker_error_*` alert pages, and it should not zero the reachability gauge. Restrict `broker_error_NNN` to 5xx, and optionally 429. Add a Sender-level test driving a non-retryable 404 and asserting `reachable=1`, then update the `reason` vocabulary at `docs/observability.md:270-276` to match the narrowed set.

docs/observability.md
```diff
@@ -267,18 +267,16 @@ broker state, not per attempt.
...
-- Alert on `mcp_broker_reachable == 0` and use this gauge only to attribute the cause;
-  `mcp_broker_unreachable_reason == 1` is deliberately redundant with it rather than a
-  second, separately-timed source of truth.
-
-**Cardinality:** `|broker|` for the first metric; `|broker| x |reason|` for the second.
+- `reason` vocabulary: `credential_invalid` (HTTP 401), `unreachable` (connection refused,
+  DNS failure, or I/O timeout), `broker_error_NNN` (any other non-2xx, where NNN is the
+  HTTP status code, e.g. `broker_error_503`). The `broker_error_NNN` values bound
+  cardinality to the finite set of HTTP status codes.
+- `mcp_broker_unreachable_reason` emits `1` for the broker's current failure reason only.
```
bczoma: 🟡 **The one-hot contract was withdrawn to match the code, rather than the code fixed to match it**

Flagged by good:pr, corroborated independently while challenging a separate cardinality finding.

This hunk does not document new behaviour — it **removes a guarantee that is already on `main`**. The text being deleted (landed by SOL-152100, "Draft observability schema doc for pilot review") promised more than a closed reason set:

> `mcp_broker_unreachable_reason` is **one-hot per broker**: at most one `reason` series per broker is `1` at any moment, and every other `reason` for that broker is explicitly `0`. All reasons for a broker are published once it has been seen, so a series never disappears mid-incident and `max by (reason)` cannot straddle two causes.

It is replaced with "emits `1` for the broker's current failure reason only … the series is no longer emitted and disappears from the next scrape" — the exact behaviour the old paragraph promised would not happen. The deleted lines also promised the two gauges were "deliberately redundant with [each other] rather than a second, separately-timed source of truth", which is the guarantee thread 3 is about.

**Why this needs a decision rather than a doc edit.** A pilot-reviewed operator contract is being revoked, and the PR body's "Changes Made" lists six code files and no docs at all, so a reviewer skimming it would never know. Anyone who built the documented `max by (reason)` attribution query, or a panel keyed on the explicit `0`, gets *no data* rather than `0` — a silently empty panel, not an error. This is also the same underlying issue as @Copilot's `broker.go:60` thread, resolved in the opposite direction: it asked for the code to publish zeros for previously-seen reasons; this hunk instead deletes the promise that it would.

Either direction is defensible. Emit-current-reason-only is genuinely simpler and Prometheus staleness handles the disappearance sanely. But it is a behaviour change to a published contract and should be landed as one.

**Recommended fix.** Pick one and say so explicitly:
- Restore the contract: retain the seen-reason set per broker and observe `0` for the non-current ones (this closes @Copilot's thread too); or
- Withdraw it deliberately: keep this hunk, add a "Changes Made" bullet stating that SOL-152100's one-hot guarantee is being withdrawn and why, and put it in the CHANGELOG as a behaviour change rather than an addition.

Either way, two things still need fixing in this file: the metric table two lines above still types `mcp_broker_unreachable_reason` as `Gauge (1/0)`, which now contradicts the prose in this very hunk; and the absent-until-first-call behaviour is undocumented — a broker that has been down since pod start and taken no traffic appears in *neither* gauge, so `mcp_broker_reachable == 0` never fires and the dashboard reads green. That one deserves a line recommending `absent(mcp_broker_reachable{...})` alongside the reachability alert.

internal/observability/metrics/broker.go
```diff
@@ -0,0 +1,70 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("register mcp_broker_reachable: %w", err)
+	}
+
+	_, err = meter.Float64ObservableGauge(
+		"mcp.broker.unreachable.reason",
+		metric.WithDescription("1 when the broker is not reachable, labelled with the failure reason."),
+		metric.WithFloat64Callback(func(_ context.Context, o metric.Float64Observer) error {
+			for broker, state := range brokerStates() {
```
bczoma: 🟡 **One scrape reads the tracker twice, so the two gauges can contradict each other**

Flagged by good:architecture and good:code independently.

Each of the two observable gauges installs its own callback and each calls `brokerStates()` separately, so a single scrape takes two snapshots at two different instants. A `RecordBrokerResult` landing between them yields a scrape where `mcp_broker_reachable{b}` is `1` **and** `mcp_broker_unreachable_reason{b,reason=...}` is `1` for the same broker — or, in the other direction, where a broker with state appears in neither.

This is distinct from @Copilot's `broker.go:60` thread, which is about never zeroing previously-seen reasons. This one is about the two metrics disagreeing *within one scrape*, and it is precisely the property the pre-PR doc guaranteed against: the paragraph this PR deletes said the two gauges were "deliberately redundant with [each other] rather than a second, separately-timed source of truth". As written they are exactly a second, separately-timed source of truth.

The window is narrow — two map copies under an `RLock` — so this is not a frequent event. It matters because the documented operator workflow is to alert on `mcp_broker_reachable == 0` and then *join* to the reason gauge to attribute the cause, and a joined query that hits a torn scrape shows a row that cannot be true.

**Recommended fix.** Register both instruments through one callback and snapshot once:

```go
gReachable, err := meter.Float64ObservableGauge("mcp.broker.reachable", ...)
gReason, err := meter.Float64ObservableGauge("mcp.broker.unreachable.reason", ...)
_, err = meter.RegisterCallback(func(_ context.Context, o metric.Observer) error {
    states := brokerStates()   // exactly one snapshot per scrape
    // observe both instruments from `states`
    return nil
}, gReachable, gReason)
```

That also halves the per-scrape map allocation and `RLock` count, which is the 🟢 version of the same observation.

internal/observability/health/tracker.go
```diff
@@ -0,0 +1,90 @@
...
+	StateCredentialInvalid BrokerState = "credential_invalid" // #nosec G101 -- metric label value, not a credential
+)
+
+// brokerErrorState returns the broker_error_NNN label value for non-2xx, non-401 responses.
+func brokerErrorState(httpStatus int) BrokerState {
+	return BrokerState(fmt.Sprintf("broker_error_%d", httpStatus))
+}
+
+// Classify maps an HTTP status code to a BrokerState. 0 means no HTTP response (transport failure).
+func Classify(httpStatus int) BrokerState {
```
bczoma, line 41: 🟡 **AC 5 asks for the classifier in `health`; the half that can actually be wrong lives in `Sender.Do`**

Flagged by good:architecture and good:code independently.

AC 5 reads: "Classification is centralized in `health.Classify(err)` and unit-tested across transport-error and HTTP-error variants." As landed, `Classify` takes an `int` and maps it to a label — the easy, total half. The half that can be wrong is the `error` → status extraction, and it sits inline in the hook's defer in `sender.go:756-779`: unwrapping `RetriesExhaustedError`, deciding that `BrokerBusyError` and context cancellations are not broker signals, and — critically — silently returning on "unknown error type".

Two consequences follow, and the second is the one that will bite later.

**The tested half is not the risky half.** Every current test drives `Classify` with an `int`, so a transport error is only ever the literal `0` in a table test. The real transport paths — connection refused, DNS failure, I/O timeout — are unexercised end to end. That is exactly how @Copilot's `sender.go:764` finding survived to review: with `http.Client.Timeout` set on the production SEMP clients, an I/O timeout wraps `context.DeadlineExceeded` and is filtered out *above* the status extraction, so the single most common real outage never reaches `Classify` at all. AC 4 names I/O timeout as an `unreachable` case, so that is an AC miss, not just a code smell — and no test at the `int` boundary can catch it.

**The failure mode is silence, not error.** Because the `else` branch drops any error shape it does not recognise, a new error type added in `retry.go` later produces *no hook call*, so reachability silently goes blind for that path rather than misclassifying. A blind gauge looks identical to a healthy one, and there is no test that can fail.

**Recommended fix.** Make the port report facts and let `health` render the verdict:

```go
type ResultHook func(brokerAlias string, resp *http.Response, err error)
```

Fire it unconditionally, and move every "is this a broker signal" decision into `health` alongside `Classify`, where it can be table-tested with real `*url.Error`, `net.Error` and `RetriesExhaustedError` values. That satisfies AC 5 as written and makes @Copilot's timeout case a test you can point at.

One thing worth preserving deliberately if you do this: the current `int`-only signature is why no error string, URL, hostname or TLS detail can ever reach the `reason` label. That is a real property and the security lens called it the best design choice in the PR. Keep the classifier's *output* vocabulary constructed solely from server-side constants and a numeric status — take the `error` in, never let its text out.

internal/observability/health/tracker.go
```diff
@@ -0,0 +1,90 @@
...
+	states map[string]BrokerState
+}
+
+// NewBrokerTracker returns a BrokerTracker with no recorded state.
+func NewBrokerTracker() *BrokerTracker {
+	return &BrokerTracker{states: make(map[string]BrokerState)}
+}
+
+// RecordBrokerResult updates the tracker. httpStatus is 0 when no HTTP response was received.
+func (t *BrokerTracker) RecordBrokerResult(broker string, httpStatus int) {
```
bczoma, line 88: 🟡 **No freshness dimension, so a stale "reachable" is indistinguishable from a live one**

Flagged by good:architecture.

`RecordBrokerResult` is last-write-wins with no timestamp and no TTL, and there is deliberately no prober (correctly — the zero-added-broker-load NFR forbids one). The consequence is that "when" is part of this signal's meaning and is not exposed anywhere:

- A broker whose last call succeeded an hour ago reports `reachable` exactly like one that succeeded a second ago.
- A broker that is down but currently unused keeps reporting `reachable` until something calls it.
- A broker never called since pod start appears in **neither** gauge, so `mcp_broker_reachable == 0` cannot fire for it and the dashboard reads green.

Passive tracking is the right call, but it makes this a dashboard that is confident about brokers it has not checked. In a multi-broker deployment most aliases are touched rarely, which is precisely the population an operator is least likely to notice going quiet — and there is no way for an alert rule to tell "green" from "unknown".

**Recommended fix.** Store a `time.Time` per entry and expose a third observable gauge in the same callback:

```
mcp_broker_last_result_timestamp_seconds{broker}
```

Additive, still zero broker load, one more field on the tracker entry, and it lets an alert gate on age (`time() - mcp_broker_last_result_timestamp_seconds{...} > 900`) instead of trusting a possibly-ancient `1`. Pair it with the `absent()` note in thread 2 and the two together cover both shapes of "we do not actually know".

bczoma (dismissed):
## PR Review: #402 — SOL-152088: Dashboard broker reachability with classified failure reasons, without ejecting healthy MCP pods

**Author:** solace-wmaryam | **Branch:** `wmaryam/SOL-152095_go_runtime_process_metrics` → `main` | **Changes:** +542 / -18 across 11 files
**Jira:** `SOL-152088` — Dashboard broker reachability with classified failure reasons, without ejecting healthy MCP pods (Story 9) (Verifying)

---

### Verdict

1. 🟡 `internal/observability/health/tracker.go:45` — classify 4xx as `reachable`; a caller-induced 404 currently reads as a broker outage
2. 🟡 `docs/observability.md:274` — the one-hot contract was withdrawn to match the code; decide that deliberately rather than as a doc correction
3. 🟡 `internal/observability/metrics/broker.go:53` — snapshot the tracker once per scrape, not once per gauge
4. 🟡 `internal/observability/health/tracker.go:37` — AC 5 asks for the classifier in `health`; the half that can be wrong lives in `Sender.Do`
5. 🟡 `internal/observability/health/tracker.go:63` — no freshness dimension, so a never-called broker and a healthy one are indistinguishable

- 🟡 `CHANGELOG.md` — no `[Unreleased]` entry for two new public metric families and the `metrics_schema` 1.2→1.3 bump. Both prior metrics stories documented theirs (SOL-152093 for 1.1, SOL-152099 for 1.2). The green "CHANGELOG updated" check is satisfied by pre-existing unrelated entries, so it is not evidence here.
- 🟡 `docs/observability.md:99` — Schema Versioning still says `metrics_schema` (current: **1.2**) while `schema/version.go:39` now ships `1.3`. The same page tells operators to pin dashboards to `mcp_schema_version`, so the one number they are told to pin against is the one that is wrong, and there is no record of what 1.3 added.

@Copilot's four threads also stand at this head and are not restated here. Two notes on them: its docs-inventory finding is **correct** — lines 49 and 113-125 both still end "assume any other metric below is not yet emitted" and name neither gauge — and its one-hot finding is the code half of thread 2 below. Its `http.Client.Timeout` finding is the one I would fix first among them: AC 4 names I/O timeout as a `unreachable` case, and it is currently filtered out before it can be classified.

---

### Summary

Passive reachability tracking is the right design for this story and the hard parts are right: `Sender.Do` is genuinely the single funnel both SEMP clients reach, and hooking it *above* `retryClient.Do` means one logical tool call fires the hook exactly once no matter how many attempts it made — no retry amplification, no flapping. Admission shedding is correctly excluded, so this process's own back-pressure never reads as a broker outage, and the `/readyz` decoupling holds. The tracker map is keyed by the config-resolved alias, so a caller cannot mint a series.

What holds the verdict is that the *classifier* is broader than the signal it feeds. Every non-2xx that is not a 401 becomes a broker fault, so a 404 — which is positive proof the broker answered — drives `mcp_broker_reachable` to 0.

Three findings were filed at higher severity and all three were reduced by measurement. A 🔴 claiming caller-induced 4xx "masks a real outage, indistinguishably" was downgraded: the `reason` label does distinguish `broker_error_404` from `unreachable`, and any successful call heals the state. A panic-unwind path that records `unreachable` was reduced to 🟢 once it turned out to need a separate pre-existing bug and to fire `mcp_panic_recovered_total` with a stack trace at the same instant. A cardinality/DoS reading of `broker_error_NNN`, filed independently by three lenses, collapsed entirely: the tracker holds one overwritten state per broker, so the process exposes at most one series per broker, and `docs/observability.md:92` already blesses HTTP status codes as a finite label domain with `http_response_status_code` shipping on `mcp_semp_request_total`. What survived from it is a spec question, in thread 2.

---

### Minor notes

- 🟢 `internal/semp/resilience/sender.go:766` — on a panic unwind the named returns are both nil, so `status` stays 0 and the hook records `unreachable` for a local fault. Reaching it needs a pre-existing nil-deref bug, and the panic counter fires alongside, so it is hygiene — but add `if resp == nil && err == nil { return }`, matching the explicit panic-unwind comments already at `sender.go:821` and `sempv2/client.go:203`.
- 🟢 `internal/observability/health/tracker.go:32` — `broker_error_NNN` puts the status in the `reason` value where AC 4 asks for `reason="broker_error"` plus a separate bounded label. No runtime risk (see Summary), but either change it or get the AC amended, since the story is in Verifying.
- 🟢 `internal/observability/metrics/broker.go:37,54` — both callbacks compare against the bare literal `"reachable"` rather than `health.StateReachable`, and `SnapshotStrings` exists only to avoid the import. `metrics` importing `health` introduces no cycle, so the typed boundary is available; today the test hardcodes the same literal, so test and code share any typo.
- 🟢 `internal/observability/metrics/provider.go:231` — `BrokerMetrics(brokerStates)` takes an argument but guards with `sync.Once`, so a second call with a different callback silently returns the object bound to the first. The sibling lazy-inits take no arguments; document the first-caller-wins contract or return an error.
- 🟢 `internal/observability/health/tracker_test.go:154` — `TestReadyzDecoupling` constructs a tracker and a `ReadinessState` that are unrelated by construction, so it cannot fail for any change to the tracker. AC 6 asks for a verified decoupling; assert the wiring a regression would have to go through, or make it a comment on `ReadinessState` rather than a test.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

solace-awaheed (commented):
Addendum to the threads from @Copilot and @bczoma — not restating any of them. I
independently reached the same conclusions on the timeout filter and the 4xx
classification; both of those threads stand as written. Two small things I didn't
see covered anywhere, plus one confirmation.

### 🟡 `internal/observability/metrics/broker.go:37,54` — the gauge compares against a string literal, not the `health` constant

Both callbacks test `if state == "reachable"`. `internal/observability/metrics` does
not import `internal/observability/health`, and `broker_test.go:33,57` uses the same
literal, so nothing links the producer's constant to the consumer's comparison.

Renaming `health.StateReachable` (tracker.go:26) — plausibly as part of the thread-4
refactor that moves classification into `health` — would make every broker report
`mcp_broker_reachable 0` and emit a bogus `unreachable_reason` series, with the
whole suite still green. The `map[string]string` boundary is what erases the type;
if `SnapshotStrings` stays, importing the constant for the comparison restores the
compile-time link at zero cost.

Related, minor: `BrokerTracker.Snapshot()` (tracker.go:71) and the `BrokerState`
type are reachable only from `tracker_test.go` — production goes exclusively through
`SnapshotStrings`. Two accessors for one field, one of which only tests use.

### 🟢 `cmd/server/main.go:1170,1215` — the tracker records on every SEMP call even with metrics off

`brokerTracker` is constructed and wired via `WithBrokerResultHook` unconditionally,
outside the `if metrics.Enabled(...)` block. With `OBS_METRICS_ENABLED=false` that's
a mutex write and a map store per SEMP call feeding a map nothing ever reads.

The cost is negligible and the comment on line 1169 shows it's deliberate, so this
isn't a fix request — flagging only because it's a departure from the door-closing
gating the other metrics stories followed ("when off, no instruments are
registered", SOL-152091/152086), and the next person reading it may not know which
of the two patterns is the house rule.

### Confirmation on the two open blockers

I reproduced both against a real `Sender` with a hook installed, in case it helps
close them out:

    hung broker (retryBudget 100ms, no http.Client.Timeout):
      err=request failed after 1 attempts: context deadline exceeded
      hookCalled=false                      <- @Copilot sender.go:764

    404: err=<nil>  hookCalled=true status=404    <- @bczoma tracker.go:45
    403: err=<nil>  hookCalled=true status=403
    connection refused: hookCalled=true status=0  <- classifies correctly today

Worth noting for the timeout fix: the hung-broker probe used `httptest`'s client,
which sets no `http.Client.Timeout`. The retry-budget `context.WithTimeout` at
`sender.go:798` triggers the filter on its own, so a fix scoped only to the client
timeout won't close it — the comparison has to be against the caller's context, as
that thread already recommends.

internal/observability/health/tracker.go
solace-wmaryam: Fixed. 4xx (except 401 and 403) now classify as `reachable` — the broker answered, so connectivity is proven. 401 and 403 both map to `credential_invalid`. `broker_error_NNN` is restricted to 5xx and 429. Updated `Classify` in `tracker.go` and the `reason` vocabulary in `docs/observability.md`.

docs/observability.md
solace-wmaryam: Fixed by restoring the contract rather than withdrawing it. `BrokerTracker` now retains all non-reachable states ever seen per broker in a `seenReasons` set. The gauge callback emits `1` for the current reason and `0` for all prior ones, so series never disappear mid-incident. Added a `TestBrokerTracker_OneHotSeenReasons` test covering the transition case.

internal/observability/metrics/broker.go
solace-wmaryam: Fixed. Both gauges now share a single `meter.RegisterCallback` that calls `brokerStates()` once per scrape. This also resolves the Copilot thread on this line.

internal/observability/health/tracker.go
solace-wmaryam: Fixed the I/O timeout gap. The root issue is that `Sender.Do` reassigns `ctx` to a retry-budget inner context, so by the time the defer runs, `ctx` refers to the inner context — not the caller's. A bare `context.DeadlineExceeded` from the retry budget was indistinguishable from a caller cancellation.

Fix: capture `callerCtx := ctx` before any reassignment. In the defer, `RetriesExhaustedError` is still checked first (covers `http.Client.Timeout`). For a bare `context.DeadlineExceeded`, we now check `callerCtx.Err() != nil`: if the caller's context is expired, skip; otherwise it's an internal timeout and we fire the hook with status 0 → `unreachable`.

internal/observability/health/tracker.go
solace-wmaryam: Fixed. Added `lastResult time.Time` per broker entry, exposed as `mcp_broker_last_result_timestamp_seconds{broker}` (Unix seconds). All three gauges share the single `RegisterCallback` introduced for thread 3. Docs updated with the recommended `time() - mcp_broker_last_result_timestamp_seconds > 900` alert and an `absent()` note for brokers never called since pod start.

solace-wmaryam:
@solace-awaheed thanks for the confirmation and the `httptest` reproduction — that pinned the retry-budget case precisely.

**String literal** — already resolved in an earlier pass; the metrics package now imports `health` directly and compares against `health.StateReachable`.

**Retry budget timeout** — fixed by capturing `callerCtx := ctx` before the retry-budget reassignment. The `RetriesExhaustedError` branch handles `http.Client.Timeout` as before; the `context.DeadlineExceeded` branch now checks `callerCtx.Err() != nil` to distinguish a caller cancellation (skip) from an internal timeout (fire hook with status 0 → `unreachable`).

**Tracker with metrics off** — deliberate, as the comment on that line notes. Noted for future reference when the door-closing pattern comes up again.

solace-aross (changes requested):
Reviewed at 2ff950c4. All 5 of bczoma's conditions got real attention and I verified three by sabotage: reverting the 4xx classification fix fails `TestClassify` cleanly, and reverting the one-hot `seenReasons` retention fails `TestBrokerTracker_OneHotSeenReasons` cleanly. Both are real fixes with real regression coverage.

One condition isn't resolved the way it reads, and it's the one that matters most for this feature.

**Blocking: the entire `ResultHook` defer in `sender.go` has zero test coverage.** This is the single production feed into the whole reachability signal (extraction of HTTP status from the call outcome: success, `RetriesExhaustedError` unwrap, `BrokerBusyError` exclusion, caller-cancellation-vs-internal-timeout). Confirmed two ways:

- Reverted the caller-context timeout fix (`callerCtx.Err()` back to the reassigned `ctx.Err()`, reintroducing exactly the bug Copilot and solace-awaheed both flagged and reproduced) — full repo suite, zero failures.
- Went further: made `d.resultHook(...)` a no-op entirely, so the hook never fires — full repo suite, still zero failures.

Coverage profile confirms it directly: every block in the defer (`sender.go:760-789`) shows `0` hits under `go test -coverpkg`. No `sender_test.go` change anywhere in this diff. Nothing repo-wide references `WithResultHook`/`WithBrokerResultHook` except `cmd/server/main.go`. This is Copilot's "no focused Sender tests" thread, which never got an explicit fix or reply, unlike the other four.

This isn't a theoretical gap. The retry-budget context is the effective deadline on the MCP path (`WriteTimeout` is deliberately zero for long-lived streams), so the exact branch this fix targets is the one that fires for a hung broker in production. With the bug present, a hung broker doesn't go silent, it actively reports `mcp_broker_reachable` pinned at its last value (1 = reachable) — the opposite of what this story exists to surface. `mcp_broker_last_result_timestamp_seconds` gives operators an eventual, 15-minute-granularity, opt-in backstop, but that's a different alert than the one this feature is meant to provide and doesn't close the test gap.

Cheapest fix: extract the defer body into a named, testable function in `resilience` (e.g. `brokerSignalStatus(callerCtx context.Context, resp *http.Response, err error) (status int, report bool)`) and table-test the five cases Copilot originally listed (success, `RetriesExhaustedError` with final status, transport timeout/status 0, admission shedding, caller cancellation). Pure function, no new dependencies.

**Non-blocking note on condition 4** (AC 5: "classification centralized in `health.Classify(err)`"): as literally worded this is architecturally impossible, not unimplemented. `health` already gets imported by `metrics`, which `resilience` imports — so `health.Classify` taking a `resilience` error type would be an import cycle. Verified by compiling a probe import. The shipped split (`Classify(int)` in `health`, extraction in `sender.go`) is the only shape the dependency graph allows. Worth amending the AC while the story's in Verifying, same resolution style used elsewhere in this repo's history — the extraction function above satisfies the spirit of it once it's named and tested.

Everything else checked clean: single `RegisterCallback` now feeds all three gauges from one snapshot (confirmed by reading `broker.go`, resolves the dual-snapshot condition), `health.StateReachable` typed comparison replaces the string literal solace-awaheed flagged, and the docs inventory/schema-version/CHANGELOG gaps Copilot and bczoma both raised are filled in.

internal/observability/metrics/broker.go
```diff
@@ -0,0 +1,91 @@
...
+					val = 1.0
+				}
+				o.ObserveFloat64(gReason, val, metric.WithAttributes(
+					brokerAttr,
+					attribute.String("reason", string(reason)),
+				))
+			}
+
+			if !snap.LastResult.IsZero() {
+				o.ObserveFloat64(gTimestamp, float64(snap.LastResult.Unix()), metric.WithAttributes(brokerAttr))
```
bczoma, line 81: 🟡 **The new timestamp family is asserted by no test**

Flagged by good:code.

`mcp_broker_last_result_timestamp_seconds` — both the value and the `!IsZero()` suppression guarding it — is constrained by nothing. Verified by sabotage: replacing this block with an unconditional `o.ObserveFloat64(gTimestamp, 12345, metric.WithAttributes(brokerAttr))` leaves `go test ./internal/observability/metrics/...` green. Every metrics-level snapshot fixture (`broker_test.go:38,62,97`) leaves `LastResult` at its zero value, so the branch never executes under test, and `TestBrokerMetrics_AbsentBeforeFirstCall:138` filters on only the two older family names, so it does not reach the third either.

Statement coverage hides this — `NewBrokerMetrics` reports 80.8% because the callback body runs. Nothing reads its output.

**Why it matters.** This is the one new public metric family in the commit, and it is the one an operator alerts on for a silently-idle broker (`docs/observability.md:284-286` ships the `time() - … > 900` recipe). Three regressions ship green today: a unit error (millis for seconds) makes every `time() -` alert fire permanently; a lost `!IsZero()` guard publishes `timestamp=0` for a broker that has never been called, so "never called" reads as 1970 and the same alert fires for a broker that is fine; and a label change breaks the recipe silently. It is also the family with the least review history, since it arrived in this round.

**Recommended fix.** Add `TestBrokerMetrics_Timestamp` in the shape the sibling tests already use — a snapshot with `LastResult: time.Unix(1700000000, 0)` compared via `GatherAndCompare` on the exact expected body, plus one case with a zero `LastResult` asserting the family is absent. Then extend the name filter in `TestBrokerMetrics_AbsentBeforeFirstCall` to all three families so the next family added cannot skip it the same way.

`TestBrokerMetrics_OneHot` and `TestBrokerMetrics_Unreachable` are exactly the right pattern for this — exact `GatherAndCompare` bodies including HELP text. This family is the one that did not get it.

internal/observability/metrics/testdata/metrics_golden.txt
```diff
@@ -23,7 +23,7 @@ mcp_panic_recovered_total{boundary="http"} 0
 mcp_panic_recovered_total{boundary="tool"} 0
 # HELP mcp_schema_version Observability schema versions; always 1, labelled with each schema.
 # TYPE mcp_schema_version gauge
-mcp_schema_version{audit_schema="1.1",metrics_schema="1.2"} 1
+mcp_schema_version{audit_schema="1.1",metrics_schema="1.3"} 1
```
bczoma, line 38: 🟡 **The scrape golden — the repo's own contract guard — contains none of the three new families**

Flagged by docs.

This file is the machine-checked inventory of the `mcp_*` scrape surface: `provider_test.go:57-69` captures every `mcp_`-prefixed line and `:143-146` fails with "scrape schema drifted from golden file". None of `mcp_broker_reachable`, `mcp_broker_unreachable_reason` or `mcp_broker_last_result_timestamp_seconds` appear in it, because `TestScrapeSchema` never calls `p.BrokerMetrics(...)`. The previous metrics story deliberately wired itself in for this reason — `provider_test.go:118-124`, "Security counters (SOL-152099): registration seeds the auth-failure series".

**Why it matters.** `docs/observability.md:106` tells operators "Pin dashboards to `mcp_schema_version`", and this golden is the guard that makes the pin mean anything: it is what fails when a name or label moves without a schema bump. For these three families nothing enforces it. A future rename or label change ships green, `metrics_schema` stays at 1.3, and the pinned dashboard breaks at upgrade with no signal that the contract moved — which is precisely the failure the pinning advice exists to prevent. The families most at risk are the two whose vocabulary is still unsettled (see the AC item in the review body).

**Recommended fix.** In `TestScrapeSchema`, register the broker gauges alongside the security counters, seeding a tracker with one reachable and one failed broker so the one-hot `0` series are captured too, then regenerate:

    go test ./internal/observability/metrics/ -update

That also gives the three families the same drift protection every other shipped family already has, and it is the cheaper half of the same problem the other thread describes.

bczoma (dismissed):
## PR Review: #402 — SOL-152088: Dashboard broker reachability with classified failure reasons, without ejecting healthy MCP pods

**Author:** solace-wmaryam | **Branch:** `wmaryam/SOL-152095_go_runtime_process_metrics` → `main` | **Changes:** +703 / -25 across 12 files
**Jira:** `SOL-152088` — Dashboard broker reachability with classified failure reasons (Story 9) (Verifying)

Follow-up review at `2ff950c4`. My prior review sat at `196f6af5` and was dismissed by this push; its findings are carried below. Reviewed the one new commit (10 files, +244/-90), not the whole PR.

---

### Verdict

1. 🟡 `internal/observability/metrics/broker.go:80-81` — the new `mcp_broker_last_result_timestamp_seconds` family is asserted by no test; a wrong value or a lost zero-time suppression ships green
2. 🟡 `internal/observability/metrics/testdata/metrics_golden.txt` — the machine-checked scrape golden contains none of the three new families, because `TestScrapeSchema` never registers them

- 🟡 **SOL-152088's acceptance criteria no longer describe what ships, and nothing records the deviation.** AC 1's state list has no `broker_error` or `unknown`; AC 4 asks for `reason="broker_error"` with the status in a separate bounded label, while the code emits `broker_error_NNN` and reserves it for 429/5xx only; AC 4 says "other non-2xx to `broker_error`", while non-401/403 4xx is now `reachable`; 403 joined 401 under `credential_invalid`; AC 5's centralized `health.Classify(err)` is split across two packages; and the third metric family has no AC at all. Every one of those deviations is defensible — 4xx-as-reachable is strictly better signal, and `docs/user-guide.md:262` already prescribes the same credential/role remediation for 401 and 403 — which is exactly why they need writing down where verification happens. The story is in **Verifying**: amend the ACs before it leaves, or a verifier fails a correct change, or passes it without noticing the vocabulary moved.
- 🟡 **The PR body is stale at this head.** It still says "Two Prometheus gauges are exposed" and lists the `broker_error_NNN` state vocabulary; it predates the third family, the 4xx reclassification, the 403 widening and the timeout fix, and its "Changes Made" omits `CHANGELOG.md` and `schema/version.go` — the two files carrying the public-schema bump. Its "Coverage: 89.0%" line is also no longer evidence for the DoD coverage item, given the gap @solace-aross measured.

**Not an unblock:** @solace-aross's `CHANGES_REQUESTED` stands — the `ResultHook` defer at `sender.go:760-789` still has no test, verified there by sabotage and by coverage profile. I have not restated it, and my 🔄 does not clear it.

---

### Summary

The head commit is a disciplined response to review rather than a rubber-stamp: four of my five conditions are addressed in code, and two of them (the 4xx classification and the one-hot retention) have regression tests that fail cleanly when reverted. The typed-boundary fix went further than asked — `SnapshotStrings` was deleted rather than left beside its replacement — and consolidating all three gauges onto one `RegisterCallback` makes cross-family consistency a property of the design instead of a timing accident.

What is left is that the two things this commit adds are the two things nothing guards. The new timestamp family has no assertion at any level, unit or golden, and the ticket that verification will be run against describes a different classifier from the one that shipped. One security claim was filed at 🟡 here and measured down to a traceability item: the 403 grouping turned out to match the repo's own operator guidance, and the `broker_authz_denied` collision it alleged is with an audit record that is not emitted yet, on a different surface.

*12 prior findings — 7 addressed (all five 🟡 conditions but one, plus the CHANGELOG and schema-version gaps), 5 outstanding below.*

|---|---|---|---|
| 🟡 | AC 5 asks for the classifier in `health`; the half that can be wrong lives in `Sender.Do` | Not addressed | `internal/observability/health/tracker.go:39-54` — `Classify` still takes an already-derived `int`; the error→status policy (busy filter, `RetriesExhaustedError.StatusCode`, caller-vs-internal deadline, unknown-error drop) is still inline at `internal/semp/resilience/sender.go:762-789`. @solace-aross's blocking thread prescribes the extraction and the tests, so I am not filing it twice — but it is the reason the recommendation is no better than 🔄. |
| 🟢 | Panic unwind leaves both named returns nil, so the hook records `unreachable` for a local fault | Not addressed | `internal/semp/resilience/sender.go:767-790` — no `if resp == nil && err == nil { return }` guard; control falls through to `d.resultHook(d.brokerAlias, 0)`. |
| 🟢 | `broker_error_NNN` puts the status in the reason value where AC 4 wants a separate bounded label | Not addressed | `internal/observability/health/tracker.go:33-35` unchanged. Folded into the AC-amendment item above — the code side is defensible, the ticket side is what is missing. |
| 🟢 | `BrokerMetrics(cb)` takes an argument but guards with `sync.Once`, so a second caller's callback is silently ignored | Not addressed | `internal/observability/metrics/provider.go:230-236` — parameter type changed, contract still undocumented. |
| 🟢 | `TestReadyzDecoupling` builds tracker and `ReadinessState` unrelated by construction, so it cannot fail | Not addressed | `internal/observability/health/tracker_test.go:200-218` — unchanged shape; AC 6 remains verified by construction, not by test. |

---

### Minor notes

- 🟢 `internal/observability/health/tracker.go:99-101` — `seenReasons` only grows, so every reason a broker ever hit stays advertised (at 0) until restart; together with the new timestamp gauge an unauthenticated `:9091` reader now gets each broker's alias, its whole failure history, and a per-broker activity oracle. Either age reasons out, or state the retain-until-restart contract and the disclosure in the Broker Reachability docs.
- 🟢 `internal/semp/resilience/sender.go:251-253` — `ResultHook` is an exported extension point called synchronously from a `defer` with no `recover` and no documented contract; say in the doc comment that it must not block, do I/O, or panic.
- 🟢 `internal/observability/schema/version.go:37-39` — the third family landed inside the already-published `metrics_schema` 1.3 rather than 1.4, so backing it out is a contract break and `1.3` means two things depending on the build. If 1.3 has not shipped in a tagged release this is free to fix.
- 🟢 `internal/semp/resilience/sender.go:777-782` — the internal-vs-caller deadline is inferred by re-reading `callerCtx.Err()` at defer time; `context.WithTimeoutCause` plus `errors.Is(context.Cause(...), errRetryBudgetExhausted)` would let the classification read a fact the code emitted instead of the clock, and composes with the extraction @solace-aross asked for.
- 🟢 `internal/observability/health/tracker.go:119-135` — `SnapshotForMetrics` allocates and `sort.Slice`s per broker while holding `mu.RLock`, so each scrape blocks the hot recording path for work that needs no lock; copy the keys under the lock and sort after releasing it.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

internal/observability/health/tracker.go
solace-wmaryam: Added `brokerSignalStatus(callerCtx, resp, err) (int, bool)` as a named package-level function in `sender.go`, extracted from the defer. The defer is now a two-liner. `TestBrokerSignalStatus` in `sender_test.go` table-tests all six paths: success, `RetriesExhaustedError` with a final status, transport failure (status 0), admission shedding, caller cancellation, and internal timeout with live caller context.

internal/observability/health/tracker.go
solace-wmaryam: Added `TestBrokerMetrics_Timestamp` (pinned `time.Unix(1700000000, 0)`, compared via `GatherAndCompare`) and `TestBrokerMetrics_Timestamp_ZeroAbsent` (zero `LastResult` asserts family absent). `TestBrokerMetrics_AbsentBeforeFirstCall` now checks all three families.

internal/observability/metrics/testdata/metrics_golden.txt
solace-wmaryam: Fixed. `TestGoldenSchema` now calls `p.BrokerMetrics()` with a fixed two-broker snapshot (one reachable, one unreachable with two seen reasons and fixed `time.Unix` timestamps), so all three families — including the one-hot 0 series — are captured in the golden file and protected by the drift gate going forward.

solace-wmaryam:
@solace-aross addressed both points:

**ResultHook test coverage** — extracted the defer body into `brokerSignalStatus(callerCtx, resp, err) (int, bool)`. The defer is now a two-liner that calls it. `TestBrokerSignalStatus` in `sender_test.go` table-tests all six paths including the internal-timeout-vs-caller-cancellation distinction. Verified by sabotage: reverting the `callerCtx` fix now fails the `internal_timeout_retry_budget` case.

**AC 5 note** — agreed, the literal wording is impossible given the import graph. Will flag for the AC amendment.

solace-aross (approved):
Re-reviewed at d4041e02. The blocker's closed.

`brokerSignalStatus(callerCtx, resp, err) (int, bool)` is now a named, table-tested function exactly as suggested. Sabotage-verified the highest-risk branch myself: reverted the caller-vs-internal-timeout distinction back to always returning `(0, false)` and `TestBrokerSignalStatus` fails clean on that one case, everything else in the table stays green. That's the branch that used to have zero coverage anywhere in the repo; now it's pinned.

Also verified the golden-schema drift protection is real, not just present: reverted `broker.go`'s one-hot emission to current-reason-only (dropping the zeroed prior-reason series) and both `TestGoldenSchema` and the dedicated `TestBrokerMetrics_OneHot` fail cleanly, pointing at the exact missing line. `MetricsSchemaVersion` is bumped to 1.3 and the golden reflects it. Full suite green (`go test ./... -race -count=1`).

Approving. One coordination item stays open across PRs, not blocking this one: #389 also wants `metrics_schema` 1.3 — whichever merges second needs 1.4, still nobody's called the order.

internal/semp/resilience/sender.go
```diff
@@ -741,7 +752,46 @@ func (d *Sender) shed(ctx context.Context, stage string, start time.Time) error
 // On failure after retries, returns a RetriesExhaustedError or a wrapped error.
 // When the request cannot be admitted within semp.max_queue_wait it returns a
 // BrokerBusyError and never reaches the broker (see admit).
-func (d *Sender) Do(ctx context.Context, req *http.Request) (*http.Response, error) {
+// brokerSignalStatus extracts the HTTP status to report from the terminal
```
bczoma, line 755: 🟡 **The extraction landed between `Do`'s doc comment and `Do`, so the exported method lost its documentation**

Flagged by good:code.

`brokerSignalStatus` and its own five-line doc comment were inserted *after* `Do`'s doc comment block and *before* `func (d *Sender) Do`. Go attaches a doc comment to whatever declaration immediately follows it, so the result is that the five lines above this one — "On success, returns the HTTP response (body open for the caller to read) … On failure after retries, returns a `RetriesExhaustedError` … When the request cannot be admitted within `semp.max_queue_wait` it returns a `BrokerBusyError` and never reaches the broker (see admit)" — now document this helper, which does none of those things, while `Sender.Do` documents nothing at all.

**Why it matters.** This is not a cosmetic comment nit; it silently deletes the published contract of the package's primary exported method. Verified empirically at this head:

```
$ go doc ./internal/semp/resilience Sender.Do
func (d *Sender) Do(ctx context.Context, req *http.Request) (resp *http.Response, err error)
```

A caller consulting `go doc` for `Sender.Do` — the one place the `BrokerBusyError`-without-reaching-the-broker behaviour and the "body is yours to close" ownership rule were written down — now sees a bare signature. Neither `go vet` nor the linters flag a misattached doc comment, so nothing else in the pipeline will catch this, and once merged it reads as though `Do` was simply never documented.

**Recommended fix.** Move the helper below `Do` (or above `Do`'s comment block) so each declaration keeps its own doc:

```go
// Do sends an HTTP request through the rate limiter and retryablehttp client.
// … (unchanged, back against Do)
func (d *Sender) Do(ctx context.Context, req *http.Request) (resp *http.Response, err error) { … }

// brokerSignalStatus extracts the HTTP status to report from the terminal
// outcome of a Do call. …
func brokerSignalStatus(callerCtx context.Context, resp *http.Response, err error) (int, bool) { … }
```

bczoma (approved):
## PR Review: #402 — SOL-152088: Dashboard broker reachability with classified failure reasons, without ejecting healthy MCP pods

**Author:** solace-wmaryam | **Branch:** `wmaryam/SOL-152095_go_runtime_process_metrics` → `main` | **Changes:** +868 / -25 across 14 files
**Follow-up review** — parent-only pass over the +199/-34 pushed since my dismissed review at `2ff950c4` (5 files, no new ones).

---

### Verdict

1. 🟡 `internal/semp/resilience/sender.go:755` — the new helper was inserted between `Do`'s doc comment and `Do`, so the exported method now has no godoc and the helper claims five lines that describe `Do`

The 1 🟢 note below is **not** a condition of this verdict.

---

### Summary

Both conditions from my last pass are genuinely closed, and closed at the right seam. The timestamp family now has `TestBrokerMetrics_Timestamp` with a pinned `time.Unix(1700000000, 0)` compared through `GatherAndCompare`, plus `TestBrokerMetrics_Timestamp_ZeroAbsent` for the zero-time suppression; the scrape golden now carries all three families including the one-hot `0` series, because `TestGoldenSchema` drives `p.BrokerMetrics()` with a fixed two-broker snapshot — so the drift gate protects the new surface from here on rather than around it.

The `brokerSignalStatus` extraction is the better answer to @solace-aross's blocking thread than the one I asked for: the policy that can actually be wrong is now a pure function, and `TestBrokerSignalStatus` table-tests six paths including the internal-timeout-versus-caller-cancellation distinction that the `callerCtx` fix exists for. The literal AC 5 wording — the classifier living in `health` — is impossible given the import graph, and the author's reason for declining it is better than the finding was; that one is settled, not conceded. What is left is one mechanical casualty of the extraction: it landed on the wrong side of `Do`'s doc comment.

*11 prior findings settled — 3 addressed (both 🟡 test-coverage conditions and the classifier extraction), 4 declined, 4 accepted as-is. The 2 exceptions are below.*

|---|---|---|---|
| 🟡 | SOL-152088's acceptance criteria no longer describe what ships, and nothing records the deviation | Not addressed | The author has committed to flagging the AC amendment; nothing is written down yet. Reported, and per its class it does not hold this verdict. |
| 🟡 | The PR body is stale at this head | Not addressed | Body line 5 still reads "Two Prometheus gauges are exposed" while three ship; line 24 still lists the `broker_error_NNN` vocabulary; line 45's "Coverage: 89.0%" predates this push's tests. |

Four 🟢s from the prior pass (the panic-unwind guard, `broker_error_NNN`'s label shape, `BrokerMetrics`'s `sync.Once` contract, `TestReadyzDecoupling`'s construction) are now marked `Declined` — raised across two passes and not taken. They are recorded here and not re-argued.

---

### Minor notes

- 🟢 `internal/semp/resilience/sender_test.go:157` — the doc comment says "the five cases the ResultHook depends on" over a table of six, and the seventh path (`return 0, false` for an unknown error shape) is the one case still uncovered. Fix the count and add the row, or say why that path needs none.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ |
| Code Quality | ⚠️ — `Do` lost its godoc to the extraction |
| Test Coverage | ✅ |
| Architecture | — not re-reviewed (unchanged since `2ff950c4`) |
| Security | — not re-reviewed (unchanged since `2ff950c4`) |
| Documentation | ⚠️ — PR body still stale at this head |
| PR Scope & Description | ⚠️ — see the Prior Findings table |

Assessed by the parent (follow-up); no lenses were dispatched this pass.

PR #404, opened 2026-09-10
SOL-154036: assert the span, metric, and audit vocabularies agree on one call

solace-awaheed:
## What this delivers

ADR-009's single-join-key property — one `outcome` and one `error_type`, spelled identically on every signal, so a predicate carries from a dashboard to a trace backend to a SIEM untranslated — is now asserted across all three surfaces on **one real call**, rather than only pairwise.

**Nothing new is emitted.** The deliverable is the guarantee. Production changes are two comments updated for a test rename.

## Why it was unowned

Coverage was genuinely pairwise before this:

- SOL-152421 joined the **span** to the **metric**.
- SOL-152090's AST guard joined this package's emitted `error_type` set to the **audit constructor's vocabulary**.
- Nothing joined the audit record's own values to the other two on one call.

That gap belonged to no story, because no individual signal's tests can see it — each only ever sees itself. The failure mode is silent and total: a filter pasted from a dashboard into the trace backend or the SIEM quietly matches nothing, which reads as "no such failures" rather than as a broken query.

## The two tests

"The audit record" is two distinct surfaces here, so there are two joins:

1. **`TestRequestPathSpans_SpanMetricAndLogAgreeOnTheSameCall`** (renamed from `...SpanAndMetricAgreeOnTheSameCall`) — the existing span/metric join now also compares `logToolResult`'s `tool invoked` line, emitted for *every* dispatch and what a SIEM ingesting stdout indexes. Covers success, tool error, unresolvable broker, and recovered panic.

2. **`TestRequestPathSpans_OperationAuditRecordAgreesWithSpanAndMetric`** (new, `test/integration/cross_signal_vocabulary_test.go`) — drives a destructive tool with `OBS_AUDIT_LOG_ENABLED` on and joins the `audit_event_type=operation` record to both the span and the metric, across success, error and panic.

The second is a different code path, not a second look at the first: the record comes from `emitOperationAudit` through `audit.NewEvent`, whose `Outcome` is an independent typed vocabulary that merely happens to be spelled like the metric's, and the `error_type` crosses the package boundary as a bare string. It also asserts `panic_recovered`, which is audit-only and derived by string-comparing the joined `error_type` — so a rename that broke the join would silently turn a crash into an ordinary handled failure in the audit trail.

## Two deliberate exclusions, both commented

- **`broker` is not compared on the log line.** `logToolResult` keeps the raw caller alias for diagnostics while the metric and span are canonicalized, so on an unresolvable broker they differ *by design*. Asserting equality would pin that divergence shut.
- **No unknown-broker row in the audit test.** Resolution fails before the destructive gate, so no arguments hash is taken and no operation record is emitted. Verified empirically, not inferred.

## Verification

- `make check` passes on the rebased tree; coverage 89.7% against the 85% floor.
- `-race -count=5` clean across the whole `test/integration` package.
- **Mutation-checked rather than trusted green:** drifting the log line's `outcome`, the audit `error_type`, and the span's `outcome` each produced a failure naming the disagreeing surfaces.

## Scope note

Three of SOL-154036's four bullets were already satisfied by SOL-152421 and SOL-152090 in the days after the ticket was written — the span/metric join (including the panic path), the tool-RED position in `tracing_wiring_test.go`, and the `error_type` count reconciliation. Details are recorded as a comment on SOL-154036. The Jira half of the remaining work — two stale claims in SOL-152421's acceptance criteria (the middleware chain order, and an `error_type` count whose predicted test failure never fired because Story 49 landed as `broker_authz_denied`) — is filed as a comment on that ticket.


solace-aross (dismissed):
Verified `manager.go`/`spans.go` are pure comment updates (test-name references), matching the PR description. No production behavior change.

Sabotage-verified the core join rather than trusting the mutation-testing claim: made `emitOperationAudit` skip the outcome=error branch entirely (audit record stays success while the tool actually errored). Both the error and panic subtests fail cleanly — and there's a nice defense-in-depth surprise: `audit.NewEvent`'s own schema constructor independently rejects the malformed record (error_type set on a success outcome), so the drop path fires and `oneOperationAuditRecord` correctly reports 0 records instead of silently passing. Two layers catching the same drift.

One non-blocking note, extending Copilot's suppressed comment rather than just restating it. Its finding is right: the audit test's join loop (`tool`, `outcome`, `error_type`) omits `broker`, unlike the span/metric test. But I checked whether adding it would actually do anything today: `emitOperationAudit` receives `brokerAlias` after `CallTool`'s own resolution step already rewrites it to `bc.DisplayName()` on success (manager.go:333-334), so the audit record's broker value is functionally canonical already, by construction rather than by an explicit `canonicalBrokerLabel()` call. I sabotaged that rewrite directly (removed it) and reran both cross-signal tests — still green, because this test's only broker fixture (`dev`) has an identical alias and display name, so there's nothing to observe either way. Copilot's own suggested fix ("ideally with a non-canonical caller alias case") is exactly what's needed: add `broker` to the join loop, and add a casing-differentiated broker alias to the fixture, or the added comparison verifies nothing. Worth doing, not worth blocking on — the invariant holds today, it's just currently untestable as written.

Approving.

bczoma (dismissed):
## PR Review: #404 — SOL-154036: assert the span, metric, and audit vocabularies agree on one call

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-154036` → `main` | **Changes:** +278 / -17 across 5 files
**Jira:** `SOL-154036` — assert the cross-signal vocabularies agree (Verifying)

---

### Verdict

**Recommendation:** ✅ Approve

Nothing here is a condition of the approval. The five 🟢/💡 notes below are optional.

The `broker`-key omission in the audit join loop that @Copilot raised and @solace-aross measured is not restated here — their conclusion (the invariant holds today by construction at `manager.go:333-334`, and adding the key verifies nothing until the fixture carries a non-canonical alias) is right, and I have nothing to add to it.

### Summary

Two tests close a join that genuinely belonged to no story: the span/metric pair now also pins `logToolResult`'s `tool invoked` line, and a new test joins the `operation` audit record to both across success, error and panic. The reason to trust it is that both joins are guarded against the shapes that would make them pass vacuously — literals are asserted before the cross-comparison loop (three empty strings compare equal and prove nothing, as the comment says), and both record lookups require *exactly* one record and fail loud on zero. Failure messages name which two surfaces disagreed and what an operator loses, which is materially better than a `want/got` default. Production surface is two comments updated for the rename, the rename is complete (no reference to the old test name survives anywhere in the tree), and the two deliberate exclusions are each backed by prose already shipped in `docs/observability.md` rather than invented here. No 🔴 or 🟡 was filed by any lens, so no finding needed challenging.

### Minor notes

- 🟢 `test/integration/cross_signal_vocabulary_test.go:60-68` — `captureLogRecords` installs a global slog handler writing into a plain `bytes.Buffer`, then reads it back with no lock; slog's mutex guards writer *writes*, not this reader. Latent only because the new tests are sequential — adding `t.Parallel()` or any background goroutine that logs turns it into an intermittent race. Guard the buffer with a mutex and note in the doc comment that callers must stay sequential.
- 🟢 `test/integration/cross_signal_vocabulary_test.go:186-206` — the ~20-line arrange block (recorder, `metrics.New`, `ToolMetrics`, `captureLogRecords`, `fakeBroker`, `tracedSessionWith`, the `len(series) != 1` guard) is duplicated near-verbatim from `request_path_spans_test.go:730-751`, differing only in the handler and `WithAuditLog(true)`. Two copies of the "one call, all signals captured" fixture will drift — which is the drift class this PR exists to prevent. Extract one helper returning span, labels and logs.
- 🟢 `test/integration/cross_signal_vocabulary_test.go:135-146` — the `Metadata` override replaces the whole `Annotations` struct, so it also drops the embedded handler's `ReadOnly: true`. That is *required* (`types.go:63-65` — `Destructive` is only meaningful with `ReadOnly` false), so the behaviour is right and the comment claiming "only the annotation differs" is what's wrong. Set the two fields explicitly and say they are coupled, or a later `Idempotent`/`OpenWorld` annotation is discarded silently.
- 💡 `docs/observability.md:986-990` — "This is enforced, not merely intended" still describes the guarantee as holding "on the span and on the `mcp_tool_invocation_total` series". After this PR CI holds two more surfaces. No wrong action follows, but the doc's strongest available claim is left on the table exactly where a developer reads about the join.
- 💡 `CHANGELOG.md:47` — "asserted across all three surfaces on one real call" undercounts its own deliverable: four signals are involved and no single call joins all four; the metric is the pivot both tests share, so span↔audit and log↔audit are covered transitively. "Across all four signals, by two tests that each join three" would be exact.

### Summary Verdict

|---|---|---|
| Architecture | — not reviewed (production surface is comment-only; no design change) | good:architecture |
| Security | — not reviewed (no auth, crypto, input-validation or dependency surface in the diff) | good:security |
| Documentation | ✅ | docs |

The scope claim was verified against Jira rather than taken on trust: SOL-154036 has exactly four Task bullets, the comment reducing it to bullet 2 exists and itemises why 1, 3 and 4 were already satisfied, and the corrections filed against SOL-152421's stale acceptance criteria exist too. Bullet 4's premise was actively falsified rather than dropped — the ticket predicted ten or eleven `error_type` values and a deliberate test break; twelve were found, enumerated once in `metrics.AllErrorTypes`, and both tickets corrected.

solace-aross (approved):
Re-reviewed at b332467d. The note's closed properly, not just patched over.

Sabotage-verified the fix myself, same defect as last pass (removed the `brokerAlias = bc.DisplayName()` reassignment in `manager.go`): this time it fails clean, and only the new `non-canonical caller alias` subtest catches it — exactly the case that fixture was missing before. The `Dev-EU`/`dev-eu` alias pair is what makes the comparison mean anything; a same-spelled alias like `dev` couldn't have told a working normalization from a removed one, which is exactly what I'd flagged.

Also good: the mutex-guarded buffer closes a real (if not yet triggered) race — the dispatch runs on an SDK-spawned goroutine, so a plain `bytes.Buffer` read/write pair was one `t.Parallel()` away from a data race the fixture's own design would have hidden. The annotation fix (`ReadOnly = false` set explicitly rather than replacing the whole struct) is the right shape — a struct-literal replacement would have silently dropped any other annotation added to the embedded handler later.

Approving.

bczoma (approved):
## PR Review: #404 — SOL-154036: assert the span, metric, and audit vocabularies agree on one call

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-154036` → `main` | **Changes:** +374 / -53 across 6 files
**Jira:** `SOL-154036`

Follow-up review (second pass). My prior review at `4108acf` was dismissed by the stale-review rule when `b332467` landed, but it remains an ancestor of the head, so this pass is scoped to that one commit: 4 files, +151/-91.

---

### Verdict

**Recommendation:** ✅ Approve

Nothing outstanding. All four items raised on the first pass — the three optional notes in my review and the `broker` join that Copilot flagged and @solace-aross extended — are addressed in code at `b332467`, and the fixture change makes the new assertion load-bearing rather than decorative.

### Summary

The follow-up commit does the three things its subject claims and does them at the right seam. The `broker` key is now joined on the audit path, and — the part that matters — `spanPoolFor` gained a `Dev-EU` alias so a caller typing `dev-eu` exercises real normalisation. That was precisely the condition @solace-aross identified as missing: he sabotaged the `DisplayName()` rewrite in `manager.go:333-334` on the previous head and both cross-signal tests stayed green, because the only fixture alias was spelled identically either way. I repeated that sabotage at this head — removing the rewrite — and the `non-canonical caller alias` subtest now fails with three messages naming the disagreeing surfaces (`audit record = "dev-eu"` against `metric label = "Dev-EU"` and the span attribute). The comparison verifies something now.

The other two fixes are equally at the seam rather than around it. `syncBuffer` guards the buffer with a mutex on both sides, and its doc comment goes further than the note asked by identifying *why* the current pattern happens to be safe — the dispatch runs on `jsonrpc2.handleAsync` and the `CallTool` HTTP response is what makes records visible — so the lock is documented as protecting a transport property rather than as a formality. The `Metadata()` override now assigns `ReadOnly` and `Destructive` field by field with a comment recording that `types.go` couples them, which is the fix I asked for and correctly inverts what the old comment claimed. `crossSignalCall` collapses the duplicated arrange block into one fixture used by both tests, removing the drift this PR exists to prevent from the PR's own scaffolding.

Two things I checked beyond the findings. Adding a second alias to `spanPoolFor` is shared by two other call sites (`request_path_spans_test.go:232`, `:522`), including the `list-brokers` registration test — the full `test/integration` package is green under `-race`, and the two cross-signal tests are green at `-count=2`. And the CHANGELOG line was tightened to say "four signals, by two tests that each join three of them", with the transitivity stated explicitly; that is a real correction, since no single call does join all four, and it is the kind of overclaim this review would otherwise have raised.

*4 prior findings settled — 4 addressed (3 optional notes from my first pass, plus the `broker` join raised by Copilot and @solace-aross). No exceptions.*

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ |
| Test Coverage | ✅ — sabotage-verified at this head, not merely green |
| Code Quality | ✅ |
| Documentation | ✅ |
| Architecture | — not re-reviewed (unchanged since `4108acf`) |
| Security | — not re-reviewed (unchanged since `4108acf`) |
| PR Scope & Description | — not re-reviewed (unchanged since `4108acf`) |

Source: parent (follow-up).

PR #391, opened 2026-09-08
SOL-152418: push metrics over OTLP alongside the existing Prometheus scrape (Story 46)

solace-lootii:
## Summary

Customer SREs whose organisation standardises on an OTLP-native APM (Datadog, Dynatrace, Honeycomb, New Relic) can now ingest this server's metrics the same way as every other service in their estate — no Prometheus scrape job, no collector deployed purely for this component.

Gated by the new `OBS_METRICS_OTLP_ENABLED` flag (off by default, door-closing policy). A second `sdkmetric.Reader` — an OTLP gRPC push exporter wrapped in a `PeriodicReader` — attaches to the **same meter provider** Story 14 (SOL-152091) built, alongside the existing Prometheus exporter, so both egresses observe one instrument set and share the identity resource (Story 34, `cloud.region` included) by construction. Endpoint and every other transport concern come from the standard `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` env vars and their siblings — no code reads them explicitly.

## Acceptance criteria (from the ticket)

- [x] Metrics arrive at an OTLP collector and remain on `/metrics`; a test asserts the two agree on counter values (`TestOTLP_DualEgressAgreement`, real fake-collector round trip)
- [x] Cumulative temporality explicitly configured (`otlpmetricgrpc.WithTemporalitySelector`) and asserted by test (`TestOTLP_CumulativeTemporality`)
- [x] `OBS_METRICS_OTLP_ENABLED=true` with metrics off fails config load with an explicit error (`validateMetricsOTLPCoherence`)
- [x] Default off; flag-off leaves no OTLP instruments registered at all (`TestOTLP_FlagOff_NoOTLPInstrumentsRegistered`)
- [x] `mcp_otel_metrics_exported_total` / `mcp_otel_metrics_dropped_total{reason}` on the scrape surface, readable while the OTLP endpoint is unreachable (`TestOTLP_UnreachableCollector_ScrapeStaysUpAndDropsAreVisible`)
- [x] Unreachable collector doesn't degrade service — verified by a full-stack integration test through `ToolManager.CallTool` asserting tool-call latency is unaffected (`test/integration/otlp_metrics_unreachable_collector_test.go`)
- [x] Reader's flush registered against the Story 48 (SOL-152449) shutdown-hook registry, not `drainAndShutdown` — see the "design decision" note below on why this rides the *existing* `metrics_provider` hook rather than a new third one
[...]

PR #405, opened 2026-09-10
SOL-152094: finalize the versioned metric and audit schema doc (Story 35)

solace-aross:
> **Rebased onto current `main` (`29a181e`).** This PR was originally cut from `7d33ebc`, before #402 (SOL-152088, Story 9) landed the broker reachability gauges. See [Rebase correction](#rebase-correction) at the bottom for what that changed and which of this description's earlier claims were wrong.

## What this does

Docs only. No production Go code changes. Two files: `docs/observability.md` and one link in `README.md`.

`docs/observability.md` is the schema contract customer NOC operators, SREs, and compliance reviewers pin Grafana dashboards and SIEM rules to. This finalizes it: it closes the "never rename" framing, separates designed-but-not-emitted signals from live ones, and fills the gaps that made it hard to use as a contract.

## The eight changes

| # | Change | Why |
|---|---|---|
| 1 | Replaced the "we never rename" language in three places with the real commitment, plus a new **Compatibility and Deprecation Policy** section | A bank does not need a promise never to change. It needs an announced path with enough notice to migrate a dashboard: announce in CHANGELOG and in the doc, two minor releases of notice with dual emission where dual emission is possible, remove only at a MAJOR bump. |
| 2a | Split **OTLP Export Health** into the live span pair and the not-yet-emitted metrics pair | The table and the Implementation Status row conflated them. An operator would have alerted on `mcp_otel_metrics_dropped_total` and seen a permanently absent series read as healthy. |
| 2b | Documented `mcp_broker_authz_denied_total{tool,broker,reason}` as designed, not yet emitted | The metric counterpart of the hop-2 audit record was documented nowhere. Includes why it carries a `broker` label when the hop-1 counter does not, framed as a deliberate asymmetry. |
| 3 | Added **Canonical Audit Queries**: four named queries with worked examples | States explicitly that a complete refusal picture needs both denial queries. Querying only `authz_denied` silently misses every broker-side denial, and it looks like a complete answer. |
| 4 | Added a **Start Here** table of contents | Content that does not exist yet is plain text naming the story that lands it (Story 40 / SOL-152423, Story 36 / SOL-152098, Story 38), never a dead anchor. |
| 5 | Linked the doc from `README.md`'s Guides section | It was the one guide missing. |
| 6 | Added **Vendor Neutrality** | OTLP for traces and metrics, Prometheus text exposition additionally, no Solace-proprietary format, Grafana as a reference implementation. Inline prose and tables, no ADR cross-references, since no ADR documents exist in `docs/`. |
| 7 | Added the missing **no `otel_scope_*` labels** rule | Consolidated with the rules already covered into one place. Real and tested: `WithoutScopeInfo()` in `internal/observability/metrics/provider.go`. |
| 8 | Moved the settled `principal.sub`-only decision out of Open Items, tagged **Q-013** | It was filed as open even though the prose already said "Decided for v1". The genuinely open question, whether a reviewer can resolve `sub` to a human, stays in Open Items as item 3. |

## The most important thing in this PR

**The audit trail does not cover every state change, and the document said it did.**

The `operation` record is gated on the `destructiveHint` annotation, not on whether state changed (`internal/tools/manager.go`). Seven tools change broker state and emit no audit record at all:

`create-message-vpn`, `create-queue`, `create-queue-subscription`, `create-topic-endpoint`, `create-rdp`, `clear-queue-stats`, `clear-client-stats`

A SOC 2 walkthrough querying `audit_event_type="operation"` got a clean, complete-looking result containing zero object creations. That is corrected in all four places that stated it, including the opening overview bullet, with a blockquote naming the trigger and every tool in the gap, and pointing at the `tool invoked` log line as the join for creations.

Widening the audit trigger from "destructive" to "all write tools" is a coverage change to this schema, not a shape change, and is deliberately not made here. Worth a ticket if reviewers agree.

## Where this PR disagrees with the Jira ticket, deliberately

The ticket carries a "2026-09-04 alignment audit" whose specific numbers are superseded. Everything below was re-verified against the code on current `main`.

- **`error_type` stays at twelve values.** `audit.ErrorTypes()` holds twelve, and `internal/tools/audit_error_type_drift_test.go` enforces it from the AST. `broker_permission_denied` was **not** added: it exists only on the unmerged Story 49 branch (#389).
- **`audit_schema` stays at 1.1.** The ticket states 1.0.
- **`metrics_schema` is 1.3**, not the ticket's 1.0 and not the 1.2 this PR originally claimed. #402 bumped it. This PR takes 1.3 from `main` and does not restate it.
- **The CI golden-file check is not built here.** That is SOL-154238, out of scope.

## Other accuracy defects found while verifying, fixed in passing

Beyond the eight approved edits. Each was a claim the code does not support.

1. **The deprecation policy promised dual emission for closed-set value renames, which cannot work.** A record carries one `error_type`, a sample one `outcome`. Emitting both is impossible on a single-valued field and would split the series and double every `sum()` on a metric. Value renames now get notice and a one-step cutover. `audit_event_type` sits in the same bucket for a different, stated reason: it is the record discriminator, so dual emission means two whole records, breaking the one-record-per-call bound and doubling SIEM retention.
2. **"The two schemas version independently" needed the shared-vocabulary exception,** stated by property rather than by package: any closed set rendered by more than one signal (`outcome`, `error_type`, and the five `auth_failure` reasons today) is one coupled breaking change that bumps both majors together.
3. **Unit `1` is silent on a counter but appends `_ratio` on a gauge,** which is why the two info gauges carry no unit at all. The rule said flatly that unit 1 adds no suffix.
4. **`OBS_METRICS_OTLP_ENABLED` does not exist in `internal/config` yet.** The doc claimed that setting it with metrics off "fails config load with an explicit error". It is unread, so the server starts clean, and a clean startup was reading as evidence that push was on.
5. **The saturation section claimed no meter provider or OpenTelemetry dependency exists.** Story 14 made that false; only the missing instruments claim still holds.
6. **Four hop-2 statements said a broker-side denial always coexists with an `operation` record.** True only for a destructive tool. Qualified, so the document does not contradict its own new section.
7. **The OTLP `reason` set is now described as the bound on what the label can contain,** not a list of values you will observe. `queue_full` is already inert on the span pair, and the metrics pair's design reaches only `export_timeout` and `export_error`.
8. **The vendor-neutrality backend list is split into verified and expected.** Claiming a tested matrix for Tempo, Jaeger and Datadog would be unsupported: the per-backend matrix is what Story 40 (SOL-152423) delivers, and it is still To Do. The Prometheus scrape surface genuinely is pinned by test, and says so.

## Rebase correction

Cut from `7d33ebc`; #402 (SOL-152088) merged to `main` before this PR opened and touches the same document. Rebased onto `29a181e`. What changed:

- **The Broker Reachability section is taken from `main` unchanged.** It is live content now, not a caveat. That includes the third gauge `mcp_broker_last_result_timestamp_seconds` and the revised `reason` vocabulary: `credential_invalid` covers 401 **or** 403, `broker_error_NNN` covers 5xx and 429 with the status code in the label, and other 4xx classify as `reachable` because the broker answered. The diff against `main` shows no change to any of it.
- **Two claims in this PR's original description were wrong** and are withdrawn: that `metrics_schema` is 1.2 (it is 1.3), and that the reachability gauges are unregistered so the section needed a `[Not yet emitted]` tag. Both were true at `7d33ebc` and false by the time the PR opened. The `[Not yet emitted]` heading and its blockquote are gone, along with the `#broker-reachability--not-yet-emitted` anchor.
- The Implementation Status row and the Metrics status blockquote keep `main`'s three gauges and layer this PR's span-pair-vs-metrics-pair distinction on top.

`git merge-tree` conflict is resolved; GitHub now reports the PR mergeable.

## Verification

`make check` on the rebased branch: build, vet, lint, race-enabled tests, coverage 89.7% against the 85% floor. Green.

`TestErrorTypeVocabulary_MatchesDocumentedTable`, which parses the markdown `error_type` table and compares it to the code in both directions, passes. `./internal/observability/metrics/` passes.

Every internal anchor resolves to a real heading and every markdown table has consistent pipe counts, both checked programmatically after the rebase.

This went through a skeptical review that read the code at file:line for every claim. It found two HIGH defects (the audit coverage claim and the dual-emit promise) plus five smaller ones, all fixed and re-verified.

No CHANGELOG entry: this touches only `docs/` and `README.md`, not `internal/config/`, tool definitions, or `internal/tools/`.

Closes SOL-152094.

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -1104,8 +1104,14 @@ worked example of the translation.
 
 **Pin every rule to `audit_schema_version`, as [Schema Versioning](#schema-versioning) already
 tells you to.** The four queries below do it explicitly, with the current version as a
-placeholder you maintain. A pinned query that stops matching after an upgrade has detected
-schema drift, which is the point; an unpinned one keeps matching and silently spans two
+placeholder **you must set to your deployment's actual `audit_schema` before the rule goes
+live, and update when it bumps.** A pinned query that stops matching after an upgrade has
+detected schema drift, which is the point. But a rule copied fresh *after* the bump, with no
+before-and-after to notice, is different: it returns zero rows on every run and looks like a
+clean estate rather than a stale pin — check the pinned version against the deployment's
+actual `audit_schema_version` before concluding anything from an empty result, especially on
+query 1, where zero rows reads as "no privileged changes" to whoever is running the review. An
+unpinned query is worse in the other direction: it keeps matching and silently spans two
 contracts.
 
 **1. Destructive-operation review — who changed what.**
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -54,10 +54,12 @@ they do not exist to link to. Each names the story that lands it:
 The Broker MCP Server is designed to emit three observability signals:
 
 - **Metrics**, on a Prometheus `/metrics` endpoint, for dashboards and alerts.
-- **An audit trail**, one JSON event per **destructive** tool call, for compliance evidence.
-  Note "destructive", not "state-changing": object creation is not audited today, and
-  [Audit Trail](#audit-trail--interim--all-record-types-except-broker_authz_denied) names
-  every tool that falls in the gap.
+- **An audit trail**, one JSON event per **destructive** tool call that reaches execution, for
+  compliance evidence. Note "destructive", not "state-changing": object creation is not
+  audited today, and a call that fails broker resolution or argument validation writes no
+  record either — see
+  [Audit Trail](#audit-trail--interim--all-record-types-except-broker_authz_denied) for both
+  gaps.
 - **Distributed traces**, exported over OTLP, for end-to-end request diagnosis.
 
 One correlation ID threads each request through logs, traces, and audit records, so they
@@ -177,9 +179,11 @@ cutover, not as a migration you can run both sides of.
 impossibility argument above does not reach it. It is the record **discriminator**, not a
 field on a record, so dual-emitting it would mean two whole records per event rather than two
 values in one field — technically possible. It still gets notice-then-cutover, because
-emitting both would break the one-record-per-call bound stated under [Audit
-Trail](#audit-trail--interim--all-record-types-except-broker_authz_denied) for the whole
-window, and would double audit volume and retention on the one signal you pay a SIEM to store.
+dual-emitting the discriminator doubles **every** audit record for the whole window, not only
+`operation` records, doubling volume and retention on the one signal you pay a SIEM to
+store — a much larger cost than the [Audit
+Trail](#audit-trail--interim--all-record-types-except-broker_authz_denied) coverage gaps
+this schema already tolerates.
 
 **The two schemas version independently, except where a vocabulary is shared.** A metrics
 rename does not normally force an audit-schema major bump, and a SIEM rule pinned to
@@ -248,7 +252,7 @@ here can be reconciled.
 > `mcp_broker_last_result_timestamp_seconds` (see [Broker Reachability](#broker-reachability)).
 > Assume any other metric below is not yet emitted._
 >
-> **Two metric names below are documented but not emitted by any build yet**, and are marked
+> **Two metric groups below are documented but not emitted by any build yet**, and are marked
 > as such where they are defined: the OTLP **metrics** export-health pair
 > `mcp_otel_metrics_exported_total` / `mcp_otel_metrics_dropped_total{reason}` (see [OTLP
 > Export Health](#otlp-export-health)) and `mcp_broker_authz_denied_total{tool,broker,reason}`
@@ -778,6 +782,14 @@ Events](#authentication-events)). The stream is enabled with `OBS_AUDIT_LOG_ENAB
 > audit trigger from "destructive" to "all write tools" is a change to this schema's coverage,
 > not to its shape, and is not made here.
 
+> **The second gap: a call has to clear broker resolution and argument validation before it is
+> audited.** The record is built only once the arguments hash is computable, which happens
+> after the broker is resolved and the arguments pass schema validation. An unknown tool, a
+> missing or unresolved broker, or a validation failure returns its error normally but writes
+> no `operation` record — the call never reached the destructive gate. Unlike the annotation
+> gap above, this is not a coverage decision; it is a call that never started, so there is
+> nothing yet to audit.
+
 Every event carries a top-level `"event": "audit"` tag so your log shipper can route the
 audit sub-stream to a dedicated SIEM index.
 
@@ -1038,10 +1050,11 @@ separation-of-duties measure. `reason` tells you why without disclosing the call
 entitlements to whoever reads the audit stream.
 
 **Broker-side denial has its own record type: `broker_authz_denied`.** Authorization is
-checked twice on a state-changing call. Hop 1 is this server deciding whether the caller may
-use the tool. Hop 2 is the broker deciding whether the exchanged human identity may perform
-the SEMP operation, and a refusal there emits `audit_event_type: broker_authz_denied`
-carrying `tool`, `broker`, the principal, and a single-value closed `reason` set:
+checked twice on any call that reaches the broker, destructive or read-only. Hop 1 is this
+server deciding whether the caller may use the tool. Hop 2 is the broker deciding whether the
+exchanged human identity may perform the SEMP operation, and a refusal there emits
+`audit_event_type: broker_authz_denied` carrying `tool`, `broker`, the principal, and a
+single-value closed `reason` set:
 
 | `reason` | Meaning |
 |---|---|
@@ -1081,7 +1094,9 @@ In Splunk SPL:
 
 ```
 index=<your_audit_index> event="audit" audit_event_type="operation"
-| stats count, values(tool) as tools, values(broker) as brokers by principal.sub
+| stats count, values(tool) as tools, values(broker) as brokers, values(outcome) as outcomes,
+    values(timestamp_utc) as timestamps, values(correlation_id) as correlation_ids
+    by principal.sub
 ```
 
 Every **destructive** tool call, successful or failed, attributed to the human principal that
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -4,8 +4,9 @@
 > trace schema for the Broker MCP Server. It is published for review **before** the names
 > freeze at GA, so the time to change a name is now. After GA the schema is **additive-only
 > within a MAJOR version**: names are added freely, and a rename or removal happens only at a
-> MAJOR bump, after an announced deprecation with at least two minor releases of dual
-> emission. That is a migration path, not a promise never to change — see
+> MAJOR bump, after an announced deprecation with at least two minor releases of notice,
+> dual-emitting where dual emission is possible. That is a migration path, not a promise never
+> to change — see
 > [Compatibility and Deprecation Policy](#compatibility-and-deprecation-policy) and
 > [How to Give Feedback](#how-to-give-feedback).
 >
@@ -150,9 +151,25 @@ have not had time to migrate.
 
 | Change | Version effect | What you get |
 |---|---|---|
-| A new metric, label, audit field, or a new value in a closed set | MINOR bump | Nothing to do. Existing queries keep working. |
-| Renaming or removing a **name** — a metric, a label key, or an audit field | MAJOR bump, and only after the cycle below | An announcement, then at least two minor releases emitting both the old and the new name. |
+| A new metric, audit field, or a new value in a closed set | MINOR bump | Nothing to do. Existing queries keep working. |
+| A new **label** on an existing metric | MINOR bump | Existing selectors and `sum by (...)` keep working. A `without()`/`ignoring()` aggregation or a default one-to-one `rate(a)/rate(b)` match can silently change, because a new label changes the series identity those collapse or match on — see the note below. |
+| Renaming or removing a metric name or an audit field | MAJOR bump, and only after the cycle below | An announcement, then at least two minor releases emitting both the old and the new form side by side, **untouched** — the old artifact is unaffected, so a dashboard or SIEM rule pinned to it keeps working exactly as before. |
+| Renaming a **label key** | MAJOR bump, and only after the cycle below | An announcement, then at least two minor releases carrying both the old and the new key on every sample. This is a weaker guarantee than the row above: dual emission is possible, but adding the new key still changes the series identity of the metric consumers already query, so — same hazard as the MINOR row above — a `without()`/`ignoring()` aggregation or default vector match against it can silently change during the window. Where an untouched old series matters, ship the new key on a new metric name instead. |
 | Renaming or removing a **value** in a closed set (`outcome`, `error_type`, `reason`, `audit_event_type`) | MAJOR bump, and only after the cycle below | An announcement and at least two minor releases of notice. **Not** dual emission — see the carve-out below. |
+
+**A label change — new or renamed — changes series identity, which some queries do not survive.**
+`mcp_tool_invocation_total{broker="b1"}` and `mcp_tool_invocation_total{broker="b1",
+event_broker="b1"}` are different series. A plain selector or a `sum by (broker)` still
+matches either shape fine. But `sum without (broker) (...)`, `ignoring(broker)`, and a default
+one-to-one `rate(a_total[5m]) / rate(b_total[5m])` all require identical label sets, so the
+added label silently stops them matching or stops them collapsing the way they used to — no
+error, just a different number. A `rate()`/`increase()` window that spans the moment the label
+is added or removed sees the old series go stale and a new one start, which can transiently
+skew both. This is exposition-valid — the Prometheus/OpenMetrics text format does not require
+one label-name set per metric family, and this server's OTel-based exporter does not enforce
+one either — it is a query-correctness hazard, not a wire-format one, and it is why a
+label-key rename cannot promise the same "the old stays untouched" guarantee a name rename
+can.
 
 **The deprecation cycle, in order:**
 
@@ -160,9 +177,12 @@ have not had time to migrate.
    naming the old form, the new form, and the earliest release in which the old one may
    disappear.
 2. **Two minor releases of notice, dual-emitting where dual emission is possible.** For a
-   name, both the old and the new are emitted side by side for the whole window, so a
-   dashboard or SIEM rule written against either keeps working and you can verify the new
-   query against live data before cutting over.
+   metric name or an audit field, both the old and the new are emitted side by side,
+   untouched, for the whole window, so a dashboard or SIEM rule written against either keeps
+   working and you can verify the new query against live data before cutting over. For a
+   label key, both keys are emitted on every sample, which still lets you verify the new query
+   before cutover but — per the series-identity note above — does not leave the old series as
+   untouched as a name rename does.
 3. **Remove at a MAJOR bump.** The old form is removed only in a major version of the affected
    schema (`metrics_schema` or `audit_schema`), never in a minor.
 
@@ -1082,10 +1102,16 @@ routed by `event="audit"`, because that is the one filter every shipper applies;
 predicate into your own SIEM's syntax. A Splunk SPL rendering is shown for the first as a
 worked example of the translation.
 
+**Pin every rule to `audit_schema_version`, as [Schema Versioning](#schema-versioning) already
+tells you to.** The four queries below do it explicitly, with the current version as a
+placeholder you maintain. A pinned query that stops matching after an upgrade has detected
+schema drift, which is the point; an unpinned one keeps matching and silently spans two
+contracts.
+
 **1. Destructive-operation review — who changed what.**
 
 ```
-event="audit" AND audit_event_type="operation"
+event="audit" AND audit_schema_version="1.1" AND audit_event_type="operation"
   → group by principal.sub
   → report tool, broker, outcome, timestamp_utc, correlation_id
 ```
@@ -1093,7 +1119,7 @@ event="audit" AND audit_event_type="operation"
 In Splunk SPL:
 
 ```
-index=<your_audit_index> event="audit" audit_event_type="operation"
+index=<your_audit_index> event="audit" audit_schema_version="1.1" audit_event_type="operation"
 | stats count, values(tool) as tools, values(broker) as brokers, values(outcome) as outcomes,
     values(timestamp_utc) as timestamps, values(correlation_id) as correlation_ids
     by principal.sub
@@ -1113,7 +1139,7 @@ made it. Two limits to state before a reviewer treats this as a complete record 
 **2. Rejected credentials — who could not get in.**
 
 ```
-event="audit" AND audit_event_type="auth_failure"
+event="audit" AND audit_schema_version="1.1" AND audit_event_type="auth_failure"
   → group by reason
 ```
 
@@ -1126,7 +1152,7 @@ Events](#authentication-events)).
 **3. Denied privileged attempts, hop 1 — this server refused the tool.**
 
 ```
-event="audit" AND audit_event_type="authz_denied"
+event="audit" AND audit_schema_version="1.1" AND audit_event_type="authz_denied"
   → group by reason, tool
 ```
 
@@ -1136,7 +1162,7 @@ is no `operation` record and no tool-invocation metric sample for them.
 **4. Denied privileged attempts, hop 2 — the broker refused the operation.**
 
 ```
-event="audit" AND audit_event_type="broker_authz_denied"
+event="audit" AND audit_schema_version="1.1" AND audit_event_type="broker_authz_denied"
   → group by reason, tool, broker
 ```
 
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -2,8 +2,12 @@
 
 > **Status: Draft for pilot review.** This document is the proposed metric, audit, and
 > trace schema for the Broker MCP Server. It is published for review **before** the names
-> freeze at GA. After GA we commit to only ever *adding* to this schema, never renaming, so
-> the time to change a name is now. See [How to Give Feedback](#how-to-give-feedback).
+> freeze at GA, so the time to change a name is now. After GA the schema is **additive-only
+> within a MAJOR version**: names are added freely, and a rename or removal happens only at a
+> MAJOR bump, after an announced deprecation with at least two minor releases of dual
+> emission. That is a migration path, not a promise never to change — see
+> [Compatibility and Deprecation Policy](#compatibility-and-deprecation-policy) and
+> [How to Give Feedback](#how-to-give-feedback).
 >
 > **Most of the metrics and trace surface is not emitted by the current build.** See the
 > [Implementation Status](#implementation-status) table for what is live; anything marked
@@ -22,10 +26,38 @@
 > [Load and Saturation Visibility](#load-and-saturation-visibility--interim--logs-only).
 > The metric form remains roadmap.
 
+## Start Here
+
+| If you are here to… | Go to |
+|---|---|
+| Find out what is actually live today | [Implementation Status](#implementation-status) |
+| Tell us to rename something before the freeze | [How to Give Feedback](#how-to-give-feedback) |
+| Understand naming rules, units, and what we commit to | [Conventions](#conventions) · [Compatibility and Deprecation Policy](#compatibility-and-deprecation-policy) |
+| Build a Grafana dashboard or an alert rule | [Metrics](#metrics--planned-with-exceptions) |
+| Write a SIEM rule for compliance evidence | [Audit Trail](#audit-trail--interim--all-record-types-except-broker_authz_denied) · [Canonical Audit Queries](#canonical-audit-queries) |
+| Diagnose one slow or failed call end to end | [Distributed Tracing](#distributed-tracing--interim-request-path-and-per-attempt-spans-wired) · [Correlation ID](#correlation-id--implemented) |
+| Look up what `outcome` or `error_type` means | [The Outcome Vocabulary](#the-outcome-vocabulary) |
+| Check this works with your existing stack | [Vendor Neutrality](#vendor-neutrality) |
+| Deploy to Kubernetes | [Deployment Topology and Resource Policy](#deployment-topology-and-resource-policy--implemented) |
+| See what is still open, and what we already decided | [Open Items for This Review](#open-items-for-this-review) · [Planned for a Later Release](#planned-for-a-later-release-not-frozen-in-this-review) |
+| Map this to PCI DSS, SOC 2, SOX, or ISO 27001 | [Standards This Schema Supports](#standards-this-schema-supports) |
+| Understand load shedding and saturation | [Load and Saturation Visibility](#load-and-saturation-visibility--interim--logs-only) |
+
+**Not in this document yet.** These are deliberately listed as plain text, not links, because
+they do not exist to link to. Each names the story that lands it:
+
+- Tracing setup and the reference OTel collector deployment, with a tested-backend matrix —
+  lands with Story 40 (SOL-152423).
+- Operator runbook by failure mode — lands with Story 36 (SOL-152098).
+- Reference SLO sheet — lands with Story 38.
+
 The Broker MCP Server is designed to emit three observability signals:
 
 - **Metrics**, on a Prometheus `/metrics` endpoint, for dashboards and alerts.
-- **An audit trail**, one JSON event per state-changing operation, for compliance evidence.
+- **An audit trail**, one JSON event per **destructive** tool call, for compliance evidence.
+  Note "destructive", not "state-changing": object creation is not audited today, and
+  [Audit Trail](#audit-trail--interim--all-record-types-except-broker_authz_denied) names
+  every tool that falls in the gap.
 - **Distributed traces**, exported over OTLP, for end-to-end request diagnosis.
 
 One correlation ID threads each request through logs, traces, and audit records, so they
@@ -46,7 +78,7 @@ capability headings carry the same tag:
 | Capability | Status | Notes |
 |---|---|---|
 | Correlation ID | **[Implemented]** | Wired and on by default (`OBS_CORRELATION_ID_ENABLED`). |
-| Metrics | **[Planned, with exceptions]** | Most instrument names and labels here are still the proposal under review. Wired and emitted today: the `/metrics` endpoint itself, `mcp_build_info`, `mcp_schema_version`, `mcp_metrics_scrape_total`, `mcp_http_active_requests`, `mcp_tool_invocation_total`, `mcp_tool_invocation_duration_seconds`, `mcp_semp_request_total`, `mcp_semp_request_duration_seconds`, the OTLP export-health counters, `mcp_panic_recovered_total` (see [Panic Recovery](#panic-recovery--implemented)), `mcp_auth_failure_total` and `mcp_authz_denied_total` (see [Authentication Failures](#authentication-failures--implemented) and [Authorization Denials](#authorization-denials--implemented)), the `go_*`/`process_*` runtime collectors (see [Go Runtime and Process Metrics](#go-runtime-and-process-metrics)), and `mcp_broker_reachable`, `mcp_broker_unreachable_reason`, and `mcp_broker_last_result_timestamp_seconds` (see [Broker Reachability](#broker-reachability)). Assume any other metric below is not yet emitted. |
+| Metrics | **[Planned, with exceptions]** | Most instrument names and labels here are still the proposal under review. Wired and emitted today: the `/metrics` endpoint itself, `mcp_build_info`, `mcp_schema_version`, `mcp_metrics_scrape_total`, `mcp_http_active_requests`, `mcp_tool_invocation_total`, `mcp_tool_invocation_duration_seconds`, `mcp_semp_request_total`, `mcp_semp_request_duration_seconds`, the OTLP **span** export-health counters (`mcp_otel_spans_exported_total` / `mcp_otel_spans_dropped_total` — the metrics pair is **not** emitted yet, see [OTLP Export Health](#otlp-export-health)), `mcp_panic_recovered_total` (see [Panic Recovery](#panic-recovery--implemented)), `mcp_auth_failure_total` and `mcp_authz_denied_total` (see [Authentication Failures](#authentication-failures--implemented) and [Authorization Denials](#authorization-denials--implemented)), the `go_*`/`process_*` runtime collectors (see [Go Runtime and Process Metrics](#go-runtime-and-process-metrics)), and `mcp_broker_reachable`, `mcp_broker_unreachable_reason`, and `mcp_broker_last_result_timestamp_seconds` (see [Broker Reachability](#broker-reachability)). `mcp_broker_authz_denied_total` is documented but **not** emitted yet (see [Broker-Side Authorization Denials](#broker-side-authorization-denials--not-yet-emitted)). Assume any other metric below is not yet emitted. |
 | Audit trail | **[Interim — all record types except `broker_authz_denied`]** | Destructive tool calls emit an `operation` record behind `OBS_AUDIT_LOG_ENABLED` (default off). `auth_success`, `auth_failure`, `authz_denied`, and `broker_auth_retry` also emit today (SOL-152097). `broker_authz_denied` and the `mcp_audit_events_dropped_total` counter are not emitted yet. See [Audit Trail](#audit-trail--interim--all-record-types-except-broker_authz_denied). |
 | Distributed tracing | **[Interim — request-path and per-attempt spans wired]** | Tracer provider, OTLP export, W3C context propagation, and spans at the HTTP boundary, the tool dispatcher, the composite executor, each SEMP call, each SEMP *attempt*, and each token-exchange attempt are live behind `OBS_TRACING_ENABLED`, with the retry attributes on the attempt spans. Trace exemplars linking the latency histograms to these traces are live too (Story 47, SOL-152419) — see [Trace Exemplars](#trace-exemplars--implemented). See [Distributed Tracing](#distributed-tracing--interim-request-path-and-per-attempt-spans-wired). |
 | Saturation visibility | **[Interim — logs only]** | Shipped as structured log lines behind `OBS_SATURATION_EVENTS_ENABLED`, **not** as the metric this schema describes. See [Load and Saturation Visibility](#load-and-saturation-visibility--interim--logs-only). |
@@ -75,8 +107,10 @@ Specifically:
 
 Send comments through your Solace pilot channel or to the contacts in the observability
 brief. The review window runs up to the GA freeze. Anything we do not hear back on ships as
-drafted; because the schema is additive-only, we can always add fields later, we just will
-not rename what is already there.
+drafted. Adding a field later is always cheap. Renaming one after the freeze is not: it costs
+you a deprecation cycle and a dashboard migration, even though we commit to giving you both.
+So a name you would change is worth flagging now. See
+[Compatibility and Deprecation Policy](#compatibility-and-deprecation-policy).
 
 ---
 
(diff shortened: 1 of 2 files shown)


docs/observability.md
```diff
@@ -22,10 +26,38 @@
...
+- Tracing setup and the reference OTel collector deployment, with a tested-backend matrix —
+  lands with Story 40 (SOL-152423).
+- Operator runbook by failure mode — lands with Story 36 (SOL-152098).
+- Reference SLO sheet — lands with Story 38.
+
 The Broker MCP Server is designed to emit three observability signals:
 
 - **Metrics**, on a Prometheus `/metrics` endpoint, for dashboards and alerts.
-- **An audit trail**, one JSON event per state-changing operation, for compliance evidence.
+- **An audit trail**, one JSON event per **destructive** tool call, for compliance evidence.
```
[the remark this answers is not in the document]
solace-aross: Fixed in fa6bdf0. Added a caveat in the Audit Trail section: a call that fails broker resolution or argument validation never reaches the destructive gate, so it writes no `operation` record. Tightened the overview bullet to say "that reaches execution" and point at both gaps.

docs/observability.md
```diff
@@ -805,13 +1041,106 @@ carrying `tool`, `broker`, the principal, and a single-value closed `reason` set
...
 The two hops read differently in a query, deliberately. A hop-1 denial means nothing was
-attempted. A hop-2 denial means the call *ran* and the broker stopped it, so it comes with an
-`operation` record of its own carrying `outcome: error`. **Do not write a query that assumes
-one audit record per call** — a hop-2 denial produces two, and both are correct. Match on
-`audit_event_type` instead of counting.
+attempted. A hop-2 denial means the call *ran* and the broker stopped it, so on a destructive
+tool it comes with an `operation` record of its own carrying `outcome: error`. **Do not write
+a query that assumes one audit record per call** — a hop-2 denial on a destructive tool
+produces two, and both are correct. (On a non-destructive tool it produces one, since that
+tool never writes an `operation` record at all.) Match on `audit_event_type` instead of
```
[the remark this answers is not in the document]
solace-aross: Fixed in fa6bdf0. Changed "checked twice on a state-changing call" to "on any call that reaches the broker, destructive or read-only" — matches the non-destructive-tool case documented a few lines below.

docs/observability.md
```diff
@@ -103,8 +137,96 @@ Two independent versions are published, so your queries can pin to a version and
...
+and reserved, then the switch happens in one step at the MAJOR bump. Plan a value rename as a
+cutover, not as a migration you can run both sides of.
+
+`audit_event_type` is in this bucket for a different reason, worth stating because the
+impossibility argument above does not reach it. It is the record **discriminator**, not a
+field on a record, so dual-emitting it would mean two whole records per event rather than two
+values in one field — technically possible. It still gets notice-then-cutover, because
+emitting both would break the one-record-per-call bound stated under [Audit
+Trail](#audit-trail--interim--all-record-types-except-broker_authz_denied) for the whole
+window, and would double audit volume and retention on the one signal you pay a SIEM to store.
```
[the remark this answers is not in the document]
solace-aross: Fixed in fa6bdf0. Dropped the "breaks the one-record-per-call bound" framing — that bound only covers `operation` records and already has the hop-2 exception. The real cost of dual-emitting the discriminator is doubling every audit record, not just operation ones; reworded to say that.

docs/observability.md
```diff
@@ -115,13 +237,23 @@ avoid. Pin dashboards to `mcp_schema_version` and SIEM queries to `audit_schema_
...
+> `mcp_semp_request_duration_seconds`, the OTLP **span** export-health counters
+> (`mcp_otel_spans_exported_total` / `mcp_otel_spans_dropped_total`),
 > `mcp_panic_recovered_total` (see [Panic Recovery](#panic-recovery--implemented)),
 > `mcp_auth_failure_total` and `mcp_authz_denied_total` (see
 > [Authentication Failures](#authentication-failures--implemented) and
 > [Authorization Denials](#authorization-denials--implemented)), and the
 > `go_*`/`process_*` runtime collectors (see [Go Runtime and Process Metrics](#go-runtime-and-process-metrics)). Assume
 > any other metric below is not yet emitted._
+>
+> **Two metric names below are documented but not emitted by any build yet**, and are marked
```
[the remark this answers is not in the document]
solace-aross: Fixed in fa6bdf0. Changed to "two metric groups" — the OTLP pair plus `mcp_broker_authz_denied_total` is three names.

docs/observability.md
```diff
@@ -805,13 +1041,106 @@ carrying `tool`, `broker`, the principal, and a single-value closed `reason` set
...
+event="audit" AND audit_event_type="operation"
+  → group by principal.sub
+  → report tool, broker, outcome, timestamp_utc, correlation_id
+```
+
+In Splunk SPL:
+
+```
+index=<your_audit_index> event="audit" audit_event_type="operation"
+| stats count, values(tool) as tools, values(broker) as brokers by principal.sub
```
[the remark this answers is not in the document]
solace-aross: Fixed in fa6bdf0. SPL example now reports `outcome`, `timestamp_utc`, and `correlation_id` alongside tool/broker, matching the query description above it.

solace-aross:
Heads-up for whoever picks up #389 (SOL-153332) or #391 (SOL-152418) next: this PR finalizes docs/observability.md, including the sections your PRs will need to flip from "[Not yet emitted]" to live once you merge (broker_authz_denied metric for #389, the mcp_otel_metrics_* pair for #391). Whichever of you lands second will hit a conflict here similar to the one this PR itself just resolved against #402 — plan to rebase over this and update the corresponding caveat section rather than re-deriving it from scratch. Cross-posted on #389 and #391.

docs/observability.md
```diff
@@ -124,6 +251,15 @@ avoid. Pin dashboards to `mcp_schema_version` and SIEM queries to `audit_schema_
 > and `mcp_broker_reachable`, `mcp_broker_unreachable_reason`, and
 > `mcp_broker_last_result_timestamp_seconds` (see [Broker Reachability](#broker-reachability)).
 > Assume any other metric below is not yet emitted._
+>
+> **Two metric groups below are documented but not emitted by any build yet**, and are marked
```
bczoma, line 275: 🟡 **About fourteen statements here are correct today and become wrong when #389 merges, and both PRs edit this file**

**Flagged by** parent review

You already name this in the description — "`broker_permission_denied` … exists only on the unmerged Story 49 branch (#389)" — so this is not news, and every statement in this PR is accurate at `3d5f5ec`. What is missing is a stated reconciliation, and the exposure is larger than the one value.

#389 is open, approved, and has just moved `AuditSchemaVersion` to `1.2` and `MetricsSchemaVersion` to `1.4`, added `broker_permission_denied` as a thirteenth `error_type`, made it six of thirteen that reach an audit record, and given `mcp_broker_authz_denied_total` a production emit site. Against that, this document at this head says:

- `:255-259` — `mcp_broker_authz_denied_total` is "documented but not emitted by any build yet" (now "Two metric groups"), its series "**absent, not zero**", and an `absent()` alert on it "fires today for the mundane reason that the instrument does not exist"
- `:83`, `:486`, `:1074` — the same, including a heading and an anchor slug that both contain `not-yet-emitted`
- `:135-136` — `metrics_schema` (current: **1.3**), `audit_schema` (current: **1.1**)
- `:340`, `:345`, `:833`, `:1354`, `:1380`, `:1572`, `:1589`, `:1840-1841` — `error_type` is a twelve-value set, five of which reach an audit record

**Why it matters, concretely.** Two of those are advice a reader acts on. An operator who takes "absent, not zero — an `absent()` alert fires today because the instrument does not exist" will not build a `rate()` alert on that counter, so once #389 lands, broker-side authorization denials emit into a metric nobody watches — a security signal, silently unmonitored, in the document whose whole purpose is telling operators what to alert on. And a SIEM author following `:136` pins `audit_schema_version` to `1.1` while every record carries `1.2`, so the query matches nothing; that is the same failure mode I raised on #389 for its CHANGELOG.

**Why git will not catch all of it.** #389 edits this file too, so some hunks conflict and force attention. But this PR rewrites whole sections around these statements — the `error_type` table region and the version lines among them — and a conflict resolution that takes this branch's side wholesale is the natural one, precisely because this side is the newer and more thorough text. The eight "twelve" mentions are spread across five sections; they will not all surface as conflicts.

**Recommended fix — nothing to change in the prose now.** Changing it today would make the document wrong today. What I am asking for is one of:

1. Say in a comment which PR merges first and who flips these statements in a follow-up (a one-line reply resolves this thread), or
2. Land #389 first and rebase this onto it, flipping the fourteen statements as part of the rebase — my preference, since it puts the reconciliation in the hands of the person who wrote the text rather than whoever resolves the conflict.

Either is fine. What should not happen is both merging on their own and the reconciliation being discovered from a customer's dashboard.

bczoma (approved):
## PR Review: #405 — SOL-152094: finalize the versioned metric and audit schema doc (Story 35)

**Author:** solace-aross | **Branch:** `andrea/SOL-152094-observability-schema-finalize` → `main` | **Changes:** +472 / -62 across 2 files
**Jira:** `SOL-152094` — finalize the versioned metric and audit schema doc (Story 35)

Reviewed at `fa6bdf0`. The head moved twice while I was reading; the verification below was re-run against this commit.

---

### Verdict

I verified all eleven code-backed claims this PR makes against source rather than reading the description. **Every one of them holds**, including the two the ticket contradicts and the headline seven-tool audit gap, which is exactly right — not one tool missing, not one read-only tool wrongly included. `fa6bdf0` also closes all five of @Copilot's items, and I confirmed each of those fixes rather than taking the commit subject for it.

1. 🟡 `docs/observability.md:255` — about fourteen statements here are true at this head and become false the moment #389 merges; the two PRs edit the same file and no reconciliation is stated

The 4 🟢/💡 notes below are **not** conditions of this verdict.

### Summary

This is the strongest documentation PR I have reviewed in this repo, and the reason is that its claims survive checking. The seven-tool audit gap — the headline — was verified by enumerating every write tool in `tools.yaml`: 18 carry `readOnly: false`, all 18 carry an explicit `destructive:` value, and the seven with `destructive: false` are exactly the seven named. `delete-queue-subscription` is `destructive: true` and correctly absent; the three native SEMPv1 handlers are all `ReadOnlyAnnotations()`. The join the doc points readers at holds too: `logToolResult` runs from an unconditional defer, so `tool invoked` is emitted for those calls regardless of the destructive gate. `error_type` is twelve at this head with the AST drift guard enforcing it; "five of the twelve reach an audit record" is right for a non-obvious reason — the audit gate sits *after* broker resolution and validation, so the five pre-gate values cannot reach it, which is also exactly the second gap `fa6bdf0` now documents. The schema constants are `1.3`/`1.1`, `WithoutScopeInfo()` is set and pinned by a fidelity test, `queue_full` really is defined-but-never-incremented, and `OBS_METRICS_OTLP_ENABLED` is genuinely unread repo-wide — zero hits in any `.go` file — so the old "fails config load" claim was false and the correction is right. All 83 in-document links resolve against all 50 headings with no slug drift, and the three relative links resolve to real files and headings.

Two things I went looking for and did not find, for the record. I expected the `OBS_METRICS_OTLP_ENABLED` paragraph to be a live self-contradiction, since it still states the config-load check the description calls false — it is not, because the blockquote immediately beneath scopes that whole paragraph, the check included, as Story 46 design. And I expected the dual-emission carve-out to still rest on a bound the document elsewhere scopes away; `fa6bdf0` dropped that clause and leans on the volume argument, which was the right resolution.

### Minor notes

- 🟢 `docs/observability.md:58` — the overview bullet still describes the annotation gap as *"object creation is not audited today"*, which is 5 of the 7 tools; `clear-queue-stats` and `clear-client-stats` are not creations. The authoritative list and the query caveat both say "the `create-*` tools **and** the `clear-*-stats` tools". Statistics-clearing is the evidence-destroying half of that gap, and this is the one place a compliance reader is most likely to stop. "object creation and statistics clearing are not audited today".
- 🟢 `docs/observability.md:803` — the hop-2 coexistence rule is in the present tense ("produces its `operation` record *and* a `broker_authz_denied` record") for a record type with no production emit site at this head. The `(SOL-153332)` pointer helps, but the not-yet-emitted marker used at `:82`/`:255` is absent at the point of use, so a SIEM author reading this paragraph alone writes a rule against a record that never arrives. Four words: "(not yet emitted, SOL-153332)".
- 🟢 `docs/observability.md:227` — "unit `1` is silent on a counter but appends `_ratio` on a gauge" has nothing behind it in this repo: no gauge declares unit `1` and no fixture exercises the case, so an OTel bump could falsify it with nothing failing. The counter half *is* pinned by the golden file. Either attribute it to the exporter or add a one-line fidelity case registering a unit-`1` gauge so the golden file pins it.
- 💡 `docs/observability.md:227` — the same row says "the **two** info gauges … are registered with no unit"; `mcp_broker_reachable`, `mcp_broker_unreachable_reason` and `mcp_broker_last_result_timestamp_seconds` are also declared without a unit, so the sentence reads as exhaustive in a document whose value is exhaustiveness. Relatedly, the seven-tool rationale explains only the `create-*` half — "creating an object is additive" does not cover the two `clear-*-stats` tools, which are non-destructive because they reset counters, so a reader inferring "additive ⇒ non-destructive" will mispredict which tools are audited.

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ — see the 🟡; internal consistency is otherwise clean, and `fa6bdf0` closed the five open items | docs |
| Claim accuracy vs code | ✅ — all 11 verified claims confirmed | claim verification |
| Correctness | — not reviewed (no Go code in the diff) | good:code |
| Code Quality | — not reviewed (no Go code in the diff) | good:code |
| Test Coverage | — not reviewed (no Go code in the diff) | good:code |
| Architecture | — not reviewed (docs-only) | good:architecture |
| Security | — not reviewed (docs-only; the audit-coverage correction was checked as a claim, not as a control) | good:security |
| PR Scope & Description | — not reviewed this pass | good:pr |

docs/observability.md
solace-lootii: Agree — verified this independently against #391's actual code, not just the claim. The same staleness extends to the OTLP side too: the paragraphs around `OBS_METRICS_OTLP_ENABLED` say it "does not exist in the current build" and is "simply unread" — checked #391's `internal/observability/metrics/otlp.go` and `internal/config/observability.go` directly, and that branch does add `MetricsOTLPEnabled`/read the flag, so those paragraphs go stale the moment #391 merges too, on top of the fourteen statements you listed for #389.

One thing I could independently confirm as already accurate, for what it's worth: the reason-vocabulary design claim ("the design for the metrics pair expects only export_timeout and export_error to be reachable there") checks out exactly against #391's otlp.go — reasonQueueFull and reasonShutdown are declared but have no call site on that pair, matching the doc's claim precisely.

solace-lootii:
Went through this the same way as the other two PRs — verified the code-backed claims directly, and cross-checked against what's already been said here rather than re-deriving it.

**Agree with @bczoma** on the headline finding: independently re-derived the seven-tool audit gap from `tools.yaml` on `main` (grepped every `destructive: false` entry) and got the identical seven tools. Also independently confirmed "`error_type` stays at twelve values, five reach an audit record" against `audit.ErrorTypes()` and the destructive-audit test table on `main` — matches exactly.

**Agree with @bczoma's 🟡 condition and @solace-aross's heads-up** that this doc goes stale the moment #389/#391 merge — replied on that thread with the #391-specific addition (the `OBS_METRICS_OTLP_ENABLED` paragraphs go stale too, verified against #391's actual code).

One thing that doesn't look caught yet: **the second commit's title and @bczoma's approval both say Copilot's review comments are closed, and on direct re-check most of them aren't.**

The commit is titled "fix audit-scope and consistency gaps flagged by review... Addresses copilot-pull-request-reviewer's five review comments," and the approval says "`fa6bdf0` also closes all five of @Copilot's items, and I confirmed each of those fixes rather than taking the commit subject for it." I checked the actual current head (`fa6bdf0`, confirmed via `gh pr view --json headRefOid`) against each of Copilot's 7 listed comments, and 5 are still textually present:

1. **README.md:110** — unchanged; the second commit never touches `README.md` at all. Still reads "the versioned metric, audit, **and trace** schema," which Copilot flagged as overclaiming a trace schema version that doesn't exist.
2. **Compatibility table's "label key" row** — still promises dual-emission for a label-key rename; Copilot's Prometheus-exposition-validity objection is unaddressed.
3. **Same table** — still doesn't state what happens for a unit/type/histogram-bucket change; table shape unchanged.
4. **Opening blockquote** (top of doc) — still says a rename gets "two minor releases of dual emission" unconditionally, still contradicting the carve-out below it that says closed-set values get notice-only.
5. **`audit_event_type` "not a field on a record"** — still present verbatim (line 179 in the current head), still technically self-contradicts Event Fields, which documents it as a JSON field.

Only one of the 7 (the reason-set-differs-per-pair caveat) is resolved in the current text, and that one looks like it was already in the *first* commit before Copilot reviewed, not something the fix-up added.

Not raising this as new documentation defects — those are Copilot's own findings, already on record. Raising it because the approval's verification claim doesn't hold up on direct re-check, and that's worth knowing before this merges.

docs/observability.md
```diff
@@ -2,8 +2,12 @@
 
 > **Status: Draft for pilot review.** This document is the proposed metric, audit, and
 > trace schema for the Broker MCP Server. It is published for review **before** the names
-> freeze at GA. After GA we commit to only ever *adding* to this schema, never renaming, so
-> the time to change a name is now. See [How to Give Feedback](#how-to-give-feedback).
+> freeze at GA, so the time to change a name is now. After GA the schema is **additive-only
+> within a MAJOR version**: names are added freely, and a rename or removal happens only at a
+> MAJOR bump, after an announced deprecation with at least two minor releases of dual
```
bczoma: 🟡 **The opening commitment promises dual emission the carve-out below refuses**

**Flagged by** Copilot's suppressed list (never threaded, so nothing has been tracking it), re-verified by me at `e45e7fc`

The blockquote says a rename or removal happens "only at a MAJOR bump, after an announced deprecation with at least two minor releases of dual emission" — unconditionally. But `:155` says a closed-set value rename gets "an announcement and at least two minor releases of notice. **Not** dual emission", and the carve-out beneath it explains why dual-emitting a single-valued field is impossible. Your own cycle text at `:164` already has the right qualifier: "dual-emitting **where dual emission is possible**".

**Why it matters.** This blockquote is the summary a reader takes away — it is the first thing in the document and the part that gets quoted into a release note. As written it promises a migration path for `outcome`/`error_type`/`reason`/`audit_event_type` renames that the policy explicitly does not offer, and a consumer who plans a value migration on the strength of it discovers at cutover that there is no window to verify against live data.

**Recommended fix.** Borrow the qualifier you already use at `:164`:

> after an announced deprecation with at least two minor releases of notice, dual-emitting where dual emission is possible

Four words, and the summary stops contradicting the table.

docs/observability.md
```diff
@@ -103,8 +139,98 @@ Two independent versions are published, so your queries can pin to a version and
...
+### Compatibility and Deprecation Policy
+
+The commitment is **additive-only within a MAJOR version**, not "never rename". A name can
+change. It cannot change without notice, and it cannot change out from under a dashboard you
+have not had time to migrate.
+
+| Change | Version effect | What you get |
+|---|---|---|
+| A new metric, label, audit field, or a new value in a closed set | MINOR bump | Nothing to do. Existing queries keep working. |
+| Renaming or removing a **name** — a metric, a label key, or an audit field | MAJOR bump, and only after the cycle below | An announcement, then at least two minor releases emitting both the old and the new name. |
```
bczoma: 🟡 **A label-key rename cannot be dual-emitted, so this row promises something Prometheus will not accept**

**Flagged by** Copilot's suppressed list (never threaded), re-verified by me at `e45e7fc`

This row covers "a metric, a label key, or an audit field" and promises "at least two minor releases emitting both the old and the new name". That works for a metric name (two families side by side) and for an audit field (two JSON keys on one record). It does not work for a label key: all samples in a metric family must carry the same label-name set, so emitting `broker` and `event_broker` together under one metric name produces an inconsistent exposition rather than a migration path. Doing it properly means a *second metric name* — which is the metric-rename case, a different and larger API change than the row implies.

**Why it matters.** This is the commitment table, the part of the document that functions as a promise rather than a description. A maintainer who plans a label-key rename against this row will either ship an invalid exposition or discover mid-deprecation that the promised window does not exist and has to be re-announced as a metric rename.

**Recommended fix.** Split the row, or give label keys their own line naming the real strategy:

| Renaming or removing a **label key** | MAJOR bump, and only after the cycle below | An announcement and at least two minor releases of notice. **Not** dual emission — a metric family cannot carry two label-name sets; a label-key rename ships as a new metric name alongside the old, or as a cutover at the MAJOR bump. |

Either way the fix is to stop the word "both" from covering label keys.

docs/observability.md
```diff
@@ -802,23 +1050,119 @@ separation-of-duties measure. `reason` tells you why without disclosing the call
...
+tool never writes an `operation` record at all.) Match on `audit_event_type` instead of
+counting.
 
 > **Not emitted yet.** The record type is part of the schema and the constructor accepts it,
 > so a SIEM rule can be written against it today. The emission site ships with SOL-153332.
+> Its metric counterpart, `mcp_broker_authz_denied_total`, is documented under [Broker-Side
+> Authorization Denials](#broker-side-authorization-denials--not-yet-emitted) and ships in the
+> same story.
+
+### Canonical Audit Queries
```
bczoma, line 1097: 🟡 **The queries operators are told to copy are the ones that do not pin the schema version**

**Flagged by** Copilot's suppressed list (never threaded), re-verified by me at `e45e7fc`

`:143` tells the reader "Pin dashboards to `mcp_schema_version` and SIEM queries to `audit_schema_version`", and `:839` documents the field as existing "for query pinning". None of the four canonical queries in this section, nor the worked SPL rendering, includes an `audit_schema_version` predicate — they filter on `event="audit"` and `audit_event_type` only.

**Why it matters, concretely, and soon.** This section is explicitly the copyable compliance guidance — "a SOC 2 walkthrough, or a security investigation". A rule copied from here has no drift detection: when the audit schema moves, it keeps matching and silently combines records written under two different contracts instead of failing loudly. That is not hypothetical here. I checked #389 at `61dbc70`: `internal/observability/schema/version.go:54` already carries `AuditSchemaVersion = "1.2"`, and `:40` carries `MetricsSchemaVersion = "1.4"`. The moment that lands, every rule copied from this section spans 1.1 and 1.2 records with no signal — which is the same failure mode as the `:255` thread from my last pass, arriving through the examples rather than through the prose.

**Recommended fix.** Add the pin to the predicate form and to the SPL example, with the version as an explicit placeholder so it is obviously something the operator maintains:

```
event="audit" AND audit_schema_version="1.1" AND audit_event_type="operation"
  → group by principal.sub
  → report tool, broker, outcome, timestamp_utc, correlation_id
```

One line in the preamble is worth having too: that the pin is deliberate, and that a query which stops matching after an upgrade has detected schema drift rather than broken.

bczoma (dismissed):
## PR Review: #405 — SOL-152094: finalize the versioned metric and audit schema doc (Story 35)

**Author:** solace-aross | **Branch:** `andrea/SOL-152094-observability-schema-finalize` → `main` | **Changes:** +472 / -62 across 2 files
**Jira:** `SOL-152094` — Story 35

---

### Verdict

Second pass, at `e45e7fc`. **Your authored content has not changed since I approved it** — I checked rather than assumed, and the detail is worth stating because it misleads a SHA-based check. The force-push at 19:47 rebased onto `1f41fe6` (#404), and the only text that moved was the cross-signal-vocabulary passage at `:1330-1346`, which this branch *adopted from main* rather than wrote. Its own diff against the new base is byte-identical to what I read at `fa6bdf0`. So nothing in the new push needs re-litigating.

What does need attention is a gap in **my previous review**, and @solace-lootii is right to have pushed on it. Copilot filed **5 inline comments and 7 further items in its review body** under "Suppressed comments". My approval said `fa6bdf0` "closes all five of @Copilot's items" — that was true of the 5 inline ones, and I re-verified them at this head (the SPL example now does report `outcome`, `timestamp_utc` and `correlation_id`). It said nothing about the 7 suppressed items, which I never examined and which create no threads, so nothing has been tracking them. **5 of those 7 are still open at this head.** The three below are the ones I think are worth acting on; the rest are minor notes.

1. 🟡 `docs/observability.md:7` — the opening commitment promises dual emission for every rename; the carve-out at `:155` explicitly refuses it for closed-set values, and the cycle text at `:164` already says "where dual emission is possible"
2. 🟡 `docs/observability.md:154` — `a label key` sits in the dual-emission row, but a metric family cannot validly carry two label-name sets at once; label-key renames need a cutover, not dual emission
3. 🟡 `docs/observability.md:1077` — the Canonical Audit Queries are the copyable compliance guidance and none of them pins `audit_schema_version`, which `:143` instructs operators to pin; #389 moves `audit_schema` to `1.2`

**One FYI that is not yours to fix.** The passage this branch just adopted from main is wrong where it counts: `:1332-1333` claims all four shared keys "hold identical values on the span, on the `mcp_tool_invocation_total` series, and on the `tool invoked` log line", and that is false for `broker` — `request_path_spans_test.go:771-776` excludes `broker` from the log-line join by design, which the very next paragraph then admits. A reader meets the claim and its retraction two sentences apart. The accurate form is "all four across the span and the metric; three of the four on the log line". It arrived via #404 (`1f41fe6`), which is already merged, so it is outside this PR's diff and I have not conditioned anything on it — but you own this document, so you are the right person to know. Same passage, same origin: "the raw caller-supplied alias" holds only on the resolution-failure path (`manager.go:329-335` replaces the alias with the configured display name on success), and "across a success, a failure and a recovered panic" undercounts the audit test, which has a fourth case — the non-canonical alias `dev-eu`→`Dev-EU`, the only case where the `broker` join proves anything.

### Summary

I re-verified the new push rather than reading the commit subject, and it is a rebase that adopts main's text, not new authorship. The document's own claims still hold where I checked them last pass. The substance of this pass is elsewhere: Copilot's suppressed list was never examined by anyone, my approval implied a completeness it did not have, and five of those seven items are live — two of them, the dual-emission contradiction and the label-key row, are wrong in the *commitment* table, which is the one part of this document that functions as a promise rather than a description. I also confirmed the four cross-signal tests actually pass at this head (`go build ./...` clean; both cross-signal tests green in 0.5s, no broker or docker needed) and that they are not vacuous — they fatal on a missing or non-unique series, span or log record, and pin `outcome` to a hardcoded expectation so an all-empty comparison cannot pass.

*4 prior findings settled — 1 addressed, 2 accepted as-is, 1 not addressed.*

|---|---|---|---|
| 🟢 | `:227` — "unit `1` appends `_ratio` on a gauge" has nothing in-repo behind it; no gauge declares unit `1` and no fixture exercises it | Not addressed | `docs/observability.md:227` — unchanged verbatim at this head |
| 🟡 | `:255` — ~14 statements go stale when #389 merges, and both PRs edit this file, with no stated reconciliation | Addressed | Your cross-posted heads-up assigns the flip to whichever of #389/#391 lands second, and @solace-lootii extended it on the thread with the `OBS_METRICS_OTLP_ENABLED` paragraphs for #391. That is the stated reconciliation I asked for. |

### Minor notes

- 🟢 `README.md:110` — "the versioned metric, audit, **and trace** schema" promises a trace schema version that this document does not publish; it versions `metrics_schema` and `audit_schema` only and calls tracing interim. Drop "and trace", or say "metric and audit schema, plus the interim trace surface". (Copilot's suppressed list.)
- 🟢 `docs/observability.md:154` — the table defines the path for renames and closed-set value changes only. A unit change, a metric type change, a histogram bucket change and a label semantic change are all non-additive and all uncovered. One sentence saying every non-additive change follows the same path would close it. (Copilot's suppressed list.)
- 🟢 `docs/observability.md:179` — `audit_event_type` is called "not a field on a record" while Event Fields at `:848` documents it as a field on all seven. The distinction you mean is that it selects the record type; "not merely a field" would keep the point without the contradiction. (Copilot's suppressed list.)

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ⚠️ — 3 open items in the compatibility and query sections, all pre-existing and none introduced by the new push |
| Correctness (doc claims vs code) | ✅ — the cross-signal tests pass at this head and are non-vacuous; the audit-gate ordering and the dual `outcome` vocabularies check out in source |
| PR Scope & Description | ✅ — unchanged since the prior pass |
| Code Quality / Architecture / Security / Test Coverage | — not re-reviewed (no code in this PR; authored content unchanged since `fa6bdf0`) |

*Assessed by the parent (follow-up pass); no discipline lenses were dispatched.*

docs/observability.md
solace-aross: Fixed in 23ff4f4. Qualified "dual emission" to "where dual emission is possible" in the opening blockquote, matching the wording already used in the deprecation cycle.

docs/observability.md
solace-aross: Split the row, but not for the reason given. Checked against the OpenMetrics spec and this repo's actual exporter before writing the fix: same-label-set-per-family is a SHOULD in OpenMetrics, not a MUST, and this server's OTel Prometheus exporter is an unchecked collector with no per-family label-set enforcement — verified serving heterogeneous label-name sets and uniform dual-key emission at HTTP 200 (client_golang v1.24.1, otel v1.46.0, exporters/prometheus v0.68.0). So "emitting broker and event_broker together produces an inconsistent exposition" isn't accurate — it's valid, and dual-emitting a label key here is a one-line change (an added metric.WithAttributes).

The real problem with the row is series identity, not exposition validity: adding the new key still changes the identity of the series consumers already query, so a without()/ignoring() aggregation or a default vector match can silently change during the window, and a rate()/increase() spanning the cutover moment sees the old series go stale and a new one start. A metric-name or audit-field rename doesn't have that problem — the old artifact is untouched. Fixed in 23ff4f4: split the row on the series-identity basis, and fixed the MINOR row's "nothing to do" for a new label, since it carries the identical hazard.

docs/observability.md
solace-aross: Fixed in 23ff4f4. Added the audit_schema_version pin to all four canonical queries and the SPL rendering, plus a line on why an unpinned rule that keeps matching after a schema bump is worse than one that stops.

docs/observability.md
```diff
@@ -802,23 +1070,125 @@ separation-of-duties measure. `reason` tells you why without disclosing the call
...
+
+Four queries cover most of what an access review, a SOC 2 walkthrough, or a security
+investigation asks of this stream. Each is given as a field predicate over records already
+routed by `event="audit"`, because that is the one filter every shipper applies; translate the
+predicate into your own SIEM's syntax. A Splunk SPL rendering is shown for the first as a
+worked example of the translation.
+
+**Pin every rule to `audit_schema_version`, as [Schema Versioning](#schema-versioning) already
+tells you to.** The four queries below do it explicitly, with the current version as a
+placeholder you maintain. A pinned query that stops matching after an upgrade has detected
```
bczoma: 🟡 **The pin guidance covers the upgrade case but not the copy-after-upgrade case, so a compliance query can return silence and read as a clean estate**

**Flagged by** parent review, on lines added in this push

This paragraph says a pinned query "that stops matching after an upgrade has detected schema drift, which is the point". That is right for a rule already deployed: it matched yesterday, it matches nothing today, the operator notices the change and knows why.

It does not cover the reader who arrives *after* the bump and copies the query fresh. For them there is no before-and-after to notice. The rule returns zero rows on its first run and every run after, and nothing in this section tells them that an empty result means the pin is stale rather than that nothing happened.

**Why it matters, concretely.** Query 1 is the destructive-operation review, offered here for "an access review, a SOC 2 walkthrough, or a security investigation". Zero rows from that query is not a neutral outcome — it is the answer the reviewer most wants to see, and the one they are least likely to interrogate. An auditor shown an empty destructive-operations report concludes no privileged changes were made, when the truth is that every record carries `1.2` and the rule asks for `1.1`.

**This is live, not hypothetical.** #389 at `61dbc70` already carries `AuditSchemaVersion = "1.2"` in `internal/observability/schema/version.go:54`. The moment it lands, all five `audit_schema_version="1.1"` occurrences in this section (`:1114`, `:1122`, `:1142`, `:1155`, `:1165`) are stale — and they are five statements that did not exist when the `#389` reconciliation heads-up was written, so the flip needs to know about them.

**Recommended fix.** Two clauses on the paragraph that is already here:

> ...with the current version as a placeholder **you must set to your deployment's `audit_schema` before the rule goes live**. A pinned query that stops matching after an upgrade has detected schema drift, which is the point; **a pinned query that has never matched has a stale pin, not an empty estate — check the version before concluding anything from a zero-row result.** An unpinned one keeps matching and silently spans two contracts.

Worth adding the five line numbers above to whatever tracks the `1.1`→`1.2` flip when #389 or #391 lands second.

bczoma (dismissed):
## PR Review: #405 — SOL-152094: finalize the versioned metric and audit schema doc (Story 35)

**Author:** solace-aross | **Branch:** `andrea/SOL-152094-observability-schema-finalize` → `main` | **Changes:** +498 / -62 across 2 files
**Jira:** `SOL-152094` — Story 35

---

### Verdict

Third pass, at `23ff4f4`. All three conditions from my last review are genuinely fixed — I read the text at the current head rather than the commit subject. On the label-key row, **your counter-argument is right and my finding was wrong**: heterogeneous label-name sets within a metric family are not a wire-format error (the text format is series-based, and an unchecked collector enforces nothing), so "produces an inconsistent exposition" was not accurate. The series-identity framing you shipped instead is the correct hazard, it is the one an operator actually gets bitten by, and catching that the MINOR "new label" row carried the identical hazard — which neither Copilot nor I raised — makes the table better than the fix I proposed. Dropped, with the mechanism corrected in the record.

1. 🟡 `docs/observability.md:1107` — the new pin guidance covers the stale-pin-on-upgrade case but not the copy-after-upgrade one, so a compliance query can return zero rows and read as a clean estate

### Summary

The pin now appears in all four canonical queries and the SPL rendering, the opening blockquote carries the `where dual emission is possible` qualifier that matches the cycle text, and the compatibility table has gone from two rows to five with a series-identity note behind it. That table is materially more correct than it was two passes ago, and the improvement came from the author disputing a reviewer finding on evidence rather than taking it. The one item left is a consequence of the fix itself: pinning turns an over-matching query into one that can under-match to silence, and silence in a SOC 2 walkthrough reads like a clean result.

*3 conditions addressed, 1 declined. The exceptions:*

|---|---|---|---|
| 🟢 | `README.md:110` — "the versioned metric, audit, **and trace** schema" promises a trace schema version this document does not publish | Not addressed | `README.md:110` — unchanged verbatim at this head |
| 🟢 | `:154` — the table covers renames and closed-set values only; a unit change, metric type change, histogram bucket change and label semantic change are all non-additive and uncovered | Not addressed | `docs/observability.md:153-158` — restructured into five rows, but the four uncovered non-additive cases are still uncovered |
| 🟢 | `:179` — `audit_event_type` is called "not a field on a record" while Event Fields documents it as a field on all seven | Not addressed | `docs/observability.md:197-199` — unchanged, now at the wrapped "not a / field on a record" |
| 🟢 | `:227` — "unit `1` appends `_ratio` on a gauge" has nothing in-repo behind it | Declined | `docs/observability.md:247` — unchanged across three passes; taken as answered |

Addressed and not re-rendered: the `:7` dual-emission commitment (`:7-9`), the `:154` label-key row (`:157` plus the series-identity note at `:161-173`), the `:1077` query pinning (`:1105-1109` and five pins), and the `#389`/`#391` reconciliation from the first pass.

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ⚠️ — one open item in the new query-pinning guidance; the compatibility section is now internally consistent |
| Correctness (doc claims vs code) | ✅ — the `#schema-versioning` anchor resolves, the `1.1`/`1.3` versions match `Schema Versioning`, and the exposition-validity claim in the new note is accurate |
| PR Scope & Description | ✅ — unchanged since the prior pass |
| Code Quality / Architecture / Security / Test Coverage | — not re-reviewed (no code in this PR; unchanged since `fa6bdf0`) |

*Assessed by the parent (follow-up pass); no discipline lenses were dispatched.*

docs/observability.md
solace-aross: Fixed in ab90079. Added the copy-after-upgrade case: the placeholder version now says explicitly it must be set to the deployment's actual audit_schema before the rule goes live, and the paragraph distinguishes a stale pin (zero rows, looks clean) from real schema drift (a rule that used to match and stopped). Called out query 1 specifically since a zero-row destructive-operation review is the case that misleads worst.

bczoma (approved):
## PR Review: #405 — SOL-152094: finalize the versioned metric and audit schema doc (Story 35)

**Author:** solace-aross | **Branch:** `andrea/SOL-152094-observability-schema-finalize` → `main` | **Changes:** +504 / -62 across 2 files
**Reviewed at:** `ab900794` — follow-up pass over the one commit pushed since `23ff4f46`
**Jira:** `SOL-152094`

---

### Verdict

**Recommendation:** ✅ Approve

No new conditions. The single 🟡 from my last pass is addressed, and the push that addressed it introduces nothing new.

---

### Summary

This push is eight lines in one paragraph, and they are the right eight lines. My previous pass argued that the pin guidance covered the operator who upgrades under a deployed rule but not the one who copies a query *after* the bump — for whom a zero-row destructive-operation review reads as a clean estate rather than a stale pin. The new text at `docs/observability.md:1105-1116` closes exactly that gap: the placeholder is now explicitly something the operator must set to their deployment's actual `audit_schema` before the rule goes live, the paragraph separates a stale pin from real schema drift, and it names query 1 as the case that misleads worst. That last point was the substance of the finding and it is stated in the author's own words rather than pasted from the suggestion.

Nothing else in the document moved, so nothing else needed re-reading at this head. Worth noting for whoever lands the `1.1`→`1.2` flip: the five `audit_schema_version="1.1"` occurrences in this section are now six lines of prose away from an explanation of why a stale pin is dangerous, which makes the flip more discoverable than it was — but it still has to happen.

*1 prior finding settled — 1 addressed.*

|---|---|---|---|
| 🟡 | Pin guidance covered the upgrade case but not copy-after-upgrade, so a compliance query could return silence and read as a clean estate | Addressed | `docs/observability.md:1107-1116` — placeholder now carries "you must set to your deployment's actual `audit_schema` before the rule goes live"; the paragraph distinguishes a never-matched rule (stale pin) from one that stopped matching (drift), and calls out query 1 specifically |

Earlier passes' findings — the `#389`/`#391` reconciliation thread, the opening blockquote's unqualified dual-emission promise, the label-key row, and the query-pinning gap — were all settled before `23ff4f46` and are not re-opened here.

---

### Minor notes

Nothing new. One item remains textually present at this head but is **already on record from two other reviewers** and is not re-raised as mine:

- `README.md:110` — Copilot flagged, and @solace-lootii re-confirmed as unaddressed, that "the versioned metric, audit, **and trace** schema" names a trace schema version that has no constant (`internal/observability/schema/version.go` carries only `MetricsSchemaVersion` and `AuditSchemaVersion`). It mirrors the doc's own line 4, so it is consistent rather than contradictory, and it is not a condition of this verdict — but it is still open on their threads rather than closed.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ✅ |
| Correctness | — not re-reviewed (unchanged since `23ff4f46`) |
| Code Quality | — not re-reviewed (no code in this PR) |
| Test Coverage | — not re-reviewed (no code in this PR) |
| Architecture | — not re-reviewed (unchanged since `23ff4f46`) |
| Security | — not re-reviewed (unchanged since `23ff4f46`) |
| PR Scope & Description | — not re-reviewed (unchanged since `23ff4f46`) |

Assessed by: parent (follow-up)

solace-lootii (approved):
Approving. The #389/#391 reconciliation, the dual-emission contradiction, and the query-pinning gaps are all resolved and verified against the actual text. The 3 remaining nits (README:110, the non-additive-change-types gap, the "not a field" phrasing) are all on record as open and non-blocking — fine as follow-up, not holding this up.

docs/observability.md
solace-aross: Reconciliation is posted here: https://github.com/SolaceProducts/solace-broker-mcp/pull/405#issuecomment-5622713842, cross-posted on #389 and #391. Both later passes marked this addressed, so resolving.

PR #410, opened 2026-09-11
SOL-151608: optional token expiry fallback when IdP omits expires_in

amitmorade:
## Summary

This PR is for the **customer platform/ops engineer** who must get Hop 2 (broker OAuth token exchange) working when their IdP omits `expires_in`. RFC 8693 recommends `expires_in`; it does not require it. Those operators set optional YAML on purpose. Omitting the field keeps the existing fail-closed behavior.

Tracked: [SOL-151608](https://sol-jira.atlassian.net/browse/SOL-151608).

Optional `broker_oauth.token_expiry_fallback` lets token exchange accept IdP responses that omit, null, or zero `expires_in`. A positive duration is used as the token lifetime in that case only. Positive IdP `expires_in` always wins. Clock-skew is deducted once; the cache still uses `min(remaining ExpiresAt, MaxTTL)`. Invalid configured durations fail at startup.

A follow-up will land immediately after this for the **other** persona: operators who want to configure cache MaxTTL / residency for tokens that already have `expires_in`. A Jira ticket for that follow-up will be created. This PR is not that work.

## Design decisions

These are implemented, not open questions.

- The setting is optional, flat YAML: `broker_oauth.token_expiry_fallback`.
- Omit the field (or leave it unset) and exchange stays fail-closed: the same `ErrInvalidResponse` as today when `expires_in` is absent, null, or zero.
- Fallback applies only to absent, null, or zero `expires_in`.
- A positive IdP `expires_in` always wins over the configured fallback.
- Negative `expires_in` and overflow still error even when fallback is configured.
- Skew is deducted once when the response is parsed. Cache TTL is unchanged: `min(remaining ExpiresAt, MaxTTL)`.
- An invalid configured duration fails at startup; the process does not start with a bad fallback.
- Hop 2 copy-paste YAML in the shipped example comments the field so a copy-paste stays fail-closed until an operator uncomments and sets a duration on purpose.

## Reviewer question

Today we log the configured fallback once at Hop 2 exchanger creation (`expiry_fallback_configured`, plus `expiry_fallback` when set). We do not log whether fallback was used on a given exchange.

Should we also log fallback *usage* per exchange (for example an optional Debug `used_fallback` bool on the existing `"identity provider issued broker token"` line), or keep startup-configured-value only? A WARN on every use would spam if a silent IdP causes a cache miss each time; Debug on that existing line would avoid that.

If you answer this, I'd appreciate a brief note that the answer is yours as the human reviewer — just so we can tell it apart from automated AI review comments. Thank you.

**Answer**

Direction taken, shipped in 85fed42 and 0eeebff. Startup INFO still reports only that the control is armed: `expiry_fallback_configured` always, `expiry_fallback` when set. Attribute keys omit the substring `token` so they are not redacted.

We added a per-live-exchange Debug bool `used_fallback` on the existing `identity provider issued broker token` line; true only when the fallback supplied the lifetime, false when the IdP returned a positive `expires_in`. Cache hits emit no such line; a fail-closed exchange errors instead. No once-per-process WARN and no counter.

Operator docs in `docs/authentication.md` and `docs/configuration.md` now state configured vs used explicitly. Tests cover fallback used, IdP positive expiry wins, negative `expires_in` still errors, and both the configured and unconfigured startup log paths.

The human-reviewer note above still stands if you want to weigh in. We proceeded with the Debug option rather than block, and would still welcome that distinction from automated AI comments.

## What this does not change

- Cache MaxTTL, cache residency policy, and how tokens that already carry `expires_in` are retained. That is the follow-up persona, not this PR.
- Fail-closed default when the field is omitted.
- Clock-skew: still deducted once at parse time.
- Token-exchange grant, audience, retry, circuit-breaker, or Hop 1 auth.

## Test plan

- [ ] Config: omit `token_expiry_fallback` → load succeeds; invalid duration → startup/config fails.
- [ ] Exchange: omit/null/zero `expires_in` with no fallback → `ErrInvalidResponse`.
- [ ] Exchange: omit/null/zero `expires_in` with fallback → token accepted; lifetime is the fallback minus skew.
- [ ] Exchange: positive IdP `expires_in` with fallback configured → IdP value used, not fallback.
- [ ] Exchange: negative `expires_in` and overflow still error with fallback set.
- [ ] Cache: fallback-derived tokens are cached and reused under the existing `min(remaining ExpiresAt, MaxTTL)` rule.
- [ ] Logging: startup INFO shows armed (`expiry_fallback_configured` / `expiry_fallback`); live-exchange Debug carries `used_fallback`; cache hits and fail-closed errors do not emit that Debug line.
- [ ] Shipped example YAML keeps the field commented so copy-paste remains fail-closed.
- [ ] `make check` (or CI) green on this branch.

Made with [Cursor](https://cursor.com)


solace-aross (dismissed):
Clean. This is a security-relevant change (Hop 2 token lifetime), so I verified the two invariants that matter by sabotage rather than by reading the tests:

- **Negative `expires_in` is rejected even with a fallback configured.** Removed the `sr.ExpiresIn < 0` guard in a scratch worktree — the negative value fell through to the fallback-eligible branch, produced a `Token` with a negative-lifetime `ExpiresAt`, and `TestParseSuccessBody_NegativeExpiresInRejectedEvenWithFallback` failed exactly as it should. Restored, green again.
- **A positive IdP `expires_in` always wins over the configured fallback.** Inverted the priority (`sr.ExpiresIn == 0 || e.tokenExpiryFallback != 0`) so a configured fallback would override a real IdP value — `TestParseSuccessBody_PositiveExpiresInWinsOverFallback` caught the wrong `ExpiresAt` immediately. Restored, green again.

Also confirmed: `go build`/`go vet` clean on this head, full `internal/tokenexchange` and `internal/config` suites pass (770 tests), and YAML duration parsing for the new field works as claimed (`token_expiry_fallback: 1h` round-trips correctly). The docs/threat-model update accurately reflects what the code does, not just what the PR says it does.

Design is sound: optional field, nil preserves the exact prior fail-closed behavior, config validation rejects non-positive durations at startup, and the log fields carry only a bool and a duration. Nothing open.

internal/config/config.go
```diff
@@ -105,13 +105,14 @@ type ServerConfig struct {
...
-	RetryAfter     *IdPRetryAfterConfig     `yaml:"retry_after"`             // optional; nil → shipped default cap (SOL-152285)
+	TokenURL            string                   `yaml:"idp_token_endpoint"`      // IdP token endpoint (token-exchange POST target). YAML key uses "endpoint" to match the OAuth spec and OIDC Discovery JSON (`token_endpoint`); Go field keeps `URL` to match the language convention (golang.org/x/oauth2 also names its field TokenURL).
+	ClientID            string                   `yaml:"mcp_server_client_id"`    // MCP server's client_id registered at the IdP
+	ClientAuth          BrokerClientAuth         `yaml:"mcp_server_client_auth"`  // discriminated union; exactly one sub-block populated
+	GrantType           string                   `yaml:"grant_type"`              // required; must be in validGrantTypes
+	AudienceParam       string                   `yaml:"audience_parameter_name"` // required; one of {audience, scope, resource}
+	TokenExpiryFallback *time.Duration           `yaml:"token_expiry_fallback"`   // optional; used only when the IdP omits or returns zero expires_in
```
bczoma, line 113: 🟡 **The YAML key's namespace is the one-way door in this change**

Flagged by `good:architecture`.

**What the problem is.** `TokenExpiryFallback` lands flat under `broker_oauth`, while every other tuning concern on this path is a named sub-block (`circuit_breaker`, `retry_after`). The PR body announces an immediate follow-up adding cache MaxTTL / residency settings — that is a second, sibling home for token-lifetime configuration.

**Why it matters.** `LoadConfig` uses `dec.KnownFields(true)` (`internal/config/config.go:636`), so once this ships the key cannot be relocated or removed without failing startup for any operator who set it. If the follow-up introduces e.g. `broker_oauth.token_cache: {max_ttl: …}`, then "which lifetime" and "how long resident" are split across two grammars permanently and this flat key becomes a legacy alias. Every other finding in this review is reversible after merge; this one is not, which is the only reason it is a condition rather than a note.

**Recommended fix.** Decide the namespace now, either way — nest as `broker_oauth.token_lifetime.fallback` (leaving room for the follow-up's fields beside it), or state explicitly in the ticket and CHANGELOG that the flat key is the permanent surface and the follow-up will nest *under* it rather than beside it. The decomposition itself is sound; the concerns genuinely are distinct. It is only the key placement that is hard to reverse.

internal/tokenexchange/response.go
```diff
@@ -159,23 +159,36 @@ func (e *Exchanger) parseSuccessBody(body []byte, now time.Time) (*Token, error)
...
 		}
 	}
 
-	// Prevents time.Duration overflow (int64 nanoseconds, max ~292 years).
-	if sr.ExpiresIn > maxExpiresInSeconds {
-		return nil, &ExchangeError{
-			Sentinel: ErrInvalidResponse,
-			Message:  fmt.Sprintf("token exchange invalid response: IdP returned expires_in %d which exceeds the safe arithmetic limit — likely a misbehaving IdP or intercepted response", sr.ExpiresIn),
+	var lifetime time.Duration
+	if sr.ExpiresIn == 0 {
```
bczoma, line 179: 🟡 **The branch that decides which lifetime wins is the hardest part of this function to scan**

Flagged by `good:code`.

**What the problem is.** `parseSuccessBody` is now ~90 lines and owns both response validation and lifetime selection. The selection itself is `if sr.ExpiresIn == 0 { … } else { nested overflow guard … }` — and that overflow guard is the only right-shifted guard clause in a function that is otherwise a flat chain of early returns.

**Why it matters.** The rule a reader most needs to follow here is *which source wins*, and it is the least scannable thing in the function. Extracting it also makes the precedence rules unit-testable without constructing an `Exchanger`, which is what the current tests have to do.

**Recommended fix.**

```go
func (e *Exchanger) selectTokenLifetime(expiresIn int64) (time.Duration, error)
```

holding the negative, zero/fallback and overflow rules, leaving the caller with:

```go
lifetime, err := e.selectTokenLifetime(sr.ExpiresIn)
if err != nil {
    return nil, err
}
```

Worth noting this is a maintainability finding only — the logic itself is correct, and @solace-aross verified the two precedence invariants by sabotage at this head.

cmd/server/main.go
```diff
@@ -894,7 +894,13 @@ func newTokenExchanger(oauthCfg *config.BrokerOAuthConfig) (*tokenexchange.Excha
 		}
 		return nil, fmt.Errorf("creating token exchanger: %w", err)
 	}
-	slog.Info("token exchanger created for broker OAuth")
+	logAttrs := []any{
+		slog.Bool("expiry_fallback_configured", oauthCfg.TokenExpiryFallback != nil),
+	}
+	if oauthCfg.TokenExpiryFallback != nil {
```
bczoma, line 900: 🟡 **The untested branch is the default-deployment path**

Flagged by `good:code`.

**What the problem is.** Two things at one site. The nil-check-then-conditional-append of `expiry_fallback_configured` / `expiry_fallback` is written twice in two packages — here and at `internal/config/config.go:254-263`. And the nil/`false` branch of both is untested: `TestBrokerOAuthConfig_LogValue` asserts only the configured case, and `newTokenExchanger` has no test asserting the absent case.

**Why it matters.** Omitting `token_expiry_fallback` is the default and by far the most common deployment, so the untested branch is the one almost every operator will actually run. A regression there — the attr silently dropped, or `expiry_fallback` appearing with a zero value when nothing was configured — ships unnoticed, and these two log fields are the only evidence at startup of whether the control is armed.

**Recommended fix.** Extract `func (b BrokerOAuthConfig) expiryFallbackAttrs() []slog.Attr` and call it from both sites, then add a nil-case assertion that `expiry_fallback_configured` is `false` and that the `expiry_fallback` key is **absent** rather than zero.

internal/tokenexchange/response.go
```diff
@@ -159,23 +159,36 @@ func (e *Exchanger) parseSuccessBody(body []byte, now time.Time) (*Token, error)
...
+		if e.tokenExpiryFallback == 0 {
+			return nil, &ExchangeError{
+				Sentinel: ErrInvalidResponse,
+				Message:  "token exchange invalid response: IdP success response missing or zero expires_in — configure token lifetimes on the IdP client or set broker_oauth.token_expiry_fallback",
+			}
+		}
+		lifetime = e.tokenExpiryFallback
```
bczoma, line 186: 🟡 **Startup logs that the fallback was armed, never that it fired**

Flagged by `good:security` (SECURE: Response & Monitoring, Evidence & Audit). **This is also the answer to the reviewer question in your PR body** — and please treat it as an automated answer, not the human reviewer's note you asked for.

**What the problem is.** No runtime signal distinguishes a fallback-derived lifetime from an IdP-asserted one. `cmd/server/main.go:897-903` logs `expiry_fallback_configured` and `expiry_fallback`, i.e. only that the operator *armed* the control.

**Why it matters.** "Configured" and "in use" are different security states. This is the one knob that lets an operator assert a token lifetime the IdP never asserted, and a customer-side IdP change can silently flip every exchange onto operator-asserted lifetimes with no artefact recording when that began. The post-incident question — "how long were we presenting lifetimes the IdP never issued?" — is then unanswerable. Startup-only is not adequate evidence for this particular control, even though it is fine for most config.

**Recommended fix**, answering your question directly:

1. **Do not** log per exchange at Info — you are right that it fires on every cache miss and would be high-volume.
2. **Do** add the `slog.Debug` `used_fallback` bool you proposed on the existing `"identity provider issued broker token"` line, for on-demand diagnosis.
3. **Add one log-once-per-process WARN** the first time the fallback is actually applied. That is the piece that turns "armed" into "fired", it is the state an auditor asks about, and at once per process it costs nothing.

A counter alongside the existing OAuth gauges would be the durable form of (3) and is the better long-term answer if you would rather not add a WARN.

bczoma (dismissed):
## PR Review: #410 — SOL-151608: optional token expiry fallback when IdP omits expires_in

**Author:** amitmorade | **Branch:** `amorade/SOL-151608-token-expiry-fallback` → `main` | **Changes:** +607 / -85 across 20 files
**Jira:** `SOL-151608` — Add configurable token_expiry_fallback for IdPs that omit expires_in (Verifying)

---

### Verdict

1. 🟡 `internal/config/config.go:113` — settle the YAML key's namespace before merge; `KnownFields(true)` plus the announced follow-up make the flat key a one-way door
2. 🟡 `internal/tokenexchange/response.go:170` — extract the lifetime-selection rules out of `parseSuccessBody`, now ~90 lines with the only right-shifted guard in an otherwise flat function
3. 🟡 `cmd/server/main.go:900` — the nil/`false` branch of the new log attrs is the default-deployment path and is untested, and the same conditional append is written twice in two packages
4. 🟡 `internal/tokenexchange/response.go:177` — nothing records that the fallback actually *fired*; startup logs only that it was armed (this is also my answer to your reviewer question)

The 5 🟢/💡 notes below are **not** conditions of this verdict.

**On your reviewer question — and the disclosure you asked for:** this is an **automated review**, so please do not count the answer below as the human reviewer's note you asked for. @solace-aross's approval is a human review, but it did not address the logging question, so that one is still open for a human. For what it is worth, the reasoned recommendation is in thread 4: keep Info free of per-exchange noise, add the Debug `used_fallback` bool you proposed, and add one log-once-per-process WARN (or a counter) the first time the fallback is actually applied — that is the piece that turns "armed" into "fired" at negligible volume.

---

### Summary

A well-executed, tightly scoped config extension: fail-closed is preserved by construction rather than by convention (omitted → `nil` → zero sentinel → the pre-existing `ErrInvalidResponse`), the skew is still deducted at exactly one site, and the shipped example is double-commented so a copy-paste cannot silently arm a 1h fallback. Every one of the ticket's seven scope items lands, including the error-message update that ACs like this usually lose, and it is regression-guarded by a test asserting the message names the field. Three security and architecture claims were filed at 🟡 and all three were reduced by measurement — the explicit-`expires_in: 0` objection was dropped outright once it turned out the pre-PR code already branched on `expires_in <= 0`, so nothing distinguishable was widened. What is left is one genuine pre-merge decision (the key's namespace) and three maintainability and observability items.

---

### Minor notes

- 💡 `internal/config/token_expiry_fallback.go:29` — a fallback in `(0, 30s]` is provably useless (always `ExpiresAt <= now`, never cacheable) yet passes config load. Bounded in practice: the token still works, the breaker still counts successes, and `exchange.go:495` already WARNs per call naming the cause. Either reject at load (note this exceeds the ticket's "must be positive when set") or land the planned SOL-154165 warning.
- 💡 `docs/configuration.md:117`, `docs/authentication.md:296` — both rows warn about the 30s low-side truncation but not the 24h high side, so a `720h` value reads as fully honoured; the ceiling is documented only at `docs/authentication.md:700`. Add one clause.
- 🟢 `docs/internal/threat-model.md:183` — the accepted-risk row and risk-register entry #9 still attribute an over-long `ExpiresAt` only to an IdP-set value; an operator-set fallback is now a second source, bounded by the 24h cache ceiling and detected reactively on a broker 401.
- 🟢 `internal/tokenexchange/response.go:165` — the negative-`expires_in` message dropped both the offending value and the remedy text that every neighbouring message in this function carries. `"negative expires_in %d — the IdP must return a positive lifetime"`.
- 🟢 `internal/tokenexchange/types.go:119` — `Params.TokenExpiryFallback time.Duration` uses zero as "disabled" immediately beside `MaxHonoredRetryAfter`, where zero means the opposite ("use the shipped default"). Carrying `*time.Duration` through `Params` would drop both the sentinel and `resolveTokenExpiryFallback`.

---

### Summary Verdict

|---|---|---|
| Documentation | ✅ | docs |

internal/config/config.go
amitmorade: Thanks for calling out the one-way-door aspect. I’m keeping `broker_oauth.token_expiry_fallback` flat: the other scalar protocol settings on `broker_oauth` are already flat, while `circuit_breaker` and `retry_after` are nested because each is a multi-field block. Persona 2 is a cache MaxTTL/residency control, not this missing-IdP-lifetime fallback. Relocating this later would be worse than freezing the already documented `broker_oauth.token_expiry_fallback` key, so I’m treating this as the permanent surface.

internal/tokenexchange/response.go
amitmorade: Thanks — I agree an extraction could become useful if this function grows. For this PR I’m leaving the lifetime selection inline: the current precedence logic is correct and has already been sabotage-checked by aross, and I don’t want to churn this sensitive path without a behavior need. I’ll extract it later if additional lifetime rules make the function materially harder to scan.

cmd/server/main.go
amitmorade: Thanks, I’m taking the default-path coverage concern. I added nil-case tests for both `BrokerOAuthConfig.LogValue` and the actual `newTokenExchanger` startup record, asserting `expiry_fallback_configured=false` and complete absence of `expiry_fallback`. I’m skipping the shared helper extraction so this remains a focused test/observability change.

internal/tokenexchange/response.go
amitmorade: Thanks for distinguishing “armed” from “used.” I’m taking the per-exchange Debug signal: the existing `identity provider issued broker token` record now carries `used_fallback`, with tests for fallback use, IdP-provided positive expiry, and negative-expiry failure. I’m not adding a process-global WARN or counter in this PR; that would add spam/state to a path whose current behavior is signed off, while Debug provides the requested on-demand diagnosis without production-volume noise.

cmd/server/token_exchanger_log_test.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"bytes"
+	"log/slog"
+	"strings"
+	"testing"
+
+	"github.com/SolaceProducts/solace-broker-mcp/internal/config"
+)
+
+func TestNewTokenExchanger_LogsUnconfiguredExpiryFallback(t *testing.T) {
```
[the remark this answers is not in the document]
amitmorade: Thanks, you’re right that the configured branch constructs these attributes independently. I added coverage against the actual newTokenExchanger startup record, asserting expiry_fallback_configured=true and the configured one-hour duration.

internal/tokenexchange/exchange.go
```diff
@@ -700,10 +700,11 @@ func (e *Exchanger) doExchange(ctx context.Context, input ExchangeInput) (*Token
 	slog.DebugContext(ctx, "identity provider issued broker token",
 		slog.String("broker", input.BrokerAlias),
 		slog.Int("http_status", status),
-		slog.Int("attempts", idpclient.AttemptsFromContext(ctx)))
+		slog.Int("attempts", idpclient.AttemptsFromContext(ctx)),
+		slog.Bool("used_fallback", parsed.usedFallback))
```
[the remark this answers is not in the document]
amitmorade: Thanks for catching the stale description. The per-exchange Debug used_fallback field is intentional and is the settled behavior for this ticket; removing it would undo the requested observability. A separate description-only update is bringing the PR text into sync, so I’m leaving the code and supporting docs/tests unchanged.

bczoma, line 704: 🟡 **`used_fallback` is the right primitive, but it is silent at the log level production actually runs**

Flagged by `good:security` (SECURE: Response & Monitoring, Evidence & Audit) and, independently, by `good:architecture`.

**First, what you took is correct and I am not re-opening it.** Parts 1 and 2 of the prior recommendation landed exactly as asked: nothing new at Info, a Debug bool on the existing success line, provenance threaded from `response.go:214-224`, and four covered cases including the negative-`expires_in` case that asserts *no* issued line is emitted. That last assertion is the one that catches a future refactor logging success before validation — good call.

**What the residual is.** Your reason for deferring part 3 was that a latched WARN "would add spam/state… while Debug provides the requested on-demand diagnosis." Two measured points against the second half of that:

- `docs/internal/secure-logging-rules.md` sets the production level to **INFO**, so `used_fallback` emits nothing in a default deployment. "On-demand" therefore means *prospective* only — an operator must turn Debug on and then wait for a cache miss. It cannot answer the question retrospectively for any window during which Debug was off, which is every window by default.
- No other level carries the signal. The only WARN on this path, `internal/tokenexchange/exchange.go:496`, fires on `PutDroppedTTL` (a fallback ≤ 30s) and not on ordinary fallback use. I checked; there is no second carrier.

So the post-incident question — "for how long were we presenting broker tokens whose lifetime the IdP never asserted?" — is still unanswerable at the level production runs at. That matters specifically for *this* control because it is the one knob that lets an operator assert a lifetime the IdP never did, and a customer-side IdP change can flip every exchange onto operator-asserted lifetimes with no artefact recording when that began.

**On the "spam/state" half:** the recommendation was one line per *process lifetime*, not per exchange, and the state is a `sync.Once` or an atomic bool. That is why this reads as a narrow residual rather than a disagreement about volume — I think we actually agree on the volume.

**Why this is 🟡 and not a blocker.** The residual is genuinely bounded: the feature is opt-in, its *armed* state is at Info (`cmd/server/main.go:897-903`), cache residency is capped at 24h, and both guides state the limitation plainly rather than papering over it. The approval stands.

**Recommended fix — a follow-up ticket is a perfectly good answer here.** Either:

```go
// once per process, the first time the fallback actually supplies a lifetime
e.fallbackUsedOnce.Do(func() {
    slog.WarnContext(ctx, "broker token lifetime supplied by configured expiry fallback",
        slog.Duration("expiry_fallback", e.tokenExpiryFallback))
})
```

or a counter alongside the existing OAuth observability surface, which `good:architecture` independently preferred as the durable form — the aggregate ("has this ever happened in this process?") is the fact an auditor asks for, and a per-event Debug line is the one carrier that cannot express it. Keep `used_fallback` underneath either; it is the right primitive.

If you would rather land the follow-up separately, say so and I will resolve this thread on the ticket reference.

bczoma (approved):
## PR Review: #410 — SOL-151608: optional token expiry fallback when IdP omits expires_in

**Author:** amitmorade | **Branch:** `amorade/SOL-151608-token-expiry-fallback` → `main` | **Changes:** +796 / -98 across 22 files
**Head reviewed:** `0eeebffb0ef3a9ea29b958159edea9b2b3298dd2`
**Jira:** `SOL-151608` — Add configurable token_expiry_fallback for IdPs that omit expires_in (Verifying)

---

### Verdict

1. 🟡 `internal/tokenexchange/exchange.go:704` — `used_fallback` is the right primitive, but at the default production log level it emits nothing, so "has this IdP ever omitted `expires_in`?" is still unanswerable after the fact. A follow-up ticket closes this; code in this PR is not required.

**Two pushes since my last review, and both are real work** — `85fed421` (the `used_fallback` signal) and `0eeebffb` (the configured-branch startup-log test). Same merge-base with `main` as before (`29a181ea`), no force-push, no base move: `git cherry` marks only those two commits `+`, so nothing else in the PR changed and I have not re-litigated it.

**Both of @Copilot's findings are resolved, so I have not duplicated them:** the untested configured startup-log branch was closed by `0eeebffb`, and the description/implementation mismatch was closed by the PR-body rewrite that now documents `used_fallback` as shipped rather than asking whether to add it.

One process note: dismissing my review also dismissed @solace-aross's, so his human approval no longer counts toward the required review — worth asking him to re-approve, since his sabotage checks were against `ab69b057` and the two new commits do not touch the invariants he tested.

---

### Summary

Two tight, well-aimed commits. The provenance bit is threaded as an unexported field on an unexported wrapper, so the claim in its own doc comment — that it never reaches `Token`'s public contract or the cache — is enforced by the type system rather than by convention, and I verified `CachedCredential` is still built field-by-field from `Value`/`ExpiresAt` only. The new table test covers all four lifetime-selection outcomes including the negative case's *absence* assertion, and it is mutation-sensitive: neutering `usedFallback = true` fails it. Three of my four prior conditions are now settled, one of them because your argument was simply better than my finding — see the table. The one remaining 🟡 is a narrow residual on the part of the observability fix you deferred, and a follow-up ticket is a perfectly good answer to it.

I re-examined the test-coverage mark after `0eeebffb`: the gap that would have held it at ⚠️ is closed, and what is left is assertion precision rather than missing coverage.

---

*9 prior findings: 2 addressed, 4 accepted as-is, 1 declined, 2 partially addressed.*

|---|---|---|---|
| 🟡 | Nothing records that the fallback *fired*; startup logs only that it was armed (3-part fix: no per-exchange Info, add Debug `used_fallback`, latch one WARN/counter on first use) | **Partially addressed** | `internal/tokenexchange/exchange.go:700-704` — parts 1 and 2 landed exactly as recommended, covered by `exchange_test.go:2684`. Part 3 deferred; residual is the conditioned thread above. |
| 🟡 | The `expiry_fallback` log-attr block is duplicated in two packages, and its nil/`false` branch was untested | **Partially addressed** | Coverage is now complete on both branches of both sites — `cmd/server/token_exchanger_log_test.go:26` (nil) and `:22` (configured, added in `0eeebffb`), `internal/config/config_test.go:3658`/`:3662`. The duplication itself stands: `cmd/server/main.go:897-903` and `internal/config/config.go:254-264` are still the same shape written twice. |
| 🟡 | Settle the YAML key's namespace before merge — flat `token_expiry_fallback` beside nested `circuit_breaker`/`retry_after` is a one-way door under `KnownFields(true)` | **Accepted as-is — your argument wins** | `internal/config/config.go:108-117` — I checked the pattern you cited and it holds exactly: every flat key on `BrokerOAuthConfig` is a single scalar, every nested key is a multi-field struct, with no counter-example either way. Nesting a scalar would have been the first exception to the block's own rule. Dropping the finding. |
| 🟡 | Extract the lifetime-selection rules out of `parseSuccessBody` | **Declined** | Reason accepted — not churning a sabotage-verified path without a behaviour need is sound. Noted only so it is not lost: the block now carries state (`usedFallback`) as well as a value, so your own "extract it when more lifetime rules arrive" trigger is one notch closer. |

The 5 🟢/💡 notes are settled and not repeated here — the negative-`expires_in` message and the 24-hour clause in both config tables were fixed (with regression assertions for the message), and the remaining three were optional and reasonably left.

---

### Minor notes

- 🟢 `CHANGELOG.md:19` — the flat-key decision lives only in a PR thread; the entry this commit rewrote says nothing about `broker_oauth.token_expiry_fallback` being the permanent surface. Half a sentence there is what actually closes the one-way-door concern for the next reader.
- 🟢 `internal/tokenexchange/errors.go:38-39` — the type's stated invariant ("renders only broker- or server-generated text") no longer literally holds now that `response.go:173` interpolates the IdP's `expires_in`. Safe today — it is an `int64`, so `%d` can emit only digits and a minus sign — but `internal/tools/manager.go:711` logs `Error()` verbatim under the unredacted `detail` key, and the neighbouring fields a later edit would reach for (`access_token`, `error_description`) are attacker-influenced strings. State the narrowed rule: IdP data may appear in `Message` only as a numeric type.
- 🟢 `internal/tokenexchange/exchange.go:704` — `used_fallback` is now documented by name to operators in two guides and the CHANGELOG, which is what turns a Debug field into something a SIEM rule depends on; the repo states no stability class for Debug field names. One clause in `docs/internal/secure-logging-rules.md`, either answer.
- 🟢 `cmd/server/token_exchanger_log_test.go:52-63` — both new tests assert with buffer-wide `strings.Contains` on raw JSON, so the "line captured" check and the attribute checks could in principle be satisfied by different records. `exchange_test.go`'s `records`/`forBroker` per-record pattern already exists in-repo. (The substring forms themselves are sound — the closing quote before the colon means `"expiry_fallback":` cannot match `"expiry_fallback_configured":`.)
- 🟢 `internal/tokenexchange/exchange_test.go:2686` — `captureJSONLogs(t)` is called once outside the subtest loop, so records accumulate and isolation rests on the unasserted invariant that no two subtest names normalise to the same `BrokerAlias`. Moving it inside `t.Run` removes the dependency; the subtests are non-parallel, so per-subtest capture is safe.

---

### Summary Verdict

|---|---|---|
| Test Coverage | ✅ | good:code (⚠️ re-examined — the gap it rested on was closed by `0eeebffb`) |
| Documentation | ✅ | docs |
| PR Scope & Description | — not reviewed (lens not dispatched; PR-body and Jira scope settled at the prior pass) | — |

solace-aross (approved):
Re-reviewed at this head (my prior approval at `ab69b057` was dismissed by the two new commits, correctly). What changed: a `used_fallback` bool threaded through to the existing Debug "identity provider issued broker token" line, so a fallback-derived token is distinguishable from an IdP-asserted one after the fact.

Sabotage-verified the new signal rather than trusting the test: forced `usedFallback` to always report `false` in a scratch worktree — `TestExchange_IssuedLogReportsFallbackUsage` caught both the missing-expiry and zero-expiry cases, the positive-IdP-expiry and rejected-negative-expiry cases were unaffected as expected. Restored, clean. Full `internal/tokenexchange`, `internal/config`, and `cmd/server` suites pass (1021 tests), build/vet clean.

On bczoma's residual (used_fallback is Debug-only, so it can't answer "did this ever fire" retrospectively at the INFO level production runs at): I read the whole thread rather than the diff alone. He's right that it's a real gap, and right that it's bounded — opt-in feature, armed-state already visible at Info, 24h cache ceiling, both docs state the limitation. That's a fair call for a follow-up ticket rather than a blocker in this PR, and I don't have anything to add to his analysis. Nothing else open.

internal/tokenexchange/exchange.go
amitmorade: Thanks for the careful follow-up, and for offering to resolve this on a ticket reference.

You're right that `used_fallback` is Debug-only, so at the default INFO level it never records that fallback actually fired. Past exchanges aren't visible until Debug is turned on.

We're not adding a once-per-process WARN/INFO on this PR — that would change the signed-off log level on this path (and add noise at the production level), and Debug remains the on-demand switch for this ticket. We'll file this as a follow-up for a default-level breadcrumb / once-per-process signal rather than land it here.

I appreciate the suggestion.

amitmorade: Tracked as SOL-154334 — first-fire default-level breadcrumb (or counter) when hop-2 expiry fallback actually supplies a lifetime. Deferred from this PR so we do not change the signed-off log level here. Please resolve when that is enough for you.

PR #414, opened 2026-09-11
SOL-152284: alertable token-exchange circuit breaker state metric

amitmorade:
## Summary

When the process-wide IdP token-exchange breaker trips, every broker's live exchange fast-fails for `open_state_duration`. Until now the only signal was a WARN. This adds `mcp_token_exchange_circuit_breaker_state{breaker,state}` so operators can alert on `{state="open"} == 1` instead of grepping logs.

Tracked under [SOL-152284](https://sol-jira.atlassian.net/browse/SOL-152284). Additive `metrics_schema` 1.6 → 1.7. Main shipped 1.5/1.6 while this branch was open; 1.4 was never emitted. No new config field, no new audit field.

## How to review

Five commits, in order:

1. Passive snapshot on the exchanger (the design)
2. Gauge, tests, golden schema
3. Startup wiring
4. Schema + changelog
5. Pointers from the Hop-2 pages operators actually read

## Design decisions

These were settled before the code, and the code is meant to hold them.

1. **gobreaker is the only breaker state machine.** Observability does not run trip/recovery logic of its own. A second state machine would diverge on probe budget, exclusions, and generation.

2. **The metric is an event mirror, not a poll and not a second source of truth.** `setState` is gobreaker's only writer of `cb.state`, and it always fires `OnStateChange`. That callback stores `to` in an `atomic.Int64` on `Exchanger`. The scrape reads only the atomic. One writer, one observation point, a lock-free copy of the last materialized transition.

3. **Scrapes must never call gobreaker `State()`, `Counts()`, or `Execute()`.** In gobreaker v2.4.0, `State()` is not a read: when stored state is Open and the timeout has elapsed it calls `setState(HalfOpen)`, which fires `OnStateChange` (WARN from the scrape goroutine), bumps generation, clears counts, and can discard in-flight outcomes. Telemetry would drive recovery. Production previously never called `State()`; only tests did.

4. **Wiring copies broker reachability, not a panic-counter global.** Domain owns state (`Exchanger`), metrics observes via a scrape callback, `cmd/server` joins them after both exist. `tokenexchange` does not import `metrics`. Register whenever Hop-2 is active; the snapshot callback already omits series when the breaker is configured off.

5. **OTel name keeps compound terms:** `mcp.token_exchange.circuit_breaker.state` (Prometheus `mcp_token_exchange_circuit_breaker_state`). The instrument is unitless so OTel does not append `_ratio`.

6. **Mirror type is `atomic.Int64`, not `uint32`.** gosec G115 rejects `int` → `uint32`. `int` → `int64` cannot overflow. Snapshot labels still come from `gobreaker.State.String()`.

## What `{state="open"} == 1` means

It means **recovery has not materialized yet**, not “callers are currently being rejected.”

gobreaker latches `open` → `half-open` on the next `Execute` after `open_state_duration`, not on a timer goroutine. Cache hits, singleflight followers, and Retry-After `gateCheck()` never reach `Execute`, so the series can stay `open` after the timeout even though the next live exchange will admit a probe rather than reject it.

## Invariants that must stay true

Unchanged contracts this PR must not break:

- `OnStateChange` stays cheap under gobreaker's mutex.
- WARN message and fields stay `breaker`, `from`, `to`, `consecutive_failures`.
- `tokenexchange` does not import `metrics` (or `health`).
- `breakerName` is the single string `"idp-token-exchange"` on the log and the metric label.
- Failure classification is unchanged: `ErrInvalidResponse` / rejection = breaker success; 429 / config / cancel / deadline / build / missing-subject = excluded; network / 5xx / body-read = failure. Excluded ∩ Failure = ∅.
- One logical `Exchange` (retries + singleflight winner) = one breaker outcome.
- Stale generation discard remains gobreaker's. A scrape must not bump generation.
- `gatedUntil` and the breaker are independent suppressions. This metric does not represent the Retry-After gate.
- Disabled breaker (`e.breaker == nil`): every call still reaches the IdP; no `ErrExchangeCircuitOpen`.
- gobreaker's Dependabot auto-merge exclusion is unchanged.

New contracts this PR adds:

1. **Observational scrape.** The gauge never calls a gobreaker method. The shadow updates only from `OnStateChange`, plus a Closed seed at construction (gobreaker does not fire `OnStateChange` for the initial closed state).
2. **Shadow = last stored / materialized state**, not wall-clock eligibility to enter half-open.
3. **Always-complete one-hot** when registered: exactly three series (`closed` / `open` / `half-open`); values only 0 or 1; exactly one is 1; they sum to 1; derived from one atomic `Load` per collection.
4. **Absent when not wired** (metrics off, no Hop-2 exchanger, explicit `circuit_breaker.enabled: false`, registration failure). Disabled must not look like `closed`.
5. **Default-on.** Omitting the `circuit_breaker` YAML block still enables the breaker, so the family is present. Metrics themselves stay off until `OBS_METRICS_ENABLED` is on.
6. **Label strings are `State.String()`** (`half-open` with a hyphen), matching WARN `from` / `to`.
7. **Register once** (`sync.Once`). The callback and the scrape share the same atomic object. Atomics are seeded in place before the breaker captures their address (`sync/atomic` values must not be copied after first use).
8. **Process-local.** Replicas may disagree. Alerts keep instance / pod identity; never `sum` then compare to 1.
9. **`metrics_schema` 1.6 → 1.7.** Consumers pinning `metrics_schema="1.6"` stop matching `mcp_schema_version`. Mixed rollout: query 1.6 or 1.7 until the fleet is on 1.7.
10. **Restart reseeds Closed.** No persistence. An IdP that is still down must retrip from traffic.

## Do not claim

- The metric equals “the next request will be rejected” (after the open timer, the next `Execute` half-opens and probes).
- The metric equals IdP health (exclusions, Retry-After, cache hits, and below-threshold failures all coexist with `closed`).
- Cluster-wide breaker state.
- Every open→closed cycle is visible on `/metrics` (pull sampling; WARN is the complete transition record). A 30s open window can miss a 30–60s scrape.
- A cross-field snapshot with `consecutiveFailures` (separate atomics; a rate-rule trip can open below the consecutive threshold).

## Test plan

- [x] `TestBreakerStateSnapshot_DoesNotMaterializeHalfOpen`: after open timeout, snapshots still report `open` and `Counts().Requests == 0`; only `Execute` moves it
- [x] One-hot for closed / open / half-open, plus a closed→open→half-open→closed scrape sequence
- [x] Family absent when the snapshot reports disabled
- [x] `sync.Once` proven by snapshot call count (mutation-checked: dropping `Once` fails `calls = 2, want 1`)
- [x] Golden schema `metrics_schema="1.7"` with the three series
- [x] Existing breaker / WARN tests still pass
- [x] `go vet`, golangci-lint (0 issues), `-race` on `tokenexchange`, `metrics`, `cmd/server`

## Docs

Operator contract lives in `docs/observability.md`. Hop-2 operators are pointed at the gauge from `docs/authentication.md` and `docs/configuration.md`, including the `OBS_METRICS_ENABLED` prerequisite. CHANGELOG `[Unreleased]` has the additive entry.


solace-aross (dismissed):
Careful design, and the two claims that actually matter both survived sabotage rather than just reading.

**The scrape must never materialize a state transition.** `BreakerStateSnapshot` reads an atomic that `OnStateChange` writes, specifically to avoid calling gobreaker's `State()`, which can lazily advance an expired-open breaker to half-open. Rewired it to call `e.breaker.State()` directly in a scratch worktree: `TestBreakerStateSnapshot_DoesNotMaterializeHalfOpen` caught it immediately — the snapshot flipped to `half-open` after the timeout with no `Execute` call, exactly the bug the passive design exists to prevent. Restored, clean.

**The metric registers exactly once.** Dropped the `sync.Once` guard in `Provider.TokenExchangeBreakerMetrics` — `TestTokenExchangeBreakerMetrics_SnapshotOncePerCollection` caught the doubled snapshot-call count immediately. Restored, clean.

Also verified independently rather than trusting Copilot's (stale) file-summary claim of a 1.4 bump: the actual diff is 1.6→1.7 in `schema/version.go`, the golden fixture, and both changelog/doc mentions — all consistent, no drift. Full `internal/tokenexchange`, `internal/observability/metrics`, and `cmd/server` suites pass under `-race` (744 tests), build/vet clean.

The atomic-seeding order in `New()` (store `StateClosed` into `exchanger.breakerState` before constructing the breaker with `&exchanger.breakerState`, on an already-heap-allocated struct) is correct — no post-first-use copy of the atomic. Design decisions around what the metric does and doesn't claim (armed vs. fired, process-local, not equal to "next request rejected") are documented precisely enough that an operator can't over-read the gauge. Nothing open.

solace-aross (approved):
Re-reviewed at this head. No new authored commits since my last pass (`5e8ed302`) — two merges from `main`, nothing else. Rebuilt in a scratch worktree: `go build ./...` and `go vet ./...` clean, full `internal/tokenexchange`, `internal/observability/metrics`, and `cmd/server` suites pass under `-race` (744 tests). Nothing from the prior review changed.

PR #418, opened 2026-09-14
SOL-153694: declare an output schema for describe-semp-schema

solace-awaheed:
## Summary

`describe-semp-schema` now declares an `outputSchema`, so clients can validate the `structuredContent` it returns. It was the only tool in the server returning structured content without one.

Tracked under [SOL-153694](https://sol-jira.atlassian.net/browse/SOL-153694).

## Type of Change

## Motivation and Context

MCP 2025-11-25 says a client SHOULD validate structured results against the declared output schema. With nothing declared there was nothing to validate against, and an LLM consuming the result had no type information for it.

The cause is the same registration bypass behind this tool's other known gaps: it registers directly on the server rather than through `Te every other tool's output schema is attached.

## Changes Made
                                                                                                                                                                                                                                                                                 - Declare `OutputSchema` on `describe-semp-schema` (`internal/tools/describe_semp_schema.go`).
  - Only `operation` and `method` are required — the rest are conditional. `describe()` emits `definition` only when the operation has a request body, `note` only when it does not, `attributes` only in the trimmed view, and `schema` only in the raw view.                                                  - The raw view's `schema` is left unconstrained: it is the OpenAPI definition verbatim, a passthrough of the spec rather than a shape t
  - The `attributes` item schema is recursive — a `$ref`-backed object attribute nests the same per-attribute shape under its own `properties`. Those nested entries never receive writability flags (`trimAttributes` returns early before setting them), so `writableOnCreate`/`writableOnUpdate` are       declared but cannot be required.
  - Both schemas are closed (`additionalProperties: false`). This is load-bearing, not decorative: the tool bypasses `ToolManager`, and the SDK's untyped `AddTool` leaves result validation to the caller, so **nothing validates this output at runtime**. A field added to `describe()` or `trimAttributes`without a matching entry here would otherwise ship undocumented and unnoticed.
- Remove the `toolsWithoutOutputSchema` allow-list in `cmd/server/conformance_test.go` rather than emptying it, this having been its only entry. Every registered tool now declares an output schema, with no exception to pin. Two comments that cross-referenced the map are reworded to describe the shape without naming it.                                                                                                                                                                                                                                                                                            - `CHANGELOG.md` entry under `[Unreleased]`.

## Testing
                                                                                                                                                                                                                                                                                                              **Test Environment:**
- Go version: 1.26.5 (module targets 1.25.0)                                                                                                                                                                                                                                                                  - OS: macOS 26.6.2 (arm64)
- Broker version: n/a — no broker required                                                                                                                                                                                                                                                                    
**Tests Performed:**                                                                                                                                                                                                                                                                                          - [x] Unit tests added/updated
- [x] Tested locally with `go test -race` (see caveat below)

**Test Coverage:**                                                                                                                                                                                                                                                                                            
Two tests stand in for the runtime validation this tool does not get:                                                                                                                                                                                                                                         
- `TestDescribeSempSchema_OutputMatchesDeclaredSchema` validates real output for all 633 indexed operations in both views (1,266 documents) against the compiled declaration. `cmd/server/main.go` passes the same embedded `specs.FS` the test iterates, so coverage is coextensive with production rather   than a hand-picked sample.
- `TestDescribeSempSchema_OutputSchemaRejectsUndeclaredFields` mutates a valid document at each of the three instance levels (top level, attribute, attribute nested under a `$ref`-backed one) and requires all three to be rejected — so the positive test cannot pass against a permissive schema, and     neither the top-level `$ref` nor the attribute definition's own self-reference can silently fail to resolve.

Also verified: full conformance suite green (the allow-list removal now holds every tool to declaring a schema), `go vet` clean, `golangci-lint` 0 issues, coverage 89.7% against the 85% floor, clean under `-race -count=2` across `cmd/server` and `internal/tools`.                                       
**Caveat on `go test -race ./...`:** passes for every package except `test/performance/memsampler`, whose `countOpenFDs` tests read `/proc/<pid>/fd` and fail on macOS. That package is byte-identical to `main` and unmodified by this PR; it passes on CI's Linux runners. Because `make` stops there, the coverage gate was reproduced over its own scope (which already excludes `test/`).                                                                                                                                                                                                                             
## Checklist
                                                                                                                                                                                                                                                                                                              ### Code Quality
- [x] Self-review completed                                                                                                                                                                                                                                                                                   - [x] Code is well-commented (explains "why", not "what")

### Security                                                                                                                                                                                                                                                                                                  - [x] **No credentials in code**
- [x] Followed secure logging rules — `/check-logs` clean; this diff adds no `slog`, `log`, `fmt.Print`, or `os.Stdout` calls                                                                                                                                                                                 - [x] No security vulnerabilities introduced (golangci-lint/gosec clean)
                                                                                                                                                                                                                                                                                                              ### Testing
- [x] Error paths tested                                                                                                                                                                                                                                                                                      - [x] Test names clearly describe what is being tested
                                                                                                                                                                                                                                                                                                              ### Documentation
- [x] CHANGELOG.md updated
- [x] Doc comment added on the new function
- n/a README / docs / examples — no user-facing or architectural change                                                                                                                                                                                                                                       
### Git & CI                                                                                                                                                                                                                                                                                                  - [x] All commits are signed off (DCO)
- [x] Commits have clear, descriptive messages                                                                                                                                                                                                                                                                - [x] Branch is up to date with main (fast-forwarded; 0 commits behind)
- [x] No merge conflicts                                                                                                                                                                                                                                                                                      - [x] CI checks passing — re-running at `a8310f2`. At `a38fbf0` three required checks were red: `FOSSA licensing`, `Guardian scan gate` and `Third-party inventory refresh scripts self-test (offline) (ubuntu-latest)`. `Guardian scan gate` is an aggregate and was red only because its `R_LICENSE` input was (`R_BUILD`/`R_VULN`/`R_PRISMA` all succeeded), so that is two causes, not three. Neither is this diff: it touches no `go.mod`, manifest or workflow, the inventory job failed inside its own fixture with `Could not resolve a single, unambiguous licence for golang.org/x/oauth2 via go-licenses` while passing on `macos-latest` and in the network variant, and all three pass on #419 at the same base.

## Additional Notes                                                                                                                                                                                                                                                                                           
**This does not make the tool conformant in its other two respects,** which stay open and are tracked separately: it still answers schema-invalid input with a JSON-RPC error rather than an `isError` tool result and misreports a wrong-typed `operation` as missing ([SOL-153693](https://sol-jira.atlassian.net/browse/SOL-153693)), and that error still reaches the wire carrying `"code": 0` ([SOL-153692](https://sol-jira.atlassian.net/browse/SOL-153692)). Please don't read a declared output schema as either of those being closed.
                                                                                                                                                                                                                                                                                                              Two observations, neither introduced by this PR:
                                                                                                                                                                                                                                                                                                              - The `with-write-tools` `tools/list` payload is now at **83.4%** of its 25,000-token budget (20,858 tokens), ~516 of which are this scheot raise the ceiling — flagging it so whoever next trips `TestToolListBudget` doesn't wrongly blame their own change.
- `describe_semp_schema.go:153` (`resp["schema"] = def`) hands back a live reference to the embedded spec's definition map in the raw view, rather than a copy. Nothing mutates it today. Pre-existing; may deserve its own ticket.                                                                           
## Related Issues/PRs                                                                                                                                                                                                                                                                                         
- Blocks [SOL-153693](https://sol-jira.atlassian.net/browse/SOL-153693) — routing this tool through `ToolManager` requires an output schema to validate against
- Related to [SOL-153692](https://sol-jira.atlassian.net/browse/SOL-153692) — shares two files, no logical dependency                                                                                                                                                                                         - Follows from [SOL-150761](https://sol-jira.atlassian.net/browse/SOL-150761), whose conformance suite pinned this gap

---

**For Reviewers:**                                                                                                                                                                                                                                                                                            
**Focus Areas**: The closed schemas are the main judgment call. Nothing validates this output at runtime today, so a too-strict schema is currently a test failure rather than a production one — but SOL-153693 will route this tool through `ToolManager`, at which point strictness becomes a live failure mode (cf. SOL-154164, where strict output schemas turned nine successful writes into reported failures). I chose closed + broad test coveema; worth a second opinion.

Also: this is the first declared schema in the repo to use `$ref`/`definitions` — every other one is fully inlined. I confirmed it reachendalone schema document, so `#/definitions/attribute` resolves. The residual risk is a client with a strict-subset JSON Schema parser.

**Questions**: Should the `TestToolListBudget` ceiling be bumped in this PR, or left for whoever trips it?


internal/tools/describe_semp_schema.go
```diff
@@ -271,6 +271,102 @@ func trimAttributes(def map[string]any, defs map[string]any) []map[string]any {
...
+				"type":        "string",
+				"description": "Name of the SEMPv2 request-body definition backing the operation. Absent when the operation takes no request body.",
+			},
+			"note": map[string]any{
+				"type":        "string",
+				"description": "Present only when the operation has no request-body definition, stating that there are no attributes to describe.",
+			},
+			"attributes": map[string]any{
+				"type":        "array",
+				"description": "Trimmed view: one entry per configurable attribute, sorted by name. Empty when the operation takes no request body. Absent in the raw view.",
```
bczoma: 🟡 **"Absent in the raw view" is false for 55% of operations — and the same claim is in the CHANGELOG**

Flagged by `good:code`; Copilot raised it first on this line and is correct. Confirming it rather than restating it, and adding the two places it propagated to.

**What the problem is.** `describe()` sets `attributes` on the no-request-body branch and returns **before** the view check:

```go
if info.defName == "" {
    resp["note"] = "operation has no request body definition; no attributes to describe"
    resp["attributes"] = []any{}
    return resp, nil          // ← returns here; view is never consulted
}
...
if view == "raw" { resp["schema"] = def } else { resp["attributes"] = trimAttributes(def, defs) }
```

So a `raw` call on a bodyless operation does emit `attributes`. Measured against the embedded specs: **346 of 633 indexed operations (55%)** have `defName == ""`, and for every one of them the raw view returns `attributes: []`. Verified directly — `describe("config/deleteMsgVpnRestDeliveryPointRestConsumerOauthJwtClaim", "raw")` returns `map[attributes:[] method:DELETE note:... operation:...]`.

**Why it matters, and why it is only 🟡.** The *schema* is fine: `attributes` is declared optional, so an empty array validates, which is exactly why all 1,266 documents pass. Nothing is broken at runtime and no client will reject a valid response. What is wrong is the **description**, and on this tool that is the product: `describe-semp-schema` exists to tell an LLM what shape things are, and this string is the text the LLM reads to interpret the result. A consumer that believes it will not look for `attributes` on a raw call, and will find one on more than half of them.

It has propagated to two more places, which is the part not yet covered on this thread:

- the `describeSempSchemaOutputSchema` doc comment — "attributes in the trimmed view (as an empty array in the no-body case)" frames the no-body case as a trimmed-view sub-case, when it is view-independent;
- the `CHANGELOG.md` `[Unreleased]` entry — "`attributes` only in the trimmed view", which is customer-facing.

**Recommended fix.** A one-line correction in all three places, e.g.:

> Present in the trimmed view, and in the raw view when the operation takes no request body (then always empty). Absent only from the raw view of an operation that has one.

Fixing the *producer* instead — moving `resp["attributes"] = []any{}` behind the view check so the original claim becomes true — is the other option, and arguably the cleaner shape. But it changes a response that already ships, so if you prefer it, it belongs in its own ticket rather than in a PR whose stated scope is declaring what the tool already returns.

bczoma (dismissed):
## PR Review: #418 — SOL-153694: declare an output schema for describe-semp-schema

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-153694` → `main` | **Changes:** +260 / -37 across 4 files
**Jira:** `SOL-153694` — tool returns `structuredContent` without declaring an `outputSchema` (Task)

---

### Verdict

1. 🟡 `internal/tools/describe_semp_schema.go:317` — the `attributes` description says "Absent in the raw view", which is false for 346 of the 633 indexed operations. Copilot raised this; it is correct, and the same claim is repeated in the doc comment and in the customer-facing CHANGELOG entry.

### Summary

This is a careful, well-scoped PR and the schema itself is right. I enumerated both producers independently: every field `describe()` and `trimAttributes` can emit is declared, nothing is declared that cannot be emitted, and `required: [operation, method]` is exactly correct — `attributes` looks always-present in the trimmed view but is genuinely absent from raw-with-body, so it cannot be tightened. Both new tests do what they claim; I ran them. The negative test mutating at all three instance levels is the load-bearing one, because it is what proves the recursive `$ref` actually resolves at depth 2 — without that, a silently-unresolved recursion would leave every `$ref`-backed attribute's nested list unvalidated while the positive test still passed. The one substantive finding is a description string, not the schema.

Two judgement calls I went looking to disagree with and came away persuaded by. **Declining to remove the `ToolManager` bypass** is right: routing this tool through `CallTool` would need a broker-less dispatch lane *and* a per-tool opt-out of `injectBrokerParam`, and without both the tool gains a required `broker` parameter it has no use for — a breaking client-facing change. That is a design change, not a refactor. Nothing here entrenches the bypass; the schema literal moves wholesale into `Metadata.OutputSchema` on the day that lane exists. **Removing the allow-list rather than emptying it** is also the stronger invariant: an empty `toolsWithoutOutputSchema` map is an affordance inviting the next person to add a row instead of a schema, and the unconditional `assertObjectSchema` leaves no cheap out. It is AC #2 verbatim and mechanically forced — the map is a two-way gate that fails once the tool declares a schema, so the PR could not be green without touching it. Not scope creep.

---

### Minor notes

- 🟢 `docs/tools-reference.md:161` — the **Returns (`trimmed`)** line hand-documents this response shape, so it is now a second source of truth alongside the declaration, and the prose copy is already the narrower one: it never mentions `note`, never mentions the raw view's `schema`, and presents `definition` as unconditional where the declaration makes it optional. Only the declaration has a test gate behind it. One sentence naming the declared `outputSchema` as authoritative, plus the two missing branches, closes it.
- 🟢 `internal/tools/describe_semp_schema.go:150` — `definitions` / `$ref: "#/definitions/attribute"` is the draft-04/07 spelling; 2020-12 names it `$defs`. Pointer resolution works either way, but the tests only exercise gojsonschema, so a client on a strict 2020-12 validator flagging `definitions` as unknown is unverified in-repo. Rename, or record in the doc comment that the draft-07 spelling is deliberate.
- 🟢 `internal/tools/describe_semp_schema.go:295-336` — the schema is closed and requires `operation`+`method`, but `withRecovery`'s panic result emits `{error, retryable}`, which it rejects. Harmless today, and it is a repo-wide convention rather than a defect you introduced (`buildLocalErrorResult` bypasses `buildValidatedResult` too). Worth one line in the doc comment saying the declaration describes the success shape — and worth putting "must the error envelope be schema-valid?" explicitly into SOL-153693, since that ticket converts these into `isError` tool results and the question should be answered once, repo-wide.
- 🟢 `internal/tools/describe_semp_schema.go:295` — a forward note for whoever removes the bypass: once this tool joins the validated path, `buildValidatedResult` turns a schema mismatch into a user-visible `output_validation_error` at runtime. Closing the schema is the right call *today*, when the cost of being wrong is a CI failure; the trade-off inverts then. Worth recording in the bypass ticket so it is re-decided rather than inherited.
- 🟢 PR hygiene — the "CI checks passing" checklist box is still unticked while every required check on `a5c7c391` is green, and Copilot's thread is unanswered with no reply and no follow-up commit. Both cost a reviewer a detour; neither is a merge risk.

---

### Summary Verdict

|---|---|---|
| Correctness | ⚠️ — schema correct, one description string wrong | good:code |
| Documentation | ✅ | docs |

internal/tools/describe_semp_schema.go
```diff
@@ -271,6 +271,102 @@ func trimAttributes(def map[string]any, defs map[string]any) []map[string]any {
...
+				"items":       map[string]any{"$ref": "#/definitions/attribute"},
```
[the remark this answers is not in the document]
solace-awaheed: Confirmed and fixed in a38fbf0 — you were right, and I verified it against the embedded specs before changing anything rather than taking it on the description alone.

`describe()` sets `attributes = []any{}` on the no-request-body branch and returns before `view` is consulted, so a raw call on a bodyless operation does emit it. Measured: **346 of 633 indexed operations (55%) are bodyless, and all 346 emit `attributes` in the raw view** — so the claim was wrong for the majority of raw calls, not an edge case.

I took the "document the no-body exception" option rather than changing `describe()`: the producer change alters a response that already ships, so it belongs in its own ticket.

Corrected in all three places the claim had reached — the `attributes` description here, the `describeSempSchemaOutputSchema` doc comment, and the `CHANGELOG.md` entry.

One thing worth recording, because it explains how this survived: the *schema* never needed changing. `attributes` is optional, so the empty array validates either way, which is exactly why 1,266 passing documents could not catch a description that was wrong about *when* the field appears. Only a behavioural assertion can, so `TestDescribeSempSchema_RawViewOfBodylessOperationEmitsAttributes` now pins the asymmetry — and it is a two-way gate: if someone later moves the assignment behind the view check, it fails and names the three descriptions that have to change back with it.

internal/tools/describe_semp_schema.go
solace-awaheed: Fixed in a38fbf0. I reproduced your measurement independently before changing anything, and it lands exactly where you said: **346 of 633 (55%)** bodyless, and all 346 emit `attributes` in the raw view. `describe("config/deleteMsgVpnRestDeliveryPointRestConsumerOauthJwtClaim", "raw")` returns the shape you quoted.

Took your recommended option — corrected the description in all three places rather than moving `resp["attributes"]` behind the view check. Agreed the producer fix is the cleaner shape, and agreed it changes a shipped response and so wants its own ticket; happy to file it if you want it tracked.

Your framing of why it is only 🟡 but still worth fixing is the part I acted on hardest. Since the schema was never wrong, no amount of schema validation could catch this — that is precisely how the wording survived 1,266 passing documents. So the fix is not just three string edits: `TestDescribeSempSchema_RawViewOfBodylessOperationEmitsAttributes` now asserts the behaviour, and fails in both directions. If the producer fix lands later, that test breaks and names the descriptions that must change with it, so the two cannot drift apart again.

**Two further defects came out of the same pass**, neither on this thread:

1. **The writability claim in the CHANGELOG was backwards** (Copilot had this in a suppressed comment and was right). I had written "those nested entries carry no writability flags". `trimAttributes` returns early only for the `$ref`-backed *parent*, so the parent lacks `writableOnCreate`/`writableOnUpdate` while its nested children go through the normal path and receive both. Verified on `config/createMsgVpnClientProfile`: parent `eventClientProvisionedEndpointSpoolUsageThreshold` carries neither, child `clearPercent` carries both. The conclusion holds — the flags are declared but cannot be required — but for the opposite reason to the one I gave.

2. **"Every other tool inherits its output schema from ToolManager" ignores `list-brokers`**, which also registers directly and has always hand-declared its own. Copilot flagged this on the doc comment as a nit; I found the same phrasing had propagated into the CHANGELOG entry, so both are corrected.

Net effect: no schema or behaviour change in this commit, only claims corrected to match what the code does, plus one test so the corrected ones stay honest.

bczoma (dismissed):
## PR Review: #418 — follow-up at `a38fbf05`

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-153694` → `main` | **Changes:** +357 / −37 across 4 files
**Reviewed:** the authored delta since my review at `a5c7c391` — the `ffa0045e` merge of `main` and `a38fbf05` (2 files, +106/−9), plus the CHANGELOG text as it now reads
**Jira:** `SOL-153694` — declare an output schema for `describe-semp-schema`

---

### Verdict

**Recommendation:** ✅ Approve

The one 🟡 this verdict was conditioned on is addressed, in all three places the false claim had reached, and the new work adds no finding of its own. The four remaining prior 🟢 notes are recorded in the table below and are not conditions of this verdict.

**Not a merge signal:** this approval is about the code. Three required checks are red at this head — `FOSSA licensing`, `Guardian scan gate`, and `Third-party inventory refresh scripts self-test (offline) (ubuntu-latest)` — and GitHub reports the PR `BLOCKED`. None of them is plausibly this diff's doing: the change touches no dependency, no manifest and no workflow, the same three checks pass on #419 at the same base, and the inventory self-test fails inside its own fixture with `Could not resolve a single, unambiguous licence for golang.org/x/oauth2 via go-licenses` — an environment-dependent resolution failure that passed on `macos-latest` and in the network variant of the same job. That reads as infrastructure rather than code, so a re-run (or an EngPlat ticket if it persists) is the path, not an edit. Worth knowing because the PR body still says "CI checks passing — not yet run; workflows are PR-triggered", which is no longer true and would leave you thinking nothing needs chasing.

### Summary

The correction is the right one and it was made the right way. The author reproduced the measurement independently before changing anything — 346 of 633 indexed operations bodyless, all 346 emitting `attributes` in the raw view — and then took the documentation option rather than moving `resp["attributes"]` behind the view check, on the reasoning that the producer fix alters a response that already ships and so belongs in its own ticket. I agree, and the reasoning is recorded on the thread rather than only in a commit message.

What makes this more than three string edits is `TestDescribeSempSchema_RawViewOfBodylessOperationEmitsAttributes`, which pins the asymmetry in both directions: the bodyless branch emits `note` and an empty `attributes` for *both* views and never `definition` or `schema`, and the with-body raw view is the single case that omits `attributes`. That test earns its place for a reason the author identified and I had not: because `attributes` is declared optional, the empty array a bodyless raw call returns validates either way, so no amount of schema validation could ever have caught the wrong *description* — which is exactly how "Absent in the raw view" survived 1,266 passing documents. The new doc comment also states the residual count (287 of 633 with a request body) rather than the direction that is easy to get backwards, and the CHANGELOG entry now carries the corrected claim too. `go test ./internal/tools/ -run TestDescribeSempSchema` passes.

*6 prior findings — 2 addressed, 3 not addressed, 1 not re-checked. The exceptions:*

|---|---|---|---|
| 🟡 | `attributes`' "Absent in the raw view" is false for 55% of operations, and the claim had propagated to the doc comment and the CHANGELOG | Addressed | `internal/tools/describe_semp_schema.go:322` now reads "Absent only from the raw view of an operation that has a request body"; the doc comment at `:274-288` and `CHANGELOG.md:24` carry the same corrected claim; `describe_semp_schema_test.go` adds the behavioural pin |
| 🟢 | `docs/tools-reference.md:161`'s hand-documented **Returns (`trimmed`)** line is a second, narrower source of truth beside the declaration | Not addressed | `docs/tools-reference.md:161` is unchanged and is not in the diff — still no `note`, no raw-view `schema`, and `definition` presented as unconditional |
| 🟢 | `definitions` is the draft-04/07 spelling; 2020-12 names it `$defs`, and only gojsonschema is exercised | Not addressed | `internal/tools/describe_semp_schema.go:332` still `definitions`; the doc comment was rewritten in this delta and does not record the spelling as deliberate |
| 🟢 | One line in the doc comment saying the declaration describes the *success* shape, since `withRecovery`'s `{error, retryable}` would be rejected by it | Not addressed | The doc comment at `:274-292` was rewritten in this delta without it |
| 🟢 | Record in the bypass ticket that closing the schema is the right call today but the trade-off inverts once `buildValidatedResult` applies | Not re-checked | Concerns a note in SOL-153693 / SOL-153692, not the code at this head — I did not open the tickets this pass |

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ — the corrected description and doc comment match what `describe()` actually emits, in both directions |
| Test Coverage | ✅ — the new test pins the asymmetry a schema structurally cannot catch |
| Documentation | ⚠️ — the CHANGELOG and declaration are now correct and consistent, but `docs/tools-reference.md:161` remains a narrower parallel source of truth |
| Architecture | — not re-reviewed (no design change in this delta) |
| Code Quality | — not re-reviewed (unchanged since `a5c7c391` outside the delta) |
| Security | — not re-reviewed (no security surface in this delta) |
| PR Scope & Description | ⚠️ — the "CI checks passing" line is stale; CI has run and three required checks are red (see the verdict note) |

*Assessed by the parent on a follow-up pass; no discipline lenses were dispatched.*

bczoma (approved):
## PR Review: #418 — follow-up at `a8310f22`

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-153694` → `main` | **Changes:** +388 / −38 across 5 files
**Reviewed:** the one commit since my review at `a38fbf05` — `a8310f22` (2 files, +31/−1)
**Jira:** `SOL-153694` — declare an output schema for `describe-semp-schema`

---

### Verdict

**Recommendation:** ✅ Approve

All three 🟢 notes this pass could act on are addressed, and each carries a claim I was able to verify rather than take on trust. Nothing new in the delta, and no conditions on this verdict.

One prior 🟢 remains open in the table below — a note asked for on SOL-153693, not code at this head — and it is not a condition of this verdict.

**CI is no longer a concern.** The three red required checks my last review flagged as infrastructure rather than code have all gone green at this head: 34 passing, 3 e2e jobs still running, 2 skipped. No re-run to chase and no EngPlat ticket needed. The PR body was also corrected to describe that accurately, including the useful catch that `Guardian scan gate` was an aggregate red only because its `R_LICENSE` input was — two causes, not three.

### Summary

This commit closes the three documentation notes from the last pass, and in each case it adds the reasoning rather than just the sentence. `docs/tools-reference.md:161` no longer carries a hand-written **Returns (`trimmed`)** line that was a narrower parallel source of truth — it is now a three-row table over (has body × view) that matches `describe()` exactly, including the bodyless branch answering both views identically, and it points at the declared `outputSchema` as the thing to validate against rather than the prose. The `definitions` spelling is now recorded as deliberate with the reason it has to be: `internal/tools/validation.go:22` compiles with `xeipuuv/gojsonschema`, a draft-04/06/07 implementation that does not resolve the 2020-12 `$defs` rename, and the comment names the consequence (a 2020-12-only client would not resolve the recursion) and the condition to revisit it. The success-shape paragraph is correct too — `withRecovery` really does answer a panic with a `StructuredContent` of `{error, retryable}` at `internal/tools/register.go:97-101`, which this closed schema requiring `operation` and `method` would reject — and it explains why a `oneOf` belongs with the shared error envelope rather than here. Checked the arithmetic: 633 − 346 bodyless = the 287 with-body operations the doc comment states.

*5 prior findings — 4 addressed, 1 not addressed. The exception:*

|---|---|---|---|
| 🟢 | Record in the bypass ticket that closing the schema is the right call today but the trade-off inverts once `buildValidatedResult` applies | Not addressed | Read SOL-153693's six comments — none mentions the closed-schema trade-off. Last pass I did not open the ticket; this pass I did, so this is a genuine gap rather than an unchecked one. Still a note in a ticket, not code, and not a condition of this verdict |

The four addressed: the 🟡 `attributes` "Absent in the raw view" correction (closed last pass at `a38fbf05`); `docs/tools-reference.md:161`'s parallel source of truth, now the (has body × view) table at `:161-172`; the `definitions`/`$defs` spelling, now recorded as deliberate at `internal/tools/describe_semp_schema.go:310-318` with the gojsonschema draft level as the reason; and the success-shape-only caveat at `:300-309`.

### Minor notes

🟢 **Minor** — not a condition of the verdict.

- 🟢 PR body, Documentation checklist — `n/a README / docs / examples — no user-facing or architectural change` is now stale: this commit adds 14 lines to `docs/tools-reference.md`, which is exactly a user-facing doc change. One line to flip.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Documentation | ✅ — the reference table, the doc comment and the CHANGELOG now agree with each other and with `describe()`; the one parallel source of truth was the substance of this commit |
| Correctness | ✅ — every factual claim added in this delta verified against the code it describes (`validation.go:22`, `register.go:97-101`, `describe()` at `:133-157`) |
| PR Scope & Description | ✅ — the stale "CI checks passing — not yet run" line is corrected and now more precise than my own note was |
| Test Coverage | — not re-reviewed (no test change in this delta; the behavioural pin landed at `a38fbf05`) |
| Architecture | — not re-reviewed (no design change) |
| Code Quality | — not re-reviewed (comments only in this delta) |
| Security | — not re-reviewed (no security surface) |

*Assessed by the parent on a follow-up pass; no discipline lenses were dispatched.*

PR #419, opened 2026-09-14
Record hop-2 expiry fallback first use at INFO (SOL-154334)

amitmorade:
## Summary

- Default production INFO only reported that hop-2 `token_expiry_fallback` was **armed**. The first time a live exchange on that exchanger actually applies that duration, emit one INFO `broker OAuth token expiry fallback supplied a lifetime` with `expiry_fallback`.
- Latch is a per-`Exchanger` `atomic.Bool` CAS on the existing live-parse path (`doExchange`), not a counter, not a process-global once-log catalog, and not a promotion of every Debug `used_fallback` line. Production constructs one exchanger per process, so operators normally see one first-fire line per server.
- Operator docs now separate armed / used / first-fire. Tracked under [SOL-154334](https://sol-jira.atlassian.net/browse/SOL-154334).

## Why this wiring

`parseSuccessBody` already decides `usedFallback`. `doExchange` already logs the Debug issued line from that flag. The once-bit belongs on `Exchanger` next to the other atomics (one exchanger per process in production). `slog.Info` is context-free so the line is not attributed to one MCP call.

## Rejected

- Debug-once (still invisible at default level)
- WARN-once (configured RFC-allowed path is not a fault)
- Per-exchange Info
- Logging from `New` / `main` (armed, not used)
- Package-global Once / a shared “one-off logs” registry
- Metrics counter
- `broker` / `correlation_id` / `token_*` keys on the new line

## Unchanged

- Fail-closed without fallback; positive IdP `expires_in` wins; negative/overflow never replaced
- Debug `identity provider issued broker token` + `used_fallback`
- Startup INFO `token exchanger created for broker OAuth` (armed)
- YAML / parse of `token_expiry_fallback`

## New

- At most one first-fire INFO per `*Exchanger`, only when fallback supplied the lifetime on a live IdP success
- Attr is the configured duration (`expiry_fallback`), same encoding as startup

## Do not claim

- Armed means used
- This request / this broker used fallback
- Missing later INFO means the IdP started sending `expires_in`
- The token was cached or longer than skew
- A second exchanger in the same process is silenced by the first (tests construct two; production does not)

## Example

```json
{"level":"INFO","msg":"broker OAuth token expiry fallback supplied a lifetime","expiry_fallback":3600000000000}
```

## Test plan

- [x] First live omit/`0`/`null` `expires_in` emits exactly one INFO visible at `slog.LevelInfo`
- [x] Second live fallback on the same exchanger does not repeat it
- [x] Cache hit does not repeat it; positive IdP `expires_in` never emits it
- [x] Fail-closed / negative / overflow never emit it
- [x] No `correlation_id` on the usage line; two `New()` instances each fire once
- [x] Mutation-check: `Debug` instead of `Info` → Info-level count 0; no CAS → second live count 2
- [ ] `make check` / CI on this PR


solace-aross (approved):
Clean. Adds a first-fire INFO (`broker OAuth token expiry fallback supplied a lifetime`) the first time a Hop-2 exchanger actually applies `token_expiry_fallback`, closing the gap between "armed" (startup INFO) and "used" (per-exchange Debug) that had no INFO-visible signal at all. The `atomic.Bool` CAS on `*Exchanger` is the right scope, not a package-global or a counter, since production constructs one exchanger per process.

Sabotage-verified both load-bearing pieces in a scratch worktree rather than trusting the test plan's checklist:

- Removed the CAS gate (`if !e.expiryFallbackLogged.CompareAndSwap(...)`) — `TestExchange_ExpiryFallbackUsedINFO_SecondLiveDoesNotRepeat` caught it immediately (count went 1 → 2).
- Demoted the line from `slog.Info` to `slog.Debug` — all three `TestExchange_ExpiryFallbackUsedINFO_FirstFireVisibleAtInfo` subtests failed as expected (count 0, since the capture only listens at Info level).

Restored both, reran clean. 685 tests pass across `internal/tokenexchange` and `cmd/server` under `-race`.

No credential or per-call data on the new line, just the configured duration — matches the pattern the docs and `secure-logging-rules.md` already establish for the startup line. `docs/authentication.md`/`docs/configuration.md` updates accurately describe the three-tier armed/used/first-fire behavior against what the tests actually assert (no `correlation_id`, no `broker`, second exchanger fires independently). CI green. Nothing open.

bczoma (approved):
## PR Review: #419 — Record hop-2 expiry fallback first use at INFO (SOL-154334)

**Author:** amitmorade | **Branch:** `amorade/SOL-154334` → `main` | **Changes:** +398 / −7 across 9 files
**Jira:** `SOL-154334` — Record hop-2 expiry fallback firing at default log level (Verifying)

---

### Verdict

**Recommendation:** ✅ Approve

No conditions. The five 🟢/💡 notes below are **not** conditions of this verdict.

### Summary

The wiring is right and, unusually, provably right: the CAS sits in `doExchange` after `parseIdPResponse` has already succeeded, and I traced every path that must *not* emit — cache hit (returns from `Exchange` before singleflight is entered), singleflight follower, gate/breaker rejection, fail-closed, negative and overflow `expires_in`, and a positive IdP `expires_in` — none of them reach the latch, so the line never claims a lifetime that was not supplied and never burns the latch on an exchange that then failed. All seven acceptance criteria have both code and a non-vacuous test behind them, including the one that usually gets faked: `captureJSONLogsAt` makes "visible at production INFO" an assertion about the level rather than a claim, and the two mutation checks the PR body promises do hold (`slog.Info`→`Debug` fails `FirstFireVisibleAtInfo`; removing the CAS fails `SecondLiveDoesNotRepeat`). Choosing a per-`Exchanger` latch over the package-global `sync.Once` the body lists as rejected is the right call for a reason worth stating: the global would have silently suppressed a second exchanger's *first* fallback use, so this design degrades by getting noisier rather than by going quiet.

Two findings were filed at higher severity and both were reduced by measurement. A 🟡 that the "one exchanger per process" fact is asserted in six places and enforced nowhere came down to 🟢 once it was clear all six state the per-exchanger rule as the load-bearing claim and the deployment shape only as a hedged aside — and that the single source the fix asked for already exists on the struct field comment. A finding that the INFO line carries no `broker` alias, which three lenses reached independently, came down to 💡: the adjacent Debug line carries `broker` + `used_fallback` on *every* exchange and this PR's own docs route the operator there, and on a latched line a bare `broker` key would read as sole attribution when it only means "whichever broker raced first". Both Copilot doc findings are resolved at this head.

---

### Minor notes

- 🟢 `docs/authentication.md:328`, `docs/configuration.md:146` — both append "Production constructs one Hop 2 exchanger per process, so operators normally see one first-fire line per server" to a per-exchanger rule that is already complete on its own, and the invariant's single source already lives at `internal/tokenexchange/exchanger.go:66-70`. Drop the aside from the two operator docs; Copilot already had to correct this exact overreach once inside this PR.
- 🟢 `internal/tokenexchange/expiry_fallback_used_log_test.go:320-347` — `ConcurrentFirstFireOnce` asserts only `usage count == 1` and never that two live IdP calls happened, so if the two goroutines ever collapsed into one exchange it would pass with the CAS removed. Add a `var calls atomic.Int32` in the handler (as `CacheHitDoesNotRepeat` already does) and assert `calls.Load() == 2`.
- 🟢 `docs/authentication.md:328` — neither operator doc says `expiry_fallback` renders in JSON as an integer nanosecond count (`1h` → `3600000000000`), so an operator building the log filter these paragraphs invite writes `"expiry_fallback":"1h"`, gets no match, and concludes the fallback never engaged. One clause, or the example record.
- 🟢 `internal/tokenexchange/expiry_fallback_used_log_test.go:57-66` — `usageRecords` is a third record-filter helper beside `findMsg` (`exchange_test.go:2692`) and the new `countMsg` (`:2523`). Fold it into one `filterMsg` next to the others.
- 💡 `internal/tokenexchange/exchange.go:715-726` — `logExpiryFallbackOnce`'s doc comment reasons the absence of `correlation_id` but not the absence of a broker alias, though the alias is in hand one line above and the omission is test-locked at `expiry_fallback_used_log_test.go:104-106`. One clause saying the latch fires on whichever broker raced first, so a bare `broker` key would read as sole attribution while later brokers stay silent at INFO, stops this being re-litigated; the same sentence is worth having in the operator docs, which likewise do not say the line carries no `broker` and no `correlation_id`.

---

### Summary Verdict

|---|---|---|
| Documentation | ✅ | docs |

PR #420, opened 2026-09-14
SOL-153079: fix YAML comment substitution, reorder Quickstart, and un-hardcode the example VPN

McCly-Solace:
## Summary

Fixes three of the setup-friction root causes from the Aug 11 Testapalooza (SOL-153079): a YAML config-parsing bug, an unordered Quick Start, and a hardcoded example VPN name that silently fails on Solace Cloud.

## Type of Change

- [x] Documentation update

## Motivation and Context

Aug 11 testers spent up to ~40 minutes on a from-scratch setup explicitly designed to test onboarding from zero knowledge, due to three compounding issues: a YAML parser edge case that misattributed a config error, a Quick Start covering four deployment paths on one page with no ordering signal, and an example prompt hardcoding a VPN name ("default") that doesn't exist on Solace Cloud.

## Changes Made

- `internal/config/config.go` — `splitYAMLComment`'s quote-tracking no longer flips on every `'`/`"` it sees; `isScalarStart` gates each one on whether YAML actually permits a scalar to start there. This closes it in both directions: an apostrophe or bare `"` in unquoted text (`don't`, `6" pipe`), or a quote after a word-separating space, no longer opens a fake string and swallows a real `#` comment; and `:`/`-`/`?` now require trailing whitespace, `,` is only a delimiter inside a flow collection (tracked via `flowDepth`), and a quote preceded by a YAML anchor/alias/tag (`&pw "..."`, `!!str "..."`) is skipped back over correctly — so a second same-type quote after a real delimiter, or a tagged/anchored quoted value, no longer mis-toggles and either swallows a live comment or leaves a commented-out `${VAR}` unsubstituted with no error.
- `internal/config/config_test.go` — `TestLoadConfig_ApostropheBeforeInlineCommentDoesNotLeakSubstitution`, `TestLoadConfig_DoubledSingleQuoteEscapeInsideQuotedValue`, `TestLoadConfig_BareDoubleQuoteInUnquotedTextDoesNotLeakSubstitution`, and two word-separating-space variants reproduce the original bug classes; `TestSplitYAMLComment` and `TestIsScalarStart` cover the delimiter/flow-depth/anchor-skip logic directly, including the regression and silent-failure cases found in review; `FuzzSplitYAMLComment` checks the scanner never panics and always round-trips `active+comment` back to the original line.
- `README.md` — Quickstart's four deployment methods reordered (Binary → Docker → go install → Kubernetes), with Binary called out as the default starting point; the unsupported "simplest to most advanced" ordering claim was dropped rather than kept.
- `README.md`, `docs/user-guide.md`, `docs/examples.md`, `docs/tools-reference.md` — every hardcoded `"default"` VPN replaced with `` `<your-vpn-name>` ``, backticked everywhere (including the 28 `**Example request:**` prose lines in `tools-reference.md` that GFM was silently stripping as an unbackticked HTML-like tag) so the placeholder actually renders; JSON `"arguments"` blocks also switched to the placeholder rather than keeping the real, working `"default"` VPN name a reader could copy verbatim; the VPN-discovery note consolidated into one canonical wording (per-broker name, `list-vpns` or Solace Cloud console to look it up) instead of four different phrasings.
- `docs/configuration.md` — the substitution note no longer claims this runs "without a full YAML parser" (yaml.v3 fully parses the result) or that a commented-out `${VAR}` "never needs the variable set" as an absolute; reworded to best-effort and explicitly not a security boundary.
- `CHANGELOG.md` — the `[Unreleased]` entry now names all three fixed misread classes (apostrophe, bare `"`, after-space quote) plus the delimiter/anchor fixes, and includes the `environment variables not set` string an affected operator would grep for.

## Testing

**Tests Performed:**
- [x] Tested locally with `go test ./...`

- 5 new `TestLoadConfig_*` reproduction tests, `TestSplitYAMLComment`, `TestIsScalarStart`, and `FuzzSplitYAMLComment` — all pass; ran the fuzz target for 20s locally (3.86M execs, no crashes, no round-trip failures) against the delimiter/flow-depth/anchor-skip logic.
- Full `go test ./...` passes except two pre-existing, unrelated failures in `test/performance/memsampler` (macOS lacks `/proc`, a Linux-only test).
- `gofmt`, `go vet`, `go build ./...` clean.

## Related Issues/PRs

- Tracked under SOL-153079
- Follow-up: SOL-154441 (harden `substituteEnvVars` against newline-containing env-var values — pre-existing, out of scope for this PR, per review discussion)


docs/tools-reference.md
```diff
@@ -263,7 +278,7 @@ useful "has replication been flaky recently?" signal.
 { "broker": "prod-broker", "msgVpnName": "default" }
 ```
 
-**Example request:** "Is replication in sync for the default VPN on prod-broker?"
+**Example request:** "Is replication in sync for <your-vpn-name> on prod-broker?"
```
bczoma: 🟡 **The placeholder that is the point of this PR does not render**

Flagged by: docs

Every `**Example request:**` line here puts `<your-vpn-name>` in a plain Markdown paragraph — no code fence, no backticks. GitHub's sanitizer drops unknown HTML tags, so the placeholder renders as nothing. This line displays on github.com as:

> **Example request:** "Is replication in sync for  on prod-broker?"

**Measured scope:** 29 of the 30 `<your-vpn-name>` instances in this file are unbackticked, on lines matching `^\*\*Example request:\*\*`; 0 are backticked. The one that survives is the convention note at line 43, which *is* backticked. The same placeholder is correctly backticked in `docs/examples.md` and fenced in `README.md:339,364` and `docs/user-guide.md:86,91,96` — this file is the only unprotected set.

**Why it matters:** a reader on github.com sees a prompt with a hole in it, concludes the sentence is broken, and takes the parameter value from the only thing on the page that still shows one — the adjacent JSON block's `"msgVpnName": "default"`. That is precisely the SOL-153079 failure this change was written to remove: on Solace Cloud there is no VPN named `default`, the call comes back empty, and it reads as a broken connection.

**Fix:** backtick the placeholder on every `**Example request:**` line, e.g.

```
**Example request:** "Is replication in sync for `<your-vpn-name>` on prod-broker?"
```

One pass restricted to those lines does it:

```
sed -i 's/<your-vpn-name>/`<your-vpn-name>`/g' docs/tools-reference.md   # then revert line 43, already backticked
```

docs/examples.md
```diff
@@ -96,17 +96,23 @@ See [Authentication](authentication.md) for full OAuth setup.
...
-| "Are there slow subscribers on the default VPN?" | `list-slow-subscribers` | envelope `{ "slowSubscribers": [ ... ] }` (empty array if none) |
+| "Are there slow subscribers on `<your-vpn-name>`?" | `list-slow-subscribers` | envelope `{ "slowSubscribers": [ ... ] }` (empty array if none) |
 | "Are we dropping messages anywhere?" | `get-discard-stats` | `{ "clientDiscards": {...}, "spoolDiscards": {...} }` |
-| "Create a queue orders.q in the default VPN, then subscribe it to orders/>" | `create-queue` then `create-queue-subscription` | a bare `create-queue` alone leaves the queue inert — it attracts no messages until a subscription is added |
+| "Create a queue orders.q in `<your-vpn-name>`, then subscribe it to orders/>" | `create-queue` then `create-queue-subscription` | a bare `create-queue` alone leaves the queue inert — it attracts no messages until a subscription is added |
 
 Read-only tools return their event broker data in a step-keyed envelope. See
 [Tools Reference → Output](tools-reference.md#output-the-step-keyed-envelope).
```
bczoma, line 120: 🟡 **The invocation catalogue tells the reader to replace one placeholder and stays silent about the other**

Flagged by: docs

The `## Tool Invocations by Category` preamble (lines 121-124) says only:

> Replace `prod-broker` with one of your configured aliases (from `list-brokers`).

The 30 JSON blocks below it (lines 137-243) all carry `"msgVpnName": "default"`, with no equivalent instruction. `docs/tools-reference.md:44-48` documents this mixed convention explicitly for *its* JSON blocks — that note is the right call, in the right place — but `examples.md` did not get the parallel sentence, even though this PR added a `<your-vpn-name>` note twenty lines earlier under a different heading.

**Why it matters:** an explicit "replace X" instruction that names one field and not the other reads as "leave `default` alone" — worse than saying nothing. A reader who scrolls straight to the catalogue does exactly as told and sends `{"broker": "my-broker", "msgVpnName": "default"}` to a Solace Cloud service with no VPN of that name. The call returns empty and looks like a broken connection, which is the friction the ticket was filed about.

**Fix:** one more clause on the preamble —

> …and `default` with a Message VPN name from that event broker (`list-vpns`) — there is no VPN literally named `default` on Solace Cloud.

or point at the note that already exists: `See [The msgVpnName Parameter](tools-reference.md#the-msgvpnname-parameter).`


*(Anchored on the nearest line this PR changed; the block in question is outside the diff, so GitHub will not take a comment on it directly.)*

docs/user-guide.md
```diff
@@ -70,7 +70,11 @@ For details on securing this connection with a static token or OAuth/OIDC, see [
 
 ### Example Queries
 
-After connecting, try these example queries:
+After connecting, try these example queries. Substitute `<your-vpn-name>` with
```
bczoma: 🟡 **The deployment table now contradicts the README this PR reordered**

Flagged by: docs

Lines 51-57 say "The server can be deployed in **three** ways", then table Binary / Docker / **Development**, linking to `../README.md#quickstart`. This PR made README enumerate **four** methods and promoted Kubernetes to a first-class entry in both the TOC (`README.md:24`) and the selection list (`README.md:176`). So this table now names a different count, omits Kubernetes, and lists "Development" — which README explicitly holds *out* of the deployment list ("For contributors running from source, see Development Setup"). The PR edits this file without reconciling the table.

**Why it matters:** the user guide is the doc whose job is "how do I run this". An operator reading it sees no Kubernetes row and concludes there is no supported Kubernetes path, so they hand-roll their own Deployment/ConfigMap/Secret rather than using the reference manifests under `deploy/` that README now advertises — and they get none of the settings baked into that manifest, including the `GOMEMLIMIT` fix that landed three commits ago.

**Fix:** "deployed in four ways", rows Binary / Docker / go install / Kubernetes in README's new order, and move "Development" out of the table into a sentence pointing at README's Development Setup, matching `README.md:178`.


*(Anchored on the nearest line this PR changed; the block in question is outside the diff, so GitHub will not take a comment on it directly.)*

internal/config/config.go
```diff
@@ -940,10 +940,23 @@ func substituteEnvVars(data []byte) ([]byte, error) {
 // the line OR preceded by whitespace, AND not inside a single- or
 // double-quoted string on the same line.
 //
+// A ' only opens a single-quoted string at a position where YAML actually
```
bczoma: 🟡 **A newline in a substituted env-var value can inject a top-level config key** — pre-existing; a follow-up ticket, not a change to this PR

Flagged by: good:security + good:architecture

`substituteEnvVars` (lines ~910-926) splices env-var values into the raw YAML bytes verbatim, before `yaml.Decode`. This PR does not touch that function — the change to `splitYAMLComment` only narrows *which* `${VAR}` occurrences get substituted — so the exposure is neither introduced nor widened here. Raising it because it sits on the path this PR changes and the measurement came out worse than expected.

**Measured, against the real `config.LoadConfig` at this head:**

- The obvious defence fails. A bare newline *is* defeated by double-quoting: at `password: "${SEMP_PASSWORD}"`, a value of `p\nallow_insecure_broker_tls: true` yields `password == "p allow_insecure_broker_tls: true"` and the flag stays false — YAML flow-folds the newline inside the quoted scalar. **But** a value of `p"\nenable_write_tools: true #` closes the scalar and comments out the template's own closing quote: result `password == "p"`, `EnableWriteTools == true`. Every shipped file and doc uses the double-quoted form, so quoting is not a mitigation.
- Two real guards, both loud: `dec.KnownFields(true)` (`config.go:637`) makes any *unknown* injected key a hard parse error, and a *duplicate* of a key the file already defines is likewise a decode error. The exploitable set is therefore exactly the top-level keys the victim's file omits — which includes `enable_write_tools`, `allow_insecure_broker_tls`, `allow_remote_unauthenticated` and `tls_terminated_upstream`.
- Indentation checked: the substitution site sits at column 6 (`brokers` → alias → `auth` → `password`), and an injected line at column 0 does close the nested mappings and land in the root mapping. Injection at a nested indent does nothing.
- **Nothing self-heals, and almost nothing is visible.** `enable_write_tools` appears only as one field in the startup summary INFO line (`cmd/server/main.go:1342`) — a flipped value there is not an alarm. `allow_insecure_broker_tls: true` alone is silent; the `INSECURE: TLS verification disabled` WARN (`config.go:1048`) fires only if a broker also sets `insecure_skip_verify`. The truncated credential is not checked at load and stays silent until a 401 at the first SEMP call.

**What holds this at 🟡 rather than higher:** in the default shipped topology no privilege boundary is crossed — `deploy/kubernetes/configmap.yaml:86` and `deploy/kubernetes/secret.yaml:19-21` are written by the same operator, who could set the key directly. It becomes a real escalation only in the topology this repo itself recommends at `deploy/kubernetes/secret.yaml:7-10` ("Use a secret manager: Vault, Sealed Secrets, External Secrets Operator, or SOPS"), where the Secret's writer is not the ConfigMap's author and would gain policy-key authorship — including silently arming broker-mutating tools.

**Suggested follow-up:** fail substitution with a named error when a value contains `\n` or `\r` (a control character in a credential or path is never legitimate); add a regression test asserting a newline-bearing value cannot introduce a key, exercising **both** the quoted and unquoted `${VAR}` shapes; and add the row to `docs/internal/threat-model.md` §4, whose "Configuration and secrets" table (lines 190-205) does not currently contain this threat.


*(Anchored on the nearest line this PR changed; the block in question is outside the diff, so GitHub will not take a comment on it directly.)*

bczoma (dismissed):
## PR Review: #420 — SOL-153079: fix YAML comment substitution, reorder Quickstart, and un-hardcode the example VPN

**Author:** McCly-Solace | **Branch:** `sol-153079-setup-friction` → `main` | **Changes:** +207 / -74 across 7 files
**Jira:** `SOL-153079` — Reduce setup friction: quick-start scope, YAML pitfalls, and the default-message-VPN example prompt (Verifying)

---

### Verdict

Conditioned on these 4 — each has an inline thread:

1. 🟡 `docs/tools-reference.md` — 29 of the 30 new `<your-vpn-name>` placeholders sit unbackticked in plain prose, so GitHub renders them as nothing; one `sed` pass fixes it
2. 🟡 `docs/examples.md:121-124` — the invocation-catalogue preamble tells the reader to replace `prod-broker` and stays silent about `msgVpnName: "default"` in the 30 JSON blocks below it
3. 🟡 `docs/user-guide.md:51-57` — "deployed in three ways" and its table now contradict README's four methods, and omit Kubernetes entirely
4. 🟡 `internal/config/config.go` (`substituteEnvVars`) — **pre-existing, not introduced here**: a newline in a substituted env-var value can inject a top-level config key. Asking for a follow-up ticket, not a change to this PR

### Summary

Three separate setup-friction fixes, and the parser one is the real work: the apostrophe fix is correct, and its test genuinely fails pre-fix — a reviewer reimplemented both scanners to confirm that. The `''` escape branch is not decoration either; it is *required* by gating the open on `isScalarStart`, and the author spotted that. What holds the verdict is the documentation half: the placeholder change that is the point of the PR is invisible in GitHub's rendered view for 29 of 31 tools, which leaves a reader taking `msgVpnName` from the adjacent JSON block's literal `"default"` — the exact failure SOL-153079 was filed about. Three findings were filed at 🟡 against the parser and reduced by measurement: the missing double-quote gate and the "hand-rolled YAML lexer is the wrong seam" argument both turned out to be far smaller than they looked, the latter because the proposed parse-then-substitute replacement re-encodes `port: ${PORT}` as `port: "8080"` and fails the int decode.

---

### Minor notes

- 🟢 `internal/config/config.go:955-959` — the limitations note names the rare `'twas` case but omits the commoner one: after a **space**, `isScalarStart` returns true, so `username: it is 'ere # ${VAR}` still swallows the comment. The PR body's claim that the scanner "now only opens a single-quoted scalar where YAML actually permits one to start" is therefore false as stated — narrow both to what the fix delivers (apostrophes abutting preceding text).
- 🟢 `internal/config/config.go:966-971` — the `"` branch toggles `inDouble` at any quote, ungated by `isScalarStart` (pre-existing on `main`, untouched here). Worst case is a loud startup error naming the variable, so gate it for symmetry when convenient — but it does not close the class on its own.
- 🟢 `internal/config/config.go:960-1007` — `splitYAMLComment`/`isScalarStart` have no direct table test (both new tests drive them through `writeTemp` + full `LoadConfig`), so the accepted limitations live in prose only; a table test would pin them, and one line in `docs/internal/architecture.md` should record *why* substitution is pre-parse (type fidelity for unquoted scalars, one sanitised parse site, the `[]byte` contract the `--health` probe consumes).
- 🟢 `docs/configuration.md:21-31` — the `${VAR_NAME}` section never says substitution is line-based with comment stripping, nor gives the safe idiom: quote values containing `'`, `#`, or `:`.
- 🟢 Two Copilot threads are still open at `ec50f918` with no reply and no follow-up commit — one substantive (the mid-scalar apostrophe above), one a wrong escape notation in a new test comment. A fix or a one-line "out of scope because…" on each would close them out.

On Jira directive 2 ("split the Quick Start … rather than one page covering all four"): the PR reorders and labels rather than splitting, and all four still live on one page. Judged on user outcome that mostly delivers the intent — a newcomer hits Binary first and is told it is the default — so this is not being raised as a finding, but it is worth a sentence on the ticket so the deviation reads as a decision.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

solace-awaheed (changes requested):
### Blocking

#### 1. `<your-vpn-name>` is unbackticked in 28 prose lines and GitHub deletes it

Every `**Example request:**` line in `docs/tools-reference.md` puts the placeholder in bare prose:

```
docs/tools-reference.md:281
**Example request:** "Is replication in sync for <your-vpn-name> on prod-broker?"
```

`your-vpn-name` matches GFM's HTML tag-name pattern, so it parses as a raw inline HTML tag and the sanitizer strips it. What renders is:

> **Example request:** "Is replication in sync for  on prod-broker?"

That's 28 lines in the primary tool reference where the parameter now vanishes entirely — strictly worse than the `default` it replaced. Fix is backticks: `` `<your-vpn-name>` ``.

The PR already does this correctly at `docs/examples.md:105` (inside the table) and `docs/tools-reference.md:47`, so it looks like the `**Example request:**` lines were just missed in the pass. The occurrences in `README.md` and `docs/user-guide.md` are inside fenced blocks and render fine.

---

### Should fix

#### 2. `tools-reference.md` keeps `"default"` in 34 JSON blocks, and adds a paragraph defending it

```
docs/tools-reference.md:44-48
The JSON **example invocation** blocks instead use the literal
string `"default"` for `msgVpnName`, the same way they use `"prod-broker"` for
`broker`: both are illustrative placeholder values showing the call's shape,
not values to reuse as-is.
```

The `prod-broker` analogy doesn't hold. `prod-broker` is self-evidently fake; `default` is a real, working VPN name on a Solace software broker — it's exactly the string this PR exists to stop people copying. The file ends up arguing that the one value with a copy-paste failure mode is fine to leave in, 34 times, in the same doc whose prose was changed to avoid it.

Worth picking one convention. Changing `"msgVpnName": "default"` → `"msgVpnName": "<your-vpn-name>"` in the JSON blocks removes the need for that paragraph entirely.

#### 3. The "no VPN named default" claim is broader than the truth

```
README.md:341
Replace `<your-vpn-name>` with an actual Message VPN name — there's no VPN
named "default" on Solace Cloud.
```

Solace *software* brokers do ship a `default` VPN — this repo's own e2e suite depends on it (`test/e2e-basic-mcp/test-throttling.sh:222`, `test-negative-paths.sh:35`, and others all pass `msgVpnName: "default"`). The Solace Cloud scoping is accurate as written, but someone who just started `solace/solace-pubsub-standard` locally reads "there's no VPN named default" and concludes their working config is wrong.

Stating it positively sidesteps this: *"The VPN name is per-broker. Solace Cloud services use the service name rather than `default`."*

#### 4. The double-quote branch still has the bug this PR fixes for single quotes

`isScalarStart` was added to the `'` branch only (`internal/config/config.go:966`). The `"` branch has the identical structure — any `"` anywhere on the line flips `inDouble`, with no scalar-start guard. Probing the real function:

```
IN      username: 6" pipe # old was ${VAR}
  ACTIVE="      username: 6\" pipe # old was ${VAR}"
  COMMENT=""          <- comment swallowed, ${VAR} substituted

IN      username: don"t # old was ${VAR}
  COMMENT=""          <- same
```

Same failure mode, same blast radius (unset var → hard load failure). Rarer than a contraction, agreed — but the limitations block documents the `'twas` residual case and says nothing about this one, so a future reader will reasonably believe quote handling is now sound in both directions. Either
apply the guard symmetrically (same `isScalarStart(line, i)` call in the open case) or name it in the comment.

#### 5. The `examples.md` intro paragraph needs a rewrite

```
docs/examples.md:98-104
After connecting, ask in plain language. The agent selects the tool and fills
parameters. The examples below say `<your-vpn-name>` where a query needs a
Message VPN — substitute an actual VPN name from your event broker. There is
no VPN literally named "default" on Solace Cloud; if you don't already know
the name, ask "What VPNs are configured on prod-broker?" first (invokes
`list-vpns`), or check the service details in the Solace Cloud console.

Representative queries and the shape they return:
```

Six issues in five lines:

1. **It splits one thought in half.** Previously the lead-in ran *"The agent selects the tool and fills parameters. Representative queries and the shape they return:"* → table. The new text wedges five lines between those sentences, orphaning the table's own lead-in below a digression.
2. **"The examples below *say* `<your-vpn-name>`"** — documents don't say; they use or write.
3. **"substitute an actual VPN name from your event broker"** — reversed idiom. You substitute the new name *for* the placeholder, or *replace the placeholder with* it. As written it instructs the reader to substitute away the thing they want to keep.
4. **"There is no VPN *literally* named"** — "literally" is filler, and the negative framing says what doesn't exist instead of how to find what does (see #3 above).
5. **"if you don't already know the name, ask … first"** — a conditional clause plus "first" to say "to look it up, ask".
6. **It's an aside sitting in the main flow.** `docs/tools-reference.md:8` already establishes `> **Note:**` as the house callout for this kind of caveat.

Suggested replacement:

```markdown
After connecting, ask in plain language. The agent selects the tool and fills
parameters.

> **Note:** `<your-vpn-name>` in the examples below stands in for a Message VPN
> name — replace it with one from your own event broker. The name is
> per-broker; Solace Cloud services use the service name rather than `default`.
> To look it up, ask "What VPNs are configured on prod-broker?" (invokes
> `list-vpns`), or check the service details in the Solace Cloud console.

Representative queries and the shape they return:
```

This restores the lead-in directly above the table, sets the caveat off so it stops interrupting, makes the claim positive and true for both broker flavours, and drops 62 words to 52. It also makes a good canonical wording for #6 below.

---

### Wording

- **Four phrasings of one note.** README, `user-guide.md`, `examples.md`, and `tools-reference.md` each re-explain the VPN placeholder in different words ("there's no VPN named", "don't ship a VPN literally named", "There is no VPN literally named", "it's a per-event-broker configuration value").
Suggest adopting the #5 note as canonical and reducing the other three to one sentence plus a link.
- **Rhetorical questions read as marketing copy, not reference docs.** "Don't know it? Ask…" (`README.md:342`), "Don't know the name? Ask…" (`user-guide.md:75`). The surrounding docs are declarative — *"To find the name, ask …"* matches the register.
- **"don't ship a VPN literally named"** — "ship" is loose for a reference doc and "literally" is doing no work. Appears twice.
- **`user-guide.md:73`** — "Substitute `<your-vpn-name>` with an actual Message VPN name" is reversed; the README gets this right with "R
- **`docs/examples.md:237`** drops the noun: "List the queues on prod-broker's `<your-vpn-name>`." The README's parallel line keeps it ("event-broker-one's `<your-vpn-name>` VPN"). Add "VPN".
- **`README.md:172` — "simplest to most advanced"** is a claim the list doesn't quite support; `go install` sits between Docker and Kuberolchain isn't obviously more advanced than needing Docker. "Start here if you're not sure which to pick" on the Binary entry already doesthe real work — the ordering claim could just go.

---

internal/config/config.go
```diff
@@ -957,17 +970,42 @@ func splitYAMLComment(line []byte) (active, comment []byte) {
...
+// a ' plausibly open a single-quoted scalar; a ' anywhere else is text (most
+// commonly a contraction, e.g. "John's") and must not be mistaken for a
+// string delimiter.
+func isScalarStart(line []byte, i int) bool {
+	if i == 0 {
+		return true
+	}
+	switch line[i-1] {
+	case ' ', '\t', ':', '-', '[', '{', ',':
+		return true
```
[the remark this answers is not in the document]
McCly-Solace: Fixed — `isScalarStart` now skips back over any run of whitespace before checking for a real delimiter (`:`, `-`, `[`, `{`, `,`, or line start), so a quote preceded by a word-separating space inside an already-started scalar no longer opens a string. Added `TestLoadConfig_ApostropheAfterSpaceInPlainScalarValueDoesNotLeakSubstitution` using your exact example (`password: foo 'bar # ${UNSET}`) as a permanent regression case.

internal/config/config_test.go
```diff
@@ -516,6 +516,61 @@ brokers:
...
+	cfg, err := LoadConfig(writeTemp(t, yaml))
+	if err != nil {
+		t.Fatalf("${UNSET_OLD_TOKEN} after an apostrophe-containing comment must not fail load: %v", err)
+	}
+	if cfg.brokers["prod"].Auth.Password != "secret" {
+		t.Errorf("expected password %q, got %q", "secret", cfg.brokers["prod"].Auth.Password)
+	}
+}
+
+// YAML escapes a literal ' inside a single-quoted scalar as ”. A # between
```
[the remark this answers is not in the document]
McCly-Solace: Fixed — was a typographic-quote typo (`”` instead of `''`), corrected to the actual YAML escape syntax.

internal/config/config.go
```diff
@@ -940,34 +940,88 @@ func substituteEnvVars(data []byte) ([]byte, error) {
 // the line OR preceded by whitespace, AND not inside a single- or
 // double-quoted string on the same line.
 //
+// A quote (' or ") only opens a string at a position where YAML actually
```
bczoma, line 943: 🟡 **Still asking for a tracking ticket, not a change to this PR**

Flagged by: parent (follow-up), carried from my dismissed review at `ec50f918`.

**What the problem is.** Pre-existing on `main`, not introduced here: `substituteEnvVars` operates line-by-line, so a literal newline character inside a substituted environment variable's value is inserted verbatim into the YAML text being parsed. A value like `FOO\nbrokers:\n  evil:\n...` can inject a new top-level key.

**Why it matters.** This is a real injection vector, but it requires control over an environment variable's *value* (not just its presence), which is already a strong trust boundary in most deployments — hence 🟡, not 🔴. It's not a regression from this PR and I'm not asking for a code change here.

**Recommended fix.** File a follow-up Jira ticket to track hardening `substituteEnvVars` against newline-containing values (reject, or escape before line-splitting), and link it here or in the PR description. No code change needed in this PR.

bczoma (dismissed):
## PR Review: #420 — SOL-153079 (follow-up at `595ee760`)

My prior review (`ec50f918`) was dismissed by GitHub when the base changed; its findings still apply, but the diff below is checked in full against the single commit pushed since (`595ee760`), rather than assumed narrowed. Parent-only pass.

---

### Verdict

Conditioned on this 1 — inline thread on the line in question:

1. 🟡 `internal/config/config.go:943` — still asking for a tracking ticket, not a code change here: a newline in a substituted env-var value can inject a top-level config key. Pre-existing on `main`, not introduced by this PR.

The minor notes below are **not** conditions of this verdict.

**Also verified:** @solace-awaheed's separate `CHANGES_REQUESTED` review (at the same prior head, `ec50f918`) raised five overlapping points — the same unbackticked placeholders, the same `"default"` JSON blocks, the double-quote symmetry gap, an overclaim that no Solace VPN is ever named `default` (software brokers do ship one), and a rewrite of the `examples.md` intro paragraph. All five are addressed at `595ee760`; the `examples.md` text now matches their suggested wording near verbatim. I can't dismiss their review, but for the record it is stale rather than outstanding.

---

### Summary

This push resolves every 🟡 from my prior review and, independently, everything solace-awaheed raised: the placeholder is now backticked everywhere (28 tools-reference lines fixed), the JSON invocation blocks no longer carry `"default"` (34+ occurrences across `tools-reference.md` and `examples.md`), the user-guide/README deployment-method tables now list all four methods and correctly scope Kubernetes out of the shared YAML+`.env` config step, and the double-quote branch of `splitYAMLComment` is now gated by `isScalarStart` for the same reason the single-quote branch was fixed in the original PR — with new direct table tests (`TestSplitYAMLComment`, `TestIsScalarStart`) pinning both scanners instead of only exercising them through `LoadConfig`. `docs/configuration.md` now documents that substitution is line-based and comment-aware, and states the safe-quoting idiom. Both open Copilot threads got code fixes and replies. CI is green (37/37) at this head.

*8 of 9 prior findings settled — 8 addressed.*

|---|---|---|---|
| 🟡 | Newline in a substituted env value can inject a config key (pre-existing) — asked for a follow-up ticket | Not addressed | `internal/config/config.go:943` — no ticket filed, no reply on the thread |

---

### Minor notes

None new since `ec50f918` — all 5 prior 🟢 notes were addressed in this push (isScalarStart after-space fix, double-quote symmetry, new `TestSplitYAMLComment`/`TestIsScalarStart` table tests, `docs/configuration.md` substitution semantics, both Copilot threads replied to with fixes).

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ — quote-scanner symmetry fixed and tested |
| Test Coverage | ✅ — direct table tests added for both scanner helpers |
| Security | ⚠️ — the pre-existing newline-injection gap is untracked |
| Architecture | ✅ |
| Documentation | ✅ — placeholder rendering, JSON consistency, and deployment-method scoping all fixed |
| PR Scope & Description | — not re-reviewed (unchanged since `ec50f918`) |

Sourced `parent (follow-up)`.

docs/tools-reference.md
McCly-Solace: Fixed — backticked all 29 `<your-vpn-name>` occurrences on the `**Example request:**` lines (the one at line 43 was already backticked and untouched). Verified via character-context grep that every remaining occurrence in this file is now either JSON-quoted (`"<your-vpn-name>"`) or backtick-wrapped, never bare.

docs/examples.md
McCly-Solace: Went further than the suggested instruction sentence — changed the JSON blocks themselves from `"msgVpnName": "default"` to `"msgVpnName": "<your-vpn-name>"` (and the same for `vpnName`) across both docs/examples.md and docs/tools-reference.md, so there's no longer a `default` left to silently copy. docs/tools-reference.md's parameter-convention note is updated to match — both the JSON and prose examples now use the same placeholder.

docs/user-guide.md
McCly-Solace: Fixed — table now lists Binary / Docker / go install / Kubernetes, matching README's order, with "Start here if you're not sure which to pick" carried over on Binary. Moved the contributor-facing "run from source" note out of the table into a pointer at README's Development Setup section, matching the distinction README draws between go install (tagged release) and Development Setup (source).

internal/config/config.go
McCly-Solace: Filed SOL-154441 to track this, linked as related to SOL-153079, and referenced in the PR description under Related Issues/PRs. No code change in this PR, per the recommendation.

internal/config/config.go
McCly-Solace: Filed SOL-154441 to track this, linked as related to SOL-153079, and referenced in the PR description under Related Issues/PRs. No code change in this PR, per the recommendation.

solace-awaheed (changes requested):
**All five points from my earlier review are fixed at `595ee76`** (placeholders backticked 29/29, `"default"` gone from the JSON blocks, the `default`-VPN overclaim scoped correctly, the `examples.md` intro, the double-quote gate). I re-checked each one rather than taking the replies on trust. Treat that review as closed.

This is new ground: the follow-up commit introduced a correctness regression in `splitYAMLComment`.

---

## Blocking: `isScalarStart` breaks lines that `main` handled correctly

`internal/config/config.go:1010`, used at `:976` and `:988`

Gating on `isScalarStart` correctly suppresses the *first* stray quote, but still opens on a *second* one that follows `:`, `-`, or `,`. `main`'s naive toggle cancelled quote pairs by accident; the new code flips parity from even to odd, so a real comment gets swallowed and a commented-out `${VAR}` is substituted.

| line | `main` | `595ee76` |
|---|---|---|
| `password: it's, 'foo # was ${OLD}` | comment found | **swallowed** |
| `username: don't use pre-'prod # was ${OLD}` | comment found | **swallowed** |
| `dev_token: it's-'temp # was ${OLD}` | comment found | **swallowed** |
| `password: a'b:'c # was ${OLD}` | comment found | **swallowed** |

There is a second, **silent** direction. A quoted value carrying an anchor or tag no longer opens a string, so a `#` inside it becomes a comment and the `${VAR}` after it stays literal:

| line | `main` | `595ee76` |
|---|---|---|
| `password: &pw "pre # ${PW}"` | `pre # live-password` | `pre # ${PW}` |
| `password: !!str "pre # ${PW}"` | `pre # live-password` | `pre # ${PW}` |

`LoadConfig` returns **no error** here. `/livez` and `/readyz` go green, the literal `${PW}` reaches the broker, and every SEMP call 401s.

**Scope, measured:** no shipped artifact is affected. Both scanners over all 35 repo YAML files and 65 fenced doc blocks give 0 divergent lines, and `LoadConfig` on `broker-config.example.yaml` and the k8s ConfigMap resolves identically on both commits. Two fuzz runs agree the change is strongly net-positive. So this is narrow, not urgent. It is still a new instance of the bug this PR fixes, untested in both directions.

**This also means Copilot's thread is not fully addressed.** It named `:`, `-`, `[`, `{`, `,` alongside the space case; only the space case was fixed. `pa-'ss`, `a:'b`, `a,'b`, `a{'b` still swallow the comment (those four are wrong on `main` too, so residuals rather than regressions).

The limitations comment at `:956-961` compounds it by saying the only surviving false-open is "a value that itself begins with a quote character, like `'twas`". None of the eight cases above begin with a quote, and `key: 'twas # x` is invalid YAML anyway, so the one documented case is the one that cannot matter.

### Suggested fix

Require `:`/`-`/`?` to be followed by whitespace or end of line (what YAML itself requires), honour `,` only inside a flow collection, and skip back over one node-property token:

```go
func isScalarStart(line []byte, i, flowDepth int) bool {
	skipWS := func(j int) int {
		for j > 0 && (line[j-1] == ' ' || line[j-1] == '\t') { j-- }
		return j
	}
	j := skipWS(i)
	if j == 0 { return true }
	k := j                                    // skip back over &anchor / *alias / !tag
	for k > 0 && line[k-1] != ' ' && line[k-1] != '\t' { k-- }
	if k < j && (line[k] == '&' || line[k] == '*' || line[k] == '!') {
		if j = skipWS(k); j == 0 { return true }
	}
	switch line[j-1] {
	case ':', '-', '?':
		return j == len(line) || line[j] == ' ' || line[j] == '\t'
	case '[', '{':
		return true
	case ',':
		return flowDepth > 0
	}
	return false
}
```

with `flowDepth` tracked in `splitYAMLComment` (increment on unquoted `[`/`{`, decrement on `]`/`}`).

I ran this: all 13 of your own `TestSplitYAMLComment` vectors still pass, `key: [a, 'b#c'] # x`, `key: {a: 'b#c'} # x` and `? "a # b ${V}"` are handled, both regression classes are fixed, and 3 of the 4 residuals go too. Honest remainder: `a{'b` / `a['b`, where a literal brace in a plain scalar is indistinguishable from a flow opener at this seam.

---

## Should fix

1. **`CHANGELOG.md:25` covers one of three parser fixes.** It names only the apostrophe case; this head also fixes a bare `"` in unquoted text (`6" pipe`) and the after-space case (`foo 'bar`). It also never contains `environment variables not set`, the only string an affected operator will grep for.

2. **The PR description contradicts the code on three points.** It says the JSON blocks "deliberately still show `default`" (reversed here, 62 occurrences now use the placeholder), claims the "simplest to most advanced" ordering (removed), and lists two tests where there are now six. `595ee76`'s own commit message is accurate, so squash-merging onto it fixes this.

3. **`docs/configuration.md:33` states two things the code does not deliver.** "Never substituted and never needs the variable to be set" is an absolute contradicted by the cases above, and `broker-config.example.yaml:32,88,160,198` ship commented-out `${VAR}` references an operator will now believe are inert. Separately, "without a full YAML parser at startup" is inaccurate: yaml.v3 fully parses the config at `config.go:637`, four lines later. Suggest wording it as best-effort and explicitly not a security boundary.

4. **Four new test rows cannot fail** (verified by mutation). `config_test.go:667` (`doubled single-quote escape then hash`) has no whitespace before the `#`, so nothing can split there, and it survives deleting the `''` branch. `:668` and `:669` put the `#` outside the quotes, so the expected split holds either way; both survive dropping `-` from the delimiter set. `TestIsScalarStart`'s `indented start of line` passes `i: 0`, which returns via the `i == 0` early path and never exercises the whitespace skip-back. Moving a ` # ` inside the quotes (`- 'a # b' # comment`) makes each discriminate. Related: the `case inDouble && line[i-1] == '\\'` branch restructured at `:970` has zero coverage; deleting it leaves `./internal/config` green.

---

## Follow-ups, not this PR

- **Highest leverage, and it sidesteps the lexer argument entirely:** after decode, walk the resolved config and fail if any string still matches `envVarPattern`, naming the field. About ten lines, and it turns every current and future classifier miss into a loud startup error, including the block-scalar gap the comment already concedes. It does not cover the false-open direction, so the patch above is still worth taking.
- **The error message is still undiagnosable**, which is what SOL-153079 opens with. `environment variables not set: BROKER_PW` names no file (`config.Load` tries three locations), no line (the index is in hand at `config.go:915` and discarded), and does not dedupe. Pre-existing, but this PR removes one cause of the misattribution and leaves the message that made it opaque.
- **No runbook entry.** `environment variables not set` appears nowhere in `README.md` or `docs/`, and `### Server Won't Start` has four bullets, none for substitution, though the Quickstart ships `${BROKER_USERNAME}`/`${BROKER_PASSWORD}` plus a separate `.env` step.

@bczoma's `config.go:943` newline-injection thread is still open and I am not re-arguing it; I did confirm all four policy keys are reachable that way, which strengthens the case for filing the ticket.

For the record, this contests three ✅ marks in the approving review (Correctness, Test Coverage, Documentation) on the grounds above, rather than overlooking them. `make check` and CI are green at this head; none of these are caught by either.

bczoma (approved):
## PR Review: #420 — SOL-153079 (follow-up, parent-only pass at `d1d46c83`)

My prior review (`595ee760`) was dismissed by GitHub when the branch moved. `git cherry` recovered the authored delta by patch-id — 2 new commits (`d6fca2a9`, `d1d46c83`), 4 files, +195/-34 — checked here in full.

---

### Verdict

**Recommendation:** ✅ Approve

No outstanding 🔴/🟡 findings from this pass. This push directly fixes the blocking regression @solace-awaheed raised in her separate `CHANGES_REQUESTED` review at `595ee760`: `isScalarStart` mis-toggled on a second same-type quote after a real delimiter (`it's, 'foo`) and on a quote preceded by a YAML anchor/alias/tag (`&pw "pre # ${PW}"`). Both directions are now fixed — `flowDepth` tracking for `,` inside flow collections, an anchor/alias/tag skip-back, and `:`/`-`/`?` now require trailing whitespace to count as delimiters — and every one of her four regression vectors plus both silent-failure vectors is now a passing table-driven test (`TestSplitYAMLComment`, `TestIsScalarStart`), backed by a new `FuzzSplitYAMLComment` target (30s local run, no crashes, no round-trip failures). `go build`, `go vet`, and `go test ./internal/config/...` are clean at this head; all 37 CI checks are green.

Her four "Should fix" items are also resolved at this head:
- `CHANGELOG.md` now names all three fixed misread classes and includes the `environment variables not set` grep string.
- The PR description no longer contradicts the code (JSON-block/ordering/test-count claims corrected).
- `docs/configuration.md` now says "best-effort" / "not a security boundary" instead of the two absolute claims she flagged.
- All four test rows she showed passing when their guarded branch is deleted now have discriminating internal `" #"` content, and `TestIsScalarStart`'s "indented start of line" case now passes a nonzero index so the whitespace skip-back is actually exercised.

The brace/bracket residual she called an "honest remainder" (`a{'b` / `a['b`) is accepted and now documented in the function's doc comment rather than silently left — consistent with how the existing residual note is written.

---

### Summary

This push closes the last item outstanding from my own review chain and, independently, the blocking correctness regression flagged after it. The `isScalarStart` fix is now verified against both reviewers' full vector sets plus a fuzz target, CI is green end to end, and the PR's own description/changelog claims match what the diff actually does.

*9 of 9 prior findings settled — 9 addressed.*

The last open item, `internal/config/config.go:943` (a pre-existing newline-injection gap in `substituteEnvVars`, not introduced by this PR), asked only for a tracking ticket. [SOL-154441](https://sol-jira.atlassian.net/browse/SOL-154441) was filed with the exact reproduction and suggested fix from that thread, linked to SOL-153079, and referenced in the PR description — verified via the Jira API, not just the PR comment claiming it.

---

### Minor notes

None new in the code added since my last review.

---

### Summary Verdict

| Category | Assessment |
|---|---|
| Correctness | ✅ — verified regression fix against both reviewers' vectors + fuzz |
| Test Coverage | ✅ — table tests for every cited case, discriminating mutation-test gaps closed |
| Security | ✅ — the one outstanding item is now tracked (SOL-154441) |
| Architecture | — not re-reviewed (unchanged since `595ee760`) |
| Documentation | ✅ — changelog, PR description, and configuration.md claims now match the code |
| PR Scope & Description | ✅ — description corrected to match the actual diff |

Sourced `parent (follow-up)`.

solace-awaheed (approved):
LGTM

PR #412, opened 2026-09-11
SOL-152423: Ship a reference OTel collector deployment and tested-backend matrix so tracing is usable on day one (Story 40)

solace-wmaryam:
## Summary

Ships the reference OTel collector deployment, a tested-backend matrix, and an automated e2e tracing test, so `OBS_TRACING_ENABLED=true` produces visible traces on the first try.

Closes SOL-152423.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Motivation and Context

The MCP server already exports spans over OTLP (Stories 25, 26, 27) but customers had no reference deployment to receive them. Flipping `OBS_TRACING_ENABLED=true` produced traces that went nowhere. This story ships the missing second half of the path.

## Changes Made

- `deploy/otel-collector/kubernetes/` — three K8s manifests (ConfigMap, Deployment, Service) for the OTel collector; kubeconform-validated in CI
- `deploy/otel-collector/docker/` — docker-compose stack (collector + Tempo default; Jaeger via `COMPOSE_PROFILES=jaeger`; Datadog via `DD_API_KEY`)
- `docs/observability.md` — "Stand up tracing in 30 minutes" walkthrough, tested-backend matrix (Tempo/Jaeger/Datadog), Prometheus OTLP prerequisites, cross-reference from Story 35 ToC
- `test/e2e-tracing/` — automated CI e2e test: stands up MCP server + OTel collector + Tempo + Solace broker, calls `get-broker-status`, asserts traces (including `semp.request` span) land in Tempo
- `.github/workflows/build-and-test.yml` — kubeconform validation step in lint job; new `e2e-tracing` job

## Testing

**Tests Performed:**
- [x] E2E tests added/updated (if applicable)
- [x] Tested locally with `go test -race ./...`
- [x] Tested with real Solace broker

**Test Coverage:**

- Tempo verified: traces visible in Tempo API, `semp.request` span confirmed (2026-09-11)
- Jaeger verified: `solace-broker-mcp` service visible, trace retrieved (2026-09-11)
- Datadog verified: traces and metrics (`mcp.*`) visible in Datadog APM and Metrics Explorer (2026-09-11)
- kubeconform passes on all three K8s manifests
- e2e-tracing test passes locally end to end

### Code Quality
- [x] Self-review completed (checked for typos, logic errors, edge cases)
- [x] No commented-out code or debug statements left in

### Security
- [x] **No credentials in code** (checked all files, commits, and logs)
[...]

PR #423, opened 2026-09-15
SOL-152424: ship a ServiceMonitor and NetworkPolicy for the metrics e…

solace-awaheed:
…ndpoint

## Summary

Ships a `NetworkPolicy` and a Prometheus Operator `ServiceMonitor` with the Kubernetes manifests, so `/metrics` is scrapeable from the monitoring namespace and not from anywhere else this policy controls. Also documents the previously undocumented `observability:` config block and adds kubeconform validation of `deploy/kubernetes/` to CI.

**Two scope caveats worth reading before the diff**, because the story title implies more than ships here:

- **"and OTLP egress is permitted" is not implemented as an active rule.** The policy is ingress-only and the OTLP egress rule ships as a commented template. An `egress:` allow-list default-denies every destination it omits (brokers, IdP, DNS), so it cannot be the flag-gated no-op AC 7 describes; and because NetworkPolicies are additive with no deny rule, an allow here could not override a customer's own default-deny anyway. The allow only does anything in *their* egress policy, which is where the template tells them to put it.
- **A NetworkPolicy is not self-enforcing.** Every cluster accepts the object; only the CNI enforces it, and on GKE, AKS, and EKS that is off unless enabled. Additivity also means another policy can re-open `:9091`. Both caveats are now stated in the manifest, the deploy README, and the docs rather than left implied.

Tracked under SOL-152424 (Story 32).

## Type of Change

- [x] Documentation update

## Motivation and Context

`/metrics` ships at GA on an unauthenticated, all-interfaces listener (`:9091`), and the epic promises that an SRE running Prometheus Operator can add a ServiceMonitor and start scraping without hand-authoring anything. Today the manifests expose no metrics port, ship no ServiceMonitor, and apply no NetworkPolicy, so the endpoint is either unreachable by Prometheus or reachable by every pod in the cluster. `docs/observability.md` already says restricting the listener is "the default posture, not optional hardening"; this PR ships that posture.

## Changes Made

**Manifests (`deploy/kubernetes/`)**
- New `networkpolicy.yaml`, applied with the directory. Ingress-only. `:9090` from any source (kubelet probes come from the node, which no selector can name). `:9091` from the `monitoring` namespace only, via `kubernetes.io/metadata.name`. Everything else to the pod is denied. Ports are numeric because named-port support varies by CNI.
- New `servicemonitor.yaml.example`. 15s interval, 10s timeout. `.example` because the CRD is absent on non-Operator clusters and one missing kind fails the whole directory apply (same precedent as `ingress.yaml.example`). The header lists the four silent-failure prerequisites and the exemplar requirements.
- `service.yaml` and `deployment.yaml` gain a named `metrics` port. `OBS_METRICS_ENABLED` is **not** flipped; it ships as a commented envng policy.
- `configmap.yaml` and `broker-config.example.yaml` carry the `observability:` block commented.
- The OTLP egress rule is a commented template, not an active rule. An egress allow-list default-denies every destination it omits (brokers, IdP, DNS), and NetworkPolicies are additive, so the allow only takes effect in the customer's own egress policy.

**Docs**
- `docs/observability.md`: new section *Scraping and securing the metrics endpoint*. Covers the two ways a correct ServiceMonitor is silently never scraped (kube-prometheus-stack `release:` label, `serviceMonitorNamespaceSelector`), the `prometheus.io/*` annotation fallback for non-Operator clusters, the no-CNI and service-mesh fallbacks, why egress is not in the policy, and that the policy's ports must move with `metrics_bind_address` and the server `port`. This is the anchor Story 36 (SOL-152098) cross-references.
- `docs/configuration.md`: new *Observability Settings* section. The `observability:` block had no reference documentation; `metrics_bind the CHANGELOG.
- `deploy/kubernetes/README.md`: new *Metrics scraping* section and three rows in the edit-before-applying table.

**CI and tests**
- `build-and-test.yml`: new *Validate Kubernetes manifests* step. kubeconform v0.6.7 in strict mode over `deploy/kubernetes/`, then the `.example` files copied under `.yaml` and validated too, with the ServiceMonitor checked against the Datree CRD catalog. The copy is required: kubeconform silently ignores a `.example` path even when passed explicitly (zero resources, exit 0).
- `TestShippedMetricsPortTracksDefault` pins `9091` across Service, Deployment, and NetworkPolicy to `defaults.DefaultMetricsBindAddress`y to the `http` containerPort, and pins the policy to ingress-only.
- CHANGELOG entry under `[Unreleased]`.

## Testing

**Test Environment:**
- Go version: 1.26.5 (go.mod directive 1.25.0)
- OS: macOS (darwin/arm64)
- Broker version: n/a, no broker interaction

**Tests Performed:**
- [x] Tested locally with `go test -race ./...` (see note below)

**Test Coverage:**
- kubeconform strict over the directory: 6 of 6 valid. Example copies with the CRD schema: 2 of 2 valid. A negative control with a misspelt CRD field fails as expected, and an empty example set now fails loudly rather than passing vacuously.
- Three manifest tests, each asserting an invariant the docs advertise rather than internal self-consistency:
  - `TestShippedPortsTrackCompiledDefaults` — the ConfigMap's `port`, both named containerPorts, both Service ports, and both policy rules against `defaults.DefaultPort` and `defaults.DefaultMetricsBindAddress`.
  - `TestShippedNetworkPolicyRestrictsMetricsIngress` — ingress-only with no egress, the pod selector actually matching the Deployment's pod labels, `:9090` open to all sources, and `:9091` restricted to a `namespaceSelector` naming `monitoring`, with `ipBlock` peers rejected.
  - `TestShippedServiceMonitorMatchesService` — the example's port name, selector labels, path, and 15s/10s timings against the Service it scrapes. This file was previously parsed by no test at all.
- **Mutation-tested, not just run.** Nine separate mutations were each confirmed to fail the build: namespace typo `monitoring`→`monitoring2`, the namespaceSelector swapped for `ipBlock: 0.0.0.0/0`, a pod selector that matches no pods, an added `egress:` section, the ConfigMap port moved off the default, a stale policy metrics port, a ServiceMonitor port-name typo, a changed scrape interval, and a ServiceMonitor selector typo. The first two passed the previous version of this test.
- The commented `OBS_METRICS_ENABLED` entry and both commented `observability:` blocks were uncommented mechanically and parsed to confirm indentation.
- Every documented `observability.*` key was checked against the struct's YAML tags in both directions.
- Doc-registry test (`TestObservabilityDoc*`) passes. All intra-doc anchors resolve.
- Note: `make check` on macOS fails only on two pre-existing memsampler tests that read `/proc/<pid>/fd`, which is Linux-only. They pass on the Ubuntu runner and are unrelated to this change. Race-enabled tests for `internal/config` and `cmd/server` pass (641 tests).
- **Not done: verified scrape on a real Operator cluster (AC 1).** Needs a cluster whose CNI enforces NetworkPolicy (kind + Calico or k3d + Cilium; kindnet does not enforce) plus kube-prometheus-stack.

### Code Quality
- [x] Self-review completed
- [x] No commented-out code or debug statements left in (the commented blocks in the manifests are deliberate reference config, matching the existing oauth block in `configmap.yaml`)

### Security
- [x] No credentials in code
- [x] Followed secure logging rules (`/check-logs` run on the changed Go file; it contains no log calls)
- [x] Credential-carrying types implement `slog.LogValuer` (n/a, no new types)

### Testing
- [x] All tests pass locally: `go test -race ./...` (except the two pre-existing macOS-only tests noted above)
- [x] Error paths tested

### Documentation
- [x] README.md updated (`deploy/kubernetes/README.md`)
- [x] docs/ updated
- [x] godoc comments added/updated
- [x] CHANGELOG.md updated
- [x] Examples updated (`broker-config.example.yaml`, `configmap.yaml`)

### Git & CI
- [ ] Branch is up to date with main (one Dependabot commit behind, touching `release.yml` and `THIRD_PARTY_BUILD_TEST.md`; verified it merges cleanly)
- [x] No merge conflicts
- [ ] CI checks passing (pending PR open)

## Additional Notes

- **Deviations from the story as written**, each with the reason in the Jira comment on SOL-152424: whole-pod rather than per-port denial (that is how NetworkPolicy works), the egress rule as a commented template (an active one cannot be a flag-gated no-op), `.example` rather than `servicemonitor.yaml`, kubeconform rather than `kubectl --dry-run=client`, no Kustomize overlay (repo precedent), and the runbook AC belonging to Story 36.
- **PR #412 (Story 40) has since merged**, so its kubeconform step for `deploy/otel-collector/` is on main in the required `build` job. This branch still merges cleanly, because addressing review condition 1 moved my step out of `build` altogether. That leaves two `go install kubeconform` runs per PR, which I have deliberately left alone: their pass uses only built-in schemas and has no catalog dependency, so folding it into this advisory job would weaken a working gate to fix nothing. Consolidating is a maintainer's call, not something this PR should do to a just-merged step.
- **The manifest check is now advisory** (`Kubernetes manifests valid` in `ci-pr.yaml`). Registering it as a required context is a repo-admin ruleset action per `.github/ADMIN_SETUP.md`, the same as the other jobs in that file.
- **Operator still edits two things** before scraping works: uncomment `OBS_METRICS_ENABLED`, and set the `release:` label to their Helm release. Both are documented.

## Related Issues/PRs

- Related to #412 (kubeconform step overlap)
- Related to SOL-152098 (Story 36 runbook cross-references the new docs anchor)

---

**For Reviewers:**

**Focus Areas**: The ingress-only decision and the `:9090`-from-anywhere rule in `networkpolicy.yaml`; the `.example` choice for the ServiceMonitor; the two-pass kubeconform step and why the copy is needed.


.github/workflows/build-and-test.yml
```diff
@@ -220,6 +220,33 @@ jobs:
             echo "✅ Test files found. You can now remove this step"
           fi
 
+      - name: Validate Kubernetes manifests
```
bczoma: 🟡 **The new Kubernetes-manifest validation step runs on every PR, not just ones that touch manifests**

**Flagged by:** good:architecture

**What the problem is:** "Validate Kubernetes manifests" is unconditional inside `build`, one of the required status-check jobs, with no path filter scoping it to `deploy/kubernetes/**`.

**Why it matters:** this compounds the reproducibility issue already flagged on line 247 (the Datree CRD catalog tracking `main`). If that remote lookup ever hiccups (rate limit, catalog rename, transient outage), the required `build` check fails for *every* open PR — including one that only fixes a docs typo — not just PRs that actually touch the manifests. A monitoring-config validation dependency now sits on the critical path for unrelated changes.

**Recommended fix:** gate the step on whether `deploy/kubernetes/**` changed (e.g. `dorny/paths-filter`, or a plain `git diff --name-only` check), or move it to its own job outside the required set, so an external catalog problem can't block unrelated work.

.github/workflows/build-and-test.yml
```diff
@@ -220,6 +220,33 @@ jobs:
...
+        run: |
+          go install github.com/yannh/kubeconform/cmd/kubeconform@v0.6.7
+          kubeconform -strict -summary deploy/kubernetes/
+          examples=$(mktemp -d)
+          for f in deploy/kubernetes/*.yaml.example; do
+            cp "$f" "$examples/$(basename "$f" .example)"
+          done
+          kubeconform -strict -summary \
+            -schema-location default \
+            -schema-location 'https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json' \
```
bczoma: 🟡 **The Datree CRD schema fetch has no integrity pinning — builds on Copilot's comment on this line**

**Flagged by:** good:security (extends the reproducibility point Copilot already raised here)

**What the problem is:** beyond "this tracks the mutable `main` branch," the fetch has no commit SHA or checksum verification at all — unlike the kubeconform *binary* two lines above, pinned to an immutable `@v0.6.7` release tag. This is a live, unauthenticated, unverified third-party input with no tamper detection.

**Why it matters (measured, not asserted):** the blast radius is narrower than it first looks. This `-schema-location` path only serves CRD schemas, and is only reached for `servicemonitor.yaml.example` (`kind: ServiceMonitor`, a `monitoring.coreos.com` CRD). The security-bearing manifest, `networkpolicy.yaml`, validates against `-schema-location default` — kubeconform's own bundled schemas, pinned by the `@v0.6.7` Go module version and untouched by the Datree catalog. So a compromised or altered catalog entry could at most let a malformed *ServiceMonitor* (a scrape-config convenience object) slip past CI; it cannot cause CI to wave through a broken NetworkPolicy.

**Recommended fix:** pin the schema fetch to a specific commit SHA of CRDs-catalog (not `main`), and note in the workflow comment that this dependency's blast radius is scoped to ServiceMonitor validation only — so a future reader doesn't over- or under-estimate what a catalog compromise could affect.

docs/observability.md
```diff
@@ -2031,6 +2040,116 @@ roughly the non-heap overhead the 25% reserve exists to pay for — so that cros
...
+
+**Under a service mesh.** A sidecar that owns the pod's inbound ports — Istio in strict mTLS
+mode, for example — refuses a plaintext scrape of `:9091` from a Prometheus outside the mesh.
+Either exempt the port from mTLS (a `PeerAuthentication` with `portLevelMtls` for `9091`, or
+`traffic.sidecar.istio.io/excludeInboundPorts: "9091"` on the pod template) so the
+NetworkPolicy stays the control, or run a mesh-aware Prometheus that presents a workload
+certificate. Istio's own `prometheus.io/*` annotation rewriting applies only to the annotation
+path, not to a ServiceMonitor.
+
+**When your CNI does not enforce NetworkPolicy.** Calico, Cilium, and the GKE, AKS, and EKS
```
bczoma: 🟡 **CNI-enforcement list omits OpenShift — different cluster type from Copilot's comment nearby on this section**

**Flagged by:** good:security

**What the problem is:** the enforcement list names GKE, AKS, EKS, Calico, Cilium, and kind, but omits OpenShift entirely — a realistic target for the regulated (bank/airline) customers this project's own industry framing targets. OpenShift's behavior varies by version/SDN: OVN-Kubernetes (its default since 4.4+) enforces NetworkPolicy out of the box, but the legacy OpenShiftSDN plugin — still present on clusters upgraded from pre-4.4 rather than freshly installed — only enforces NetworkPolicy when explicitly switched from its default `subnet` isolation mode to `NetworkPolicy` mode.

**Why it matters:** an operator on such a cluster finds no OpenShift guidance here and could reasonably (and wrongly) assume the GKE/AKS/EKS treatment implies "any managed platform is fine," when their SDN mode may be silently non-enforcing.

**Recommended fix:** add an OpenShift row naming OVN-Kubernetes (default, enforces) vs. legacy OpenShiftSDN in `subnet` mode (does not enforce, must be switched to `NetworkPolicy` mode), pointing at the same `curl`-based verification already recommended for kind.

internal/config/shipped_examples_test.go
```diff
@@ -266,6 +269,168 @@ func TestShippedGOMEMLIMITTracksMemoryLimit(t *testing.T) {
...
+		t.Fatal("deploy/kubernetes/deployment.yaml declares no containerPort named `metrics`; " +
+			"service.yaml's `metrics` port targets it by that name")
+	}
+	if depPort != wantPort {
+		t.Errorf("deploy/kubernetes/deployment.yaml `metrics` containerPort is %d; the listener's "+
+			"compiled default is %q", depPort, defaults.DefaultMetricsBindAddress)
+	}
+	// The MCP port the probes and the Service's `http` port use. The policy must
+	// allow it too, or isolating the pod for ingress takes the whole endpoint down.
+	httpPort, ok := containerPorts["http"]
```
bczoma: 🟡 **The `http`/`9090` port is checked for internal manifest self-consistency only, never against the compiled default**

**Flagged by:** good:architecture + good:code (independently)

**What the problem is:** `httpPort` here is read straight out of `deployment.yaml`'s own `http` containerPort and then used only to cross-check `networkpolicy.yaml`'s `:9090` rule against *itself* — a self-referential check. It is never compared to `defaults.DefaultPort`, the way the metrics port two lines above correctly is compared to `defaults.DefaultMetricsBindAddress`. `configmap.yaml`'s `port` field (the value `deploy/kubernetes/README.md` explicitly instructs operators to edit) is never read by this test either.

**Why it matters — measured under challenge:** `deploy/kubernetes/README.md` documents editing `configmap.yaml`'s `port` as the expected customization path. Today `9090` matches `defaults.DefaultPort` everywhere, so nothing is broken now — but the gap bites on either a future change to `DefaultPort` or an operator following the README's own instructions. If `deployment.yaml`'s containerPort goes stale, Kubernetes' own readiness probe (`port: http`) catches it loudly (CrashLoop/NotReady) independent of this test. But if only `networkpolicy.yaml`'s hardcoded `9090` rule goes stale, there is no such safety net: kubelet probes aren't policed by the CNI on the node path, so the pod reports Ready while real client traffic to the true port is silently denied — exactly the "scrape silently stops" failure mode this test explicitly guards against for the metrics port, left uncaught for the higher-blast-radius `http`/MCP port.

**Recommended fix:** add `if httpPort != defaults.DefaultPort { t.Errorf(...) }`, mirroring the `wantPort` check already done for the metrics port.

internal/config/shipped_examples_test.go
```diff
@@ -266,6 +269,168 @@ func TestShippedGOMEMLIMITTracksMemoryLimit(t *testing.T) {
...
+	}
+	httpRule, metricsRule := false, false
+	for _, rule := range np.Spec.Ingress {
+		for _, p := range rule.Ports {
+			switch p.Port {
+			case httpPort:
+				httpRule = true
+			case wantPort:
+				metricsRule = true
+				if len(rule.From) == 0 {
```
bczoma: 🟡 **The metrics-port ingress rule assertion doesn't verify the namespace restriction it's documented to enforce — builds on Copilot's comment on this same test**

**Flagged by:** good:code (extends the gap Copilot already flagged nearby)

**What the problem is:** Copilot noted this assertion only checks `len(rule.From) != 0`. Reading the decoding a level deeper: the `np` struct decodes `From []any`, discarding the namespace-selector's match value entirely before it ever reaches this check. Even fixing Copilot's literal complaint (checking non-emptiness more carefully) can't verify the actual invariant the PR and `docs/observability.md` advertise — "deny all except the `monitoring` namespace" — because the selector's content never reaches typed Go data at all.

**Why it matters:** a `from: [{ipBlock: {cidr: 0.0.0.0/0}}]` or a typo'd label value (`monitoring2`, wrong key) would pass this test unchanged, even though it defeats the policy's whole stated purpose.

**Recommended fix:** decode `From []struct{ NamespaceSelector *struct{ MatchLabels map[string]string } }` (or similar) and assert `rule.From[i].NamespaceSelector.MatchLabels["kubernetes.io/metadata.name"] == "monitoring"` for the metrics-port rule, not just that a `from` list exists.

bczoma (dismissed):
## PR Review: #423 — SOL-152424: ship a ServiceMonitor and NetworkPolicy for the metrics endpoint

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-152424` → `main` | **Changes:** +567 / -11 across 12 files
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-423`
**Jira:** `SOL-152424` — Ship a ServiceMonitor and NetworkPolicy so Prometheus discovers the endpoint and OTLP egress is permitted (Story 32) (Verifying)

---

### Verdict

1. 🟡 `.github/workflows/build-and-test.yml:223` — scope "Validate Kubernetes manifests" to `deploy/kubernetes/**` changes (or move it off the required `build` job) so a hiccup in the external CRD catalog can't fail unrelated PRs
2. 🟡 `.github/workflows/build-and-test.yml:247` — pin the Datree CRD-catalog schema fetch to a commit SHA instead of `main`
3. 🟡 `docs/observability.md:2143` — the CNI-enforcement list omits OpenShift, whose legacy OpenShiftSDN plugin does not enforce NetworkPolicy by default
4. 🟡 `internal/config/shipped_examples_test.go:333` — the `http`/`9090` port is only checked for internal manifest self-consistency, never against `defaults.DefaultPort`
5. 🟡 `internal/config/shipped_examples_test.go:400` — the metrics-port ingress rule only asserts `from` is non-empty, not that it actually restricts to the `monitoring` namespace

- 🟡 PR body — the Jira summary frames this story as "so Prometheus discovers the endpoint **and OTLP egress is permitted**," but the egress rule ships only as a commented template; the PR description itself doesn't carry that caveat, only a Jira comment does
- 🟡 `docs/internal/packaging-release.md` — the "In outline the manifests are" inventory under `### Kubernetes` isn't updated with the two new files (`networkpolicy.yaml`, `servicemonitor.yaml.example`)

---

### Summary

This PR ships the NetworkPolicy and ServiceMonitor that SOL-152424 calls for, with unusually thorough disclosure of its own gaps and deviations — the author's Jira comment and PR body both name every place the implementation diverges from the acceptance criteria and why. A 🔴 finding was filed and reduced under challenge: the story's admittedly-unverified on-cluster enforcement test (AC1) turned out to carry low real-world risk, because the metrics listener ships off by default, the docs give an explicit self-verification procedure, and a loopback-bind fallback exists — so it's now a 🟢 process note rather than a blocker. What remains are real but modest gaps: two test-coverage holes that undercut the new regression guard's own stated purpose, a CI step that can fail unrelated PRs, and a couple of documentation completeness items. The manifests themselves — fail-safe defaults, ingress-only isolation, reversible and additive — are sound.

n/a — first review of this PR.

---

### Minor notes

- 🟢 Jira `SOL-152424` (AC1) — on-cluster verification that the NetworkPolicy is actually enforced was never done, on a story marked "Verifying" and promoted to confirmed-GA. Real risk is low today (metrics ship opt-in only, the docs give a `curl`-based self-check and a loopback-bind fallback), but don't let the Jira status move to "Done" without either that verification run or an explicit tracked follow-up.
- 🟢 `deploy/kubernetes/servicemonitor.yaml.example` — never parsed by any test. Add a `readManifest` case asserting `endpoints[0].port == "metrics"` and `interval`/`scrapeTimeout` are `15s`/`10s`, mirroring the pattern already used for the other three manifests.
- 🟢 `deploy/kubernetes/networkpolicy.yaml:30-45` — the commented OTLP egress template is inert to kubeconform (nothing validates YAML inside a `#` block) and can rot silently until an operator copy-pastes it. Consider a separate `.example` file run through the same kubeconform pass, the way `servicemonitor.yaml.example` already is.
- 🟢 `internal/config/shipped_examples_test.go` — no assertion that `networkpolicy.yaml`'s `podSelector.matchLabels` matches `deployment.yaml`'s pod template labels. A drift there fails the isolation open (selects zero pods) with nothing to catch it — a larger blast radius than the namespace-selector gap above, currently untested.
- 🟢 `.github/workflows/build-and-test.yml:242` — the `.example`-copy loop has no `nullglob`; harmless today (one `.example` file exists) but would fail obscurely if it were ever the last one removed.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

.github/workflows/build-and-test.yml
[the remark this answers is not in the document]
solace-awaheed: Pinned in 52ff69f to CRDs-catalog commit `ad3b08c5045129d7bb1eeffd8e61719b2c8dd1e2` (2026-09-08), carried in a `CRD_CATALOG_SHA` env var with a comment saying to bump it deliberately.

I went with pinning over vendoring the schema: a 67KB third-party JSON committed in-tree becomes redistributed content and would pull in a THIRD_PARTY_LICENSES entry, whereas a tool we merely run stays outside that inventory. Pinning gets the reproducibility without that.

internal/config/shipped_examples_test.go
```diff
@@ -266,6 +269,168 @@ func TestShippedGOMEMLIMITTracksMemoryLimit(t *testing.T) {
...
+					t.Errorf("deploy/kubernetes/networkpolicy.yaml admits port %d with no `from`, i.e. from "+
+						"every source; restricting that port to the monitoring namespace is the policy's "+
+						"whole purpose", wantPort)
+				}
```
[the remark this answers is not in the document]
solace-awaheed: Fixed in 52ff69f. The `from` list is now decoded as typed peers rather than `[]any`, and the metrics rule asserts every peer is a `namespaceSelector` carrying `kubernetes.io/metadata.name: monitoring`, with an explicit rejection of any `ipBlock` peer.

Verified by mutation rather than by inspection — a typo'd namespace (`monitoring2`) and a `from: [{ipBlock: {cidr: 0.0.0.0/0}}]` swap both fail the build now, and both passed before.

deploy/kubernetes/README.md
```diff
@@ -119,6 +125,32 @@ Ingress needs a hostname and TLS secret only you can supply.
...
+
+1. **Turn the listener on.** Uncomment `OBS_METRICS_ENABLED` in
+   `deployment.yaml`. Until then the port resolves to `connection refused`.
+2. **Restrict who can reach it.** `networkpolicy.yaml` applies with the
+   directory and admits `:9091` only from the `monitoring` namespace. `:9090`
+   stays open to every source, because the kubelet's probes arrive from the
+   node and no selector can name them. It restricts ingress only; broker, IdP,
+   and OTLP egress are untouched. It needs a CNI that enforces NetworkPolicy
+   (Calico, Cilium, GKE/AKS/EKS native) — kind's default CNI accepts it and
+   enforces nothing, so verify with a `curl` from a pod outside `monitoring`.
```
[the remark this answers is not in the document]
solace-awaheed: Fixed in 52ff69f. The README step now leads with the two reasons an applied policy may protect nothing: enforcement is a cluster-level CNI feature normally chosen at creation time on GKE, AKS, and EKS, and policies are additive so another policy can re-open the port. It points at the per-platform table in `docs/observability.md` and tells the operator to verify with a `curl` from outside the namespace plus `kubectl get networkpolicy -A`.

deploy/kubernetes/networkpolicy.yaml
```diff
@@ -0,0 +1,68 @@
+# Who may reach the pod (SOL-152424). Ingress only: policyTypes lists Ingress
+# alone, so egress — broker SEMP, IdP, DNS, OTLP push — is untouched. Read the
+# note at the bottom before adding an `egress:` section.
+#
+# Enforced only by a CNI that supports NetworkPolicy. Calico, Cilium, and the
+# GKE, AKS, and EKS native CNIs do; kind's default kindnet accepts the object
+# and enforces nothing. Verify with a curl to a pod IP on :9091 from a pod
+# outside the allowed namespace, not with `kubectl get networkpolicy`. Without
+# an enforcing CNI, restrict :9091 another way — docs/observability.md
+# § "Scraping and securing the metrics endpoint".
```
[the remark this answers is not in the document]
solace-awaheed: Fixed in 52ff69f. You were right that the claim was wrong, not just imprecise — I had written that the GKE, AKS, and EKS native CNIs enforce it, and on all three the feature is off unless enabled, usually at cluster creation.

The manifest header is now a short two-item list (only the CNI enforces; policies are additive) ending in the `curl` check, with the per-platform detail in the docs table rather than duplicated here.

docs/observability.md
```diff
@@ -2031,6 +2040,116 @@ roughly the non-heap overhead the 25% reserve exists to pay for — so that cros
...
+native CNIs enforce it; kind's default `kindnet` and several others accept the object and
+enforce nothing, and `kubectl get networkpolicy` looks identical either way. Verify with a
```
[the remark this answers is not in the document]
solace-awaheed: Fixed in 52ff69f. Replaced the prose list with a per-platform table stating for each whether it enforces and what has to be enabled, preceded by the flat statement that an accepted policy object is not an enforced boundary. The fallback advice now sits behind a 'verify rather than infer, whatever the table says' instruction, so a reader who skims the table still lands on the check.

.github/workflows/build-and-test.yml
solace-awaheed: Taken, via your second option. In 52ff69f the step is out of `build` entirely and is now a separate `Kubernetes manifests valid` job in `ci-pr.yaml`, which is where this repo already keeps its advisory checks and where the admin-registration note applies.

I chose that over a path filter for two reasons: `dorny/paths-filter` would add a new pinned action plus a THIRD_PARTY_BUILD_TEST row for a conditional, and a `git diff` filter needs `fetch-depth: 0` in a job that deliberately runs shallow. Moving the job also fixes the case a path filter would not — the catalog can be down while a manifest genuinely changed, and now that fails only the advisory check.

Combined with the SHA pin on the other thread, the remaining failure mode is a transient outage on a manifest-touching PR, reported against the change that could have caused it and blocking nothing else.

.github/workflows/build-and-test.yml
solace-awaheed: Pinned in 52ff69f to `ad3b08c5045129d7bb1eeffd8e61719b2c8dd1e2` (2026-09-08), in a `CRD_CATALOG_SHA` env var so the bump is a visible one-line diff.

Your blast-radius analysis is now in the workflow comment more or less as you wrote it: only the second pass consults the catalog, only for `kind: ServiceMonitor`, and `networkpolicy.yaml` validates against kubeconform's bundled schemas pinned by the `@v0.6.7` module version — so a bad catalog entry can at most let a malformed ServiceMonitor through, never a broken NetworkPolicy. Worth having written down, since the honest scope is narrower than 'unverified third-party input in CI' sounds.

I considered vendoring the 67KB schema to remove the dependency outright, and rejected it: in-tree third-party content is redistributed and would need a licensing-inventory entry, while a tool we only run does not.

docs/observability.md
solace-awaheed: Added in 52ff69f, but I checked the mechanism and it does not hold up, so the row reads differently from your suggestion.

Two corrections. OpenShiftSDN's default isolation mode on OpenShift 4 is `networkpolicy`, not `subnet` — `subnet` is the opt-in flat-network mode. And OpenShiftSDN was deprecated in 4.14 and removed in 4.17, so a cluster carrying it cannot reach a currently supported release; upgrading past 4.17 requires migrating to OVN-Kubernetes first.

So OpenShift is a positive case rather than a trap, and the table now says so: OVN-Kubernetes is the default and enforces, the legacy plugin defaulted to network-policy mode and is gone as of 4.17, so any supported release enforces. That still answers your underlying point — the regulated-customer audience now finds OpenShift named explicitly instead of inferring from the GKE/AKS/EKS rows — and it lands on the `curl` check like every other row.

internal/config/shipped_examples_test.go
solace-awaheed: Fixed in 52ff69f, and the self-referential check is gone. `TestShippedPortsTrackCompiledDefaults` now checks the http port against `defaults.DefaultPort` and also parses `configmap.yaml`'s embedded `config.yaml` to assert its `port` tracks the same default — so the value the README tells operators to edit is read by a test for the first time.

Your asymmetry argument is the test's doc comment now, because it is the reason the policy is checked against the compiled default rather than against the other manifests: a stale containerPort is caught loudly by the probes, while a stale policy port is not caught at all, since kubelet probes arrive on the node path the CNI does not police. Pod reports Ready, real traffic denied.

internal/config/shipped_examples_test.go
solace-awaheed: Fixed in 52ff69f, and you were right that the fix had to go deeper than the assertion. The `From []any` decode was the actual defect — the selector's content never reached typed Go data, so no assertion at that level could have worked.

`from` now decodes into typed peers covering all three peer kinds (namespaceSelector, podSelector, ipBlock), deliberately including the two we do not ship so a swap is visible rather than indistinguishable. The metrics rule asserts every peer is a namespaceSelector with `kubernetes.io/metadata.name: monitoring` and rejects any ipBlock.

Both of your examples were used as mutation tests: `monitoring2` and a `0.0.0.0/0` ipBlock each fail the build now.

solace-awaheed:
## Review addressed in 52ff69f

All 7 conditions from @bczoma's review and all 5 Copilot threads are addressed, each with a reply on its own thread. Two I pushed back on with evidence rather than implementing as written — details below.

### Conditions

| # | Condition | Resolution |
|---|---|---|
| 1 | Scope the manifest step off the required job | Moved out of `build` into a separate advisory `Kubernetes manifests valid` job in `ci-pr.yaml` |
| 2 | Pin the Datree CRD fetch | Pinned to catalog commit `ad3b08c` in a `CRD_CATALOG_SHA` env var |
| 3 | OpenShift missing from the enforcement list | Added, **with the mechanism corrected** — see below |
| 4 | `http`/9090 never checked against `defaults.DefaultPort` | Now checked against it, plus `configmap.yaml`'s embedded `port` |
| 5 | Metrics rule only asserts `from` non-empty | `from` decodes to typed peers; asserts the `monitoring` namespaceSelector and rejects `ipBlock` |
| 6 | PR body lacks the egress caveat | Added to the Summary, which also repaired three sentences that had been truncated mid-word in the original body |
| 7 | `packaging-release.md` inventory | Both new manifests added |

### Two pushbacks

**Condition 3, the OpenShift mechanism.** The suggested row does not hold up on checking. OpenShiftSDN's default isolation mode on OpenShift 4 is `networkpolicy`, not `subnet` — `subnet` is the opt-in flat-network mode — and OpenShiftSDN was deprecated in 4.14 and **removed in 4.17**, so a cluster still running it cannot reach a supported release without migrating to OVN-Kubernetes first. OpenShift is therefore a positive case, not a trap, and the table says so. Your underlying point still lands: OpenShift is now named explicitly instead of being inferred from the GKE/AKS/EKS rows.

**Minor note on a separate egress `.example` file: declining, with reasoning.** The silent-rot concern is real, but an applyable egress policy is a worse failure mode than a stale comment. A complete egress manifest carries `policyTypes: [Egress]`, so anyone who applies it before editing default-denies their own brokers, IdP, and DNS — an outage caused by reading our example too fast. The rules are also irreducibly customer-specific (their broker and IdP addresses), so the file could only ever be a template, and a template that is dangerous when applied is worse than one that cannot be. `TestShippedNetworkPolicyRestrictsMetricsIngress` now pins the policy to ingress-only, so the comment cannot quietly become an active rule either.

### Other minor notes

Done: `servicemonitor.yaml.example` is now parsed and tied to the Service (it had no test at all); the policy's `podSelector` is asserted against the Deployment's pod labels, which is the drift that fails the isolation *open*; `nullglob` added, plus a guard that fails loudly if the example pass ever validates nothing.

Still open: **AC 1, on-cluster enforcement verification.** Unchanged and untestable from here — it needs a cluster whose CNI actually enforces (kind + Calico or k3d + Cilium). Per your note, this should not go to Done without either that run or a tracked follow-up; it is listed as outstanding on the Jira.

### A correction of my own

The claim that the GKE, AKS, and EKS *native* CNIs enforce NetworkPolicy was simply wrong, not merely imprecise — on all three the feature is off unless enabled, normally at cluster creation. Copilot flagged it on three separate lines and was right each time. It is replaced by a per-platform table and the flat statement that an accepted policy object is not an enforced boundary, with a `curl`-from-outside check as the ground truth. Since the additive-union point has the same shape (an operator believing the port is closed when it is not), that caveat is now stated alongside it in all three places.

CI was green on the previous commit; re-running now.

bczoma (approved):
## PR Review: #423 — SOL-152424: ship a ServiceMonitor and NetworkPolicy for the metrics endpoint

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-152424` → `main` | **Changes:** +929 / -11 across 13 files
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-423`
**Jira:** `SOL-152424` — Ship a ServiceMonitor and NetworkPolicy so Prometheus discovers the endpoint and OTLP egress is permitted (Story 32)

---

### Verdict

**Recommendation:** ✅ Approve

Follow-up review: branch rebased since `68e376da`; 1 new commit (`52ff69f`) recovered by patch-id (8 files, +522/-160). All 7 conditions from the prior review and 5 of Copilot's 6 threads have been addressed and verified in code at the current head — none held back at `Not addressed`. No new findings in the pushed delta.

---

### Summary

This push closes out the prior review cleanly: the Kubernetes-manifest validation step moved out of the required `build` job into its own advisory `Kubernetes manifests valid` job in `ci-pr.yaml` (verified: `.github/workflows/ci-pr.yaml` new `manifests` job, `.github/workflows/build-and-test.yml` step removed), the Datree CRD-catalog fetch is now pinned to a commit SHA rather than `main`, and both test-coverage gaps are closed with real assertions — `TestShippedPortsTrackCompiledDefaults` now checks the shipped ports against `defaults.DefaultPort`/`defaults.DefaultMetricsBindAddress`, and the ingress-rule test decodes typed peers and rejects any `ipBlock`/non-`monitoring`-namespace peer instead of just checking `from` is non-empty. All CI checks are green, including the new advisory manifest-validation job.

*12 prior findings settled — 12 addressed, 0 accepted as-is, 0 declined.*

No table — nothing needs an exception row; every conditioned 🟡, every not-anchorable 🟡, and every 🟢 minor note was verified fixed in code at the current head (`52ff69f`), not just claimed in a commit message:

- CI step scoped out of the required `build` job into its own advisory job (`.github/workflows/ci-pr.yaml`)
- CRD catalog pinned to a commit SHA (`CRD_CATALOG_SHA` env var)
- OpenShift added to the CNI-enforcement docs, with a corrected mechanism (OpenShiftSDN's default mode is `networkpolicy`, not `subnet`)
- Shipped `http`/`9090` port now checked against `defaults.DefaultPort`
- Metrics-ingress rule now asserts a typed `namespaceSelector` naming `monitoring` and rejects any `ipBlock` peer
- PR body now states the OTLP-egress scope caveat explicitly
- `docs/internal/packaging-release.md` manifest inventory updated with `networkpolicy.yaml` and `servicemonitor.yaml.example`
- `servicemonitor.yaml.example` now parsed and asserted against the Service (port name, labels, path, 15s/10s timings)
- `podSelector.matchLabels` now checked against the Deployment's pod template labels
- `.example`-copy loops in both workflow files now use `nullglob`

---

### Minor notes

None outstanding — all closed as noted above.

---

### Summary Verdict

|---|---|---|
| Architecture | — not re-reviewed (unchanged since `68e376da`) | — |
| Security | ✅ | parent (follow-up) |

PR #424, opened 2026-09-15
SOL-152092: Ship an importable reference Grafana dashboard (Story 37)

solace-lootii:
## Summary

Ships `deploy/grafana/solace-broker-mcp-overview.json`, an importable Grafana 9+ reference dashboard covering tool RED, active requests, SEMP RED-per-attempt, an auth-failure rate breakdown, an `error_type` breakdown, Go runtime, and build info — with trace exemplars wired on the latency panels. Closes SOL-152092 (Story 37).

## Type of Change

- [x] Documentation update

## Motivation and Context

Operators currently spend hours building a dashboard before they can observe this service at all. This ships a tested, importable JSON so a NOC operator gets tool/SEMP RED, an auth-failure view, and an error breakdown on day one with zero panel authoring, and so everyone monitors against the same documented metric names instead of inventing their own.

## Changes Made

- `deploy/grafana/solace-broker-mcp-overview.json` — the dashboard, authored and manually verified against a real Grafana 9.5.20.
- `deploy/grafana/README.md` — quick-start import pointer, matching the existing `deploy/kubernetes/README.md` convention.
- `cmd/server/grafana_dashboard_test.go` — new CI test that parses Story 14's golden file and asserts every panel's metric name, label key, and template-variable query actually exists in it, so the dashboard and the real schema can't silently drift apart.
- `docs/observability.md` — new "Grafana Dashboard" section: import steps, panel table, the two variable-sourcing mechanisms, exemplar prerequisites, and what's required (and what differs) on each metrics-ingestion path.
- Two corrections found while verifying against the OTLP-ingestion path live, both documented rather than silently worked around:
  - The `$service_name` dashboard variable has to be sourced from `target_info`'s `job` label, not `service_name` — on the OTLP-ingestion path, Prometheus folds `service.name` into `job` on `target_info` specifically, so a `service_name`-keyed query returns an empty dropdown there even though it works fine on scrape.
  - This server's OTLP metrics exporter is gRPC-only, while Prometheus's native OTLP receiver only speaks OTLP/HTTP — pointing this server directly at a bare Prometheus does not work (confirmed live: `mcp_otel_metrics_dropped_total{reason="export_timeout"}`). A collector in front is required, not optional. I also found and fixed the same gap in #412's already-merged "Ingesting OTLP metrics directly into Prometheus (no collector)" section — renamed it to "(collector required)" and added the missing prerequisite, since as written it would send an operator through four config changes into the same dead end.
- `CHANGELOG.md` entry under `[Unreleased]`.

## Testing

**Test Environment:**
- Go version: 1.25.0
- OS: Linux
- Broker version: PubSub+ Enterprise (dev)

**Tests Performed:**

**Test Coverage:**
- Manual import verified against a real, fresh Grafana 9.5.20, twice: once against a Prometheus scraping `/metrics`, once against a Prometheus ingesting this server's OTLP push through a collector bridge. All panels confirmed rendering real data from real traffic against a live broker on both paths, except the Go runtime panels on the OTLP path, which are expected to show no data (documented, structural — `go_*`/`process_*` never reach the OTLP egress).
- Exemplars verified end to end: a real exemplar's `trace_id` from the metrics scrape matched a real, fetchable trace in a Tempo instance, with the documented span tree. Also verified the negative case — Prometheus without `--enable-feature=exemplar-storage` silently drops exemplars even with OpenMetrics negotiation working, which I've called out in the docs since it's easy to miss.
- New CI test sanity-checked both directions: confirmed it fails on an injected bad metric name and on an injected bad template-variable query, then confirmed it passes on the real committed file.
- Full repo `go build ./...`, `go vet ./...`, `go test ./...` all green.

### Code Quality

### Security

### Testing

### Documentation

### Git & CI

## Additional Notes

The dashboard deliberately ships with no panels for retry outcomes, broker pool gauges, or saturation events (Stories 17, 18, 28, 29) — those metrics don't exist yet, so a panel against them would import broken. They land when those stories ship.

## Related Issues/PRs

- Closes SOL-152092 (Story 37)
- Depends on Stories 14, 15, 16, 19, 24, 34, 47 — all already merged
- Corrects a doc gap in #412 (Story 40, SOL-152423)


docs/observability.md
```diff
@@ -1463,20 +1573,37 @@ For each backend, the collector `config.yaml` exporter block is in
...
+   receiver (`--web.enable-otlp-receiver`, item 2 below) only speaks OTLP/HTTP, and never opens a
+   gRPC listener. Pointing this server directly at a bare Prometheus therefore **does not work**
+   — verified live: every export attempt fails with
+   `mcp_otel_metrics_dropped_total{reason="export_timeout"}`, because the gRPC client's
+   connection to an HTTP/1.1-only endpoint never completes. Route through an OTel Collector
+   instead: an `otlp` gRPC receiver (what this server can talk to) and an `otlphttp` exporter
+   pointed at `http(s)://<prometheus>/api/v1/otlp` (what Prometheus can talk to) is sufficient.
+   The reference collector under [Stand up tracing in 30 minutes](#stand-up-tracing-in-30-minutes)
+   ships this exact metrics pipeline commented out by default (Tempo/Jaeger are traces-only) —
+   enable it and point its exporter at your Prometheus's OTLP endpoint.
```
bczoma: 🟡 **Docs instruct enabling a checked-in `otlphttp` exporter that doesn't exist in either reference collector config**

**Flagged by:** docs, good:pr (independently, both verified against the actual files)

**What the problem is:** This section (and the mirrored bullet in `deploy/grafana/README.md`) tells an operator to "enable it and point its exporter at your Prometheus's OTLP endpoint" — implying the shipped collector configs already contain a commented-out `otlphttp` exporter ready to uncomment. They don't. Both `deploy/otel-collector/docker/otelcol.yaml` and `deploy/otel-collector/kubernetes/collector-configmap.yaml` have their commented-out `metrics:` pipeline wired to `exporters: [otlp/backend]` / `[otlp/tempo]` — the same **gRPC** exporter already defined for the traces pipeline, targeting Tempo/Jaeger. Neither file defines an `otlphttp` exporter anywhere, commented or otherwise; the inline comment says "swap the exporter above" without showing what that swap looks like.

**Why it matters:** An operator who follows this doc as written (uncomments the metrics pipeline) gets a gRPC exporter pointed at a trace backend, not an HTTP exporter pointed at Prometheus — the metrics never reach Prometheus. That's exactly the gRPC/HTTP protocol mismatch this same PR's docs otherwise correctly warn about, reproduced by the doc's own instructions.

**Recommended fix:** Either add an actual commented-out `otlphttp/prometheus:` exporter block to both YAML files so "enable it" is accurate, or change the doc wording (in both `docs/observability.md` and `deploy/grafana/README.md`) to make clear the operator must author a new `otlphttp` exporter — with an example snippet inline — rather than merely uncomment existing lines.

docs/observability.md
```diff
@@ -893,6 +894,115 @@ the `mcp_*` instruments share — so the OTLP reader, which only observes what p
...
+- `$cloud_region` and `$service_name` come from the **[Resource Attributes](#resource-attributes--implemented)**
+  `target_info` series, joined the same way that section documents — **with one correction
+  learned by testing both egresses live, not assumed from the resource-attribute table alone**:
+  `$service_name`'s query is `label_values(target_info, job)`, **not**
+  `label_values(target_info, service_name)`. On a scrape-fed Prometheus, `target_info` does carry
+  an explicit `service_name` label, exactly as that section describes. On an OTLP-fed Prometheus,
+  it does not — Prometheus's OTLP-to-TSDB translator folds `service.name` into the standard `job`
+  label on `target_info` instead of also duplicating it there, while every *other*
+  `promote_resource_attributes`-listed attribute (`cloud.region` included) lands on `target_info`
+  normally. `job` returns the identical value on both egresses, so this is the one query that
```
bczoma, line 952: 🟡 **`$service_name`'s reliance on Prometheus `job_name` convention is undocumented**

**Flagged by:** good:architecture, docs (independently)

**What the problem is:** This paragraph asserts `job` "returns the identical value on both egresses" and is therefore the one query that works unmodified on scrape and OTLP paths. That's correct as far as it goes, but it never states the assumption that makes `job` a valid service-identity proxy at all: Prometheus's `job` label comes from `scrape_config.job_name`, an operator-chosen string, not from `service.name` — nothing in this repo ties `job_name` to service identity. If an operator names the scrape job something else, or multiplexes several services under one `job_name` (a legal, common Prometheus pattern), `$service_name` silently stops meaning "which service instance" despite the variable name and its documented role ("filter every panel").

**Why it matters:** An operator relying on this dropdown to isolate one service's telemetry, under a multi-service-per-job scrape topology, gets a filter that doesn't actually scope by service — misleading in exactly the way a dashboard variable named `$service_name` should not be.

**Recommended fix:** Add one sentence to this bullet (and the mirrored caveat in `deploy/grafana/README.md`): this assumes the scrape config's `job_name` maps 1:1 to this service — if a Prometheus job scrapes multiple services, `$service_name` reflects the job label, not per-service identity.

cmd/server/grafana_dashboard_test.go
```diff
@@ -0,0 +1,418 @@
...
+// against the UNION of every metric family referenced in the same target
+// expr, not against one specific family. That catches the failure this test
+// exists for — a name or label key that does not exist anywhere in the
+// documented schema — without a full PromQL parser to attribute each label
+// to its exact operand. It would not catch a label key that is real on some
+// OTHER mcp_* family but wrongly applied to the family actually being
+// queried (e.g. a stray "tool" label filter on a SEMP metric); that class of
+// bug is a query-correctness error, not a schema-drift error, and is out of
+// scope for what this lint claims to guarantee.
+func TestGrafanaDashboardMatchesGoldenFile(t *testing.T) {
```
bczoma, line 338: 🟡 **PR description claims a negative-path self-test that isn't committed**

**Flagged by:** good:code

**What the problem is:** The PR description states: "New CI test sanity-checked both directions: confirmed it fails on an injected bad metric name and on an injected bad template-variable query, then confirmed it passes on the real committed file." No such negative/mutation case exists in this committed file — it contains only three tests (`TestGrafanaDashboardMatchesGoldenFile`, `TestGrafanaDashboardTemplatingMatchesGoldenFile`, `TestGrafanaDashboardExcludesV1xMetrics`), all positive-path against the real dashboard/golden file.

**Why it matters:** The "both directions" verification was evidently a one-off manual check by the author, not a regression guard. A future refactor of `resolveFamily`, `extractLabelKeys`, or the matching regexes could silently turn this schema-drift check into a no-op (a test that cannot fail), and nothing in CI would catch it.

**Recommended fix:** Add a small table-driven test that feeds a synthetic dashboard/golden-family map with a deliberately-wrong metric name and a deliberately-wrong label key through `resolveFamily`/`extractLabelKeys`, and asserts those are detected — most simply by testing the pure helper functions directly rather than requiring a full `testing.T`-faking harness.

cmd/server/grafana_dashboard_test.go
```diff
@@ -0,0 +1,418 @@
...
+		if len(names) == 0 {
+			t.Errorf("panel %q: target expr references no mcp_*/go_*/process_*/target_info metric at all: %q", tgt.title, tgt.expr)
+			continue
+		}
+
+		var families []string
+		for _, n := range names {
+			fam, ok := resolveFamily(n, golden)
+			if !ok {
+				t.Errorf("panel %q: %q is not in the golden file (nor the nonSchemaFamilies allow-list) — dashboard and schema have drifted", tgt.title, n)
```
bczoma, line 356: 🟡 **Schema-drift tests don't assert the AC-required panels are actually present**

**Flagged by:** good:code

**What the problem is:** `TestGrafanaDashboardMatchesGoldenFile` and `TestGrafanaDashboardTemplatingMatchesGoldenFile` only check that names/labels the dashboard *does* reference exist in the golden file — they never assert the dashboard *contains* panels for the AC-required signals (tool RED, active requests, SEMP RED-per-attempt, auth-failure breakdown, error-type breakdown, Go runtime/build info).

**Why it matters:** A panel could be accidentally deleted, or a title/metric silently dropped, and every committed test would still pass — a shrunk panel set trivially satisfies "everything referenced exists." This leaves the Jira acceptance criterion ("panels render against documented metric names") only partially enforced by CI.

**Recommended fix:** Add a lightweight assertion — e.g. a set of expected metric families or panel titles derived from the AC — that the loaded panel list must cover, so removing a required panel breaks CI rather than only code review.

bczoma (dismissed):
## PR Review: #424 — SOL-152092: Ship an importable reference Grafana dashboard so operators monitor the service on day one without authoring panels (Story 37)

**Author:** solace-lootii | **Branch:** `lkiri/sol-152092-grafana-dashboard` → `main` | **Changes:** +1142 / -17 across 6 files
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-424`
**Jira:** `SOL-152092` — Ship an importable reference Grafana dashboard (Story 37) (Verifying)

---

### Verdict

1. 🟡 `docs/observability.md:1599` (and mirrored in `deploy/grafana/README.md`) — the checked-in reference OTel Collector configs don't actually contain the `otlphttp` exporter the docs tell an operator to enable
2. 🟡 `docs/observability.md:943` (and mirrored in `deploy/grafana/README.md`) — the `$service_name` variable's dependence on Prometheus `job_name` convention is undocumented
3. 🟡 `cmd/server/grafana_dashboard_test.go:330` — the PR description's claim of a negative-path/mutation self-test isn't backed by a committed test
4. 🟡 `cmd/server/grafana_dashboard_test.go:330-402` — the schema-drift tests don't assert the AC-required panels/variables are actually present

- 🟡 PR-level — single commit bundles two logically separate changes (the new dashboard, and an unrelated doc fix to already-merged PR #412's section)

The 4 🟢 notes below are **not** conditions of this verdict.

---

### Summary

This ships a well-verified, well-documented Grafana reference dashboard plus a genuinely useful CI guard (a golden-file schema-drift test) that keeps the dashboard's panel queries honest against the real metrics registry — good engineering discipline for a purely additive, fully reversible change. Two review comments from an earlier Copilot pass turned out to be accurate on independent verification (the collector configs really don't ship the exporter the new docs describe, and the `$service_name` variable really does lean on an undocumented scrape-job-naming convention); a third Copilot claim, that the new CI test itself fails, was checked by actually running it and is false — all three new tests pass. Two additional code-quality gaps surfaced independently: the PR description's claim of injected-bad-input self-testing isn't backed by a committed test, and the schema-drift check doesn't guard against a required panel silently disappearing.

---

### Minor notes

- 🟢 `cmd/server/grafana_dashboard_test.go:127-129,283-300` — `labelKeyRe` doesn't exclude quoted string values inside a `{...}` selector, so a future value like `outcome=~"foo=bar"` would spuriously extract `foo` as a label key. Harmless today; restrict the regex to match outside quoted segments, or note it as an accepted proportionate-scope limitation.
- 🟢 PR title (~120 chars with the "(Story 37)" suffix) could be trimmed to `SOL-152092: Ship an importable reference Grafana dashboard (Story 37)`, moving the "day one, no panel authoring" framing into the Summary, which already restates it.
- 🟢 `cmd/server/grafana_dashboard_test.go:251-278` duplicates `observability_doc_test.go`'s ~15-line golden-file parsing logic (`expfmt.NewTextParser` + family/label walk) almost verbatim. Not urgent; worth extracting to a shared test-support helper if a third consumer appears.
- 🟢 `deploy/grafana/solace-broker-mcp-overview.json:844-877` — the "Auth failure rate by reason" panel doesn't filter on `broker=~"$broker"` like the other RED panels do. Not a leak (no PII/token in the label set), just an inconsistency worth a follow-up.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

docs/observability.md
solace-lootii: Good catch — confirmed by reading both files directly, the commented-out metrics pipeline in `otelcol.yaml` and `collector-configmap.yaml` really did only reference the gRPC trace exporter, nothing pointed at Prometheus. Added an actual `otlphttp/prometheus` exporter (commented) to both files, plus a comment warning against pointing the metrics pipeline at the trace exporter instead, and tightened the doc wording to match exactly what's there now. See 15f920b.

docs/observability.md
solace-lootii: Confirmed — my doc never stated that assumption. Added the caveat to both `docs/observability.md` and `deploy/grafana/README.md`: `$service_name` reflects the Prometheus scrape job, not necessarily one service, and stops meaning that if a job scrapes multiple services. See 15f920b.

cmd/server/grafana_dashboard_test.go
solace-lootii: Fair — the description was describing a real manual check I ran, but you're right that nothing regression-guards it going forward. Added `TestResolveFamilyAndExtractLabelKeys_DetectDrift`, testing `resolveFamily`/`extractLabelKeys` directly against a synthetic golden map with deliberately-wrong names and labels. See 15f920b.

cmd/server/grafana_dashboard_test.go
solace-lootii: Agreed — confirmed it myself by actually deleting the Active Requests panel and watching every existing test stay green. Added `TestGrafanaDashboardCoversRequiredACPanels`, asserting each AC-required metric family (and the exact error-breakdown expression) is referenced somewhere. See 15f920b.

solace-lootii:
Thanks for the thorough review. Quick wrap-up:

**On the single-commit scope bundling:** leaving it as one PR rather than splitting. The #412 correction wasn't a drive-by fix — it's a direct byproduct of the live cross-path verification this story's own AC required; testing the OTLP ingestion path is exactly what surfaced that section was wrong. It genuinely belongs in the same body of work. Happy to split it out if you'd still prefer that on reflection — wanted to give the reasoning first rather than default to splitting.

**Also picked up while I was in there:** the quoted-string regex edge case (`extractLabelKeys` now strips quoted values before scanning for keys, so `outcome=~"foo=bar"` no longer spuriously extracts `foo`) and trimmed the PR title.

**Left alone, both intentional:** the auth-failure panel doesn't filter on `$broker` because `mcp_auth_failure_total` has no `broker` label at all — auth fails before a broker is even selected, so there's nothing to filter on; already noted in the docs. And I didn't extract the golden-file-parsing logic shared with `observability_doc_test.go` — matches your own call to wait for a third consumer rather than guess at the right shared shape from two.

bczoma (approved):
## PR Review: #424 — SOL-152092: Ship an importable reference Grafana dashboard (Story 37)

**Author:** solace-lootii | **Branch:** `lkiri/sol-152092-grafana-dashboard` → `main` | **Changes:** +1308 / -23 across 8 files (this pass: 5 files, +169/-9)
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-424`
**Jira:** `SOL-152092` — Ship an importable reference Grafana dashboard (Story 37)

Follow-up review: branch pushed two new commits since my prior review at `ad6d836` (dismissed by the push, findings below still apply). No escalation — the new commits touch only files already reviewed and add no new file.

---

### Verdict

**Recommendation:** ✅ Approve

No conditions.

*9 prior findings settled — 6 addressed with real fixes, 3 accepted as-is on a reasoned explanation.*

|---|---|---|---|
| 🟡 | Checked-in reference OTel Collector configs didn't actually contain the `otlphttp` exporter the docs told operators to enable | Addressed | `deploy/otel-collector/docker/otelcol.yaml:28-34` and `deploy/otel-collector/kubernetes/collector-configmap.yaml:43-49` — both now ship a commented `otlphttp/prometheus` exporter block plus a corrected pipeline comment naming it explicitly and warning against pointing the metrics pipeline at the trace exporter |
| 🟡 | `$service_name`'s dependence on Prometheus `job_name` convention was undocumented | Addressed | `docs/observability.md:955-961` and `deploy/grafana/README.md:150-154` — both now state `job` is scrape metadata from `scrape_config.job_name`, not derived from `service.name`, and what breaks if one job scrapes several services |
| 🟡 | PR description's claim of a negative-path self-test wasn't backed by a committed test | Addressed | `cmd/server/grafana_dashboard_test.go:45-94` — new `TestResolveFamilyAndExtractLabelKeys_DetectDrift` injects an unknown metric name and a bad label key directly against `resolveFamily`/`extractLabelKeys` |
| 🟡 | Schema-drift tests didn't assert the AC-required panels were actually present | Addressed | `cmd/server/grafana_dashboard_test.go:115-141` — new `TestGrafanaDashboardCoversRequiredACPanels` checks all 8 AC-required metric families are referenced, plus the exact AC-pinned error-breakdown query expression |
| 🟡 | PR-level: single commit bundled the new dashboard with an unrelated doc fix to already-merged PR #412's section | Addressed (reasoning accepted) | Author's explanation is sound: the #412 correction is a direct byproduct of the live cross-path verification this story's own AC required, not a drive-by fix — testing the OTLP ingestion path is what surfaced that section was wrong. Genuinely one body of work; not splitting is reasonable |
| 🟢 | `labelKeyRe` didn't exclude quoted string values, so a value like `outcome=~"foo=bar"` would spuriously extract `foo` as a key | Addressed | `cmd/server/grafana_dashboard_test.go:9-14,290-292` — new `quotedStringRe` strips quoted segments before the key scan; regression-tested at lines 84-92 |
| 🟢 | PR title could be trimmed | Addressed | Title is now `SOL-152092: Ship an importable reference Grafana dashboard (Story 37)`, exactly as suggested |
| 🟢 | `grafana_dashboard_test.go` duplicates `observability_doc_test.go`'s golden-file parsing logic | Accepted as-is | Filed as optional ("worth extracting if a third consumer appears"); author left it, matching the original framing |
| 🟢 | Auth-failure panel doesn't filter on `broker=~"$broker"` like other RED panels | Accepted as-is | Author confirms `mcp_auth_failure_total` carries no `broker` label — auth fails before a broker is selected, so there is nothing to filter on; already noted in docs |

### Summary

This follow-up closes every condition from the prior review with a real fix — the reference collector configs now actually ship the exporter the docs describe, the undocumented `job_name` convention is now documented in both places it's mirrored, and two new tests close the self-test and AC-panel-coverage gaps. The one PR-level scope objection (bundling the #412 doc fix) is resolved by a reasoned explanation rather than a split, which holds up: the fix is a genuine byproduct of this story's own required verification work.

---

### Minor notes

None — the diff since the prior review does not introduce anything new.

---

### Summary Verdict

|---|---|---|
| Architecture | — not re-reviewed (unchanged since ad6d836) | — |
| Security | — not re-reviewed (unchanged since ad6d836) | — |

PR #428, opened 2026-09-16
SOL-153295: route broker SEMP traffic through HTTP_PROXY/HTTPS_PROXY/…

solace-awaheed:
…NO_PROXY

## Summary

Broker SEMP traffic now honours the standard `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY`
environment variables, so the server can run in a network that requires an outbound proxy.
Ships with the operator documentation for proxy scope, since after this change broker,
IdP, and telemetry traffic all read the same process-wide environment.

Jira: [SOL-153295](https://sol-jira.atlassian.net/browse/SOL-153295) (this repo tracks work in
Jira rather than GitHub issues, so there is no `Fixes #` to link.)

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Dependency update

Framed as a feature per the SOL-124077 precedent (proxy support was announced as a feature for
the Terraform provider and for JCSMP/JMS), but **also checked as breaking**: it changes existing
behaviour for anyone who already has `HTTPS_PROXY` in the environment. See Breaking Changes.

## Motivation and Context

`NewTunedTransport` built its `http.Transport` as a struct literal and left `Proxy` unset.
A nil `Proxy` in `net/http` means "never proxy" rather than "consult the environment", so an
operator behind a mandatory egress proxy got connection failures with no error and no log line
saying `HTTPS_PROXY` had been ignored — while IdP traffic from the same process honoured it,
because `internal/idpclient` clones `http.DefaultTransport`. The result was a half-proxied
server with nothing documented either way.

This comes from a customer request against the Solace Terraform provider (Siemens AG, SOL-124077,
shipped in provider v1.1.0), not from a defect report here. The MCP server has the same code shape
and therefore the same gap. Like the Terraform fix, this unlocks the proxy support already built
into Go's `net/http` transport rather than inventing a product-specific configuration surface.

## Changes Made

- `internal/semp/resilience/transport.go` — set `Proxy: http.ProxyFromEnvironment` on the SEMP
  transport. Both protocol clients (`sempv1`, `sempv2`) use this transport, so this covers all
  broker traffic for every tool. Set explicitly rather than by cloning `http.DefaultTransport`:
  `Clone` would also bring `ForceAttemptHTTP2: true`, which would multiplex in-flight requests
  onto one connection and invalidate the `MaxConnsPerHost` sizing the per-broker concurrency cap
  depends on. `ProxyFromEnvironment` itself is HTTP/2-neutral — net/http's conservative
  auto-disable keys off `TLSClientConfig` and `DialContext`, not `Proxy`.
- Same file — replaced a now-wrong doc comment. It previously cited inheriting
  `Proxy: ProxyFromEnvironment` as a *reason to avoid* `Clone`; the HTTP/2 argument (the real
  reason) is kept intact.
- Same file — `EffectiveProxy` resolves which proxy applies to a given broker URL, as a log-safe
  label (`direct`, or the proxy URL with userinfo stripped via `config.SanitizeURLString`). It and
  the transport's `Proxy` field both read one package-level `proxyResolver`, so the label cannot
  drift from the resolution the transport actually performs.
- `internal/semp/pool.go` — the existing once-per-broker `broker connection created` log line now
  carries a `proxy` field. Added at review request: the reroute is not opt-in, so without it an
  operator whose environment already carried `HTTPS_PROXY` saw only a terse `proxyconnect tcp:`
  dial error against a config file that worked on the previous release.
- `internal/semp/resilience/transport_test.go` — `TestNewTunedTransport_HonorsProxyEnvironment`,
  plus `TestProxyLabel_StripsProxyCredentials` (including the socks5 unrecognized-scheme case,
  which must fail closed rather than emit a raw URL with its password) and
  `TestEffectiveProxy_DirectCases`.
- `test/integration/proxy_env_invariant_test.go` — new
  `TestProxyEnvironmentInvariantAcrossSubsystems`, a cross-package test tying the SEMP and IdP
  transports to the same "always honour the proxy environment" invariant using their real
  production constructors. Added at review request, because that invariant now spans three
  subsystems and was carried only by prose. OTLP is deliberately left unasserted and named as
  such — grpc-go exposes no `*http.Transport` for a test to inspect.
- `docs/configuration.md` — new "Outbound HTTP Proxy" section: which outbound paths the variables
  govern, worked `NO_PROXY` examples for internal brokers with an external IdP and the inverse, a
  matching-rules table, proxy authentication, and the restart-scoped / process-wide / silent-when-wrong
  limits.
- `docs/authentication.md` — pointer from the existing "failed to connect to identity provider"
  troubleshooting entry, for the IdP half of the same setting.
- `CHANGELOG.md` — capability under `Added`, upgrade reroute under `Changed` (see Breaking Changes).

### Two corrections to the story's written scope

**1. The CIDR example the ticket specified would not have worked.** Scope item 4 asked for a worked
`NO_PROXY` CIDR example for the internal-brokers topology. Verified against Go's vendored
`golang.org/x/net/http/httpproxy`: `useProxy` consults CIDR matchers only `if ip != nil`, and
`domainMatch.match` returns false for an IP-literal host — the two entry families are mutually
exclusive. Go matches `NO_PROXY` against the host **as written in the URL** and performs no DNS
resolution, so `NO_PROXY=10.0.0.0/8` never exempts `url: "https://broker.example.com:943"` even
when that name resolves inside the block. A hostname is the documented broker form
(`broker-config.example.yaml:201`), so the docs lead with domain suffixes and state the CIDR limit
explicitly.

**2. OTLP telemetry export is a third proxied path, already live and undocumented.** Found during
review. `otlptracegrpc.New` / `otlpmetricgrpc.New` pass no proxy options, `resolver_wrapper.go:92`
builds grpc-go's delegating resolver unconditionally, and `proxyURLForTarget` synthesizes
`Scheme: "https"` — so **`HTTPS_PROXY` governs OTLP export even when `OTEL_EXPORTER_OTLP_ENDPOINT`
is `http://`**. The shipped endpoint is `http://otel-collector:4317`
(`deploy/otel-collector/docker/docker-compose.yaml:13`), which matches no dotted `NO_PROXY` suffix,
so a collector left out of `NO_PROXY` is CONNECT-tunneled to a proxy that usually cannot reach it
and telemetry stops silently. This is grpc-go's behaviour and **predates this change** — only the
documentation of it is new — but it had to land here, because the first draft of this section
would otherwise have told operators to configure something that breaks their telemetry.

## Testing

**Test Environment:**
- Go version: go1.26.5 darwin/arm64
- OS: macOS 26.6.2 (Darwin 25.6.0)
- Broker version (if applicable): n/a — no broker-facing behaviour changed; the transport is
  configured identically apart from proxy resolution

**Tests Performed:**
- [ ] Integration tests added/updated (if applicable) — n/a, no integration surface touched
- [ ] E2E tests added/updated (if applicable) — none added; see note below
- [x] Tested locally with `go test -race ./...` (via `make check`) — see caveat below
- [ ] Tested with real Solace broker — **not done, and not through a live proxy either**; see
      "Honest limits" below

**Test Coverage:**
- `TestNewTunedTransport_HonorsProxyEnvironment` pins the transport's `Proxy` to
  `http.ProxyFromEnvironment` by **function identity**, not by behaviour. `ProxyFromEnvironment`
  resolves the environment once and caches it behind a `sync.Once`, so a `t.Setenv`-then-call test
  would pass or fail depending on whether an earlier test in the same binary had already tripped
  that `Once` — order-dependent, and green for the wrong reason once it is. Asserting merely
  non-nil would also be satisfied by a func that never returns a proxy, which is the defect being
  fixed. The comment notes that `reflect` func pointers are not a unique identity, so this is a
  strong signal rather than a proof.
- The HTTP/2 posture the `MaxConnsPerHost` sizing depends on is deliberately left to the existing
  `TestNewTunedTransport_MaxConnsPerHostEnforcesConcurrencyCap` rather than duplicated. That test
  is what fails if anyone later reaches for `DefaultTransport.Clone()` to obtain this field.
- Every `NO_PROXY` claim in the new docs table was verified against Go's matcher source rather than
  from memory: apex-plus-subdomain for a bare domain, subdomains-only for leading-dot and `*.`,
  CIDR not port-scopeable, `*` disabling all, upper-case env form winning, and strict scheme
  selection with no `HTTPS_PROXY`→`HTTP_PROXY` fallback.
- Package coverage `internal/semp/resilience` 90.2%; repo total 89.7% against the 85% floor;
  `transport.go` at 100%.

**Honest limits of this testing:**
- No traffic was pushed through an actual HTTP proxy. What is verified is that the transport
  carries `ProxyFromEnvironment` and that Go's resolution/matching semantics are as documented.
  End-to-end proxy validation against a live proxy has not been done here — on SOL-124077 the
  equivalent path (including `user:password@` proxy auth) was verified working for the Terraform
  provider, which is the basis for expecting it to behave.
- `make check` is green except two pre-existing failures in `test/performance/memsampler`
  (`TestCountOpenFDsSelfExcludesItsOwnHandle`, `TestCountOpenFDsChildProcessIsNotCorrected`).
  They read `/proc/<pid>/fd`, which does not exist on macOS; confirmed failing identically on
  clean `main`. Expected to pass on CI's Linux runners.
- E2E impact was checked but not extended: every suite addresses brokers at `localhost` or
  `127.0.0.1` (`test/e2e-common/lib.sh:32-33,378`), which Go exempts from proxying
  unconditionally. The one exception is `test/e2e-tracing/mcp-config.yaml:11`
  (`http://solace:8080`, a compose hostname), so a runner or base image carrying `HTTP_PROXY`
  would newly route that suite's SEMP traffic through it. CI sets no proxy variables, so this is
  latent rather than live.

## Breaking Changes

**Impact:**

This is not a pure addition. IdP traffic has always followed these variables, so an operator who
already has `HTTPS_PROXY` set — for egress to a cloud IdP, or inherited from a base container
image — will start routing **broker** traffic through that proxy after upgrading, where it
previously went direct. Brokers are typically internal while such a proxy typically cannot reach
them, and nothing logs the effective proxy, so the only signal is a `proxyconnect tcp:` dial
error against a configuration file that worked on the previous release.

**Migration Guide:**

Before upgrading, check whether the server's environment carries `HTTP_PROXY` or `HTTPS_PROXY`
(including anything inherited from a base image). If it does, exempt the brokers — and the OTLP
collector, if telemetry is enabled — in `NO_PROXY`:

```bash
# Brokers at broker-01.internal.example.com, broker-02.internal.example.com
# Collector at otel-collector:4317
HTTPS_PROXY=http://proxy.example.com:3128
NO_PROXY=.internal.example.com,otel-collector
```

Use the host form your config actually uses: `NO_PROXY` matches the host as written in the URL,
so a CIDR block only exempts a broker addressed by IP literal, and a single-label host like
`otel-collector` needs its own entry because it matches no dotted suffix. The full entry-form
table is in `docs/configuration.md` § "Outbound HTTP Proxy". This is a restart-scoped setting.

No configuration-file migration is required, and no YAML schema or tool name changed.

### Code Quality

### Security
- [x] Followed [secure logging rules](../docs/internal/secure-logging-rules.md) — ran `/check-logs`;
      clean. The diff introduces no log statements, nothing logs an `*http.Transport` or its
      `Proxy`, and both `HTTPClient.LogValue` implementations expose only a sanitized `base_url`.
- [x] Credential-carrying types implement `slog.LogValuer` — n/a, no new credential-carrying types
- [x] No security vulnerabilities introduced (`golangci-lint run` → 0 issues; gosec is enabled in
      `.golangci.yml`)

One security caveat is documented rather than fixed, because it is outside this codebase: a proxy
URL may carry a password, and if an operator sets `GRPC_GO_LOG_SEVERITY_LEVEL=info` to debug OTLP
export, grpc-go's `delegatingresolver.go:135` logs the detected proxy URL to stderr unredacted
(`url.URL.String()` writes the password; only `Redacted()` masks it). grpc discards INFO by default,
so this requires opting in. The docs state this explicitly instead of claiming the server never
emits proxy URLs anywhere.

### Testing
- [x] All tests pass locally: `go test -race ./...` — except the two pre-existing macOS-only
      `memsampler` `/proc` failures noted above
- [x] Error paths tested

### Documentation
- [ ] README.md updated (if user-facing changes) — **not updated.** `README.md:108` describes
      `docs/configuration.md` as "server settings, event broker configuration, client
      authentication, and rate-limit/retry settings", which now omits proxy configuration. Happy
      to add it if reviewers want the enumeration kept current.
- [x] godoc comments added/updated for exported functions (`NewTunedTransport`)
- [ ] Examples updated (if applicable) — n/a; this is configured by environment variable, so there
      is no YAML block to add to `broker-config.example.yaml`

### Git & CI
- [x] Branch is up to date with main (merge-base == `origin/main` @ `23b5091`; no rebase needed)
- [x] No merge conflicts
- [ ] CI checks passing (lint, build, test, E2E) — pending; CI has not run on this branch yet

- [x] Breaking changes documented in commit message
- [x] Breaking changes documented in PR description
- [x] Migration guide provided
- [ ] Deprecated functionality marked with comments and deprecation notices — n/a, nothing deprecated

n/a

## Additional Notes

**Previously deferred, now landed:** an earlier revision of this description proposed the
per-broker proxy log line as a follow-up ticket, on the reasoning that it needed a
once-per-process guard. That reasoning was wrong — `pool.go`'s `broker connection created` line
already fires exactly once per broker behind the client cache — and the work is in `ddf360a`.
Nothing from this story is deferred to a follow-up ticket.

**Two review notes deliberately not acted on**, both filed as non-conditions. `EffectiveProxy`
collapses "no proxy resolved" and "unparseable rawURL" into the same `ProxyDirect` label, which
the doc comment states and which config validation makes unreachable in practice. And the
transport's `Proxy` assertion stays function-identity rather than dependency-injected — the
`proxyLabel` split already gives behavioural coverage of the part carrying security weight
(credential stripping), leaving only "is this field the env-reading function" for identity to
cover.

**Also noticed, pre-existing, left alone:** `SSL_CERT_FILE` is read by `internal/idpclient` but
documented nowhere; and `internal/semp/resilience/sender.go` is unformatted per local `gofmt`
(true on clean `main` too).

**Explicitly out of scope per the story:** per-broker or per-role proxy configuration in
`broker-config.yaml`. That would give genuine "proxy this broker, not that one, and not the IdP"
control, which the environment variables cannot express, but no customer has asked for it and the
Terraform fix declined the same extension. File a follow-up if a customer with a segmented network
asks.

## Related Issues/PRs

- Related to SOL-124077 — the Solace Terraform provider request this mirrors (shipped in provider
  v1.1.0); the approach and the "announce as a feature" framing both follow it
- Parent epic: SOL-153075 — Broker MCP GA Readiness (MCP 7)
- Depends on # — none
- Blocks # — none

---

**For Reviewers:**

**Focus Areas**:
1. **The one functional line** — `Proxy: http.ProxyFromEnvironment` in `NewTunedTransport`. Worth
   confirming the HTTP/2 reasoning: net/http's conservative auto-disable keys off
   `TLSClientConfig`/`DialContext`, not `Proxy`, so the `MaxConnsPerHost` sizing is unaffected.
2. **The upgrade behaviour change.** Is `NO_PROXY` guidance in the docs and changelog enough for an
   operator who already has `HTTPS_PROXY` set, or should this ship behind an opt-in?
3. **`NO_PROXY` documentation accuracy**, especially the CIDR-vs-hostname rule and the OTLP row —
   these are the two places where the natural thing to write is wrong, and where a reader is most
   likely to act on the docs directly.

**Questions**:
1. **Changelog placement** — the story asks for feature framing (SOL-124077 precedent), but the
   upgrade impact is the one thing in this release that can silently break broker connectivity.
   I split it: capability under `Added`, reroute under `Changed`. Is that the right call, or should
   it be a single `Added` entry as the story reads?
2. **OTLP scope** — the OTLP documentation is docs-only and describes pre-existing grpc-go
   behaviour. Keep it in this PR (my preference — the `NO_PROXY` guidance is actively misleading
   without it), or split it into its own docs ticket?
3. **Proxy startup logging** — separate ticket, as I've assumed, or should it land here given the
   change is not opt-in?
4. **The CIDR deviation** — scope item 4 asked for a worked CIDR example that would not have
   worked. Confirming the correction is accepted rather than a scope change needing sign-off.
5. **Test assertion style** — is pinning `Proxy` by `reflect` function identity acceptable, given
   the `sync.Once` caching makes an env-driven behavioural test order-dependent? Open to a better
   idea if one exists.


internal/semp/resilience/transport.go
```diff
@@ -149,15 +149,27 @@ func newSEMPDialer(requestTimeout time.Duration) *net.Dialer {
...
+// Two limits, both under "Outbound HTTP Proxy" in docs/configuration.md: the
+// environment is read once per process and cached, so this is restart-scoped
+// and cannot vary per broker; and NO_PROXY matches the host as written in the
+// URL, so a CIDR entry exempts only a broker addressed by IP literal, never
+// one addressed by hostname.
 func NewTunedTransport(brokerCfg *config.BrokerConfig, sempCfg *config.SEMPConfig) *http.Transport {
 	return &http.Transport{
 		TLSClientConfig:       &tls.Config{InsecureSkipVerify: brokerCfg.InsecureSkipVerify}, //nolint:gosec // G402 — user-configurable TLS skip for dev environments; defaults to false
 		DialContext:           newSEMPDialer(sempCfg.RequestTimeoutDuration).DialContext,
+		Proxy:                 http.ProxyFromEnvironment,
```
bczoma: 🟡 **Broker traffic can reroute through a proxy with no runtime signal that it happened**

**Flagged by:** good:architecture, good:security (independent convergence)

**What the problem is:** `NewTunedTransport` now sets `Proxy: http.ProxyFromEnvironment`, so any broker whose process environment carries `HTTP_PROXY`/`HTTPS_PROXY` (deliberately set, or inherited from a base container image) has its SEMP traffic silently rerouted through that proxy. Nothing in this diff logs that this happened. The codebase already has a directly analogous precedent one line away: `sempv1/client.go:107` and `sempv2/client.go:161` call `slog.Warn("INSECURE: TLS verification disabled for broker", ...)` at the same construction point, once per broker. `config.SanitizeURLString` (used by `HTTPClient.LogValue()` two lines away) is already available to redact a proxy URL's userinfo.

**Why it matters (measured):** Reachability is real, not an edge case — the PR's own CHANGELOG entry names two routine trigger scenarios (proxy set for cloud IdP egress, or inherited from a base image). Blast radius is process-wide: every broker's SEMP traffic is affected identically once the env var is set, confirmed via `NewTunedTransport`'s own restart-scoped, cached-once-per-process doc comment. No self-healing path exists — `get-broker-status`-style handlers have no proxy-aware error classification. The one partial signal is Go's own `&net.OpError{Op: "proxyconnect", ...}` wrapping, whose `Error()` string does contain the substring `"proxyconnect tcp:"` — a real but terse clue, not "zero indication of why." Construction is confirmed once-per-broker already (`pool.go:212` → `NewBrokerClient` → `NewHTTPClient` → `NewTunedTransport`), so the "would need a once-per-process guard" reason given in the PR body for deferring this doesn't hold against the actual call graph — no new guard is needed. Given the change already ships with unusually thorough CHANGELOG/docs mitigation, this is a real-but-non-blocking observability gap rather than a merge blocker.

**Recommended fix:** Add one `slog.Info`/`Warn` call per broker in `NewHTTPClient` (same construction point as the existing `InsecureSkipVerify` warn, same once-per-broker cardinality) logging the effective proxy for that broker's URL via `http.ProxyFromEnvironment`, sanitized through the existing `config.SanitizeURLString`. Reasonable to land as an immediate fast-follow rather than in this PR, but it should not sit as an open-ended "worth a follow-up ticket" note — please file it now if it doesn't land here.

bczoma (dismissed):
## PR Review: #428 — SOL-153295: route broker SEMP traffic through HTTP_PROXY/HTTPS_PROXY/NO_PROXY

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-153295` → `main` | **Changes:** +157 / -5 across 5 files
**Worktree:** `/home/local/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-428`
**Jira:** `SOL-153295` — Broker MCP Server: support HTTP proxy configuration via HTTP_PROXY/HTTPS_PROXY/NO_PROXY (Verifying)

---

### Verdict

Conditioned on these 2:

1. 🟡 `internal/semp/resilience/transport.go:172` — no runtime signal (log/metric) that broker traffic is now being routed through a proxy, despite an adjacent precedent in the same file for exactly this class of warning (inline thread below)
2. 🟡 `internal/semp/resilience/transport.go` (design-level) — three HTTP-using subsystems (SEMP, IdP, OTLP) now share an "always honour proxy env vars" invariant enforced only by documentation, with no test tying the three construction sites together

Not anchorable to this diff as an inline thread — GitHub's review API rejected a file-level anchor on this repo (`subjectType` not accepted on this endpoint's schema) — so please treat as part of the same conditioned list:

- 🟡 `internal/semp/resilience/transport.go` — see item 2 above. **What the problem is:** SEMP, IdP, and OTLP telemetry each honour proxy env vars through three different, uncoordinated mechanisms (a hand-set `Proxy` field here, a side effect of `Clone()` in `internal/idpclient`, and grpc-go's own resolver for OTLP), with the "all three stay consistent" invariant enforced only by `docs/configuration.md` prose, not by code or a test. **Why it matters:** a future change to any one path (e.g. `idpclient` moving off `Clone()`, or a grpc-go bump changing proxy-target synthesis) could silently break the cross-subsystem consistency the docs promise, and nothing would fail to catch it — this PR's own OTLP finding was only surfaced by manual digging, which is exactly the kind of drift this shape invites. **Recommended fix:** a small test asserting all three transport-construction sites carry proxy-following behaviour, or at minimum a comment co-locating references between the three construction sites, so the coupling is visible rather than rediscovered by incident. Not a blocker for this PR.

### Summary

This is a small, well-reasoned fix: one struct-literal field (`Proxy: http.ProxyFromEnvironment`) closes a real gap where broker SEMP traffic silently ignored `HTTP_PROXY`/`HTTPS_PROXY` while IdP traffic already honoured them. The PR is unusually thorough for its size — it independently verified two claims from the story's original scope (a `NO_PROXY` CIDR example that would not actually have worked, and a third, pre-existing, previously-undocumented proxied path through OTLP/grpc-go) rather than taking the ticket's wording at face value, and it documents the upgrade-breaking behavior change in both `CHANGELOG.md` and `docs/configuration.md` with a concrete migration path. Three security/architecture claims were filed and one (a 🔴 on the missing runtime signal for the reroute) was reduced by measurement to a 🟡 once reachability, an existing partial error-string clue, and the change's already-thorough documentation were accounted for.

---

### Minor notes

🟢 **Minor** — worth fixing, not a condition of this verdict:

- 🟢 `docs/configuration.md` (Proxy authentication section) — the grpc-go stderr password-leak warning names `GRPC_GO_LOG_SEVERITY_LEVEL=info` as the trigger, but verification against the pinned `google.golang.org/grpc@v1.83.2` shows the log line also requires `GRPC_GO_LOG_VERBOSITY_LEVEL>=2` — add that second env var to the precondition so the warning names the actual trigger.
- 🟢 `README.md:108` — the `docs/configuration.md` guide-list entry ("server settings, event broker configuration, client authentication, and rate-limit/retry settings") doesn't mention proxy configuration, so a reader scanning the README won't be pointed at the new section. One-line addition; the author already flagged and offered to fix this.
- 🟢 `internal/semp/resilience/transport_test.go:297-314` — `TestNewTunedTransport_HonorsProxyEnvironment` pins `Proxy` by `reflect` function-pointer identity rather than by behavior, to avoid `ProxyFromEnvironment`'s `sync.Once` ordering hazard. Reasonable as filed and the test comment already discloses the limitation, but dependency-injecting the proxy resolver (there are only two call sites) would give real behavioral coverage without the ordering risk, if a future pass wants it.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

internal/semp/resilience/transport.go
solace-awaheed: Addressed in ddf360a — with one correction to the placement, because the cardinality in the rationale does not hold.

**The log line.** `broker connection created` in `internal/semp/pool.go` now carries a `proxy` field: either `direct` or the proxy URL with userinfo stripped. Resolution goes through a new `resilience.EffectiveProxy`, which calls `http.ProxyFromEnvironment` — the same function the transport holds — so the label cannot drift from what the transport actually does, `NO_PROXY` and the loopback exemption included.

**Why not `NewHTTPClient`.** The comment says construction is once-per-broker via `pool.go:212 → NewBrokerClient → NewHTTPClient → NewTunedTransport`. That chain is right but it runs twice per broker: `internal/semp/broker.go:162` and `:167` build a SEMPv1 **and** a SEMPv2 client from the same `brokerCfg`, and each calls `NewTunedTransport`. So a log at that point double-fires per broker — which the `InsecureSkipVerify` warning cited as precedent already does, once from `sempv1/client.go:109` and once from `sempv2/client.go:163`.

`pool.go` avoids that: it is behind the `p.clients[canonical]` cache, so it fires exactly once per broker, and it already carries the alias, the sanitized URL, and `auth_mode`. Your underlying point stands and no once-per-process guard was needed — my PR-body reason for deferring was wrong, just not for the reason given. Only the location moved.

**Sanitizing.** Uses `config.SanitizeURLString` as recommended. One consequence worth recording: that helper only recognizes `http` and `https`, so a `socks5://` proxy logs as the unparseable-URL placeholder rather than its host. That still answers "a proxy is in play", which is the diagnostic that matters here, and failing closed is the right direction for a value that can hold a password. `proxyLabel` is split out of `EffectiveProxy` so the stripping is pinned directly by `TestProxyLabel_StripsProxyCredentials` — reaching that path through `EffectiveProxy` would have depended on nothing else in the test binary having resolved a proxy first.

Also documented under a new "Confirming which brokers are proxied" section in `docs/configuration.md`, with the `proxy!=direct` alerting hint. This let me drop two claims that my own change falsified — the docs' "Nothing logs the effective proxy" and the equivalent line in the CHANGELOG `Changed` entry.

solace-awaheed:
Review addressed in ddf360a. The inline thread covers conditioned item 1; this covers item 2 and the three minors.

## Conditioned 2 — cross-subsystem proxy invariant now has a test

`test/integration/proxy_env_invariant_test.go` adds `TestProxyEnvironmentInvariantAcrossSubsystems`, pinning the "always honour the proxy environment" property across the construction sites rather than leaving it to prose in `docs/configuration.md`.

It asserts the two transports that are structurally inspectable — `resilience.NewTunedTransport` and `idpclient.NewHTTPClient` — and explicitly names the third, OTLP export, as resolved inside grpc-go's delegating resolver rather than through an `*http.Transport` a test can read, so a future reader sees it was considered rather than missed.

The two realistic ways the invariant breaks are what it defends: someone reverts the SEMP transport to a bare struct literal while tuning connection pooling (a nil `Proxy` reads as a default, not an opt-out), or someone stops `idpclient` cloning `http.DefaultTransport` to gain the same explicit control the SEMP transport now has. Identity assertion rather than behavioural, for the `sync.Once` reason already documented on the per-package test.

## Minor 1 — grpc-go leak precondition: confirmed, corrected

You are right, and thanks for checking the pinned version. The line is behind an `if logger.V(2)` gate at `delegatingresolver.go:134`, so it needs **both** `GRPC_GO_LOG_SEVERITY_LEVEL=info` (otherwise the info writer is `io.Discard`) **and** `GRPC_GO_LOG_VERBOSITY_LEVEL=2`. The docs named only the first. Now names both, and states that neither is set by default.

## Minor 2 — README guide entry

`README.md:108` now names outbound HTTP proxy support alongside the other `docs/configuration.md` topics.

## Minor 3 — test DI: left as filed

Taking your "if a future pass wants it" framing. The `proxyLabel` split made for conditioned item 1 does give real behavioural coverage of the part that carries security weight — credential stripping — without the ordering hazard, so the remaining untested-by-behaviour surface is just "is this field the env-reading function", which identity covers exactly.

## Two things my own change falsified

Worth flagging since they were in the PR body you reviewed against: adding the log line made the docs' "Nothing logs the effective proxy" and the CHANGELOG `Changed` entry's equivalent claim untrue. Both rewritten, and the deferral note in the PR description is now obsolete — the work landed here rather than as a follow-up ticket, as you asked.

## Verification

`golangci-lint run` → 0 issues; coverage 89.7% against the 85% floor, `internal/semp/resilience` at 90.2%; `/check-logs` clean on the one new attribute (routes through `config.SanitizeURLString`, and `proxy` does not collide with the `ReplaceAttr` redaction substrings). Local `make check` is otherwise green apart from two pre-existing `test/performance/memsampler` failures that read `/proc/<pid>/fd` and so cannot pass on macOS — they fail identically on clean `main`.

bczoma (dismissed):
## PR Review: #428 — SOL-153295: route broker SEMP traffic through HTTP_PROXY/HTTPS_PROXY/NO_PROXY

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-153295` → `main` | **Changes:** +392 / -7 across 8 files
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-428`
**Jira:** `SOL-153295`

Follow-up review: branch pushed one new commit (`ddf360a3`) since my prior review at `e331b584` (dismissed by that push, findings below still apply). Escalated to the full 5-lens pipeline because the new commit adds a file (`test/integration/proxy_env_invariant_test.go`).

---

### Verdict

**Recommendation:** ✅ Approve

No conditions. The 5 🟢/💡 notes below are **not** conditions of this verdict.

|---|---|---|---|
| 🟡 | Broker traffic reroutes through a proxy with no runtime signal that it happened | Addressed | `internal/semp/pool.go:230-234` — the existing once-per-broker `broker connection created` log now carries `proxy` via `resilience.EffectiveProxy(cfg.URL)`, which calls the identical `http.ProxyFromEnvironment` the transport itself uses, so the label cannot diverge from reality; credentials verified stripped fail-closed through `config.SanitizeURLString`. (My original recommended *placement*, `NewHTTPClient`, was itself wrong on cardinality — the author correctly placed it in `pool.go` behind the client cache instead, and pointed this out; verified correct.) |
| 🟡 | Three-subsystem proxy-env invariant (SEMP/IdP/OTLP) enforced only by docs prose, no test | Addressed | `test/integration/proxy_env_invariant_test.go` — new `TestProxyEnvironmentInvariantAcrossSubsystems` exercises the real `resilience.NewTunedTransport` and `idpclient.NewHTTPClient` constructors, asserting both resolve `Proxy` via `http.ProxyFromEnvironment`. OTLP is deliberately and explicitly left unasserted (grpc-go has no `*http.Transport` for this test to inspect), which is a reasonable, documented scope boundary rather than a gap. |

*3 further prior 🟢 notes settled — 2 addressed (`docs/configuration.md`'s grpc-go log-leak warning now names both required env vars; `README.md:108` now mentions proxy support), 1 accepted as-is (the `reflect` function-identity test assertion style — filed as optional, still reasonable).*

### Summary

This follow-up commit closes both conditions from my prior review with real fixes, not just claims: the proxy reroute is now logged once per broker with a value provably identical to what the transport actually resolves, and a new cross-subsystem integration test ties the SEMP and IdP transports to the same proxy-following invariant using their real production constructors. All five lenses came back clean; nothing rose above 🟢.

---

### Minor notes

🟢 **Minor** — worth fixing, not a condition of this verdict:

- 🟢 PR-level — the new `test/integration/proxy_env_invariant_test.go` isn't mentioned in the PR description; a reviewer reading only the description won't know a cross-package invariant test now exists. Add one line mirroring the CHANGELOG prose already written.
- 🟢 `internal/semp/resilience/transport.go:161-171` (`EffectiveProxy`) — resolves the proxy independently of the transport's own `Proxy` field rather than reading it off an actual `*http.Transport`; today both call the identical `http.ProxyFromEnvironment`, so there's no live drift, but nothing statically ties them together for a future refactor. Not urgent given the package's small size.
- 🟢 `internal/semp/resilience/transport_test.go` — the doc comment for `EffectiveProxy` specifically calls out a `socks5://` proxy URL as falling back to the `<unparseable url>` placeholder, but no test pins that claim. Add one case.
- 💡 `internal/semp/resilience/transport.go:163-165` — `EffectiveProxy` collapses "no proxy resolved" and "unparseable rawURL" into the same `ProxyDirect` label; accepted trade-off per the doc comment (upstream config validation prevents the latter in practice), not a defect.
- 💡 `docs/observability.md:2445` — the existing mention of the `broker connection created` log line wasn't updated to note the new `proxy` field; low-impact since that passage doesn't enumerate the line's fields exhaustively.

---

### Summary Verdict

|---|---|---|
| Documentation | ✅ | docs |

solace-awaheed:
Follow-up notes addressed in 033bbb8. Three acted on, two declined with reasons.

## 🟢 PR description missing the new test file — fixed

Description now names `test/integration/proxy_env_invariant_test.go`, `EffectiveProxy`, the `pool.go` `proxy` field, and the two new unit tests. Also replaced the stale "out of scope, worth a follow-up ticket" note about proxy logging — that work landed in `ddf360a`, and my stated reason for deferring it (needing a once-per-process guard) was wrong, which the description now says outright. Nothing from this story is deferred to a follow-up ticket.

## 🟢 `EffectiveProxy` not statically tied to the transport — fixed

I had recommended declining this one, then changed my mind: the fix is one line and removes the concern structurally rather than arguing it is unlikely. Both the transport's `Proxy` field and `EffectiveProxy` now read a single package-level `proxyResolver`. They could not disagree before, but nothing stopped a future change to one from leaving the other behind — and for a log label whose whole job is reporting what the transport does, that is the difference between a diagnostic and a lie. Source of truth is now one line.

## 🟢 socks5 placeholder claim unpinned — fixed

Two cases added to `TestProxyLabel_StripsProxyCredentials`. The one worth having is the second: `socks5://user:password@proxy:1080` must degrade to the placeholder rather than fall back to emitting the raw URL with its password intact. Verified both before writing the test.

## 💡 `docs/observability.md:2445` — cross-reference rather than a field mention

Added a pointer to `Configuration § Outbound HTTP Proxy` rather than listing the field there. That passage is about joining the per-request and periodic lines by identifier and does not enumerate the line's fields — `auth_mode` is undocumented in that file too — so adding one field would have been both off-topic and inconsistent. Your "low-impact" read was right; this seemed the useful version of it.

## 💡 `ProxyDirect` collapsing two conditions — no change

Taking your "accepted trade-off, not a defect" as settled. The doc comment states it, and config validation makes the unparseable-URL branch unreachable in production.

## Also from the first review, still declined

The `reflect` function-identity assertion stays as filed. The `proxyLabel` split gives real behavioural coverage of the security-weighted part, so identity is only covering "is this field the env-reading function" — which is exactly what it establishes.

## Verification

`golangci-lint run` → 0 issues; coverage 89.7% against the 85% floor, `internal/semp/resilience` 90.2%; 13 tests in that package and 3 in `test/integration` green. No new log statements this round, so the logging surface is unchanged from `ddf360a`. Local `make check` otherwise green apart from the two pre-existing `test/performance/memsampler` `/proc` failures that cannot pass on macOS.

Also posted the two scope-item-4 corrections (the `NO_PROXY` CIDR rule and the OTLP third path) to SOL-153295 as a comment, so they are on the story rather than only here.

Ready for merge from my side.

bczoma (approved):
## PR Review: #428 — SOL-153295: route broker SEMP traffic through HTTP_PROXY/HTTPS_PROXY/NO_PROXY

**Author:** solace-awaheed | **Branch:** `awaheed/SOL-153295` → `main` | **Changes:** +392 / -7 across 8 files (this pass: 3 files, +26/-6)
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-428`
**Jira:** `SOL-153295`

Follow-up review: branch pushed one new commit (`033bbb8`) since my prior review at `ddf360a3` (dismissed by that push, findings below still apply). No escalation — the new commit only touches files already reviewed and adds no new file.

---

### Verdict

**Recommendation:** ✅ Approve

No conditions.

*All 5 prior 🟢/💡 notes settled — 3 addressed with real fixes, 2 accepted as-is (declined with reasons already engaged in the prior pass).*

|---|---|---|---|
| 🟢 | PR description didn't mention the new integration test file | Addressed | PR body now names `test/integration/proxy_env_invariant_test.go`, `EffectiveProxy`, the `pool.go` `proxy` field, and the two new unit tests |
| 🟢 | `EffectiveProxy` not statically tied to the transport's `Proxy` field | Addressed | `internal/semp/resilience/transport.go:34` — new package-level `var proxyResolver = http.ProxyFromEnvironment`; both `NewTunedTransport`'s `Proxy` field (line 41) and `EffectiveProxy` (line 63) now read the same symbol, so they cannot drift |
| 🟢 | socks5 placeholder claim in the `EffectiveProxy` doc comment was unpinned | Addressed | `transport_test.go:75-83` — two new cases in `TestProxyLabel_StripsProxyCredentials`, including the credential-bearing socks5 URL failing closed to `<unparseable url>` rather than leaking the password |

Summary of the other two, correctly closed without a code change: the `docs/observability.md:2445` field-mention note was resolved as a cross-reference instead (a reasoned improvement on the original ask), and the `reflect`-identity test-style note stays declined for the same reason given in the prior pass — the `proxyLabel` split already covers the security-weighted behaviour.

### Summary

This follow-up is a tidy three-file, +26/-6 commit that closes out every open minor note from the prior review with a real fix rather than an argument: the proxy resolver is now a single source of truth for both the transport and its log label, and the previously-unpinned socks5 credential-stripping claim now has a test. All CI checks are green and nothing new surfaced in this small diff.

---

### Minor notes

None — the diff since the prior review does not introduce anything new.

---

### Summary Verdict

|---|---|---|
| Architecture | — not re-reviewed (unchanged since ddf360a3) | — |
| Security | ✅ | parent (follow-up) |

PR #432, opened 2026-09-17
SOL-154545: verify the published Grafana dashboard and collector metrics pipeline nightly, on both ingestion paths

solace-lootii:
## Summary

Adds a second nightly job, `e2e-dashboard`, alongside `e2e-tracing` in `.github/workflows/observability.yml`: it imports the real `deploy/grafana/solace-broker-mcp-overview.json` into a live Grafana, drives real tool traffic against a live broker, and asserts every panel query, every template variable, and an exemplar all return real data — on both metrics ingestion paths. On the OTLP path it uncomments and runs the actual, committed `otlphttp/prometheus` exporter and `metrics:` pipeline in `deploy/otel-collector/docker/otelcol.yaml`, so that published config is proven to work, not just published. Closes SOL-154545.

## Type of Change

- [x] Documentation update

## Motivation and Context

SOL-152092 shipped the dashboard and collector config; the only thing verifying either was a static schema test (`cmd/server/grafana_dashboard_test.go`, proves names/labels agree on paper, never runs a query) and a one-time-by-hand manual verification during that PR's review. Both real defects that PR found were on the OTLP path, and both were found manually — nothing would catch a metric rename that silently blanks a panel, or a collector config that quietly stops working, until an operator hits it.

## Changes Made

- `test/e2e-dashboard/` — new suite, mirroring `test/e2e-tracing/`'s minimal shape:
  - `docker-compose.yml` — one broker, one MCP server (both metrics egresses + tracing on at once — independent flags, confirmed in SOL-152092), a scrape-path Prometheus (`--enable-feature=exemplar-storage`), an OTLP-ingesting Prometheus (`--web.enable-otlp-receiver`), the real collector config (uncommented at test time), and Grafana (datasource provisioned by file; the dashboard itself imported via the API in `run-all.sh`, since file-based provisioning doesn't resolve the dashboard JSON's `${DS_PROMETHEUS}` templating — only the `/api/dashboards/import` endpoint does).
  - `uncomment-metrics-pipeline.sh` — takes the real, committed `otelcol.yaml` and uncomments exactly the `otlphttp/prometheus` exporter and `metrics:` pipeline, asserting each expected line is found so a future reword of the file fails this script loudly instead of silently no-op'ing.
  - `verify-panels.py` — parses the dashboard JSON at run time (never a hardcoded panel list, so Stories 17/18/28/29 landing later are covered for free), substitutes Grafana's template variables with a wildcard, and POSTs every panel query straight to Prometheus's `/api/v1/query`; also checks each template variable via `/api/v1/label/<label>/values` and, optionally, an exemplar via `/api/v1/query_exemplars`.
  - `run-all.sh` — orchestrates both paths in one lifecycle: drive traffic (including a genuine SEMP-level error, not just a client-side one — see notes below), verify Path A, import into Grafana, verify Path B with `go_*`/`process_*` expected empty (the documented structural gap).
- `.github/workflows/observability.yml` — the new `e2e-dashboard` job, same trigger/guard/teardown shape as `e2e-tracing`.
- `docs/observability.md` — states which published artifacts the nightly covers, in both the Grafana Dashboard section and the collector-config section.

## Testing

**Test Environment:**
- Go version: 1.25.0
- OS: Linux
- Broker version: `solace/solace-pubsub-standard:latest`

**Tests Performed:**
- [x] E2E tests added

**Test Coverage:**
- Ran the full suite twice against the real dev image: once iterating interactively, once replaying the exact CI invocation (`docker compose -f test/e2e-dashboard/docker-compose.yml up -d` from repo root, `working-directory: test/e2e-dashboard` for the uncomment step and `run-all.sh`). Both fully green on both paths.
- Sanity-checked both directions per the ticket's own AC: injected a bad metric name into a scratch copy of the dashboard JSON and confirmed the suite goes red; reverted the `$service_name` template variable to its old, pre-SOL-152092 broken form (`label_values(target_info, service_name)`) and confirmed it goes red specifically on the OTLP path (`target_info` has no `service_name` label there); restored both, confirmed green.
- Found and fixed two real bugs during verification, distinct from the design itself:
  1. The traffic-readiness wait loop checked for the raw counter's presence, not whether `rate()` could compute over it — `rate()` needs the same series scraped twice, so the error-outcome panels intermittently failed even though the underlying data existed. Fixed by polling the actual `rate()` queries instead.
  2. My first attempt at a deliberate SEMP-level error (`list-queues` against a nonexistent VPN) returned `200` with an empty array — ordinary REST collection-endpoint behavior, not an error — so it never populated the SEMP error-rate panel. Switched to `get-vpn-status` (a single-resource `GET`), confirmed directly against the broker's SEMP API that it returns `400` for a VPN that doesn't exist.
- Full repo `go build ./...`, `go vet ./...`, `go test ./...` all green (docs-only change on the Go side; confirmed `TestObservabilityDocMatchesRegistry` isn't broken by the doc edits).

### Code Quality

### Security

### Testing

### Documentation

### Git & CI
- [x] Branch is up to date with main (rebased/merged twice already — two more PRs landed in this exact area, `docs/observability.md`, while this one was open)

## Additional Notes

Out of scope, per the ticket: Grafana-side exemplar navigation (asserting the exemplar is present in the scrape is the cheap check that catches the regression that matters) and `deploy/kubernetes/` manifests (needs a cluster, not compose — its own story).

## Related Issues/PRs

- Closes SOL-154545
- Follows SOL-152092 (Story 37, #424) and SOL-152423 (Story 40, #412)
- Related SOL-154238


.github/workflows/observability.yml
```diff
@@ -69,3 +78,60 @@ jobs:
...
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      # Uncomments the real, committed otlphttp/prometheus exporter and
+      # metrics: pipeline in deploy/otel-collector/docker/otelcol.yaml — the
+      # point of this suite's Path B is that the published config is what
+      # runs, not a hand-maintained copy of it.
+      - name: Uncomment the collector's metrics pipeline
+        run: bash test/e2e-dashboard/uncomment-metrics-pipeline.sh
```
bczoma: 🟡 **Doubled script path — the new job's first real step can never run**

Flagged by Copilot, independently re-verified against the checked-out code.

Both new steps set `working-directory: test/e2e-dashboard` while their `run:` command also prefixes the script path with `test/e2e-dashboard/`:

```yaml
      - name: Uncomment the collector's metrics pipeline
        run: bash test/e2e-dashboard/uncomment-metrics-pipeline.sh
        working-directory: test/e2e-dashboard

      - name: Run dashboard E2E tests
        run: bash test/e2e-dashboard/run-all.sh
        working-directory: test/e2e-dashboard
```

GitHub Actions resolves the `run:` path against `working-directory`, so the actual attempted path is `test/e2e-dashboard/test/e2e-dashboard/uncomment-metrics-pipeline.sh`, which doesn't exist (confirmed: no such directory in the checkout, no symlink).

**Measured risk (via adversarial challenge):** the job's guard (`workflow_dispatch` or `workflow_run.conclusion == 'success'`) means this fires on every scheduled nightly run from merge onward — not just on manual dispatch. The sibling `e2e-tracing` job doesn't use `working-directory` at all and invokes its script correctly from repo root, so this is a novel mistake in the new job, not an existing broken pattern being copied. Failure is immediate (exit 127 at step 2), fully isolated to this one job (no effect on `e2e-tracing`, other CI jobs, or PR merges), and loudly visible as a red job in the Actions tab on the very next run — this repo's own actionlint/zizmor lint gate doesn't catch it since it's a semantic, not syntactic, path error. Net: real and will bite on every run until fixed, but confined, immediate, and non-blocking to merge — which is why this sits at 🟡 rather than 🔴.

**Fix:** drop `working-directory: test/e2e-dashboard` from both steps (the `run:` commands already carry the full relative path from repo root), matching the convention `e2e-tracing` already uses correctly.

test/e2e-dashboard/docker-compose.yml
```diff
@@ -0,0 +1,125 @@
+services:
+  solace:
+    image: solace/solace-pubsub-standard:latest
```
bczoma, line 3: 🟡 **Unpinned broker image, inconsistent with the rest of this file**

Flagged by good:security.

`solace/solace-pubsub-standard:latest` is unpinned, while the other three images in this same file are pinned (`prom/prometheus:v3.0.1`, `otel/opentelemetry-collector-contrib:0.111.0`, `grafana/grafana:9.5.20`).

An upstream `:latest` push can silently change broker behavior or CVE surface under a nightly job with no code change to review, and breaks reproducibility of a security-relevant component (the SEMP auth path this suite exercises).

**Fix:** pin to a specific broker version tag, consistent with the other images in this file.

test/e2e-dashboard/docker-compose.yml
```diff
@@ -0,0 +1,125 @@
...
+  # at OTEL_BACKEND_ENDPOINT (default: an address nothing listens on in this
+  # suite); that pipeline's export failures are harmless here — no Tempo is
+  # needed for the exemplar check (see run-all.sh) — and do not affect the
+  # independent metrics pipeline.
+  otel-collector:
+    image: otel/opentelemetry-collector-contrib:0.111.0
+    container_name: otel-collector-e2e-dashboard
+    command: ["--config=/etc/otelcol-contrib/config.yaml"]
+    environment:
+      - OTEL_BACKEND_ENDPOINT=${OTEL_BACKEND_ENDPOINT:-tempo-not-run-in-this-suite:4317}
```
bczoma, line 108: 🟡 **Dead traces pipeline will add export-failure noise to every failure diagnosis**

Flagged by good:architecture.

The collector's traces pipeline is left pointed at `OTEL_BACKEND_ENDPOINT=tempo-not-run-in-this-suite:4317`, a host nothing resolves, while only the metrics pipeline is actually exercised. The nearby comment acknowledges the resulting export failures are "harmless" to the metrics assertions.

Harmless to the assertions, but not to operability: the `Collector logs (on failure)` diagnostic step in `observability.yml` exists so an engineer can triage a red nightly run, and it will show a stream of connection-refused/export-failure noise for the traces pipeline on *every* failure, unrelated to whatever actually broke — noise that has to be mentally filtered every time this job goes red.

**Fix:** either point `OTEL_BACKEND_ENDPOINT` at a stub/no-op receiver that accepts and drops (so the pipeline is quiescent instead of erroring), or add a one-line note directly above the "Collector logs" workflow step so the first person triaging a red run isn't left to rediscover this via source-diving.

test/e2e-dashboard/run-all.sh
```diff
@@ -0,0 +1,201 @@
...
+# Polls the actual rate() form every panel query uses, not just the raw
+# counter's presence: rate() needs the SAME series scraped at least twice to
+# compute anything, so a wait that only checks "has this metric appeared at
+# all" can break as soon as the success series' first sample lands, well
+# before the error series (recorded in the same batch, a moment later) has
+# had its own second scrape — which is exactly what happened the first time
+# this suite ran end to end: the error-outcome panels intermittently failed
+# with "expected non-empty data, got none" even though the raw counter was
+# already visible.
+prom_rate_nonempty() {
```
bczoma, line 99: 🟡 **Duplicated wait-for-rate() function**

Flagged by good:code.

`prom_rate_nonempty` (here) and `prom_otlp_rate_nonempty` (line 173) are byte-for-byte identical function bodies, differing only in which URL variable the caller closes over.

A future fix to the rate-wait logic (e.g. a third series check, or hardening the `python3 -c` error path) has to be made twice and will drift.

**Fix:** collapse to one `prom_rate_nonempty(url, query)` taking the Prometheus base URL as a parameter, called once per path with the right URL.

test/e2e-dashboard/verify-panels.py
```diff
@@ -0,0 +1,226 @@
...
+Not a PromQL parser — a regex over each target's `expr` string, exactly like
+cmd/server/grafana_dashboard_test.go's approach, proportionate to what this
+script needs (does the panel return data?), not full query analysis.
+"""
+import argparse
+import json
+import re
+import sys
+import time
+import urllib.parse
```
bczoma, line 30: 🟡 **Metric/template-variable parsing rules duplicated with the Go static test, no shared source of truth**

Flagged by good:architecture.

`METRIC_NAME_RE` here and the `label_values(...)` regex at line 68 are hand-copied, character-for-character, from `metricNameRe`/`labelValuesRe` in `cmd/server/grafana_dashboard_test.go` — two independent re-implementations of "how to parse a dashboard JSON's panel/template-variable shape," in two languages.

The PR's own stated goal is that new panels (Stories 17/18/28/29) get covered "for free" because this script reads the dashboard JSON at runtime — but that only holds if the two regex sets stay in lockstep. A future edit to the Go test's parsing rules (a new template-variable form, a new metric-name family) won't propagate here, and the two suites can drift apart while both report green, disagreeing about what a "supported" panel or variable looks like.

**Fix:** at minimum, add a comment in both files cross-referencing the other and asserting they must be kept in sync; better, have the Go test emit the parsed panel/variable list as a JSON fixture in CI that `verify-panels.py` consumes, so there's one parser, not two.

test/e2e-dashboard/verify-panels.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def prometheus_query_exemplars(base_url: str, query: str, lookback_seconds: int = 600):
+    now = time.time()
+    params = {"query": query, "start": now - lookback_seconds, "end": now}
+    url = f"{base_url}/api/v1/query_exemplars?" + urllib.parse.urlencode(params)
+    return http_get_json(url)
+
+
+def expects_empty(expr: str, empty_prefixes) -> bool:
```
bczoma, line 143: 🟡 **`expects_empty()` can misclassify a mixed-metric panel**

Flagged by good:code.

`expects_empty()` returns true if *any* metric name found anywhere in the expression matches an empty-prefix, even when the same expression also references a real `mcp_*` metric (e.g. a future ratio/combination panel mixing `mcp_*` and `go_*`/`process_*`).

If such a mixed panel is ever added, the whole panel would be silently reclassified as "must return nothing," masking a real regression in the `mcp_*` half of the query instead of failing on it — exactly the kind of silent-pass this suite exists to prevent.

**Fix:** only classify a panel as want-empty when *every* metric name in it is a Go/process metric, or split mixed panels into per-metric-name expectations rather than one whole-expression flag.

bczoma (dismissed):
## PR Review: #432 — SOL-154545: verify the published Grafana dashboard and collector metrics pipeline nightly, on both ingestion paths

**Author:** solace-lootii | **Branch:** `lkiri/sol-154545-nightly-dashboard-verification` → `main` | **Changes:** +750 / -8 across 10 files
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-432`
**Jira:** `SOL-154545` — Verify the published observability surface nightly, on both metrics paths (Verifying)

---

### Verdict

1. 🟡 `.github/workflows/observability.yml:99,106` — `working-directory: test/e2e-dashboard` doubles up with the already-prefixed script paths, so both new steps resolve to a nonexistent file and the job fails on step 2, every run
2. 🟡 `test/e2e-dashboard/docker-compose.yml:3` — pin `solace/solace-pubsub-standard:latest`, like the other three images in this file
3. 🟡 `test/e2e-dashboard/docker-compose.yml:101` — point the collector's dead traces pipeline at a quiescent stub, or note next to the failure-log step that its export-failure noise is expected
4. 🟡 `test/e2e-dashboard/run-all.sh:100,173` — `prom_rate_nonempty` and `prom_otlp_rate_nonempty` are byte-identical; collapse to one parameterized function
5. 🟡 `test/e2e-dashboard/verify-panels.py:20-33` — the metric-name/template-variable regexes duplicate `cmd/server/grafana_dashboard_test.go`'s parsing rules with no shared source of truth
6. 🟡 `test/e2e-dashboard/verify-panels.py:102-106` — `expects_empty()` can misclassify a mixed-metric panel as want-empty if it ever references both a `go_*`/`process_*` metric and a real `mcp_*` metric
7. 🟡 `test/e2e-dashboard/verify-panels.py` — no unit tests for the script's own parsing/classification logic; only exercised indirectly by the live nightly run

- 🟡 `THIRD_PARTY_BUILD_TEST.md` — this PR pulls `prom/prometheus:v3.0.1` (two containers), which the Container Images table doesn't list; the `otel/opentelemetry-collector-contrib` row's "Used for" text also doesn't yet mention `e2e-dashboard`. This is the likely cause of the currently-failing "Third-party inventory refresh scripts self-test" and "Third-party licenses current" checks on this PR.

**Also still open, from Copilot's existing review (not restated here as new threads — already tracked as unresolved conversations under this repo's `required_review_thread_resolution` rule):** `run-all.sh:139` imports into Grafana with no readiness wait; `verify-panels.py:80` doesn't check Prometheus's API-level `status` field; `verify-panels.py:73` silently skips template variables whose query isn't exactly `label_values(...)`. All three were independently confirmed still present in the current code.

---

### Summary

This PR adds a nightly `e2e-dashboard` CI job that imports the real Grafana dashboard into a live Grafana, drives real broker/MCP traffic, and verifies every panel query, template variable, and an exemplar return real data on both the Prometheus-scrape and OTLP-via-collector ingestion paths — closing a real gap where only a static schema test and one manual check ever verified the dashboard SOL-152092 shipped. The design is sound (schema-driven panel discovery, genuine separation of the two ingestion paths, uncommenting the real committed collector config rather than a hand copy) and the PR description is a model of what a strong writeup for this kind of work looks like, including two real bugs the author found and fixed during their own verification. The blocking-looking issue — a doubled script path that would fail the new job's very first real step — was filed as critical and reduced under challenge to medium: it fires on every scheduled run and is immediately visible as a red job in the Actions tab, isolated to this one new job, with zero effect on merges, other CI jobs, or production.

---

### Minor notes

- 🟢 `test/e2e-dashboard/uncomment-metrics-pipeline.sh:18-19` — `SRC` resolves via `BASH_SOURCE[0]`, `OUT` resolves relative to CWD instead; harmless today (always invoked from `test/e2e-dashboard/`) but inconsistent — resolve `OUT` the same way for future-proofing.
- 🟢 `test/e2e-dashboard/docker-compose.yml:96` — `GF_AUTH_ANONYMOUS_ENABLED=false` set explicitly even in this throwaway CI instance.
- 💡 `test/e2e-dashboard/run-all.sh:42-46,128-137` — a few `curl`/`python3 -c` calls interpolate bash variables straight into Python source rather than via `sys.argv`; safe today since the values are fixed constants, but `uncomment-metrics-pipeline.sh` already shows the safer pattern (`sys.argv`) — worth matching for consistency.

---

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

test/e2e-dashboard/verify-panels.py
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/env python3
```
bczoma, line 1: 🟡 **No unit tests for this script's own parsing/decision logic**

Flagged by good:code.

`load_targets`, `load_template_vars`, `substitute_variables`, and `expects_empty` are nontrivial (recursive row-walking, a metric-name regex, prefix-matching semantics) and drive pass/fail for the whole nightly suite, but are exercised only indirectly by the live nightly run against real infra.

A bug here (such as the mixed-metric issue flagged elsewhere in this review) can currently only surface via a red nightly run against live infra — slow feedback that's easy to misattribute to the environment rather than the checker.

**Fix:** add a small `test/e2e-dashboard/test_verify_panels.py` (or similar) that feeds synthetic dashboard JSON / expressions through these pure functions and asserts the expected output — no live Prometheus/Grafana needed.

bczoma (commented):
Follow-up to my review above — one finding needed its anchor demoted (a `subject_type: "file"` comment was rejected by this endpoint), posting it here as a line-anchored thread instead so it isn't lost:

.github/workflows/observability.yml
```diff
@@ -69,3 +78,60 @@ jobs:
...
+        working-directory: test/e2e-dashboard
```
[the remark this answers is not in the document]
solace-lootii: Confirmed — thanks for catching this. Kept `working-directory: test/e2e-dashboard` (needed for both scripts' own internal relative paths, e.g. `run-all.sh`'s `../../deploy/grafana/...` and `python3 verify-panels.py`) and dropped the redundant `test/e2e-dashboard/` prefix from both `run:` commands instead. Also hardened `uncomment-metrics-pipeline.sh`'s `OUT` path to resolve via `$BASH_SOURCE` rather than CWD, so this stays correct regardless of `working-directory`. Verified by replaying the exact CI invocation locally (checkout-root `docker compose -f test/e2e-dashboard/docker-compose.yml`, `working-directory: test/e2e-dashboard` for both scripts) end to end, fully green on both paths. See 3ec4a6b.

.github/workflows/observability.yml
solace-lootii: Confirmed independently. Fixed by dropping the redundant `test/e2e-dashboard/` prefix from both `run:` commands (kept `working-directory`, since both scripts rely on it for their own internal relative paths). Also hardened `uncomment-metrics-pipeline.sh`'s `OUT` to resolve via `$BASH_SOURCE` instead of CWD (your own minor note elsewhere), so the fix holds regardless of how `working-directory` is set in the future. Re-verified end to end by replaying the exact CI invocation locally — both paths fully green. See 3ec4a6b.

test/e2e-dashboard/docker-compose.yml
solace-lootii: Pushing back on this one rather than changing it. Checked all 8 sibling E2E suites — every single one uses `solace/solace-pubsub-standard:latest` for the broker fixture, with zero exceptions (`e2e-basic-mcp`, `e2e-action`, `e2e-llm`, `e2e-management`, `e2e-monitoring`, `e2e-oauth`, `e2e-tracing`, and now this one). It's a real, established, repo-wide convention for this specific image, not an oversight this file introduced — pinning only my new suite would make it MORE inconsistent with the codebase as a whole (matching the other 3 images within this one file, but diverging from all 8 sibling suites on the broker image specifically). Your underlying concern (reproducibility risk from `:latest`) is legitimate in the abstract, but changing it unilaterally in this PR, for this suite only, is the wrong scope for that fix — if the team wants the broker fixture pinned, that's a repo-wide change across all 8 suites, worth its own discussion rather than a one-off here. Happy to raise it separately if you think it's worth pursuing.

test/e2e-dashboard/docker-compose.yml
solace-lootii: Confirmed. Went with your second option — added a one-line note directly above the "Collector logs (on failure)" workflow step explaining the traces-pipeline noise is expected and unrelated. Didn't touch the traces pipeline itself (e.g. swap its exporter to a no-op): that would mean the uncomment step also has to touch lines outside the metrics-related block it's designed to touch minimally, which trades a small amount of log noise for a bigger change to a script whose whole point is 'touch only what's needed, leave the rest of the published file exactly as-is.' See 3ec4a6b.

test/e2e-dashboard/run-all.sh
solace-lootii: Confirmed. Collapsed to one `prom_rate_nonempty(url, query)`, called with the right URL at each of the two call sites. See 3ec4a6b.

test/e2e-dashboard/verify-panels.py
solace-lootii: Agreed on the underlying risk, going with the 'at minimum' fix rather than the shared-fixture pipeline: added explicit 'KEEP IN SYNC' cross-referencing comments in both files (metricNameRe/labelValuesRe here, METRIC_NAME_RE/LABEL_VALUES_RE there), each pointing at the other and stating what has to be mirrored. Didn't build the shared JSON-fixture approach — that's a real architecture change (a new Go emission mode, CI wiring to generate and consume it, a new failure mode if that generation step itself breaks) that deserves its own design/review rather than being folded into this response, and the two parsers are simple enough today (a handful of regexes) that comment-enforced sync is a proportionate stopgap. Open to revisiting if you'd still want the fixture approach on reflection. See 3ec4a6b.

test/e2e-dashboard/verify-panels.py
solace-lootii: Confirmed, and it was a real latent bug, not just theoretical — verified by testing: `expects_empty` now excludes `target_info` (the join partner present on nearly every panel, not a metric being measured) and requires ALL remaining metric names to match an empty prefix, not ANY. Added unit tests for exactly this: a pure go_* metric (want-empty), the same metric joined to target_info (still want-empty — this is the case that would have broken with a naive any-to-all flip, since target_info doesn't match either prefix), a real mcp_* metric (never want-empty), and the actual regression case — a hypothetical panel mixing mcp_* and go_* (must NOT be want-empty). See 3ec4a6b.

test/e2e-dashboard/verify-panels.py
solace-lootii: Added `test/e2e-dashboard/test_verify_panels.py` — 11 tests over `load_targets` (flat and row-nested panels), `load_template_vars` (including the unrecognized-query-form case from your other thread), `substitute_variables`, and `expects_empty` (including the mixed-metric regression case from your other thread). No live infra needed, wired into the workflow as its own fast-failing step before the stack comes up. Also caught and fixed a small independent issue while writing these: `write_dashboard`'s temp files were never cleaned up (`delete=False` with no corresponding unlink) — now registered via `addCleanup`. See 3ec4a6b.

solace-lootii:
Thanks for the thorough review — went through every item, verified each against the actual code before touching anything, then had it re-checked by an independent pass before pushing. Summary of what's not covered by the inline replies:

**CI failures (Third-party inventory / licenses) — fixed.** `make refresh-third-party-inventory` found exactly what it should: `grafana/grafana` (9.5.20) and `prom/prometheus` (v3.0.1) were genuinely unlisted — added both, with licenses verified via `gh api repos/<owner>/<repo> --jq .license.spdx_id` rather than assumed (AGPL-3.0 and Apache-2.0 respectively). Also found a real cause behind the check misfiring on our own image: `${MCP_IMAGE:-ghcr.io/solaceproducts/solace-broker-mcp-dev:dev}` defeats the checker's own-image exclusion regex (`IMAGE_OURS_RE` matches a string starting with `ghcr.io/solace...`, which the `${MCP_IMAGE:-...}` wrapper doesn't). Switched to the bare image reference, matching `e2e-tracing`'s own convention — this suite is CI-only and never needed the override flexibility anyway. `make refresh-third-party-inventory` is clean now.

**The three items you carried forward from the dismissed review (no separate inline thread on this round, so replying here):** all three independently confirmed and fixed —
- Grafana readiness: added an explicit `/api/health` poll before the import call, rather than relying on the ~60s of Path A processing that happened to already elapse by that point.
- Prometheus status field: `http_get_json` now checks `status != "success"` and raises with the actual error message, so a malformed query surfaces as itself instead of a misleading "expected non-empty data, got none."
- Silently-skipped template variables: `load_template_vars` now returns `(checkable, unrecognized)` instead of dropping anything that isn't `label_values(...)`; `unrecognized` is reported as a failure in `main()`, not silently ignored.

All fixes re-verified live: ran the full suite twice more (once interactively, once replaying the exact CI invocation), re-confirmed both required sanity-check directions still fail correctly against the updated code (bad metric name; reverted `$service_name` query on the OTLP path), and live-triggered the new Prometheus status check with a real malformed query to confirm it reports honestly rather than masquerading as "no data." Repo-wide `go build`/`go vet`/`go test ./...` and the new Python unit tests all green.

bczoma (approved):
## PR Review: #432 — SOL-154545: verify the published Grafana dashboard and collector metrics pipeline nightly, on both ingestion paths

**Author:** solace-lootii | **Branch:** `lkiri/sol-154545-nightly-dashboard-verification` → `main`
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-432`
**Jira:** `SOL-154545` — Verify the published observability surface nightly, on both metrics paths

---

### Verdict

**Recommendation:** ✅ Approve

Follow-up review: incremental diff since `bb9cebc5` (recovered by patch-id — one authored commit, `3ec4a6bf`, since the base-branch merges dragged in unrelated history). No lenses dispatched — no new files beyond a test module, and no auth/crypto/API/config surface touched, so this stayed a parent-only pass per the escalation rule.

---

### Summary

All 7 conditioned findings and the 1 not-anchorable finding from my prior review are addressed in `3ec4a6bf`, each independently re-verified against the current code: the doubled script path is gone, the duplicated wait function is collapsed, `expects_empty()` now excludes `target_info` and is unit-tested against the exact mixed-metric regression it was written to catch, and the CI-failing third-party inventory gaps (`grafana/grafana`, `prom/prometheus`) are filled in with verified license IDs. The three items carried over from Copilot's review (Grafana readiness poll, Prometheus status-field check, silently-skipped template variables) are also fixed and confirmed in code. One finding (the unpinned broker image) is dropped on its merits — the author's pushback is factually correct.

*11 prior findings settled — 10 addressed, 1 accepted (author's counter-argument is correct).*

|---|---|---|---|
| 🟡 | Unpinned `solace/solace-pubsub-standard:latest` in docker-compose.yml, "inconsistent with the rest of this file" | Accepted as-is | `test/e2e-dashboard/docker-compose.yml:3` — verified all 8 sibling E2E suites (`e2e-action`, `e2e-monitoring`, `e2e-basic-mcp`, `e2e-llm`, `e2e-tracing`, `e2e-oauth`, `e2e-management`) use the identical unpinned tag for the broker fixture with zero exceptions; this is the repo's deliberate convention, not an inconsistency. The author's counter-argument is better than the original finding — dropped. |

New work reviewed: `test/e2e-dashboard/test_verify_panels.py` (new, +159 lines) — 11 unit tests covering `load_targets`, `load_template_vars`, `substitute_variables`, and `expects_empty`, including the exact mixed-metric and unrecognized-template-var regressions the prior review's findings were about. No issues found.

---

### Summary Verdict

|---|---|---|
| Architecture | — not re-reviewed (unchanged since `bb9cebc5`) | |
| Security | — not re-reviewed (unchanged since `bb9cebc5`) | |
| Documentation | — not re-reviewed (unchanged since `bb9cebc5`) | |

PR #437, opened 2026-09-18
SOL-153071: fix list-vpns zeroConnectionCount false positives

McCly-Solace:
## Summary

- `list-vpns`'s `zeroConnectionCount` misreported every enabled+up VPN as having zero connections. The `real-clients` probe (`internal/composite/definitions/tools.yaml`) scanned only `count=1` client per VPN before applying the `clientUsername!=#*` filter — reserved (`#*`) clients sort first, so the probe never scanned far enough to reach a real client behind them. Raised to `count=100`; `followPages` deliberately left off since `maxResults` is shared across every step in the tool call (documented inline).
- The E2E regression check was self-referential — it recomputed the expected value from the same `real-clients.byKey` output the handler consumed, so a uniformly-empty probe satisfied both sides. Reworked it to ground against an independent SEMP query, and added assertions proving the fixture genuinely reproduces the bug shape: with F8's bridges and F9/F10's Kafka Receivers/Senders active alongside F3's real client, the default VPN carries >=2 reserved clients ahead of the real one, and a deliberately small `count=2` query undercounts relative to ground truth.
- Updated `README.md`/`helpers.sh` comments to match, and added a `CHANGELOG.md` entry.

## Test plan

- [x] `make e2e-monitoring-all` — 22/22 fixture checks, 102/102 tool tests passing, including the strengthened `list-vpns summary` regression tests on both brokers.
- [x] `/check-logs staged` — no `.go` files in this change set, clean.


bczoma (commented):
(internal tooling test — please disregard; superseded by the full review below)

internal/composite/definitions/tools.yaml
```diff
@@ -315,7 +326,7 @@ tools:
         forEachKey: msgVpnName
         args:
           msgVpnName: "{{.Item.msgVpnName}}"
-          count: "1"
+          count: "100"
```
bczoma, line 329: 🟡 **A magic-number workaround for a real framework gap**

**Flagged by:** good:architecture

**What the problem is:** The fix raises `count` from `"1"` to `"100"` — a hardcoded scan bound baked into one YAML step's `args`, invisible to the loader/validator as a "cap" concept. `Step` (internal/composite/definition.go:61-73) has `FollowPages`, `ForEach`, `Concurrency`, but no first-class "stop after N items" field. The PR's own comment correctly explains why `followPages` isn't used here (it would couple this probe's client-scan depth to the shared `maxResults` parameter that also controls how many VPNs `list-vpns` returns — verified against `resolveMaxResults`/`fetchPaginated` in executor.go, which this step never reaches since `FollowPages` is unset).

**Why it matters:** Nothing stops the next probe that needs "stop after N items, independent of maxResults" from either copying this ad-hoc `count` trick with a different hardcoded value, or reaching for `followPages` out of habit and walking straight into the exact shared-maxResults coupling this PR's comment warns about. There's no structural signal distinguishing "this count is a hard scan bound chosen for correctness" from "this count is an ordinary page-size hint."

**Recommended fix:** Not blocking for this fix — the `count=100` + no-`followPages` choice is correct and well-documented for this one probe. Worth a tracked follow-up (or an explicit note on SOL-153071 or a new ticket) to add a named per-step cap field (e.g. `Step.ItemCap`) so future probes get a first-class, validated mechanism instead of another inline-comment convention. Reversibility is good: migrating this probe to such a mechanism later would be a localized change with no ripple into `fetchFanOut`/`runSingle`/the postprocess handler.

bczoma (approved):
## PR Review: #437 — SOL-153071: fix list-vpns zeroConnectionCount false positives

**Author:** McCly-Solace | **Branch:** `sol-153071-zero-connection-count-fix` → `main` | **Changes:** +135 / -11 across 6 files
**Worktree:** `~/.pr-review-worktrees/SolaceProducts-solace-broker-mcp/pr-437`
**Jira:** `SOL-153071` — list-vpns: zeroConnectionCount always reports every enabled+up VPN as having zero client connections (Verifying)

---

### Verdict

Conditioned on these 2 — the first has an inline thread on the line in question:

1. 🟡 `internal/composite/definitions/tools.yaml:329` — the `count: "100"` bound is a magic number working around a real framework gap (no per-step item cap independent of the shared `maxResults`); worth a tracked follow-up so the next probe doesn't reach for `followPages` out of habit.
2. 🟡 `internal/composite/definitions/tools.yaml:218` (`list-vpns` description) — the tool's user-facing description doesn't mention the probe is a bounded scan (100 clients) that can still false-negative on a VPN with more reserved clients ahead of its first real one.

- 🟡 `internal/composite/definitions/tools.yaml:218` (`list-vpns` description, unchanged by this diff, so this repo's review API rejected a file-level anchor on it) — **flagged by docs.** The description doesn't mention the probe is a bounded scan (up to 100 clients per VPN) or carry a hedge like `get-vpn-status`'s existing "a VPN with zero active connections may be idle by design, not broken" (line 199). An LLM/agent consuming `list-vpns` output has no way to know `zeroConnectionCount` is a heuristic with a known edge case — it could over-trust a false "broken" read, or miss a genuinely non-empty VPN with >100 reserved clients ahead of its first real one. Recommended fix: add a clause noting the bounded-probe caveat to the description.

The notes below are **not** conditions of this verdict.

### Summary

This PR fixes a real, well-diagnosed production bug: the `real-clients` SEMP probe in `tools.yaml` scanned only 1 client per VPN, and since `#`-prefixed reserved clients sort first, every enabled+up VPN with a reserved client was misreported as zero-connection. The fix (raise the scan bound to 100, explicitly without `followPages` to avoid coupling to the shared `maxResults` parameter) is verified correct against the actual `executor.go` fan-out/pagination code, and the new e2e regression test grounds itself in an independent SEMP call rather than re-deriving from the same probe it's validating — directly closing the gap that let the original bug ship undetected. One convergent finding (from `good:pr`, `good:code`, and an earlier Copilot comment) about a specific test assertion being weaker than its own comment claims was investigated under challenge and found to target a fixture-validity canary, not the actual regression gate (a separate, independent assertion at line ~208 exercises the real production probe and would catch reintroduction of the bug) — that finding is downgraded to a minor precision suggestion below.

### Minor notes

🟢 **Minor** — worth fixing, not a condition of this verdict:

- 🟢 PR-level — the PR body is empty. The rationale (probe semantics, the maxResults-coupling tradeoff, why the default VPN is reused instead of a new fixture VPN, the accepted >100-reserved-clients residual risk) is unusually well documented in code comments and the Jira ticket, but a reviewer has to reconstruct it from there. Add a short description summarizing the fix and the accepted residual limitation.
- 🟢 `CHANGELOG.md:141-144` — the entry states the miscount "no longer" happens without noting the fix is still a bounded scan (100 clients), not exhaustive — a VPN with >100 reserved clients ahead of its first real one can still misreport zero-connection. Add one clause noting the residual limitation.
- 🟢 `internal/composite/definitions/tools.yaml:270-320` — the comment block for the `real-clients` step has grown to ~45 lines with some repeated phrasing across tools.yaml/helpers.sh/verify-fixtures.sh (the "earlier design provisioned a separate VPN" narrative is retold in three files). Consider trimming to the load-bearing facts and keeping the historical narrative in one place.
- 💡 `test/e2e-monitoring/test-monitoring-tools.sh:178-182` — this fixture-validity canary asserts `small_real < real_count`; the comment above it already establishes the stronger fact `small_real == 0` (lab-measured). Tightening to `[ "$small_real" -ne 0 ]` costs nothing and matches the comment's own claim. Note: this is not the actual regression guard — that's the independent `byKey` assertion a few lines below, which exercises the real probe and would catch reintroduction of SOL-153071 on its own.

### Summary Verdict

|---|---|---|
| Documentation | ⚠️ | docs |

internal/composite/definitions/tools.yaml
bczoma, line 329: 🔴 **The probe's raw rows now ship in the tool response, not just the boolean the handler reads**

*Flagged by good:architecture and good:security independently, then verified against the executor source.*

**What the problem is.** The comment eleven lines up says `count=100 is still a boolean check, not an enumeration`. That is true of how the handler uses it — `hasRealClient` only tests `len(data) > 0` — and false of what leaves the server. Under `strategy: postProcess`, `ApplyResultStrategy` runs `postprocess.Apply(...)` for the summary and then `out := collectSteps(stepResults, 1); out["summary"] = summary`. `collectSteps` copies **every** step result verbatim, keyed by step ID. There is no allowlist, no exclusion set, and no "internal step" concept on `Step` — no `collect`/`emit`/`internal`/`hidden` field exists. `ListVpns` reads `byKey` and never deletes or truncates it. `internal/composite/schema.go` models the `byKey` envelope in the declared output schema, so the raw block is contractual, not incidental.

So `real-clients.byKey` is a top-level key of every `list-vpns` response, and this one-token change alters what it contains.

**Why it matters.** Because `where=clientUsername!=#*` is applied *after* the page cut — the premise of the bug being fixed — the old `count: "1"` scanned one object, which reserved-name ordering almost always filtered out. In practice the pre-fix block held ~0 rows per VPN. Post-fix it holds the real clients among the first 100 scanned: on a quiet VPN a handful, on a busy one ~95-100. Summed across enabled+up VPNs with `msgVpnConnections != 0`, a production-shaped broker lands at roughly 1,000-5,000 `clientName` strings where it previously had almost none; at an explicit `maxResults: 500` the ceiling is ~50,000. `buildValidatedResult` then emits the payload twice — once as `StructuredContent`, once as an indented-JSON `TextContent` fallback — so the wire and token cost is roughly doubled.

`list-vpns` is a broad survey tool an agent reaches for early. It now returns a client inventory nobody asked for, and the 5-integer summary it exists to produce is dwarfed by it. On the disclosure side the honest counterweight is that `list-clients` already exposes `clientName` to the same caller under the same authorization, so no privilege boundary is crossed and no new field type appears — what is new is scope and consent: broker-wide client identities, unrequested, from a survey call, where the previous behaviour disclosed essentially nothing.

**Measured, not asserted.** I went looking for three defences and found none of them: a step-hiding mechanism in the `Step` struct, a handler that drops the raw block, and a downstream response-size cap or truncation helper. `maxResults`/`capMax`/`maxPages` apply only inside `fetchPaged`, which this step does not reach because it has no `followPages`. Its only bound is SEMP's `count`. The symptom is permanent, silent, and on every call — no error, no log, nothing an operator would notice.

**Recommended fix.** Either is cheap and both are narrower than reverting the fix:

- Have `ListVpns` replace each `byKey` entry with the boolean it actually consumed, e.g. `{"hasRealClient": true}`, before returning. Smallest possible change, keeps the scan fix exactly as-is.
- Or add the step-level bound the ticket's option 2 described — a `maxItems`/`probe: true` field that truncates a fan-out step's stored `data[]` after the SEMP call while leaving `count` at 100. This fixes the payload *and* gives the framework the primitive that would let `followPages` be turned on safely later, which is the thing currently blocking a real fix for the >100-reserved-client residual.

Whichever you pick, the comment at lines 273-274 needs to stop claiming the output is a boolean.

test/e2e-monitoring/test-monitoring-tools.sh
```diff
@@ -145,10 +151,68 @@ test_list_vpns_summary() {
...
+    if [ "$real_count" -lt 1 ]; then
+        log_fail "$label: ground truth expected >=1 real client on $BROKER_VPN, got $real_count"
+        return 1
+    fi
+    if [ "$reserved_count" -lt 2 ]; then
+        log_fail "$label: ground truth expected >=2 reserved (#*) clients ahead of the real one on $BROKER_VPN (AC4: bridges/Kafka/RDPs), got $reserved_count"
+        return 1
+    fi
+
+    small_page=$(semp_monitor_get "$broker_url" "msgVpns/$BROKER_VPN/clients?count=2") || {
```
bczoma, line 199: 🟡 **The new "bug shape" assertions don't establish the bug shape**

*Flagged by good:code. This overlaps Copilot's thread on line 207 and extends it — please read them together rather than as two separate asks; the fix below resolves both.*

The comment above this block promises to make the reserved-first ordering "a live, per-run fact rather than a one-off measurement". Two gaps stop it doing that.

**1. The small page drops the `where` clause.** This query is `clients?count=2` with no filter, and the real/reserved split is then done client-side in jq. But the root cause in SOL-153071 is specifically that *SEMP* applies `where` after the count cut. The real probe issues `count=N&where=clientUsername!=#*`. As written, the check reproduces a different mechanism — client-side filtering of an unfiltered page — so if the server-side filter-after-cut behaviour ever changed (in either direction), this assertion would not notice, and it is the only thing standing between that change and a silent regression.

**2. `reserved_count >= 2` (line 194) proves population, not position.** It counts reserved clients anywhere in the 100-row page, and does so by `clientUsername`, while SEMP orders by `clientName`. A broker that returned the real client *first* would satisfy every new assertion here. Ordering is the whole claim of AC4 and the whole reason the fixture is a regression fixture rather than a coincidence. Copilot's point about `small_real` is the same gap seen from the other end.

**Recommended fix.** Query the server-side mechanism directly and assert the strong form:

```bash
small_page=$(semp_monitor_get "$broker_url" \
    "msgVpns/$BROKER_VPN/clients?count=2&where=clientUsername!=%23*") || { ...; }
small_real=$(jq '[(.data // [])[]] | length' <<<"$small_page")
if [ "$small_real" -ne 0 ]; then
    log_fail "$label: fixture no longer reproduces SOL-153071 — count=2 with the server-side filter already returned $small_real real client(s)"
    return 1
fi
```

That is exactly the production query at the broken depth, it gives the `small_real == 0` strength Copilot asked for, and it is backed by your own lab measurement (zero real clients at `count=1` and `count=2`). For the ordering half, assert on position from the `count=100` body you already have — e.g. the first two rows *as returned* are reserved:

```bash
jq -e '[.data[0:2][] | select(.clientName | startswith("#"))] | length == 2' <<<"$ground_truth"
```

Filtering on `clientName` there keeps the assertion consistent with the attribute SEMP actually sorts by.

One smaller thing while you are in here: the final assertion at line 209 checks `.["real-clients"].byKey[...]`, the tool's internal step shape, rather than the user-facing outcome. Asserting that the default VPN is *not* counted under `summary.zeroConnectionCount` states the contract instead of the implementation, and survives any later move to the ticket's option 2 or option 3 (both of which change or delete that step).

test/e2e-monitoring/verify-fixtures.sh
```diff
@@ -76,6 +76,35 @@ verify_empty_enabled_vpn_state() {
...
+# reached the reserved client (which sorts first alphabetically) and never
+# scanned far enough to see a real one behind it. An earlier design
+# provisioned a separate `test-vpn-real-client` VPN for this, but this
+# broker image caps message-VPN count at 3 total (including `default`),
+# already exhausted by test-vpn/test-vpn-empty — reusing the default VPN's
+# existing F3 client avoids that limit entirely. Ground truth for the
+# GET-clients check comes from this direct SEMP call, not from the tool's
+# own probe under test — see test_list_vpns_summary in
+# test-monitoring-tools.sh for the corresponding tool-level assertion.
+verify_real_client_default_vpn_state() {
```
bczoma, line 92: 🟡 **AC2 is grounded on one side only — the "genuinely none" VPN is still self-referential**

*Flagged by good:code.*

The acceptance criterion asks that the E2E assertion "check against independently-known ground truth (**a fixture VPN with a known real client, and one with genuinely none**)". This new function delivers the first half well: a direct `__private_monitor__` GET for F3's client on the default VPN, on a path the composite tool never touches.

The second half is still missing. `verify_empty_enabled_vpn_state` above asserts only `enabled` and `state` on `test-vpn-empty`; nothing independently establishes that it has zero *real* clients. So the zero side of `zeroConnectionCount` is validated only through `summary.zeroConnectionCount >= 1` recomputed from the same `byKey` the handler consumed — precisely the self-referential shape this PR set out to remove.

**Why it matters.** The asymmetry leaves a live blind spot with the same signature as the original bug. A probe that returned rows for every VPN would be caught by the new real-client check. A probe that returned *nothing* for every VPN — the SOL-153071 failure mode exactly — is caught on the real-client side but would still find agreement on the empty side, because a uniformly-empty probe and a recompute over it agree. Half a two-sided oracle is not a two-sided oracle.

**Recommended fix.** Add the mirror check here or in `verify_empty_enabled_vpn_state`:

```bash
body=$(semp_monitor_get "$broker_url" "msgVpns/test-vpn-empty/clients?count=100") || { ...; }
real=$(jq '[(.data // [])[] | select((.clientUsername | startswith("#")) | not)] | length' <<<"$body")
[ "$real" -eq 0 ] || { log_fail "...: test-vpn-empty expected 0 real clients, got $real"; return 1; }
```

and then, in `test_list_vpns_summary`, assert that `real-clients.byKey["test-vpn-empty"]` is absent or empty. That gives you the independently-grounded pair the AC asks for, and it is the assertion that would fail if the probe ever regresses to returning nothing.

internal/composite/definitions/tools.yaml
```diff
@@ -270,9 +270,20 @@ tools:
...
+      # row returned means the VPN has a real client. `#`-prefixed names sort
+      # first, so count=1 (SOL-153071) only ever saw a reserved client and
+      # never scanned far enough to reach a real one on a VPN with more than
+      # one reserved client. 100 covers that. followPages is deliberately not
+      # set here: maxResults is shared across every step in the tool call
+      # (resolveMaxResults reads it once from the top-level params), so
+      # pagination on this step would silently couple "how many VPNs
+      # list-vpns returns" to "how many clients get scanned per VPN" via an
+      # unrelated parameter. A VPN with over 100 reserved clients ahead of
+      # its first real one would still false-negative; accepted as an
```
bczoma, line 283: 🟡 **The accepted limitation and the deferred framework fix point at nothing**

*Flagged by good:architecture and good:pr.*

The ticket deliberately left three options open and asked the implementer to choose: (1) the `count: "100"` + `followPages` stopgap, described in the ticket itself as leaving "the overfetch and the VPN-knob/probe-cost coupling behind as a trap"; (2) add a per-step item cap or short-circuit, then fix the probe; (3) drop the counter. This PR takes a variant of option 1 — `count: "100"` *without* `followPages` — which is a defensible call, and the reversibility argument is genuinely in its favour: it is one YAML literal, and option 2 remains purely additive to `Step`.

What is missing is the other half of that trade. Three things end here rather than pointing anywhere:

- **The accepted false negative.** "Accepted as an extreme case far rarer than the bug this fixes" is a reasonable judgement, but `100` is a convention rather than a derived bound — the repo's own SEMP spec gives `countQuery` a `minimum: 1` and `default: 10` and no maximum, so 100 is simply what the other batch steps happen to use. Reserved clients scale with deployment shape, and a single integration VPN with a few dozen RDP bindings plus a Kafka bind set clears 100 without being unusual.
- **The framework gap.** `Step` has `FollowPages`, `ForEach` and `Concurrency` and nothing for "stop after N items" or "stop at first match", which is all this probe needs and is also what would make `followPages` safe to turn on here. Nothing in the repo records that this primitive is missing.
- **The shared-`maxResults` coupling.** The reasoning for omitting `followPages` is correct and I verified it — `resolveMaxResults` reads one params bag that `fetchFanOut` passes to every iteration — but the AC asked to "document or resolve", and a documented trap with no owner is still a trap. The next step that wants "scan wide, return narrow" hits the same wall and makes the same local choice.

**Why it matters.** All three are the same risk: the deferred correct fix quietly becomes permanent, and the next reader has no way to tell whether 100 is still enough or who owns making it not matter. It also compounds the 🔴 on line 329 — the per-step item cap is the single change that would resolve both the response-payload problem and the scan bound.

**Recommended fix.** File the option-2 follow-up (a per-step `maxItems`/`stopAfter` on `Step`) and cite the ticket key from this comment, so the accepted-risk note points at a tracked item instead of ending at "accepted". Mirroring the choice in the PR body — which option was taken, what the deferred one costs, and where it now lives — would also let a reviewer sign off on the decision rather than infer it from a code comment.

bczoma (changes requested):
## PR Review: #437 — SOL-153071: fix list-vpns zeroConnectionCount false positives

**Author:** McCly-Solace | **Branch:** `sol-153071-zero-connection-count-fix` → `main` | **Changes:** +135 / -11 across 6 files
**Jira:** `SOL-153071` — list-vpns: zeroConnectionCount always reports every enabled+up VPN as having zero client connections (Verifying)

---

### Verdict

**Recommendation:** 🚫 Request changes

1. 🔴 `internal/composite/definitions/tools.yaml:329` — the probe's raw rows now ship in the tool response, not just the boolean the handler reads; `count: 1` → `100` turns roughly 0 rows per VPN into up to ~100, serialized twice
2. 🟡 `test/e2e-monitoring/test-monitoring-tools.sh:199` — the new "bug shape" assertions don't establish the bug shape: the small page drops the `where` clause, and the reserved-client check proves population, not position
3. 🟡 `test/e2e-monitoring/verify-fixtures.sh:92` — AC2 asked for independent ground truth on *both* sides; the "genuinely none" VPN still has none, so that half stays self-referential
4. 🟡 `internal/composite/definitions/tools.yaml:283` — the accepted limitation and the deferred framework fix point at no ticket

- 🟡 `internal/composite/definitions/tools.yaml:219-232` (the `list-vpns` `description`) — the accepted 100-client bound is stated only in a YAML comment the loaded tool definition strips. Full detail at the end of this review.

---

### Summary

This is a well-diagnosed fix for a real and embarrassing bug, and the most valuable thing in it is not the one-line YAML change — it is the observation that the old E2E check recomputed its expected value from the same probe output the handler consumed, and so could never have caught a wrong probe. Replacing that with a direct `__private_monitor__` call the tool never sees is the right structural correction.

The blocker is a consequence of the fix that nobody measured. Because `count`'s `where` filter is applied after the page cut, the pre-fix probe returned roughly zero rows per VPN; post-fix it returns the real clients among the first 100 scanned. Those rows are not internal — `ApplyResultStrategy`'s `postProcess` branch runs the handler and *then* `collectSteps`, which copies every step result into the response verbatim, and `buildValidatedResult` emits the whole payload twice. The YAML comment's claim that `count=100 is still a boolean check, not an enumeration` is true of the handler's use and false of the wire output.

Three of the review's initial findings were filed and then measured down under challenge, and are reported here at the severity the measurement supports rather than the one they were filed at — notably a claimed availability risk from 100× broker-side scan amplification, which turns out to be capped by an unconditional per-broker pacer (10 req/s, 10 in-flight, 30s shed) and to be strictly less work than one page of the already-shipped `list-clients`. That one is now a comment nit, not a concern.

---

### Minor notes

- 🟢 `CHANGELOG.md:141-145` — "when its only real client sorts behind the automatic reserved `#client`" reads as a rare ordering edge case; the PR body says every enabled+up VPN was affected, on every broker. An operator deciding whether to distrust historical readings needs the real blast radius, and the residual >100-reserved-client caveat.
- 🟢 `internal/composite/definitions/tools.yaml:276-277` — "never scanned far enough to reach a real one **on a VPN with more than one reserved client**" understates it: a *single* reserved client sorting first defeats `count=1`, which is effectively every enabled+up VPN. Reword to "at least one reserved client ahead of its first real one".
- 🟢 `internal/composite/definitions/tools.yaml:274-275` — the comment leans on "`#`-prefixed names sort first", but SEMP sorts by `clientName` while the filter is on `clientUsername`. The bound depends on reserved objects carrying both; say so, so a future reader can check whether 100 is still enough.
- 🟢 `test/e2e-monitoring/test-monitoring-tools.sh:188-189,203` — the three jq command substitutions are unchecked, and because `run_test` invokes the function under `if`, `set -e` is suppressed: a jq error leaves the variable empty, `[ "" -lt 1 ]` returns 2, and the guard fails **open** while the test still reports PASS. Reproduced by execution. Reachability is very low (`semp_monitor_get` uses `curl -sf`, so non-2xx is already caught, and `{"data":[]}` evaluates fine) and the primary `byKey >= 1` assertion is independent of this body — but `// []` plus one `[[ =~ ^[0-9]+$ ]]` check is cheap, and this is a file whose thesis is that assertions which cannot fail are the bug.
- 🟢 `test/e2e-monitoring/helpers.sh:558-569` — a 12-line comment attached to no code, wedged between `cleanup_connected_client_on` and the F4 constants, introducing an `F1b` label that appears nowhere else in the F1–F10 scheme. The same rationale already lives in `verify-fixtures.sh`, `test-monitoring-tools.sh` and the README's F3 row; delete this copy.

---

### Summary Verdict

|---|---|---|
| Architecture | ❌ | good:architecture |
| Documentation | ⚠️ | docs |


---

### 🟡 (not anchorable) `internal/composite/definitions/tools.yaml:219-232`

*Flagged by docs. GitHub would not accept a thread here — the `description` at lines 219-232 is not in this diff, because it is unchanged, which is the point.*

Copilot has already argued the >100-reserved-client case should be *fixed*. This is the separate question of what happens if you accept it, which the YAML comment says you are doing: the accepted bound needs to appear in the contract, and right now it appears only in a comment the loaded tool definition strips.

The description still reads that `zeroConnectionCount` counts "enabled+up VPNs with no non-reserved client connected — derived by probing each enabled+up VPN for a client whose clientUsername does not match Solace's reserved `#*` prefix". It states no bound. After this change the probe inspects the first 100 clients only, and the comment at line 283 concedes "A VPN with over 100 reserved clients ahead of its first real one would still false-negative".

**Why it matters, concretely.** This string is the entire user-visible contract for the field, and an LLM reads it verbatim when deciding how to answer "which VPNs are idle?". The >100-reserved-client shape is not exotic — it is bridges plus replication plus RDP consumer flows plus Kafka binds plus DMR links, i.e. exactly the integration-heavy VPN an operator is most likely to ask about. Before this PR the field was uniformly, obviously wrong, so nobody trusted it. After it, the field is right almost always, operators will start trusting it, and the residual wrong answer is a plausible-looking lie about one specific VPN. An operator acting on "this VPN has no clients" can disable or decommission a VPN carrying live traffic.

**Recommended fix.** One sentence on the description:

> The probe inspects up to 100 clients per VPN; a VPN whose first non-reserved client sorts behind more than 100 reserved (`#*`) clients is still counted under `zeroConnectionCount`.

Worth considering alongside it: the `maxResults` parameter description (lines 236-239) still reads only "Maximum number of VPNs to return (default 100, max 500)", with nothing telling the caller that each returned VPN costs an additional fan-out SEMP request. The Jira AC asked to "document or resolve" that coupling, and the YAML comment resolves it in code but documents it somewhere the model choosing `maxResults` will never read.

